PR #2404, opened 2021-09-23
Implement the metrics dispose/shutdown logic

reyang:
With #2389 merged, now we can fix the shutdown logic.

## Changes

* Updated the metric readers' logic to make Shutdown actually work.
* Updated the getting-started docs (no loops anymore).


docs/metrics/getting-started/Program.cs
```diff
@@ -31,9 +31,11 @@ public static void Main(string[] args)
 
         var counter = MyMeter.CreateCounter<long>("MyCounter");
```
cijothomas, line 32:
```suggestion
        var counter = MyMeter.CreateCounter<long>("MyFruitCounter");
```

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/MetricTests.cs
```diff
@@ -84,9 +84,6 @@ void ProcessExport(Batch<Metric> batch)
             // giving some breezing room for the End callback to complete
             await Task.Delay(TimeSpan.FromSeconds(1));
 
-            // Invokes the TestExporter which will invoke ProcessExport
-            metricReader.Collect();
```
cijothomas, line 88: just confirming:
if we remove explicit Collect(), then we are relying on the provider dispose... which has a hard-coded 5000 msec to finish flushing.
On the other hand, if we keep explicit Collect(), we can pass an explicit timeout to it. (we were not doing it now and was relying on Infinite timeout).

Net effect of this change in test code:
Previously we would wait indefinitely for Collect() to be over.
With this PR, we hope Collect() will be over in 5 secs.

If we ever need more than 5 sec, we need to bring back Collect(maxTimeInMilliSecs..).

reyang: Yes. If we need more time, what we need to do in the test cases:
1. never rely on provider Dispose for validation.
2. always use Collect explicitly, and validate the result after Collect.
3. after all the validation is done, use provider Dispose to release the resources.

src/OpenTelemetry/Metrics/BaseExportingMetricReader.cs
```diff
@@ -97,7 +99,22 @@ protected override bool OnCollect(int timeoutMilliseconds)
...
+            {
+                result = this.Collect(Timeout.Infinite) && result;
+                result = this.exporter.Shutdown(Timeout.Infinite) && result;
+            }
+            else
+            {
+                var sw = Stopwatch.StartNew();
+                result = this.Collect(timeoutMilliseconds) && result;
+                var timeout = timeoutMilliseconds - sw.ElapsedMilliseconds;
+                result = this.exporter.Shutdown((int)Math.Max(timeout, 0)) && result;
```
cijothomas, line 114: we'll need to modify PrometheusExporter to override shutdown, and shut its httpserver, right? 
And obviously, for pull based exporters, any pending data is simply lost on shutdown right?

reyang: Yes and yes.

cijothomas (approved):
LGTM.
PrometheusExporter need to override onshutdown to cleanup httpserver. but can be done after this

docs/metrics/getting-started/Program.cs
```diff
@@ -29,11 +29,13 @@ public static void Main(string[] args)
...
-        for (int i = 0; i < 20000000; i++)
-        {
-            counter.Add(1, new("tag1", "value1"), new("tag2", "value2"));
-        }
+        counter.Add(1, new("name", "apple"), new("color", "red"));
+        counter.Add(2, new("name", "lemon"), new("color", "yellow"));
+        counter.Add(1, new("name", "lemon"), new("color", "yellow"));
+        counter.Add(2, new("name", "apple"), new("color", "green"));
+        counter.Add(5, new("name", "apple"), new("color", "red"));
+        counter.Add(4, new("name", "lemon"), new("color", "yellow"));
```
mic-max, line 39: I like this new example topic!

PR #2406, opened 2021-09-23
Simplify tutorials

reyang:
## Changes

* Changed the metrics tutorials so we only have **one** "getting-started" rather than 4.
* Changed ConsoleExporter so by default we don't do periodic export.


src/OpenTelemetry.Exporter.Console/ConsoleExporterOptions.cs
```diff
@@ -26,9 +27,9 @@ public class ConsoleExporterOptions
         public ConsoleExporterOutputTargets Targets { get; set; } = ConsoleExporterOutputTargets.Console;
 
         /// <summary>
-        /// Gets or sets the metric export interval in milliseconds. The default value is 1000 milliseconds.
+        /// Gets or sets the metric export interval in milliseconds. The default value is <c>Timeout.Infinite</c>.
         /// </summary>
-        public int MetricExportIntervalMilliseconds { get; set; } = 1000;
+        public int MetricExportIntervalMilliseconds { get; set; } = Timeout.Infinite;
```
cijothomas, line 32: need to modify the getting started readme to reflect that it wont show metrics every sec.

cijothomas (approved):

src/OpenTelemetry.Exporter.Console/ConsoleExporterOptions.cs
cijothomas: nvm. the previous PR had modified it, so no more changes required.

Issue #2361, opened 2021-09-17
InMemoryExporter for Metric is buggy


cijothomas:
Unlike Traces, the Batch<Metric> provided to the exporter is re-used by SDK to keep metrics in memory. This means that, after the Export(Batch<Metric>), returns, the memory must not be read by the exporters. If they do so, they might be reading incorrect values. 

InMemoryExporter is buggy in this way - the fix would be for the InMemoryExporter to "deep copy" the Metric, MetricPoints, before Export() call returns, so that it has own copy, unaffected by further aggregations.

TimothyMothra:
@cijothomas, I'm interested in working on this. 
I stumbled into a repro of this issue.

cijothomas:
@utpilla might know more but i think we tried to fix this.. but hit some blockers.

reyang:
Here are the principles I would suggest:
1. Perf - it shouldn't affect the performance of the main story (e.g. anyone who is not using InMemoryExporter will not see perf drop).
2. Simplicity - folks who are using InMemoryExporter should not do extra things. No extra public APIs introduced.

TimothyMothra:
Questions:
- Are we committed to deep-copying Metric?
- Can we change the behavior of InMemoryExporter for Metrics only?




### Regarding deep-copy Metric

The original suggestion was to deep-copy the `Metric`, but there are challenges to this approach:
1. Cannot use MemoryStream/Binaryormatter because classes are not marked [Serializeable].
2. Cannot use XmlSerializer or JsonSerializer because classes do not have parameterless constructors.
3. A Reflection-based deep-copy might be possible but will have issues w/ parameterless constructor. This could be worked around, but this would be expensive to maintain because any change in any class could affect this.
4. An approach using dotnet's [MemberwiseClone](https://docs.microsoft.com/en-us/dotnet/api/system.object.memberwiseclone?view=net-6.0#examples) pattern might work. 
   ```
   MemberwiseClone Metric
   MemberwiseClone Metric.AggregatorStore
   Deep-Copy Metric.AggregatorStore.MetricPoint[]
   ```
    This would require some new public methods.
    Currently, both the AggregatorStore and MetricPoint array are readonly fields.
    Either the readonly would need to be removed or would need to use reflection to override the fields.


### Can we change the behavior of InMemoryExporter?

[...]

PR #2408, opened 2021-09-23
Unit test for metric stream duplication

cijothomas:
Fixes #.

## Changes

### DIFF
--- a/test/OpenTelemetry.Tests/Metrics/MetricAPITest.cs
+++ b/test/OpenTelemetry.Tests/Metrics/MetricAPITest.cs
@@ -39,6 +39,60 @@ namespace OpenTelemetry.Metrics.Tests
         }
 
         [Theory]
+        [InlineData(AggregationTemporality.Cumulative)]
+        [InlineData(AggregationTemporality.Delta)]
+        public void StreamNamesDuplicatesAreNotAllowedTest(AggregationTemporality temporality)
+        {
+            var metricItems = new List<Metric>();
+            int metricCount = 0;
+            var metricExporter = new TestExporter<Metric>(ProcessExport);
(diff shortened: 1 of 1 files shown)


test/OpenTelemetry.Tests/Metrics/MetricAPITest.cs
```diff
@@ -38,6 +38,60 @@ public MetricApiTest(ITestOutputHelper output)
...
         }
 
+        [Theory]
+        [InlineData(AggregationTemporality.Cumulative)]
+        [InlineData(AggregationTemporality.Delta)]
+        public void StreamNamesDuplicatesAreNotAllowedTest(AggregationTemporality temporality)
+        {
+            var metricItems = new List<Metric>();
+            int metricCount = 0;
+            var metricExporter = new TestExporter<Metric>(ProcessExport);
```
reyang, line 48: Could we use InMemoryExporter?

Main reason is that we build InMemoryExporter for folks to use in their unit tests, I think we need to use it in our own tests, too.

cijothomas: Yes. Once https://github.com/open-telemetry/opentelemetry-dotnet/issues/2361 is fixed.

reyang: Does #2361 block this change?

cijothomas: no. this particular test is not blocked. 
Want to switch all the test in this class to use InMemoryExporter together once #2361 is ready.

reyang (approved):

Issue #728, opened 2020-06-15
Pooling StringBuilder instances


alexeyzimarev:
OpenTelemetry for .NET extensively uses `StringBuilder` that is continuously instantiated and, therefore, allocated and then loads the GC.

The object pooling ASP.NET Core extensions allow pooling object instances, in particular, `StringBuilder` even has a default policy. It is quite easy to use but it also introduces a dependency. Would it be a good idea to implement a simplified string builder pool inside the opentelemetry-net project and use it?

I worry especially about exporters that will be extensively used.

cijothomas:
Will the dependency be required only for exporter projects? or OT SDK itself?

eddynaka:
@cijothomas , i think @alexeyzimarev  was talking about ObjectPool https://docs.microsoft.com/en-us/aspnet/core/performance/objectpool?view=aspnetcore-3.1, which we would need to install https://www.nuget.org/packages/Microsoft.Extensions.ObjectPool.

Basically, we would need to inject:
```csharp
services.TryAddSingleton<ObjectPool<StringBuilder>>(serviceProvider =>
        {
            var provider = serviceProvider.GetRequiredService<ObjectPoolProvider>();
            var policy = new StringBuilderPooledObjectPolicy();
            return provider.Create(policy);
        });
```

and, then, use it:
```csharp
public async Task InvokeAsync(ObjectPool<StringBuilder> builderPool)
{          
    // Request a StringBuilder from the pool.
    var stringBuilder = builderPool.Get();
    // some code
} 
```

alexeyzimarev:
That's right. I think it can significantly decrease allocations but it introduces a dependency. Maybe it will be built-in in .NET 5, who knows.

cijothomas:
Thanks for clarifying. @alexeyzimarev could you point couple of places where we use the string-builder and are candidates for this? We need to evaluate if using ObjectPool, or own implementation of similar would be justified.

cijothomas:
Not sure if this is still relevant after all recent perf improvements in the exporters. Will check if we need this.

mic-max:
![image](https://user-images.githubusercontent.com/15165028/134570695-e6f31aad-4381-40f4-98e3-507bfca05ed2.png)
These are all 13 places we call the `StringBuilder` constructor

The 6 uses we care about
- `PrometheusMetricBuilder.cs` function `GetSafeName()`
- `HttpInListener.cs` function `GetUri()`
[...]

PR #2410, opened 2021-09-23
Update ConsoleMetricExporter to add a space between the attributes

utpilla:
## Changes
- Add space between metric attributes when printing to console
- Update `StringBuilder.Append` method calls to use character when possible: [CA1834: Use StringBuilder.Append(char) for single character strings](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1834)

### DIFF
--- a/docs/metrics/getting-started/README.md
+++ b/docs/metrics/getting-started/README.md
@@ -33,11 +33,11 @@ output from the console, similar to shown below:
 <!-- markdownlint-disable MD013 -->
 ```text
 Export MyFruitCounter, Meter: MyCompany.MyProduct.MyLibrary/1.0
-2021-09-23T03:17:30.6198292Z, 2021-09-23T03:17:30.6356517Z] color:redname:apple LongSum
+2021-09-23T21:34:11.7728070Z, 2021-09-23T21:34:11.7843300Z] color:red name:apple LongSum
 Value: 6
-2021-09-23T03:17:30.6198292Z, 2021-09-23T03:17:30.6356517Z] color:yellowname:lemon LongSum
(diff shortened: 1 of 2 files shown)


reyang (approved):
LGTM.
I think the console output needs to be redesigned, currently it's bit messy. (and the time range is missing the left `(`, and we use so many StringBuilders...)

utpilla:
> LGTM.
> I think the console output needs to be redesigned, currently it's bit messy. (and the time range is missing the left `(`, and we use so many StringBuilders...)

Added "(" to the time range.

reyang:
> > LGTM.
> > I think the console output needs to be redesigned, currently it's bit messy. (and the time range is missing the left `(`, and we use so many StringBuilders...)
> 
> Added '(' to the time range.

The interesting thing is that I did fix this previously https://github.com/open-telemetry/opentelemetry-dotnet/pull/2285/files, then it came back. I haven't looked into what happened 😵‍💫

cijothomas (approved):

mic-max:
> LGTM.
> I think the console output needs to be redesigned, currently it's bit messy. (and the time range is missing the left `(`, and we use so many StringBuilders...)

Related issue #728. I left a comment on it recently, any thoughts on going through with the pooling change? I'd be willing to implement it.

reyang:
> > LGTM.
> > I think the console output needs to be redesigned, currently it's bit messy. (and the time range is missing the left `(`, and we use so many StringBuilders...)
> 
> Related issue #728. I left a comment on it recently, any thoughts on going through with the pooling change? I'd be willing to implement it.

Ideally, if the purpose is to format a string and then send it to the output, I think we don't need StringBuilder at all. What we can explore is to use stack-allocated memory and perform in-place formatting/serialization; If the purpose is to return a string for other components to use, a pool of pre-allocated StringBuilder would be great.

Issue #1258, opened 2020-09-10
Proposal to create public Logging APIs and log writer for easier Self Diagnostics experience


xiang17:
## Current Problems

Every project needs to create its own EventSource implementation, most of which is repeated boilerplate code to achieve the same purpose. Currently there are 14 EventSource implementations for emitting errors, yet it's not straight-forward to capture and log these EventSource events in case of need for troubleshooting.

## Proposal

To solve these problems, I'm trying to provide a cross-platform logging API and self diagnostics module, which provides two key features:

1. easy-to-use logging APIs for all OpenTelemetry modules to emit log messages/errors for troubleshooting/self diagnostics.
2. a self diagnostics module that write the log message to disk, when enabled. When disabled (default), the logging APIs would not write anything to disk.

### Logging APIs

The Logging APIs are some generic methods like LogVerbose or LogError, located in a public namespace/class for all OpenTelemetry to call for self diagnostics or troubleshooting. Once it's done, the use of other EventSources can be migrated to this. These Logging APIs still use EventSource as the underlying mechanism.

A sample use of the API I want to make is like this:
```
OpenTelemetrySelfDiagnostics.Log.LogInfo("Null pointer in exporter.");
OpenTelemetrySelfDiagnostics.Log.LogError("Null pointer in exporter.", exceptionObject);
```
(The exact namespace is to be figured out later.)

When the Self Diagnostics Module is enabled, the log message including timestamp, error level, class name, error message (and optionally exception stack) would be written to a local file as such:

```
2020-08-14T20:33:24.4788109Z ERROR OpenTelemetry.Exporter.Zipkin.ZipkinExporter: Null pointer in exporter. ExceptionStack:
...stack details...
```

### Self Diagnostics Module

1. The self diagnostics module captures and write log messages to disk, when enabled.
2. It should have minimal overhead to the current process so the user can feel confident to enable it.
3. It should be accessible to use in most environments/cases. For example, we want to use the least dependencies and/or features of the OS. It doesn't require administrator permissions. It can be enabled/disabled without restarting the process.
4. Bounded resource usage. For example, it should not consume ever-growing memory or generate an ever-growing file.

## Implementation Plan

1. The logger object sits in a public location as a static variable for everyone to call.
2. The logger provide generic log methods, which internally emits EventSource events.
3. Self Diagnostics Module listens to the events with EventListener.
[...]

PR #2421, opened 2021-09-28
Support View part1 - rename an instrument

cijothomas:
Currently, the Metric streams use the same name as that of the instrument. This PR allows user to rename a metric stream by using Views.

Option1 - for simple renames
`.AddView(instrumentName: "MyCounter", name: "MyCounterRenamed")`

Option2 (overkill for simple renaming,)
`.AddView(instrumentName: "MyFruitCounter", new MetricStreamConfiguration() { Name = "MyCounterRenamed" })`

Option3 (overkill for simple renaming)
```
.AddView((instrument) =>
            {
                if (instrument.Name.Equals("MyFruitCounter"))
                {
                    return new MetricStreamConfiguration() { Name = "MyCounterRenamed" };
                }

                return null;
            })
```

Advanced usage: Here user is configuring a view to rename every instrument, to prefix with the meter name:

```
.AddView((instrument) =>
            {
                return new MetricStreamConfiguration() { Name = instrument.Meter.Name + ":" + instrument.Name };
            })
```

Another example where the user is renaming a particular instrument from a particular meter.
```
.AddView((instrument) =>
            {
                if (instrument.Meter.Name.Equals("MyCompany.MyProduct.MyLibrary") &&
instrument.Name.Equals("MyFruitCounter"))
                {
                    return new MetricStreamConfiguration() { Name = "RenamedFruitCounter" };
                }

                return null;
            })
```


docs/metrics/customizing-the-sdk/Program.cs
```diff
@@ -60,7 +60,7 @@ public static void Main(string[] args)
             // turn off the above default. i.e any
             // instrument which does not match any views
             // gets dropped.
-            // .AddView(instrumentName: "*", new DropAggregationConfig())
+            // .AddView(instrumentName: "*", new MetricStreamConfiguration() { Aggregation = Aggregation.Drop })
```
reyang, line 63: Consider making a constant (e.g. `MetricStreamConfiguration.Drop`).

cijothomas: yea. will work on it next step.
(This PR doesn't handle Aggregation part, only name change is covered)

test/OpenTelemetry.Tests/Metrics/MetricViewTests.cs
```diff
@@ -0,0 +1,56 @@
...
+        }
+
+        [Fact]
+        public void ViewToRenameMetric()
+        {
+            var exportedItems = new List<Metric>();
+            var inMemoryReader = new BaseExportingMetricReader(new InMemoryExporter<Metric>(exportedItems));
+            using var meterProvider = Sdk.CreateMeterProviderBuilder()
+            .AddSource("TestMeter1")
+            .AddReader(inMemoryReader)
```
reyang, line 41: Use `AddInMemoryExporter` extension method.

test/OpenTelemetry.Tests/Metrics/MetricViewTests.cs
```diff
@@ -0,0 +1,56 @@
...
+            this.output = output;
+        }
+
+        [Fact]
+        public void ViewToRenameMetric()
+        {
+            var exportedItems = new List<Metric>();
+            var inMemoryReader = new BaseExportingMetricReader(new InMemoryExporter<Metric>(exportedItems));
+            using var meterProvider = Sdk.CreateMeterProviderBuilder()
+            .AddSource("TestMeter1")
```
reyang, line 40: Indent.

src/OpenTelemetry/Internal/OpenTelemetrySdkEventSource.cs
```diff
@@ -340,6 +340,24 @@ public void ExistsDroppedExportProcessorItems(string exportProcessorName, string
             this.WriteEvent(32, exportProcessorName, exporterName, droppedCount);
         }
 
+        [Event(33, Message = "Measurements from instrument '{0}' is ignored as maximum metric streams limit reached.", Level = EventLevel.Warning)]
```
reyang, line 343: Curious, do we need these separate events, or one general event would work fine? (my worry is that we keep adding events and we don't know how to manage them)

src/OpenTelemetry/Metrics/MeterProviderBuilderBase.cs
```diff
@@ -88,19 +89,51 @@ internal MeterProviderBuilder AddReader(MetricReader reader)
...
+            {
+                var needRegex = false;
+                if (instrumentName.IndexOf('*') != -1)
+                {
+                    needRegex = true;
+                }
+
+                if (needRegex)
+                {
+                    var pattern = '^' + Regex.Escape(instrumentName).Replace("\\*", ".*");
```
reyang, line 108: This will create string for every single instrument, which is very costly.
Need to extract the logic we don't repeat things.

test/OpenTelemetry.Tests/Metrics/MetricViewTests.cs
```diff
@@ -0,0 +1,106 @@
...
+            .AddReader(inMemoryReader)
+            .AddView("name1", "renamed")
+            .Build();
+
+            using var meter1 = new Meter("TestMeter1");
```
reyang, line 45: Consider using full qualified meter name to avoid conflicts with other concurrent test runs.

test/OpenTelemetry.Tests/Metrics/MetricViewTests.cs
```diff
@@ -0,0 +1,106 @@
...
+
+            // Expecting one metric stream.
+            var counterLong = meter1.CreateCounter<long>("name1");
+            counterLong.Add(10);
+            inMemoryReader.Collect();
```
reyang, line 50: Use `meterProvider.ForceFlush(1000)` which is a more common/convenient way.

src/OpenTelemetry/Internal/OpenTelemetrySdkEventSource.cs
cijothomas: would be good to have separate event for instrument drop, for each unique reason. i.e dropped due to max Metric limit and dropped due to conflict are separate events.

reyang: I wish to understand more about this, will discuss with you offline.

src/OpenTelemetry/Metrics/MeterProviderBuilderBase.cs
```diff
@@ -88,19 +89,53 @@ internal MeterProviderBuilder AddReader(MetricReader reader)
...
+        internal MeterProviderBuilder AddView(string instrumentName, MetricStreamConfiguration metricStreamConfiguration)
         {
-            // TODO: Actually implement view.
-            return this;
+            var pattern = '^' + Regex.Escape(instrumentName).Replace("\\*", ".*");
+            var regex = new Regex(pattern, RegexOptions.Compiled | RegexOptions.IgnoreCase);
+
+            Func<Instrument, MetricStreamConfiguration> viewConfig = (instrument) =>
+            {
+                var needRegex = false;
```
cijothomas, line 103: move out everything not required in the callback.

src/OpenTelemetry/Metrics/MeterProviderSdk.cs
```diff
@@ -98,27 +99,84 @@ internal sealed class MeterProviderSdk : MeterProvider
                 {
                     if (meterSourcesToSubscribe.ContainsKey(instrument.Meter.Name))
                     {
-                        lock (this.instrumentCreationLock)
+                        var viewConfigCount = this.viewConfigs.Count;
```
cijothomas, line 102: todo: add benchmark for instrument creation

src/OpenTelemetry/Internal/OpenTelemetrySdkEventSource.cs
cijothomas: Removed the additional logging from this PR. I intend to bring it back after most Metric functionalities are in good shape, Need to also consider the original proposal about self-diagnostics https://github.com/open-telemetry/opentelemetry-dotnet/issues/1258.
It might be good enough to use generic Log messages for majority of the cases.
For important things like telemetry loss (for variety is reasons like cap, invalid name, conflict, unsubscribed source etc.), we need a dedicated Event, with "reason" as one of the parameter.
eg:
LogInstrumentIgnored(instrumentname, metername, reason)
LogMeasurementIgnored(instrumentname,metername,metricname,reason)

src/OpenTelemetry/Metrics/MeterProviderBuilderBase.cs
```diff
@@ -88,19 +89,41 @@ internal MeterProviderBuilder AddReader(MetricReader reader)
...
+            Regex regex = null;
+            if (instrumentName.IndexOf('*') != -1)
+            {
+                var pattern = '^' + Regex.Escape(instrumentName).Replace("\\*", ".*");
+                regex = new Regex(pattern, RegexOptions.Compiled | RegexOptions.IgnoreCase);
+            }
+
+            Func<Instrument, MetricStreamConfiguration> viewConfig = (instrument) =>
+            {
+                if (regex != null && regex.IsMatch(instrument.Name))
```
reyang, line 107: Move the if-else out of the Func.

src/OpenTelemetry/Metrics/MeterProviderBuilderBase.cs
```diff
@@ -88,19 +89,50 @@ internal MeterProviderBuilder AddReader(MetricReader reader)
...
+        internal MeterProviderBuilder AddView(string instrumentName, MetricStreamConfiguration metricStreamConfiguration)
         {
-            // TODO: Actually implement view.
-            return this;
+            Func<Instrument, MetricStreamConfiguration> viewConfig;
+            if (instrumentName.IndexOf('*') != -1)
+            {
+                var pattern = '^' + Regex.Escape(instrumentName).Replace("\\*", ".*");
+                var regex = new Regex(pattern, RegexOptions.Compiled | RegexOptions.IgnoreCase);
+                viewConfig = (instrument) =>
```
reyang, line 103: Maybe:

```csharp
return this.AddView(instrument => regex.IsMatch(instrument.Name) ? metricStreamConfiguration : null);
```

src/OpenTelemetry/Metrics/MeterProviderBuilderBase.cs
```diff
@@ -88,19 +89,50 @@ internal MeterProviderBuilder AddReader(MetricReader reader)
...
+                    {
+                        return null;
+                    }
+                };
+            }
+
+            return this.AddView(viewConfig);
         }
 
         internal MeterProviderBuilder AddView(Func<Instrument, MetricStreamConfiguration> viewConfig)
```
reyang, line 133: AggressivelyInline

test/OpenTelemetry.Tests/Metrics/MetricViewTests.cs
```diff
@@ -0,0 +1,166 @@
...
+        }
+
+        [Fact]
+        public void ViewToProduceMultipleStreamsWithDuplicatesFromInstrument()
+        {
+            using var meter1 = new Meter("ViewToProduceMultipleStreamsWithDuplicatesFromInstrumentTest");
+            var exportedItems = new List<Metric>();
+            using var meterProvider = Sdk.CreateMeterProviderBuilder()
+                .AddSource(meter1.Name)
+                .AddInMemoryExporter(exportedItems)
```
reyang, line 148: nit: move this line after views.

test/OpenTelemetry.Tests/Metrics/MetricViewTests.cs
```diff
@@ -0,0 +1,166 @@
...
+            using var meter2 = new Meter("ViewToRenameMetricConditionallyTest2");
+            var exportedItems = new List<Metric>();
+            using var meterProvider = Sdk.CreateMeterProviderBuilder()
+                .AddSource(meter1.Name)
+                .AddSource(meter2.Name)
+                .AddInMemoryExporter(exportedItems)
+                .AddView((instrument) =>
+                {
+                    if (instrument.Meter.Name.Equals(meter2.Name, StringComparison.OrdinalIgnoreCase)
+                    && instrument.Name.Equals("name1", StringComparison.OrdinalIgnoreCase))
```
reyang, line 68: indent?

reyang (approved):

src/OpenTelemetry/Metrics/MeterProviderBuilderBase.cs
cijothomas: not in the perf-critical instrument creation path. so won't be needed.

PR #2429, opened 2021-09-29
Support View Part2 - specify tags to be used for aggregation

cijothomas:
Note on perf:
Hotpath Perf is affected by this change, as there are lot of "if" checks now added in hotpath, even when there are no views. This can be fixed by following a totally different path for Views/No Views scenario, so as to keep maximum perf for no views scenario. To follow up in subsequent PRs *after* finishing more View features (change Aggregation etc).

PR Description:
Currently, all the tags provided with measurement is used for aggregation. This PR allows user to configure View, to configure TagKeys, which will be used for aggregation. Typically used for cost control, by dropping unwanted tags.

For example, consider the following snippet of measurements being recorded (assume console exporter is configured):

```
counter.Add(1, new("name", "apple"), new("color", "red"));
counter.Add(2, new("name", "lemon"), new("color", "yellow"));
counter.Add(1, new("name", "lemon"), new("color", "yellow"));
counter.Add(2, new("name", "apple"), new("color", "green"));
counter.Add(5, new("name", "apple"), new("color", "red"));
counter.Add(4, new("name", "lemon"), new("color", "yellow"));
```

The output is as follows:
Output:
```
Export MyFruitCounter, Meter: MyCompany.MyProduct.MyLibrary/1.0
(2021-09-29T23:31:23.7786143Z, 2021-09-29T23:31:23.7972342Z] color:red name:apple LongSum
Value: 6
(2021-09-29T23:31:23.7786143Z, 2021-09-29T23:31:23.7972342Z] color:yellow name:lemon LongSum
Value: 7
(2021-09-29T23:31:23.7786143Z, 2021-09-29T23:31:23.7972342Z] color:green name:apple LongSum
Value: 2
```

Below snippet uses view with both tags. (i.e same as default behavior)

```
.AddView(
                "MyFruitCounter",
                new MetricStreamConfiguration()
                {
                    TagKeys = new string[] { "color", "name" },
                })
```

Output:
```
Export MyFruitCounter, Meter: MyCompany.MyProduct.MyLibrary/1.0
(2021-09-29T23:31:23.7786143Z, 2021-09-29T23:31:23.7972342Z] color:red name:apple LongSum
Value: 6
(2021-09-29T23:31:23.7786143Z, 2021-09-29T23:31:23.7972342Z] color:yellow name:lemon LongSum
Value: 7
(2021-09-29T23:31:23.7786143Z, 2021-09-29T23:31:23.7972342Z] color:green name:apple LongSum
Value: 2
```

Below snippet uses view to only pick single tag "color"
```
.AddView(
                "MyFruitCounter",
                new MetricStreamConfiguration()
                {
                    TagKeys = new string[] { "color" },
                })
```

Output:
```
Export MyFruitCounter, Meter: MyCompany.MyProduct.MyLibrary/1.0
(2021-09-29T23:34:42.3563688Z, 2021-09-29T23:34:42.3722304Z] color:red LongSum
Value: 6
(2021-09-29T23:34:42.3563688Z, 2021-09-29T23:34:42.3722304Z] color:yellow LongSum
Value: 7
(2021-09-29T23:34:42.3563688Z, 2021-09-29T23:34:42.3722304Z] color:green LongSum
Value: 2
```

Below snippet uses view to only pick single tag "name"
```
.AddView(
                "MyFruitCounter",
                new MetricStreamConfiguration()
                {
                    TagKeys = new string[] { "name" },
                })
```

Output:
```
Export MyFruitCounter, Meter: MyCompany.MyProduct.MyLibrary/1.0
(2021-09-29T23:35:38.0528475Z, 2021-09-29T23:35:38.0775305Z] name:apple LongSum
Value: 8
(2021-09-29T23:35:38.0528475Z, 2021-09-29T23:35:38.0775305Z] name:lemon LongSum
Value: 7
```

Below snippet uses view to drop all tags.
```
.AddView(
                "MyFruitCounter",
                new MetricStreamConfiguration()
                {
                    TagKeys = new string[] {  },
                })
```

Output:
```
Export MyFruitCounter, Meter: MyCompany.MyProduct.MyLibrary/1.0
(2021-09-29T23:36:54.2341730Z, 2021-09-29T23:36:54.2440246Z] LongSum
Value: 15
```

Below snippet uses view to specify some non-existent tag. (similar as no tags!)
```
.AddView(
                "MyFruitCounter",
                new MetricStreamConfiguration()
                {
                    TagKeys = new string[] {  "foo" },
                })
```

Output:
```
Export MyFruitCounter, Meter: MyCompany.MyProduct.MyLibrary/1.0
(2021-09-29T23:36:54.2341730Z, 2021-09-29T23:36:54.2440246Z] LongSum
Value: 15
```

### DIFF
--- a/src/OpenTelemetry/Metrics/AggregatorStore.cs
+++ b/src/OpenTelemetry/Metrics/AggregatorStore.cs
@@ -70,7 +70,8 @@ namespace OpenTelemetry.Metrics
             int maxLength = tags.Length;
             if (maxLength == 0 || this.dropAllTags)
             {
-                return this.GetInitializedZeroTagPoint();
+                this.InitializeZeroTagPointIfNotInitialized();
+                return 0;
             }
 
             var storage = ThreadStaticStorage.GetStorage();
@@ -85,7 +86,8 @@ namespace OpenTelemetry.Metrics
 
             if (actualLength == 0)
             {
-                return this.GetInitializedZeroTagPoint();
+                this.InitializeZeroTagPointIfNotInitialized();
+                return 0;
             }
 
             if (actualLength > 1)
@@ -238,7 +240,7 @@ namespace OpenTelemetry.Metrics
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        private int GetInitializedZeroTagPoint()
+        private void InitializeZeroTagPointIfNotInitialized()
         {
             if (!this.zeroTagMetricPointInitialized)
             {
@@ -252,8 +254,6 @@ namespace OpenTelemetry.Metrics
                     }
                 }
             }
-
-            return 0;
         }
     }
 }
--- a/src/OpenTelemetry/Metrics/AggregatorStore.cs
+++ b/src/OpenTelemetry/Metrics/AggregatorStore.cs
@@ -17,6 +17,7 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Runtime.CompilerServices;
 using System.Threading;
 
 namespace OpenTelemetry.Metrics
@@ -26,6 +27,8 @@ namespace OpenTelemetry.Metrics
         internal const int MaxMetricPoints = 2000;
         private static readonly ObjectArrayEqualityComparer ObjectArrayComparer = new ObjectArrayEqualityComparer();
         private readonly object lockZeroTags = new object();
+        private readonly HashSet<string> tagKeysInteresting;
+        private readonly bool dropAllTags;
 
         // Two-Level lookup. TagKeys x [ TagValues x Metrics ]
         private readonly ConcurrentDictionary<string[], ConcurrentDictionary<object[], int>> keyValue2MetricAggs =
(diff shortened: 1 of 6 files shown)


src/OpenTelemetry/Metrics/AggregatorStore.cs
```diff
@@ -39,40 +42,55 @@ internal class AggregatorStore
...
-                }
-
+                this.InitializeZeroTagPointIfNotInitialized();
                 return 0;
             }
 
             var storage = ThreadStaticStorage.GetStorage();
 
-            storage.SplitToKeysAndValues(tags, out var tagKey, out var tagValue);
+            if (this.tagKeysInteresting != null
```
reyang, line 79: This is readonly, do we need to check it every time, or we can hard-wire it?

src/OpenTelemetry/Metrics/ThreadStaticStorage.cs
```diff
@@ -49,30 +49,44 @@ internal static ThreadStaticStorage GetStorage()
...
+            }
+            else
             {
-                throw new InvalidOperationException("There must be atleast one tag to use ThreadStaticStorage.");
+                tagKeys = new string[maxLength];
+                tagValues = new object[maxLength];
             }
 
-            if (len <= MaxTagCacheSize)
+            if (tagKeysInteresting != null)
```
reyang, line 65: Maybe we can step back and put some design principle like this - if the user does not specify a custom list of tags, we should spend absolute zero CPU cycle related to the `tagKeysInteresting` checks.

cijothomas: Pushed some changes.. close to it, but still not absolute zero. We do one if() check.

src/OpenTelemetry/Metrics/AggregatorStore.cs
```diff
@@ -39,54 +42,178 @@ internal class AggregatorStore
...
+                this.tagKeysInteresting = hs;
+                this.tagsKeysInterestingCount = hs.Count;
+            }
+        }
+
+        internal void UpdateLong(long value, ReadOnlySpan<KeyValuePair<string, object>> tags)
+        {
+            try
+            {
+                var index = this.tagKeysInteresting != null ? this.FindMetricAggregatorIndexUserConfigured(tags) : this.FindMetricAggregatorIndexDefault(tags);
```
reyang, line 71: We don't have check `this.tagKeysInteresting` each time because it can be determined ahead of time?

src/OpenTelemetry/Metrics/AggregatorStore.cs
```diff
@@ -39,54 +42,178 @@ internal class AggregatorStore
...
+                        var dt = DateTimeOffset.UtcNow;
+                        this.metrics[0] = new MetricPoint(this.aggType, dt, null, null);
+                        this.zeroTagMetricPointInitialized = true;
                     }
                 }
+            }
+        }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private int FindMetricAggregatorIndexDefault(ReadOnlySpan<KeyValuePair<string, object>> tags)
```
cijothomas, line 156: inside this method, no extra cycles spent checking if user configured tagkeys.

src/OpenTelemetry/Metrics/AggregatorStore.cs
cijothomas: The approach used here : https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/Trace/TracerProviderSdk.cs#L35, was not doable here, as the argument to the function is a ref struct (ReadOnlySpan), so it cannot be declares as part of instance variable..

Need to some other tricks. Keeping this comment open now.

cijothomas: @CodeBlanch shared an idea of using our own Delegate to overcome the limitation of ref structs..Exploring it now.

src/OpenTelemetry/Metrics/ThreadStaticStorage.cs
```diff
@@ -49,30 +49,63 @@ internal static ThreadStaticStorage GetStorage()
...
-                tagKeys = new string[len];
-                tagValues = new object[len];
+                tagKeys = new string[actualLength];
+                tagValues = new object[actualLength];
             }
 
-            for (var n = 0; n < len; n++)
+            // Iterate again (!) to assign the actual value.
+            i = 0;
+            for (var n = 0; n < tags.Length; n++)
```
cijothomas, line 100: @CodeBlanch and I discussed this offline.
Alternate to the dual iteration is to use a List with initial capacity of maxlength.. We'd be "wasting" some memory, but it'll avoid the need of multiple iterations.
Will do benchmarks and pick.

src/OpenTelemetry/Metrics/AggregatorStore.cs
cijothomas: Using delegate approach to avoid this perf hit : https://github.com/open-telemetry/opentelemetry-dotnet/pull/2442/files

cijothomas:
Closing in favor of https://github.com/open-telemetry/opentelemetry-dotnet/pull/2442

PR #2432, opened 2021-09-30
Support View Part2 - specify custom buckets for Histogram

cijothomas:
Similar perf concern as https://github.com/open-telemetry/opentelemetry-dotnet/pull/2429, but mostly restricted to Meter creation and not in the measurement recording.

A very simple example showing custom bounds for Histogram

```
            .AddSource("MyCompany.MyProduct.MyLibrary")
           .AddView("MyHistogram", new HistogramConfiguration() { BucketBounds = new double[] { 10, 20 } })
            .AddConsoleExporter()
            .Build();

        var random = new Random();
        var histogram = MyMeter.CreateHistogram<long>("MyHistogram");
        for (int i = 0; i < 1000; i++)
        {
            histogram.Record(random.Next(1, 1000));
        }
```

The following can make the histogram just do Count,Sum with no buckets. (This needs to be adjusted once MinMax spec is ready)
` .AddView("MyHistogram", new HistogramConfiguration() { BucketBounds = new double[] {  } })`

### DIFF
--- a/src/OpenTelemetry/Metrics/AggregationType.cs
+++ b/src/OpenTelemetry/Metrics/AggregationType.cs
@@ -57,5 +57,10 @@ namespace OpenTelemetry.Metrics
         /// Histogram.
         /// </summary>
         Histogram = 6,
+
+        /// <summary>
+        /// Histogram with no sum,count only.
+        /// </summary>
+        HistogramSumCount = 7,
     }
 }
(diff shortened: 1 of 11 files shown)


src/OpenTelemetry/Metrics/AggregationType.cs
```diff
@@ -57,5 +57,10 @@ internal enum AggregationType
         /// Histogram.
         /// </summary>
         Histogram = 6,
+
+        /// <summary>
+        /// Histogram with no sum,count only.
```
reyang, line 62:
```suggestion
        /// Histogram with no sum, count only.
```

reyang (approved):

PR #2433, opened 2021-09-30
Prometheus fixes

cijothomas:
example to use localhost instead of "*", as it seems like higher privilege needed to run "*".
For HttpListener, modify path to end in "/".

### DIFF
--- a/examples/Console/TestPrometheusExporter.cs
+++ b/examples/Console/TestPrometheusExporter.cs
@@ -52,7 +52,7 @@ namespace Examples.Console
                 .AddPrometheusExporter(opt =>
                 {
                     opt.StartHttpListener = true;
-                    opt.HttpListenerPrefixes = new string[] { $"http://*:{port}/" };
+                    opt.HttpListenerPrefixes = new string[] { $"http://localhost:{port}/" };
                 })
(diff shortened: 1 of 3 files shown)


reyang (approved):

examples/Console/TestPrometheusExporter.cs
```diff
@@ -52,7 +52,7 @@ internal static object Run(int port, int totalDurationInMins)
                 .AddPrometheusExporter(opt =>
                 {
                     opt.StartHttpListener = true;
-                    opt.HttpListenerPrefixes = new string[] { $"http://*:{port}/" };
+                    opt.HttpListenerPrefixes = new string[] { $"http://localhost:{port}/" };
```
CodeBlanch, line 55: Interesting! Happen to catch an exception message or anything I can look into?

CodeBlanch (approved):

examples/Console/TestPrometheusExporter.cs
cijothomas: image

Separately, we want to ensure no exception are thrown from the Worker task, as it is not observed anywhere.

PR #2436, opened 2021-09-30
Add warning when running stress test with DEBUG build

ThomsonTan:
## Changes

Running stress test with DEBUG build may not be expected. Add warning message in this scenario before running the stress test.

### DIFF
--- a/test/OpenTelemetry.Tests.Stress/Skeleton.cs
+++ b/test/OpenTelemetry.Tests.Stress/Skeleton.cs
@@ -83,7 +83,7 @@ public partial class Program
                     var cntCpuCyclesOld = GetCpuCycles();
 
                     watch.Restart();
-                    Thread.Sleep(2000);
+                    Thread.Sleep(200);
                     watch.Stop();
 
                     var cntLoopsNew = (ulong)statistics.Sum();
--- a/test/OpenTelemetry.Tests.Stress/Skeleton.cs
+++ b/test/OpenTelemetry.Tests.Stress/Skeleton.cs
@@ -30,6 +30,10 @@ public partial class Program
 
     public static void Stress(int concurrency = 0)
     {
+#if DEBUG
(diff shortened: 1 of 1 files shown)


cijothomas (approved):

test/OpenTelemetry.Tests.Stress/Skeleton.cs
```diff
@@ -30,6 +30,10 @@ public partial class Program
 
     public static void Stress(int concurrency = 0)
     {
+#if DEBUG
+        Console.WriteLine("***WARNING*** The current build is DEBUG which may affect timing!\n");
```
reyang, line 34: Why do we need `\n` if we already use `WriteLine`?

ThomsonTan: I'd like to add an empty line after the current message, to make it more visible instead of combining it with other benchmark output.

reyang: Need to use https://docs.microsoft.com/en-us/dotnet/api/system.environment.newline?view=net-5.0

ThomsonTan: Fixed it in https://github.com/open-telemetry/opentelemetry-dotnet/pull/2437

ThomsonTan: Just curious with one more thought, new line representation is indeed platform dependent, but `Console.WriteLine("\n")` should go to C library as `printf("\n")`. It is unintuitive if `printf("\n")` doesn't work consistently on all platforms, and `printf()` usually does the translation of `\n` to the platform specific new line.

Anyway, `Console.WriteLine("\n")` could choose its own implementation without going to `printf()`.

reyang: I guess that's why C++ introduced `std::endl`.

PR #2438, opened 2021-09-30
Allow description override using View

cijothomas:
Example snippet:

```
AddView("MyFruitCounter", new MetricStreamConfiguration() { Name = "new name", Description = "modified description" })
```

### DIFF
--- a/src/OpenTelemetry/Metrics/Metric.cs
+++ b/src/OpenTelemetry/Metrics/Metric.cs
@@ -29,17 +29,8 @@ namespace OpenTelemetry.Metrics
             Instrument instrument,
             AggregationTemporality temporality,
             string metricName,
-            string metricDescription)
-            : this(instrument, temporality, metricName, metricDescription, DefaultHistogramBounds)
-        {
-        }
-
-        internal Metric(
-            Instrument instrument,
-            AggregationTemporality temporality,
-            string metricName,
             string metricDescription,
-            double[] histogramBounds)
+            double[] histogramBounds = null)
         {
             this.Name = metricName;
             this.Description = metricDescription;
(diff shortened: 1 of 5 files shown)


src/OpenTelemetry/Metrics/MetricStreamConfiguration.cs
```diff
@@ -35,8 +35,6 @@ public class MetricStreamConfiguration
 
         public string Description { get; set; }
 
-        public string Unit { get; set; }
```
cijothomas, line 38: Unit never made it to spec...so removing.

src/OpenTelemetry/Metrics/Metric.cs
```diff
@@ -25,15 +25,24 @@ public class Metric
...
+            string metricDescription)
+            : this(instrument, temporality, DefaultHistogramBounds, metricName, metricDescription)
         {
         }
 
-        internal Metric(Instrument instrument, AggregationTemporality temporality, double[] histogramBounds, string metricName = "")
+        internal Metric(
+            Instrument instrument,
+            AggregationTemporality temporality,
+            double[] histogramBounds,
```
utpilla, line 40: Move `histogramBounds` to the end make its default value as `DefaultHistogramBounds`? That way we wouldn't need the other ctor?

cijothomas: i tried that, but then realized the default must be a const, and arrays cannot be used for that purpose!

utpilla: Have the default value of `histogramBounds` as `null` and then assign that as `histogramBounds` inside the method?

reyang (approved):

PR #2441, opened 2021-10-01
Use instrument state as single metric instead of list with size one

cijothomas:
With views, one instrument could result in one or more Metric streams. The state object associated with an instrument is currently a List<Metric>. In case where no View is registered at all, or if the View results in a single stream, this List will be of size one.

This PR is trying to optimize the hot path by storing a single Metric instance as the state, when we know for sure that the instrument will result in a single stream at most. In those cases, the hot path is a simpler metric.Update(), as opposed to doing a foreach on the List<Metric> (with one element).

Note: Even if *any* view is registered (which may not even be affecting a given instrument), we'll fallback to the List<Metric>. This is because of the fact that, we cannot know ahead of time whether an instrument will result in single stream or not. We'll know it once the instrument is published... But by that time, we have already registered Measurement callbacks.. 
Even in this case, we do special casing for List with size one, to avoid foreach.

Trying the stress test..Numbers after 200 M loops.
before:
Loops/Second: 4,841,491, CPU Cycles/Loop: 3,304
after:
Loops/Second: 5,561,622, CPU Cycles/Loop: 2,972


src/OpenTelemetry/Metrics/MeterProviderSdk.cs
```diff
@@ -287,6 +288,36 @@ internal void MeasurementRecordedLong(Instrument instrument, long value, ReadOnl
             }
```
cijothomas, line 288: if (metrics.Count ==1)
{
metrics[0].Update()
}
else
{
do foreach..
}

src/OpenTelemetry/Metrics/MeterProviderSdk.cs
```diff
@@ -178,6 +178,16 @@ internal sealed class MeterProviderSdk : MeterProvider
                         }
                     }
                 };
+
+                // Everything double
+                this.listener.SetMeasurementEventCallback<double>((instrument, value, tags, state) => this.MeasurementRecordedDouble(instrument, value, tags, state));
```
reyang, line 183:
```suggestion
                this.listener.SetMeasurementEventCallback<double>(this.MeasurementRecordedDouble);
```

reyang (approved):

PR #2442, opened 2021-10-01
View Part 3 - Support for configuring tagkeys to be used for aggregation

cijothomas:
Mostly same as #2429 but also addresses the perf concern from https://github.com/open-telemetry/opentelemetry-dotnet/pull/2429/files#r719040932 

PR Description:

Currently, all the tags provided with measurement is used for aggregation. This PR allows user to configure View, to configure 
TagKeys, which will be used for aggregation. Typically used for cost control, by dropping unwanted tags.

For example, consider the following snippet of measurements being recorded (assume console exporter is configured):

```
counter.Add(1, new("name", "apple"), new("color", "red"));
counter.Add(2, new("name", "lemon"), new("color", "yellow"));
counter.Add(1, new("name", "lemon"), new("color", "yellow"));
counter.Add(2, new("name", "apple"), new("color", "green"));
counter.Add(5, new("name", "apple"), new("color", "red"));
counter.Add(4, new("name", "lemon"), new("color", "yellow"));
```

The output is as follows:
Output:
```
Export MyFruitCounter, Meter: MyCompany.MyProduct.MyLibrary/1.0
(2021-09-29T23:31:23.7786143Z, 2021-09-29T23:31:23.7972342Z] color:red name:apple LongSum
Value: 6
(2021-09-29T23:31:23.7786143Z, 2021-09-29T23:31:23.7972342Z] color:yellow name:lemon LongSum
Value: 7
(2021-09-29T23:31:23.7786143Z, 2021-09-29T23:31:23.7972342Z] color:green name:apple LongSum
Value: 2
```

Below snippet uses view with both tags. (i.e same as default behavior)

```
.AddView(
                "MyFruitCounter",
                new MetricStreamConfiguration()
                {
                    TagKeys = new string[] { "color", "name" },
                })
```

Output:
```
Export MyFruitCounter, Meter: MyCompany.MyProduct.MyLibrary/1.0
(2021-09-29T23:31:23.7786143Z, 2021-09-29T23:31:23.7972342Z] color:red name:apple LongSum
Value: 6
(2021-09-29T23:31:23.7786143Z, 2021-09-29T23:31:23.7972342Z] color:yellow name:lemon LongSum
Value: 7
(2021-09-29T23:31:23.7786143Z, 2021-09-29T23:31:23.7972342Z] color:green name:apple LongSum
Value: 2
```

Below snippet uses view to only pick single tag "color"
```
.AddView(
                "MyFruitCounter",
                new MetricStreamConfiguration()
                {
                    TagKeys = new string[] { "color" },
                })
```

Output:
```
Export MyFruitCounter, Meter: MyCompany.MyProduct.MyLibrary/1.0
(2021-09-29T23:34:42.3563688Z, 2021-09-29T23:34:42.3722304Z] color:red LongSum
Value: 6
(2021-09-29T23:34:42.3563688Z, 2021-09-29T23:34:42.3722304Z] color:yellow LongSum
Value: 7
(2021-09-29T23:34:42.3563688Z, 2021-09-29T23:34:42.3722304Z] color:green LongSum
Value: 2
```

Below snippet uses view to only pick single tag "name"
```
.AddView(
                "MyFruitCounter",
                new MetricStreamConfiguration()
                {
                    TagKeys = new string[] { "name" },
                })
```

Output:
```
Export MyFruitCounter, Meter: MyCompany.MyProduct.MyLibrary/1.0
(2021-09-29T23:35:38.0528475Z, 2021-09-29T23:35:38.0775305Z] name:apple LongSum
Value: 8
(2021-09-29T23:35:38.0528475Z, 2021-09-29T23:35:38.0775305Z] name:lemon LongSum
Value: 7
```

Below snippet uses view to drop all tags.
```
.AddView(
                "MyFruitCounter",
                new MetricStreamConfiguration()
                {
                    TagKeys = new string[] {  },
                })
```

Output:
```
Export MyFruitCounter, Meter: MyCompany.MyProduct.MyLibrary/1.0
(2021-09-29T23:36:54.2341730Z, 2021-09-29T23:36:54.2440246Z] LongSum
Value: 15
```

Below snippet uses view to specify some non-existent tag. (similar as no tags!)
```
.AddView(
                "MyFruitCounter",
                new MetricStreamConfiguration()
                {
                    TagKeys = new string[] {  "foo" },
                })
```

Output:
```
Export MyFruitCounter, Meter: MyCompany.MyProduct.MyLibrary/1.0
(2021-09-29T23:36:54.2341730Z, 2021-09-29T23:36:54.2440246Z] LongSum
Value: 15
```

### DIFF
--- a/src/OpenTelemetry/Metrics/AggregatorStore.cs
+++ b/src/OpenTelemetry/Metrics/AggregatorStore.cs
@@ -27,6 +27,8 @@ namespace OpenTelemetry.Metrics
         internal const int MaxMetricPoints = 2000;
         private static readonly ObjectArrayEqualityComparer ObjectArrayComparer = new ObjectArrayEqualityComparer();
         private readonly object lockZeroTags = new object();
+        private readonly HashSet<string> tagKeysInteresting;
+        private readonly int tagsKeysInterestingCount;
 
         // Two-Level lookup. TagKeys x [ TagValues x Metrics ]
         private readonly ConcurrentDictionary<string[], ConcurrentDictionary<object[], int>> keyValue2MetricAggs =
@@ -63,33 +65,20 @@ namespace OpenTelemetry.Metrics
             {
                 this.updateLongCallback = this.UpdateLongCustomTags;
                 this.updateDoubleCallback = this.UpdateDoubleCustomTags;
+                var hs = new HashSet<string>(StringComparer.Ordinal);
+                foreach (var key in tagKeysInteresting)
+                {
+                    hs.Add(key);
+                }
+
+                this.tagKeysInteresting = hs;
+                this.tagsKeysInterestingCount = hs.Count;
             }
         }
 
         private delegate void UpdateLongDelegate(long value, ReadOnlySpan<KeyValuePair<string, object>> tags);
 
         private delegate void UpdateDoubleDelegate(double value, ReadOnlySpan<KeyValuePair<string, object>> tags);
-
-        internal int FindMetricAggregators(ReadOnlySpan<KeyValuePair<string, object>> tags)
-        {
-            int tagLength = tags.Length;
-            if (tagLength == 0)
-            {
-                this.InitializeZeroTagPointIfNotInitialized();
-                return 0;
-            }
-
-            var storage = ThreadStaticStorage.GetStorage();
-
-            storage.SplitToKeysAndValues(tags, tagLength, out var tagKey, out var tagValue);
-
-            if (tagLength > 1)
-            {
-                Array.Sort<string, object>(tagKey, tagValue);
-            }
-
-            return this.LookupAggregatorStore(tagKey, tagValue, tagLength);
-        }
 
         internal void Update(long value, ReadOnlySpan<KeyValuePair<string, object>> tags)
         {
@@ -249,7 +238,7 @@ namespace OpenTelemetry.Metrics
         {
             try
             {
-                var index = this.FindMetricAggregators(tags);
+                var index = this.FindMetricAggregatorsCustomTag(tags);
                 if (index < 0)
                 {
                     // TODO: Measurement dropped due to MemoryPoint cap hit.
@@ -287,7 +276,7 @@ namespace OpenTelemetry.Metrics
         {
             try
             {
-                var index = this.FindMetricAggregators(tags);
+                var index = this.FindMetricAggregatorsCustomTag(tags);
                 if (index < 0)
                 {
                     // TODO: Measurement dropped due to MemoryPoint cap hit.
@@ -300,6 +289,48 @@ namespace OpenTelemetry.Metrics
             {
                 // TODO: Measurement dropped due to internal exception.
             }
+        }
+
+        private int FindMetricAggregatorsDefault(ReadOnlySpan<KeyValuePair<string, object>> tags)
+        {
+            int tagLength = tags.Length;
+            if (tagLength == 0)
+            {
+                this.InitializeZeroTagPointIfNotInitialized();
+                return 0;
+            }
+
+            var storage = ThreadStaticStorage.GetStorage();
+
+            storage.SplitToKeysAndValues(tags, tagLength, out var tagKey, out var tagValue);
+
+            if (tagLength > 1)
+            {
+                Array.Sort<string, object>(tagKey, tagValue);
+            }
+
+            return this.LookupAggregatorStore(tagKey, tagValue, tagLength);
+        }
+
+        private int FindMetricAggregatorsCustomTag(ReadOnlySpan<KeyValuePair<string, object>> tags)
+        {
+            int tagLength = tags.Length;
+            if (tagLength == 0 || this.tagsKeysInterestingCount == 0)
+            {
+                this.InitializeZeroTagPointIfNotInitialized();
+                return 0;
+            }
+
+            var storage = ThreadStaticStorage.GetStorage();
+
+            storage.SplitToKeysAndValues(tags, tagLength, out var tagKey, out var tagValue);
+
+            if (tagLength > 1)
+            {
+                Array.Sort<string, object>(tagKey, tagValue);
+            }
+
+            return this.LookupAggregatorStore(tagKey, tagValue, tagLength);
         }
     }
 }
--- a/src/OpenTelemetry/Metrics/AggregatorStore.cs
+++ b/src/OpenTelemetry/Metrics/AggregatorStore.cs
@@ -326,14 +326,24 @@ namespace OpenTelemetry.Metrics
 
             var storage = ThreadStaticStorage.GetStorage();
 
-            storage.SplitToKeysAndValues(tags, tagLength, out var tagKey, out var tagValue);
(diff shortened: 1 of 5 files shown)


src/OpenTelemetry/Metrics/AggregatorStore.cs
```diff
@@ -233,5 +225,81 @@ private int LookupAggregatorStore(string[] tagKey, object[] tagValue, int length
...
+
+                this.metrics[index].Update(value);
+            }
+            catch (Exception)
+            {
+                // TODO: Measurement dropped due to internal exception.
+            }
+        }
+
+        private void UpdateLongCustomTags(long value, ReadOnlySpan<KeyValuePair<string, object>> tags)
```
cijothomas, line 248: this method can assume interestingTagKeys is not null and act accodingly.

src/OpenTelemetry/Metrics/AggregatorStore.cs
```diff
@@ -40,73 +42,52 @@ internal class AggregatorStore
...
         {
-            try
-            {
-                var index = this.FindMetricAggregators(tags);
-                if (index < 0)
-                {
-                    // TODO: Measurement dropped due to MemoryPoint cap hit.
-                    return;
-                }
+            this.updateLongCallback(value, tags);
```
reyang, line 85: How much time would this dispatch take?

cijothomas: I tried a benchmark below and based on that, the difference between doing if check in hot path vs using delegate is very small. Delegate is slightly faster.

Below is .NetCore3.1

|                                Method |     Mean |    Error |   StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|-------------------------------------- |---------:|---------:|---------:|------:|------:|------:|----------:|
|             Delegate_ConditionCheckOnce | 17.71 ns | 0.481 ns | 1.358 ns |     - |     - |     - |         - |
| Direct_ConditionCheckOnEveryIteration | 17.30 ns | 0.370 ns | 0.426 ns |     - |     - |     - |         - |

Below is .Net5
|                                Method |     Mean |     Error |    StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|-------------------------------------- |---------:|----------:|----------:|------:|------:|------:|----------:|
|             Delegate_ConditionCheckOnce| 4.225 ns | 0.1111 ns | 0.1279 ns |     - |     - |     - |         - |
| Direct_ConditionCheckOnEveryIteration | 6.410 ns | 0.1596 ns | 0.2018 ns |     - |     - |     - |         - |

and below is net462
|                                Method |     Mean |    Error |   StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|-------------------------------------- |---------:|---------:|---------:|------:|------:|------:|----------:|
|             Delegate_ConditionCheckOnce | 28.64 ns | 0.553 ns | 0.543 ns |     - |     - |     - |         - |
| Direct_ConditionCheckOnEveryIteration | 34.01 ns | 0.667 ns | 1.333 ns |     - |     - |     - |         - |


```
public class MetricGeneralBenchmarks
    {
        private UpdateDelegate updateDelegate;
        private Meter meter;
        private Counter<long> counter;
        private int randomNumber;

        private delegate void UpdateDelegate(long value);

        [GlobalSetup]
        public void Setup()
        {
            this.randomNumber = new Random().Next(1, 100);
            if (this.randomNumber % 5 == 0)
            {
                this.updateDelegate = this.UpdateA;
            }
            else
            {
                this.updateDelegate = this.UpdateB;
            }

            this.meter = new Meter("TestMeter");
            this.counter = this.meter.CreateCounter<long>("counter");
        }

        [Benchmark]
        public void Delegate_ConditionCheckOnce()
        {
            this.updateDelegate(100);
        }

        [Benchmark]
        public void Direct_ConditionCheckOnEveryIteration()
        {
            if (this.randomNumber % 5 == 0)
            {
                this.UpdateA(100);
            }
            else
            {
                this.UpdateB(100);
            }
        }

        private void UpdateA(long value)
        {
            this.counter.Add(value);
        }

        private void UpdateB(long value)
        {
            this.counter.Add(value);
        }
    }
```

reyang: > I tried a benchmark below and based on that, the difference between doing if check in hot path vs using delegate is very small. Delegate is slightly faster.

This confirms that doing delegate+dispatch vs. doing a check in each loop is equally bad.
Sounds like both approaches won't work.

Maybe we should step back to something like https://github.com/open-telemetry/opentelemetry-dotnet/blob/bd189aa654948e79203d66d4b6d33fb93557d7e9/test/Benchmarks/Trace/TraceBenchmarks.cs#L114, while adding additional things (starting from no listener, to a listener which simply drops data on the floor, to a listener which passes data to processors, to a listener which passes data to a chain of processors), measuring what do we have to pay for.

I wrote a simple stress test to verify how bad an extra dispatch would be:

```csharp
public partial class Program
{
    private static readonly Meter TestMeter = new Meter("TestMeter", "1.0.0");
    private static readonly Counter<long> TestCounter = TestMeter.CreateCounter<long>("TestCounter");

    public static void Main()
    {
        var listener = new MeterListener();

        // listener.SetMeasurementEventCallback<long>(MeasurementEventCallback<long>);

        listener.SetMeasurementEventCallback<long>((instrument, value, tags, state) =>
        {
            // Console.WriteLine($"MeasurementEventCallback(\"{instrument.Name}\", {value}, {state})");
        });

        listener.EnableMeasurementEvents(TestCounter, null);
        Stress();
    }

    protected static void Run()
    {
        TestCounter.Add(100, new("key1", "value1"), new("key2", "value2"), new("key3", "value3"));
    }

    protected static void MeasurementEventCallback<T>(Instrument instrument, T value, ReadOnlySpan<KeyValuePair<string, object>> tags, object state)
    {
        // Console.WriteLine($"MeasurementEventCallback(\"{instrument.Name}\", {value}, {state})");
    }
}
```

cijothomas: Stress test run results after ~200M loops (ran thrice)

main branch:
* Total Loops: 204,065,746
* Average Loops/Second: 4,635,856
* Average CPU Cycles/Loop: 3,399
----------------------------------------
* Total Loops: 202,408,337
* Average Loops/Second: 4,550,956
* Average CPU Cycles/Loop: 3,464
----------------------------------------
* Total Loops: 222,133,483
* Average Loops/Second: 4,283,082
* Average CPU Cycles/Loop: 3,639

this PR:
* Total Loops: 279,531,225
* Average Loops/Second: 4,566,384
* Average CPU Cycles/Loop: 3,449
----------------------------------------
* Total Loops: 205,510,447
* Average Loops/Second: 4,675,048
* Average CPU Cycles/Loop: 3,377
----------------------------------------
* Total Loops: 209,301,412
* Average Loops/Second: 4,778,571
* Average CPU Cycles/Loop: 3,278

reyang (approved):
LGTM.

PR #2444, opened 2021-10-02
Add basic wildcard support for listening to legacy activities.

### LINKED ISSUE #2183: Add wildcard support to AddLegacySource
Azure SDKs produce activities that have tags and links set and don't need extra enrichment. 

For client level spans we use the client and method name as activity name (i.e. `BlobClient.Download`) so it's impractical to use the existing `AddLegacySource(string name)` to collect them as customers would need to add every individual method they use.

Feature request:

Add wildcard support to AddLegacySource the same way it exists in AddSource so customers can write

```
.AddLegacySource("Azure.*");

### PR DESCRIPTION
**Pull request by BlakeWills**
Fixes #2183.

## Changes

Add support for wildcards for legacy (DiagnosticSource) events.

### DIFF
--- a/src/OpenTelemetry/Trace/TracerProviderSdk.cs
+++ b/src/OpenTelemetry/Trace/TracerProviderSdk.cs
@@ -34,8 +34,6 @@ namespace OpenTelemetry.Trace
         private readonly Sampler sampler;
         private readonly Action<Activity> getRequestedDataAction;
         private readonly bool supportLegacyActivity;
-        private readonly bool legacyActivityWildcardMode;
-        private readonly Regex legacyActivityWildcardModeRegex;
         private BaseProcessor<Activity> processor;
 
         internal TracerProviderSdk(
@@ -50,12 +48,14 @@ namespace OpenTelemetry.Trace
             this.sampler = sampler;
             this.supportLegacyActivity = legacyActivityOperationNames.Count > 0;
 
+            bool legacyActivityWildcardMode = false;
+            Regex legacyActivityWildcardModeRegex = null;
             foreach (var legacyName in legacyActivityOperationNames)
             {
                 if (legacyName.Key.Contains('*'))
                 {
-                    this.legacyActivityWildcardMode = true;
-                    this.legacyActivityWildcardModeRegex = GetWildcardRegex(legacyActivityOperationNames.Keys);
+                    legacyActivityWildcardMode = true;
+                    legacyActivityWildcardModeRegex = GetWildcardRegex(legacyActivityOperationNames.Keys);
                     break;
                 }
             }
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/OpenTelemetry/Trace/TracerProviderSdk.cs [hunk 1]
```diff
@@ -172,13 +184,13 @@ internal sealed class TracerProviderSdk : TracerProvider
                     if (name.Contains('*'))
                     {
                         wildcardMode = true;
+                        break;
```
reyang, line 187: 💯

src/OpenTelemetry/Trace/TracerProviderSdk.cs
```diff
@@ -71,7 +83,7 @@ internal sealed class TracerProviderSdk : TracerProvider
                     if (this.supportLegacyActivity && string.IsNullOrEmpty(activity.Source.Name))
```
reyang, line 83: Given `supportLegacyActivity` is something we can determine ahead of time, consider extracting the logic and create the corresponding delegate with hard-wired logic, so we don't have to check it in every Activity start/stop.

```csharp

if (this.supportLegacyActivity)
{
    listener.ActivityStarted = (activity) =>
    {
        if (string.IsNullOrEmpty(activity.Source.Name))
        {
            // legacy logic
            // similar here, we already know whether wildcard mode should be used or not, we can hard-wire the logic
            return;
        }
        // new logic
    }
}
else
{
    listener.ActivityStarted = (activity) =>
    {
        // new logic
    }
}

Comment (reyang) [reply src/OpenTelemetry/Trace/TracerProviderSdk.cs:83]: Similar approach has been used here to optimize the hot path:
* https://github.com/open-telemetry/opentelemetry-dotnet/blob/bd189aa654948e79203d66d4b6d33fb93557d7e9/src/OpenTelemetry/Trace/TracerProviderSdk.cs#L185
* https://github.com/open-telemetry/opentelemetry-dotnet/blob/bd189aa654948e79203d66d4b6d33fb93557d7e9/src/OpenTelemetry/Trace/TracerProviderSdk.cs#L206

File: src/OpenTelemetry/Trace/TracerProviderSdk.cs [hunk 3]
```diff
@@ -34,6 +34,8 @@ internal sealed class TracerProviderSdk : TracerProvider
         private readonly Sampler sampler;
         private readonly Action<Activity> getRequestedDataAction;
         private readonly bool supportLegacyActivity;
+        private readonly bool legacyActivityWildcardMode;
```
reyang, line 37: I guess making this local and capture it in lambda/delegate would be faster than accessing the value via this pointer.
I haven't tried it myself, a [benchmark](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/test/Benchmarks/Trace/TraceBenchmarks.cs) or [stress test](https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/test/OpenTelemetry.Tests.Stress) should give us the numbers.

reyang (approved):
LGTM, I put some perf suggestions.

reyang:
@BlakeWills thanks for the contribution! Would you clear the EasyCLA so we can run the CI?

src/OpenTelemetry/Trace/TracerProviderSdk.cs
BlakeWills: That makes sense and thanks for the tip. I've had a go at an initial optimization, let me know if this is what you expected.

Benchmark results (the top table is pre-optimisation, bottom is post), they're pretty marginal gains:

image

reyang (approved):
LGTM :shipit:

src/OpenTelemetry/Trace/TracerProviderSdk.cs
reyang: 👍  :shipit:

cijothomas (approved):

PR #2396, opened 2021-09-21
Add view (configuration only)

cijothomas:
Adds ability to add Views.
Documents/example.

In next PRs: Actual implementation.

PR #2449, opened 2021-10-05
View - support for dropping instrument

cijothomas:
More View capabilities:
This PR allows user to configure View to indicate that an instrument must be dropped.
example usages:

```
// Drop all instruments, whose name starts with server
.AddView("server*", new MetricStreamConfiguration() { Aggregation = Aggregation.Drop })

// Drop a particular instrument with the name "counternotinteresting"
.AddView("counterNotInteresting", new MetricStreamConfiguration() { Aggregation = Aggregation.Drop })
```

### DIFF
--- a/src/OpenTelemetry/Metrics/DropConfiguration.cs
+++ b/src/OpenTelemetry/Metrics/DropConfiguration.cs
@@ -1,4 +1,4 @@
-// <copyright file="HistogramConfiguration.cs" company="OpenTelemetry Authors">
+// <copyright file="DropConfiguration.cs" company="OpenTelemetry Authors">
 // Copyright The OpenTelemetry Authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
--- a/src/OpenTelemetry/Metrics/DropConfiguration.cs
+++ b/src/OpenTelemetry/Metrics/DropConfiguration.cs
@@ -1,0 +1,35 @@
+// <copyright file="HistogramConfiguration.cs" company="OpenTelemetry Authors">
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+
+using System;
+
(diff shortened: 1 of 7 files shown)


test/OpenTelemetry.Tests/Metrics/MetricViewTests.cs
```diff
@@ -292,6 +292,80 @@ public void ViewToSelectTagKeys()
...
+        }
+
+        [Fact]
+        public void ViewToDropAndRetainInstrument()
+        {
+            using var meter = new Meter("ViewToDropAndRetainInstrumentTest");
+            var exportedItems = new List<Metric>();
+            using var meterProvider = Sdk.CreateMeterProviderBuilder()
+                .AddSource(meter.Name)
+                .AddView("server.requests", new MetricStreamConfiguration() { Aggregation = Aggregation.Drop })
```
reyang, line 353: Instead of having to create a new object, could we use something like `MetricStreamConfiguration.Drop`?

cijothomas: Not very clean, but this commit attempts this : https://github.com/open-telemetry/opentelemetry-dotnet/pull/2449/commits/921449597a2aa03a51a9bf238057211a96ca125b
cc: @alanwest you raised this point in one of the early View PRs as well I think.

reyang (approved):

src/OpenTelemetry/Metrics/DropConfiguration.cs
```diff
@@ -0,0 +1,35 @@
...
+// limitations under the License.
+// </copyright>
+
+using System;
+
+namespace OpenTelemetry.Metrics
+{
+    internal class DropConfiguration : MetricStreamConfiguration
+    {
+        private Aggregation aggregation = Aggregation.Drop;
```
alanwest, line 23: Unrelated to this PR, but I can't recall the reason why we have `Aggregation.None` and `Aggregation.Drop`.

cijothomas: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#none-aggregation Spec has "None". We thought its easier if we add `Drop` (point to None itself), as its more clear.. Now that I read it more, I think we can simply remove None, and have Drop only.

test/OpenTelemetry.Tests/Metrics/MetricViewTests.cs
alanwest: Yea, I made the comment here https://github.com/open-telemetry/opentelemetry-dotnet/pull/2396#discussion_r717121867.

The somewhat wonky thing about this pattern is that a user can mutate the static instance. I see you protected the Aggregation property from being set by throwing an exception. If we create more static helpers like this we could consider overriding all of the properties and throw exceptions as necessary.

alanwest (approved):
LGTM

PR #2451, opened 2021-10-05
Minor doc fix

cijothomas:
Follow up on https://github.com/open-telemetry/opentelemetry-dotnet/pull/2449/files#r722728583

PR #2453, opened 2021-10-06
Improve metrics tutorial

reyang:
Related to #2451
* Added extension method `AddMyExporter`.
* Improved the reader output.
* Simplified `Main` by putting Instruments in class initialization code (which helps people to understand that instruments are something that should be reused).

### DIFF
--- a/docs/metrics/extending-the-sdk/MyReader.cs
+++ b/docs/metrics/extending-the-sdk/MyReader.cs
@@ -15,6 +15,7 @@
 // </copyright>
 
 using System;
+using System.Text;
 using OpenTelemetry;
 using OpenTelemetry.Metrics;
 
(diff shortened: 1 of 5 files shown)


docs/metrics/extending-the-sdk/MyReader.cs
```diff
@@ -29,7 +30,18 @@ public MyReader(string name = "MyReader")
...
+        {
+            if (sb.Length > 0)
+            {
+                sb.Append(", ");
+            }
+
+            sb.Append($"{record}");
+        }
+
+        Console.WriteLine($"{this.name}.ProcessMetrics(metrics=[{sb}], timeoutMilliseconds={timeoutMilliseconds})");
```
pellared, line 44: nit: If we care about readability we could simply use `string.Join`. If we care about performance we could use some `bool` variable instead of checking `sb.Length > 0` each time.

pellared (approved):

docs/metrics/extending-the-sdk/MyReader.cs
reyang: Good point!

I think here we care about readability, not perf.
I guess we might need to go through each metric and visit all the points, so `string.Join` probably won't work well here? (e.g. this is what we have for logs https://github.com/open-telemetry/opentelemetry-dotnet/blob/f05b05861397adfb38f5425dd3cc8c2f42a2ac21/docs/logs/extending-the-sdk/MyExporter.cs#L38)

Given we don't have such logic right now, I'll simply the code in this PR.

reyang: Just tried and I couldn't figure out a simpler way 🤷.

`string.Join(", ", batch);` would not iterate through `batch`, and if we try to cast it / convert it, the result is actually less readable (e.g. with LINQ).

pellared: I was thinking about something like:

```csharp
var metricsDump = string.Join(", ", metrics.Select(x => x.ToString()).ToArray());
Console.WriteLine($"{this.name}.ProcessMetrics(metrics=[{metricsDump}], timeoutMilliseconds={timeoutMilliseconds})");
```

what is more readable? I have no idea :smile:

feel free to resolve the comment and leave it as it is

PR #2454, opened 2021-10-06
Rename AddSource to AddMeter to better reflect intent

cijothomas:
`AddSource` was copied from TracerProviderBuilder where it made sense as it was adding `ActivitySource`. For MeterProviderBuilder, more correct name would be `AddMeter` (as there is no MeterSource , but only `Meter`)


alanwest (approved):
Yes! Very helpful name change.

PR #2457, opened 2021-10-06
Handle exception from observable instruments

cijothomas:
If any instrument callback threw, we'd lose the entire Metrics. This PR addresses the issue.
The entire EventSource logging require a revisit later.


test/OpenTelemetry.Tests/Metrics/MetricAPITest.cs
```diff
@@ -38,6 +40,77 @@ public MetricApiTest(ITestOutputHelper output)
...
+                metricPoints.Add(mp);
+            }
+
+            Assert.Single(metricPoints);
+            var metricPoint = metricPoints[0];
+            Assert.Equal(100, metricPoint.LongValue);
+            Assert.NotNull(metricPoint.Keys);
+            Assert.NotNull(metricPoint.Values);
+
+            metric = exportedItems[1];
```
cijothomas, line 103: even if no MetricPoints are there, we still emit the Metric. This is something to be fixed along with https://github.com/open-telemetry/opentelemetry-dotnet/issues/2360

PR #2466, opened 2021-10-07
Reuse unused MetricPoints

alanwest:
Related to #2360.

Introduces `MetricPointStatus` to facilitate the ability to free up unused `MetricPoint`s. Currently, only works for `AggregationTemporatliy = Delta`

background: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/supplementary-guidelines.md


src/OpenTelemetry/Metrics/AggregatorStore.cs
```diff
@@ -167,29 +209,25 @@ private int LookupAggregatorStore(string[] tagKeys, object[] tagValues, int leng
...
-                            // unused points (typically with delta)
-                            // we can re-claim them here.
-                            return -1;
+                            if (this.temporality == AggregationTemporality.Cumulative)
+                            {
+                                // sorry! out of data points.
+                                return -1;
+                            }
+
+                            this.FreeUnusedMetrics();
```
cijothomas, line 225: since this call is done *inside* a lock and in hot path, could FreeUnusedMetrics() return as soon as it frees a single point, enough to unblock this measurement?

src/OpenTelemetry/Metrics/MetricPointStatus.cs
```diff
@@ -0,0 +1,50 @@
...
+// </copyright>
+
+namespace OpenTelemetry.Metrics
+{
+    internal enum MetricPointStatus
+    {
+        /// <summary>
+        /// Uninitialized <see cref="MetricPoint"/>.
+        /// </summary>
+        Unset,
```
cijothomas, line 24: Suggestion for alternate names:
Unused
Free

src/OpenTelemetry/Metrics/MetricPointStatus.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        /// <summary>
+        /// UpdatePending <see cref="MetricPoint"/>.
+        /// </summary>
+        UpdatePending,
```
cijothomas, line 29: Please add a comment to indicate that the point is just created, and hasn't received the very 1st update.

src/OpenTelemetry/Metrics/MetricPointStatus.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// Collect will move it to <see cref="NoPendingUpdate"/>.
+        /// </summary>
+        CollectPending,
+
+        /// <summary>
+        /// This status is applied to <see cref="MetricPoint"/>s with status <see cref="CollectPending"/> after a Collect.
+        /// It will be moved to <see cref="CandidateForRemoval"/> during the next Collect cycle.
+        /// If an update occurs, status will be moved to <see cref="CollectPending"/>.
+        /// </summary>
+        NoPendingUpdate,
```
cijothomas, line 42:
```suggestion
        NoPendingCollect,
```

src/OpenTelemetry/Metrics/AggregatorStore.cs
```diff
@@ -94,17 +106,48 @@ internal void Update(double value, ReadOnlySpan<KeyValuePair<string, object>> ta
 
         internal void SnapShot()
         {
-            var indexSnapShot = Math.Min(this.metricPointIndex, MaxMetricPoints - 1);
-
-            for (int i = 0; i <= indexSnapShot; i++)
+            this.batchSize = 0;
+            for (int i = 0; i < MaxMetricPoints; i++)
```
cijothomas, line 110: if we start filling from the beginning, then we might not have to navigate the entire MaxMetricPoints of this array.

src/OpenTelemetry/Metrics/AggregatorStore.cs
```diff
@@ -37,22 +38,33 @@ internal sealed class AggregatorStore
...
         private readonly double[] histogramBounds;
         private readonly UpdateLongDelegate updateLongCallback;
         private readonly UpdateDoubleDelegate updateDoubleCallback;
-        private int metricPointIndex = 0;
+        private long batchSize = 0;
         private bool zeroTagMetricPointInitialized;
         private DateTimeOffset startTimeExclusive;
         private DateTimeOffset endTimeInclusive;
 
+        private ConcurrentStack<int> metricPointFreeList = new ConcurrentStack<int>(Enumerable.Range(1, MaxMetricPoints - 1));
```
cijothomas, line 52: we could populate it in reverse order. so the index 1 will be the first to be used, then index 2.

src/OpenTelemetry/Metrics/AggregatorStore.cs
```diff
@@ -278,21 +347,66 @@ private void UpdateDoubleCustomTags(double value, ReadOnlySpan<KeyValuePair<stri
...
+
+            metricPointLock.EnterReadLock();
+
+            if (metricPoint.MetricPointStatus == MetricPointStatus.Unset)
+            {
+                metricPointLock.ExitReadLock();
+                return false;
+            }
+
+            metricPoint.Update(value);
```
cijothomas, line 381: haven't fully analyzed the implications, but I think we should move the status management to be entirely within the MetricPoint class, and avoid any locks in `AggregatorStore`.

1. MetricPoint already has `lock` for Histogrampoints now. So it can safely do Update() and SnapShot() adjusting the MetricPointStatus accordingly.
2. For SUM, there is no lock. But the following could achieve what we need, without the need of AggregatorStore entering ReadLocks

Update:
Do the update as before, and **after** that, always set status to `CollectPending`.
SnapShot:
Read the value as before. 

If the value has not changed since it read, then set status to `NoPendingUpdate`. (i.e no other sync instrument.update has occurred). If an update ever comes before next SnapShot, the next snapshot will include it. If no update, then cleanup time.

Else (i.e some other sync instrument update happened since snapshot did the "snapshotting(interlocked.read)"), do not modify the status. This means the update which just came, will be collected in the next cycle..

cijothomas: to avoid the issue where an update gets lost when a CandidateForRemoval is cleanedup, we might need to do a lock in Update() if the previous status is CandidateForRemoval.

src/OpenTelemetry/Metrics/AggregatorStore.cs
```diff
@@ -204,7 +242,7 @@ private int LookupAggregatorStore(string[] tagKeys, object[] tagValues, int leng
 
                         ref var metricPoint = ref this.metricPoints[aggregatorIndex];
                         var dt = DateTimeOffset.UtcNow;
-                        metricPoint = new MetricPoint(this.aggType, dt, seqKey, seqVal, this.histogramBounds);
+                        metricPoint = new MetricPoint(this.aggType, MetricPointStatus.UpdatePending, dt, seqKey, seqVal, this.histogramBounds);
```
cijothomas, line 245: when we re-use a MetricPoint, the MetricPoint will internally create fresh arrays for the BucketCounts. Given the fact that the buckets don't change, we should reuse the same array (after resetting all values to zero).
The arrays can be constructed one time (at AggregatorStore ctor), and reused afterwards. Else, it'd result in a lot of arrays being created/thrown away for histograms.

(will address this separately, once this is done)

src/OpenTelemetry/Metrics/AggregatorStore.cs
alanwest: Just an update, I haven't pushed it up yet, but I'm pretty close to implementing this feedback.

I've moved the status management down into the MetricPoint class itself, and I agree it's cleaner. I'm just carefully reviewing the status management before pushing it up.

I'll address your other feedback on this PR once I'm comfortable with things here. I think I've successfully simplified a few things like removing the `metricPointFreeList` as it may no longer be necessary. I've also reduced the set of `MetricPointStatus` enumeration values.

cijothomas: yea. the other part about re-using histogram arrary is definitely separate. (As i was thinking about it, I think we might have more optimizations possible there for string[] object[] as well, when View is configured) Will chat about it after the cleanup part is stable.

src/OpenTelemetry/Metrics/MetricPointStatus.cs
alanwest: I think this status may be unnecessary, I'm attempting to remove it. If it ends up staying I'll update the comment.

src/OpenTelemetry/Metrics/AggregatorStore.cs
alanwest: I've removed the free list.

src/OpenTelemetry/Metrics/AggregatorStore.cs
alanwest: Agree, pretty sure I can optimize this...

src/OpenTelemetry/Metrics/AggregatorStore.cs
```diff
@@ -217,18 +231,55 @@ private int LookupAggregatorStore(string[] tagKeys, object[] tagValues, int leng
             return aggregatorIndex;
         }
 
+        private bool TryGetUnusedMetricPoint(out int index)
+        {
+            if (this.metricPointIndex < MaxMetricPoints)
+            {
+                index = Interlocked.Increment(ref this.metricPointIndex);
```
cijothomas, line 238: Could you add a comment like to explain why we still need Interlock here.
"Interlocked.Increment required as the lock is only taken on the value2metrics, so another thread with different tag keys could be running in parallel."

cijothomas: @utpilla FYI - Now i recollect why we still needed Interlocked.

src/OpenTelemetry/Metrics/AggregatorStore.cs
```diff
@@ -217,18 +231,55 @@ private int LookupAggregatorStore(string[] tagKeys, object[] tagValues, int leng
...
+                if (index < MaxMetricPoints)
+                {
+                    return true;
+                }
```
cijothomas, line 242: we need an else block here, and attempt TryFree

cijothomas: nvm.

src/OpenTelemetry/Metrics/MetricPoint.cs
```diff
@@ -288,6 +305,60 @@ internal void TakeSnapShot(bool outputDelta)
                         break;
                     }
             }
+
+            this.MetricPointStatus = MetricPointStatus.NoPendingCollect;
```
cijothomas, line 309: This might be not thread-safe. If the snapshot(value read) was done, and before SnapShot updates the status to NoPendingCollect, another Update might have update the value,status, then this will overwrite the status, and we'll miss that update, as it wont be covered in next snapshot.

cijothomas: For histogram - this is straightforward - move the `this.MetricPointStatus = MetricPointStatus.NoPendingCollect;` inside the lock we already take anyway for Update/Snapshot.

cijothomas: For Gauge, I think the following might do the trick:
Update - same as now
SnapShot  - 
1. Read the value.
2. Update Status to NoPendingCollect
3. Check if value changed since step 1. (i.e some other update(s) occurred since step 1)
4. If yes, revert status to PendingCollect. (this will ensure that the updates we missed, will be caught in next cycle)
5. Else, move on.


I think the same can be used for SUM as well.

src/OpenTelemetry/Metrics/MetricPoint.cs
```diff
@@ -288,6 +305,60 @@ internal void TakeSnapShot(bool outputDelta)
...
+                    }
+                }
+            }
+
+            return false;
+        }
+
+        private bool TrySetCollectPending()
+        {
+            if (this.MetricPointStatus == MetricPointStatus.CandidateForRemoval || this.MetricPointStatus == MetricPointStatus.Unused)
```
cijothomas, line 345: we need to take lock for NoPendingCollect as well. Else MarkStale might just override this to CandidateForRemoval

alanwest: Yea, you're right. I was hoping there'd be a way to keep the update path lock free.

src/OpenTelemetry/Metrics/MetricPointStatus.cs
```diff
@@ -0,0 +1,52 @@
...
+        /// Collect will move it to <see cref="NoPendingCollect"/>.
+        /// </summary>
+        CollectPending,
+
+        /// <summary>
+        /// This status is applied to <see cref="MetricPoint"/>s with status <see cref="CollectPending"/> after a Collect.
+        /// It will be moved to <see cref="CandidateForRemoval"/> during the next Collect cycle.
+        /// If an update occurs, status will be moved to <see cref="CollectPending"/>.
+        /// </summary>
+        NoPendingCollect,
```
utpilla, line 44: nit: rename this to `NoCollectPending` or rename `CollectPending` to `PendingCollect`

src/OpenTelemetry/Metrics/AggregatorStore.cs
```diff
@@ -217,18 +231,55 @@ private int LookupAggregatorStore(string[] tagKeys, object[] tagValues, int leng
...
+            }
+
+            if (this.temporality == AggregationTemporality.Delta)
+            {
+                for (int i = 1; i < MaxMetricPoints; i++)
+                {
+                    ref var metricPoint = ref this.metricPoints[i];
+                    if (metricPoint.MetricPointStatus == MetricPointStatus.CandidateForRemoval)
```
cijothomas, line 250: I think the approach of freeing up space on Update path will lead to incorrect behavior.
say K1 is the key to the dictionary, pointing to MP P1 which is candidateforremoval.

Following two threads:
Thread 1: Update(K1,100) 
Looks up dictionary and finds K1->P1 mapping, so decides to Update P1.
Thread: 2: Update(K2, 135) - K2 is not found. out of MPs. So try to free up, and it can decide to free up P1. It frees up P1, and removes (K1->P1) and put K2->P1. Updates P1 with 135, status to PendingCollect.
Thread1 - is unaware that P1 is taken up by someone else, and falsely stores its update to P1.

alanwest: Ugh, yep agreed this is a problem, I've taken the route you suggested of validating in the update path that the MP keys/values are the ones expected. It might not be the prettiest solutions 🥴. I believe the fix will prevent Thread 1 from updating the wrong MP, but there may still be the possibility that the update on Thread 1 gets lost, but I think this is not nearly as bad.

src/OpenTelemetry/Metrics/AggregatorStore.cs
```diff
@@ -278,33 +342,72 @@ private void UpdateDoubleCustomTags(double value, ReadOnlySpan<KeyValuePair<stri
...
         }
 
-        private int FindMetricAggregatorsDefault(ReadOnlySpan<KeyValuePair<string, object>> tags)
+        private bool TryUpdateMetricPoint(int index, long value, string[] tagKeys, object[] tagValues)
+        {
+            ref var metricPoint = ref this.metricPoints[index];
+
+            // If the MetricPointStatus = CandidateForRemoval, it is possible that another thread freed and repurposed the MetricPoint.
+            // We must check that the point being updated is the expected one.
+            if (index == 0 || (StringArrayComparer.Equals(metricPoint.Keys, tagKeys) && ObjectArrayComparer.Equals(metricPoint.Values, tagValues)))
```
cijothomas, line 371: after thinking a bit more on this, this one also doesn't guarantee safety, as this check + actual update - are not atomic. For example, if this thread got context switched and never got the chance to run again until a long time, some other thread could have freed up this spacem causing this thread to write into some else's MP :(

alanwest:
Closing for now. See https://github.com/open-telemetry/opentelemetry-dotnet/issues/2360#issuecomment-971825501

PR #2470, opened 2021-10-08
Add validation to instrument and view name

### LINKED ISSUE #2424: Instrument name validation
Instrument name has to be validated as per the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument). The .NET API itself does not validate this, so the SDK has to validate the instrument name (upon instrument creation), and ignore the instrument (after logging that we are ignoring an instrument due to ... reason).

For Views, which provide new name for the Metric, the validation can be done in the `AddView` itself and can throw/fail fast on AddView.

joaopgrassi:
## Changes

This PR adds validation to the instrument names as described in the [Instrument spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument).

Summary of changes:

- Instruments with invalid names are ignored and an event is logged
- Views added with a name also are validated in the same way with some particularities: 
   - Overloads of `AddView` that accepts the `name` or `MetricStreamConfiguration` fail-fast if the name provided is invalid
   - Overloads where the `Func` is specified cannot be validated right away. They are only validated on instrument publish, and if invalid, are ignored with a log.
- Added new events in `OpenTelemetrySdkEventSource` to log the above cases
- Tests

Closes #2424

* [x] Design discussion issue #2424


joaopgrassi:
Rebase on main and fixed new conflicts. Ready to merge 😊

src/OpenTelemetry/Metrics/MeterProviderBuilderExtensions.cs
```diff
@@ -71,6 +76,16 @@ public static MeterProviderBuilder AddView(this MeterProviderBuilder meterProvid
         /// <remarks>See View specification here : https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#view.</remarks>
         public static MeterProviderBuilder AddView(this MeterProviderBuilder meterProviderBuilder, string instrumentName, MetricStreamConfiguration metricStreamConfiguration)
         {
+            if (metricStreamConfiguration == null)
+            {
+                throw new ArgumentNullException($"Metric stream configuration cannot be null.", nameof(metricStreamConfiguration.Name));
```
randomnoise, line 81: Sorry, I am new here but this line caught my eye. Shouldn't this be `throw new ArgumentNullException($"Metric stream configuration cannot be null.", nameof(metricStreamConfiguration));` ?

joaopgrassi: Yes, you're right @randomnoise :). Fixed and thanks!

Issue #2483, opened 2021-10-13
OTLP HttpExporter does not allow overriding path


cijothomas:
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#otlphttp-request We seem to be appending /v1/traces by default, including if user provided custom endpoints.
https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpExporterOptionsExtensions.cs#L106

Should we be doing this only if the default endpoint is used, so that users provided endpoint is not appended with the /v1/traces path.?

cijothomas:
@rypdal @alanwest -Could you help with the qn?

rypdal:
According to the latest (specification)[https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#endpoint-urls-for-otlphttp] an endpoint created from the OTEL_EXPORTER_OTLP_ENDPOINT env.var. is appended either with v1/traces or v1/metrics in case of http/protobuf. Same rule is applied to Endpoint property set explicitly from code. If user wants to have custom traces, metrics endpoint he has to use signal specific env. vars. OTEL_EXPORTER_OTLP_{Signal}_ENDPOINT. Another thing is that we are not supporting any specific env. vars. yet. For this we need dedicated issue, because signal specific env. vars. has to be supported for the case of gRPC as well. Moreover, we need to support signal specific env. vars, for many env. vars. like Headers, Timeout, Protocol, etc (full table of env. vars.)[https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#opentelemetry-protocol-exporter]. I think we need issue like "Support signal specific environment variables".

joaopgrassi:
Hey all,
Should we close this issue? I think the original problem is solved, even though we still need to address the signal-specific env vars.

PR #2492, opened 2021-10-18
Otlp http exporter: allow endpoint override

rypdal:
Fixes #.

## Changes

Implemented the following behaviour:

[default path](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#endpoint-urls-for-otlphttp) is only appended when the Options.Endpoint property wasn't set by the user, the protocol is HttpProtobuf and the OTEL_EXPORTER_OTLP_ENDPOINT environment variable is present. If the user provides a custom value for Options.Endpoint that value is taken as is.


rypdal:
> @rypdal I have a few questions regarding spec compliance here . The [spec ](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#endpoint-urls-for-otlphttp) mentions two scenarios for constructing URLs:
> 
> > 1. For the per-signal variables (OTEL_EXPORTER_OTLP__ENDPOINT), the URL MUST be used as-is without any modification. The only exception is that if an URL contains no path part, the root path / MUST be used (see Example 2).
> 
> It looks like currently we are not reading from `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` or `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT`. Is this planned for a future PR? Also, these variables should allow one to override the URL for sending a specific signal as shown in [Example 3](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#example-3) of the spec. With the changes in this PR, the `OtlpExporterOptions.TracesEndpoint` and `OtlpExporterOptions.MetricsEndpoint` cannot have any value different from `OtlpExporterOptions.Endpoint`. Is this not expected to be covered by this PR?
> 
> > 1. If signals are sent that have no per-signal configuration from the previous point, OTEL_EXPORTER_OTLP_ENDPOINT is used as a base URL and the signals are sent to these paths relative to that:
> > 
> > 
> > * Traces: v1/traces
> > * Metrics: v1/metrics.
> > 
> > Non-normatively, this could be implemented by ensuring that the base URL ends with a slash and then appending the relative URLs as strings.
> 
> When someone sets the `OtlpExporterOptions.Endpoint` property, do we not want to append "v1/traces" or "v1/metrics" to `OtlpExporterOptions.TracesEndpoint` and `OtlpExporterOptions.MetricsEndpoint` if the `Protocol` is `HttpProtobuf`?

@utpilla , please, see my answers/last changes below:

1. I decided to remove traces and metrics endpoint properties from exporter options as it should be considered along with per-signal env. vars. Moreover, other per-signal env. vars. like OTEL_EXPORTER_OTLP_{Signal}_HEADERS, OTEL_EXPORTER_OTLP_{Signal}_PROTOCOL, etc should be considered in dedicated PR as well. The scope of this PR is only allow endpoint override. AFAIK PR for support of signal specific env. vars. not exists yet. 
2. Current implementation is changed in the way that traces and metrics relative paths append logic is moved to corresponding AddOtlpExporter methods where call scope (traces or metrics) is known.
3. Specification is clear about appending of "v1/traces" or "v1/metrics" only for case when endpoint is read from OTEL_EXPORTER_OTLP_ENDPOINT. If user overwrites Endpoint then it stays as is without any changes. I think the main subject of [this](https://github.com/open-telemetry/opentelemetry-dotnet/issues/2483) issue is to allow user to overwrite with it's own export endpoint.

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/OtlpExporterOptionsExtensionsTests.cs
```diff
@@ -131,5 +131,70 @@ public void AppendPathIfNotPresent_TracesPath_AppendsCorrectly(string inputUri,
...
+
+            options.AppendExportPath(options.Endpoint, "test/path");
+
+            Assert.Equal("http://localhost:4317/", options.Endpoint.AbsoluteUri);
+        }
+
+        [Fact]
+        public void AppendExportPath_EndpointNotSet_EnvironmentVariableDefined_Adjusted()
+        {
+            ClearEndpointEnvVar();
```
mic-max, line 150:
```suggestion
```

The clearing of the environment is unnecessary as it is immediately set to a value

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/OtlpExporterOptionsExtensionsTests.cs
```diff
@@ -131,5 +131,70 @@ public void AppendPathIfNotPresent_TracesPath_AppendsCorrectly(string inputUri,
...
+
+            Assert.Equal("http://test:8888/test/path", options.Endpoint.AbsoluteUri);
+
+            ClearEndpointEnvVar();
+        }
+
+        [Fact]
+        public void AppendExportPath_EndpointSetEqualToEnvironmentVariable_EnvironmentVariableDefined_NotAdjusted()
+        {
+            ClearEndpointEnvVar();
```
mic-max, line 165:
```suggestion
```

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/OtlpExporterOptionsExtensionsTests.cs
rypdal: Removed.

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/OtlpExporterOptionsExtensionsTests.cs
rypdal: Removed.

alanwest (approved):
LGTM, thanks @rypdal!

rypdal:
@cijothomas , @pellared , @alanwest ,  @utpilla , @joaopgrassi. @mic-max , @reyang thank you all for the review and discussions.

PR #2562, opened 2021-11-04
Replace use of TestExporter with InMemoryExporter

### LINKED ISSUE #2534: Consider removing TestExporter
`InMemoryExporter` is designed for unit test purpose.

If OpenTelemetry .NET unit tests have some common need that cannot be achieved by `InMemoryExporter`, we would want to fix it in `InMemoryExporter` rather than inventing another exporter for tests.

The ask is to evaluate whether it is possible to get rid of https://github.com/open-telemetry/opentelemetry-dotnet/blob/0c93744d4ba51676b61ab1be38f7764190cfcec0/test/OpenTelemetry.Tests/Shared/TestExporter.cs and use `InMemoryExporter`.

mic-max:
Fixes #2534

### DIFF
--- a/test/OpenTelemetry.Instrumentation.AspNetCore.Tests/OpenTelemetry.Instrumentation.AspNetCore.Tests.csproj
+++ b/test/OpenTelemetry.Instrumentation.AspNetCore.Tests/OpenTelemetry.Instrumentation.AspNetCore.Tests.csproj
@@ -16,12 +16,15 @@
   </ItemGroup>
 
   <ItemGroup>
-    <ProjectReference Include="$(RepoRoot)\src\OpenTelemetry.Instrumentation.AspNetCore\OpenTelemetry.Instrumentation.AspNetCore.csproj" />
     <ProjectReference Include="$(RepoRoot)\src\OpenTelemetry.Exporter.InMemory\OpenTelemetry.Exporter.InMemory.csproj" />
+    <ProjectReference Include="$(RepoRoot)\src\OpenTelemetry.Instrumentation.AspNetCore\OpenTelemetry.Instrumentation.AspNetCore.csproj" />
+  </ItemGroup>
+  
+  <ItemGroup>
     <Compile Include="$(RepoRoot)\test\OpenTelemetry.Tests\Shared\EventSourceTestHelper.cs" Link="EventSourceTestHelper.cs" />
+    <Compile Include="$(RepoRoot)\test\OpenTelemetry.Tests\Shared\InMemoryEventListener.cs" Link="InMemoryEventListener.cs" />
     <Compile Include="$(RepoRoot)\test\OpenTelemetry.Tests\Shared\TestEventListener.cs" Link="TestEventListener.cs" />
     <Compile Include="$(RepoRoot)\test\OpenTelemetry.Tests\Shared\TestExporter.cs" Link="TestExporter.cs" />
-    <Compile Include="$(RepoRoot)\test\OpenTelemetry.Tests\Shared\InMemoryEventListener.cs" Link="InMemoryEventListener.cs" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'net6.0'">
--- a/test/OpenTelemetry.Instrumentation.AspNetCore.Tests/OpenTelemetry.Instrumentation.AspNetCore.Tests.csproj
+++ b/test/OpenTelemetry.Instrumentation.AspNetCore.Tests/OpenTelemetry.Instrumentation.AspNetCore.Tests.csproj
@@ -17,6 +17,7 @@
 
   <ItemGroup>
     <ProjectReference Include="$(RepoRoot)\src\OpenTelemetry.Instrumentation.AspNetCore\OpenTelemetry.Instrumentation.AspNetCore.csproj" />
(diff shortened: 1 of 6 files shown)


mic-max:
There still are 5 unaddressed uses of the `TestExporter` remaining that I would like some guidance on how to replace them.

This one is a benchmark of iterating through the metric points by using `ref` or not using it in the `foreach` loop.
- MetricCollectBenchmarks.cs

The following all use the `TestExporter` in a similar way passing in a delegate function to the constructor named `RunTest` which includes code that will be run during the `TestExporter.Export()` method and contains `Assert` statements.
- OtlpHttpTraceExportClientTests.cs
- OtlpMetricsExporterTests.cs
- OtlpTraceExporterTests.cs
- PrometheusExporterExtensionsTests.cs

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/OpenTelemetry.Instrumentation.AspNetCore.Tests.csproj
```diff
@@ -17,6 +17,7 @@
 
   <ItemGroup>
     <ProjectReference Include="$(RepoRoot)\src\OpenTelemetry.Instrumentation.AspNetCore\OpenTelemetry.Instrumentation.AspNetCore.csproj" />
+    <ProjectReference Include="$(RepoRoot)\src\OpenTelemetry.Exporter.InMemory\OpenTelemetry.Exporter.InMemory.csproj" />
```
mic-max, line 20: Should I make a new `ItemGroup` for the `ProjectReference` tags instead of combining them with the `Compile` tags?

test/OpenTelemetry.Tests/Metrics/MetricAPITest.cs
```diff
@@ -144,25 +136,24 @@ void ProcessExport(Batch<Metric> batch)
...
-            Assert.Equal(1, metricCount);
+            Assert.Single(metricItems);
 
             // The following will be ignored as
             // metric of same name exists.
             // Metric stream will remain one.
+            // TODO: Fix comments and Assert statements to actually ensure only 1 metric stream is present.
             var anotherCounterSameName = meter1.CreateCounter<long>("name1");
             anotherCounterSameName.Add(10);
-            metricCount = 0;
```
mic-max, line 154: How would I replicate the `metricCount = 0;` statement when using the `InMemoryExporter`?
I modified the test in a way so that it would pass, but I believe all the `Assert` statements should be `Assert.Single(metricItems)` rather than the count increasing from 1 to 2 to 3.

test/OpenTelemetry.Tests/Metrics/MetricAPITest.cs
```diff
@@ -445,34 +423,25 @@ void ProcessExport(Batch<Metric> batch)
...
 
             // These updates would be dropped.
+            // TODO review this
             counterLong.Add(10, new KeyValuePair<string, object>("key", "valueA"));
             counterLong.Add(10, new KeyValuePair<string, object>("key", "valueB"));
             counterLong.Add(10, new KeyValuePair<string, object>("key", "valueC"));
-            metricPointCount = 0;
             metricReader.Collect();
-            Assert.Equal(AggregatorStore.MaxMetricPoints, metricPointCount);
+            Assert.Equal(AggregatorStore.MaxMetricPoints * 3, MetricPointCount());
```
mic-max, line 437: Is using these multiples of 2 and 3 for the subsequent assertions fine, or should the metrics be flushed in some way?

test/OpenTelemetry.Tests/Metrics/MetricAPITest.cs
```diff
@@ -409,18 +384,21 @@ void ProcessExport(Batch<Metric> batch)
         public void TestMetricPointCap(AggregationTemporality temporality)
         {
             var metricItems = new List<Metric>();
-            int metricPointCount = 0;
-            var metricExporter = new TestExporter<Metric>(ProcessExport);
+            var metricExporter = new InMemoryExporter<Metric>(metricItems);
 
-            void ProcessExport(Batch<Metric> batch)
+            int MetricPointCount()
```
mic-max, line 389: Thoughts on moving this function to the `InMemoryExporter` class?

test/OpenTelemetry.Tests/Metrics/MetricAPITest.cs
cijothomas: we can use metricItems.Count, right?

cijothomas: metricItems.Clear() should be done, in lieu of metriCount=0;

test/OpenTelemetry.Tests/Metrics/MetricAPITest.cs
cijothomas: wont clearing metricItems help?

test/OpenTelemetry.Tests/Metrics/MetricAPITest.cs
reyang: I would vote for keeping it non-public for now. And if we've found sufficient use cases, make it public later.

reyang (approved):
LGTM.

cijothomas (approved):

PR #2565, opened 2021-11-05
View - additional tags - user API only

cijothomas:
Just the user facing config only to get some feedback.

### DIFF
--- a/src/OpenTelemetry/Metrics/MetricStreamConfiguration.cs
+++ b/src/OpenTelemetry/Metrics/MetricStreamConfiguration.cs
@@ -14,8 +14,13 @@
 // limitations under the License.
 // </copyright>
 
+using System;
+using System.Collections.Generic;
+
 namespace OpenTelemetry.Metrics
 {
+    public delegate ReadOnlySpan<KeyValuePair<string, object>> AdditionalTagsCallback();
+
     // TODO: can be optimized like MetricType
     public enum Aggregation
     {
@@ -37,6 +42,8 @@ namespace OpenTelemetry.Metrics
         public string Description { get; set; }
 
         public string[] TagKeys { get; set; }
+
+        public AdditionalTagsCallback AdditionalTagsCallback { get; set; }
 
         public virtual Aggregation Aggregation { get; set; }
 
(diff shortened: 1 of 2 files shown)


src/OpenTelemetry/Metrics/MetricStreamConfiguration.cs
```diff
@@ -14,8 +14,16 @@
...
 
+using System;
+using System.Collections.Generic;
+
 namespace OpenTelemetry.Metrics
 {
+    // TODO: Clarify if spec allows adding anything from context
+    // or only Baggage entries. If the later, we just need to accept the
+    // Baggage key names only, and SDK can internally do Baggage.GetBaggage(name)
+    public delegate ReadOnlySpan<KeyValuePair<string, object>> AdditionalTagsCallback();
```
cijothomas, line 25: @reyang need some input from you here.

reyang: I think it is `extraDimensionsFromBaggage` and `extraDimensionsFromContext`.

reyang: I suggest we focus on baggage for now. The OpenTelemetry Context is using opaque key rather than string, which gives additional challenge.

src/OpenTelemetry/Metrics/MetricStreamConfiguration.cs
```diff
@@ -38,6 +38,18 @@ public class MetricStreamConfiguration
...
+        /// Gets or sets the names of <see cref="Baggage"/> keys, which,
+        /// if available, will be used as additional tags for Metric
+        /// aggregation.
+        /// <see cref="Baggage.GetBaggage(string)"/> will be used by the SDK
+        /// internally to obtain the additional tags.
+        /// </summary>
+        /// <remarks>
+        /// It is invalid to specify this for Observable instruments.
+        /// </remarks>
+        public string[] BaggageKeysForAdditionalTags { get; set; }
```
reyang, line 51: I'm not sure if these should be called Tags or Dimensions?

It seems the word "tag" could mean two different things:
1. The .NET way of calling an Attribute.
2. The Prometheus way of calling a "dimension".

reyang: Given Baggage only take string not Attribute, I guess "tag" will introduce confusion.

PR #2578, opened 2021-11-09
Add unit as metric name suffix in PrometheusExporter

reyang:
This is to prototype https://github.com/open-telemetry/opentelemetry-specification/pull/2017#discussion_r746103164 and https://github.com/open-telemetry/opentelemetry-specification/pull/2017#discussion_r746105004 so we can hammer out the OpenTelemetry->Prometheus data model spec.

### DIFF
--- a/src/OpenTelemetry.Exporter.Prometheus/Implementation/PrometheusSerializer.cs
+++ b/src/OpenTelemetry.Exporter.Prometheus/Implementation/PrometheusSerializer.cs
@@ -178,7 +178,7 @@ namespace OpenTelemetry.Exporter.Prometheus
                     (ordinal >= (ushort)'a' && ordinal <= (ushort)'z') ||
                     (ordinal >= (ushort)'0' && ordinal <= (ushort)'9'))
                 {
-                    cursor = WriteUnicodeNoEscape(buffer, cursor, ordinal);
+                    buffer[cursor++] = unchecked((byte)ordinal);
                 }
                 else
                 {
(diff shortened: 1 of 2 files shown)


src/OpenTelemetry.Exporter.Prometheus/Implementation/PrometheusSerializer.cs
```diff
@@ -249,21 +249,42 @@ public static int WriteMetricName(byte[] buffer, int cursor, string metricName)
...
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static int WriteHelpText(byte[] buffer, int cursor, string metricName, string metricDescription = null)
+        public static int WriteHelpText(byte[] buffer, int cursor, string metricName, string metricUnit = null, string metricDescription = null)
         {
             cursor = WriteAsciiStringNoEscape(buffer, cursor, "# HELP ");
-            cursor = WriteMetricName(buffer, cursor, metricName);
+            cursor = WriteMetricName(buffer, cursor, metricName, metricUnit);
 
-            if (metricDescription != null)
+            if (!string.IsNullOrEmpty(metricDescription))
```
reyang, line 287: This is to align with the OpenTelemetry Metrics spec:

_If the description is not provided or the description is null, the API and SDK MUST make sure that the behavior is the same as an empty description string._

reyang: @cijothomas I wonder if the SDK guarantees that or we need to cover it here?

cijothomas: No. We should handle it in SDK itself, so exporters need not repeat this.

reyang: Got it, I take it as:
1. The SDK needs to convert null description to empty string.
2. We can keep the logic here since the perf difference between `string.IsNullOrEmpty` and the empty string check is unnoticeable.

CodeBlanch (approved):
LGTM

cijothomas (approved):
add a changelog entry to indicate we append unit to metric name.

PR #2580, opened 2021-11-10
Metric description and unit to be treated as empty string when not provided

cijothomas:

If the unit is not provided or the unit is null, the API and SDK MUST make sure that the behavior is the same as an empty unit string.

If the description is not provided or the description is null, the API and SDK MUST make sure that the behavior is the same as an empty description string.


reyang (approved):
An alternative approach is to test if the string is empty and set it to null.

utpilla (approved):

Issue #2425, opened 2021-09-29
Handle Instrument disposal


cijothomas:
`MeasurementsCompleted` callback is made by .NET Runtime for each instrument, when the associated Meter has been disposed. We should leverage this callback, and clean up the Metric list maintained by the MeterProvider. Note that we cannot simply remove the entry, as there could be pending updates which are not yet collected. So need to ensure that anything pending is collected, before removing the Metric from Provider.

Sub issues/leftovers:
Prometheus specific, but generally applicable: https://github.com/open-telemetry/opentelemetry-dotnet/pull/2471#discussion_r726621580

Deal with potentiall timeout in Collect: https://github.com/open-telemetry/opentelemetry-dotnet/pull/2585#discussion_r746705512

cijothomas:
https://github.com/open-telemetry/opentelemetry-dotnet/pull/2471#discussion_r726621580 - this is not handled. Will create a separate issue.

cijothomas:
> [#2471 (comment)](https://github.com/open-telemetry/opentelemetry-dotnet/pull/2471#discussion_r726621580) - this is not handled. Will create a separate issue.

No longer relevant since Prometheus exporter implemented has changed since the original comment.

cijothomas:
The issue of potentially losing metrics(and its metric point snapshots), when Collect() is used with timeout exists, and should be solved separately from instrument disposal. 

Closing this issue as the instrument disposal (i.e Meter disposal) is handled now.

PR #2585, opened 2021-11-10
Handle instrument disposal

cijothomas:
Towards #2425 
This does not reclaim the position occupied by Metric in the Metric[], so the current hardcoded limit of 1000 remains even after disposing Meters (thereby instruments from it). This can be a follow up.

This will face similar challenges as https://github.com/open-telemetry/opentelemetry-dotnet/pull/2466 when reclaiming is incorporated, but likely a lot easier to fix as Instrument creation/disposal is not a hot path (instruments are typically created once and used throughout), and can afford to protect with plain lock statements.

This also does not re-allow the metric stream name even after the original instrument is disposed. There are some spec updates likely coming in this area, and this restriction might be going away. Hence not touching it in *this* PR.

### DIFF
--- a/docs/metrics/customizing-the-sdk/Program.cs
+++ b/docs/metrics/customizing-the-sdk/Program.cs
@@ -60,7 +60,7 @@ public class Program
             // turn off the above default. i.e any
             // instrument which does not match any views
             // gets dropped.
-            // .AddView(instrumentName: "*", new MetricStreamConfiguration() { Aggregation = Aggregation.Drop })
+            // .AddView(instrumentName: "*", MetricStreamConfiguration.Drop)
             .AddConsoleExporter()
             .Build();
 
--- a/src/OpenTelemetry/Metrics/MeterProviderSdk.cs
+++ b/src/OpenTelemetry/Metrics/MeterProviderSdk.cs
@@ -30,6 +30,7 @@ namespace OpenTelemetry.Metrics
         internal const int MaxMetrics = 1000;
         internal int ShutdownCount;
         private readonly Metric[] metrics;
+        private readonly Metric[] metricsCurrentBatch;
         private readonly List<object> instrumentations = new List<object>();
         private readonly List<Func<Instrument, MetricStreamConfiguration>> viewConfigs;
         private readonly object collectLock = new object();
@@ -50,6 +51,7 @@ namespace OpenTelemetry.Metrics
             this.Resource = resource;
             this.viewConfigs = viewConfigs;
             this.metrics = new Metric[MaxMetrics];
+            this.metricsCurrentBatch = new Metric[MaxMetrics];
 
             AggregationTemporality temporality = AggregationTemporality.Cumulative;
 
@@ -135,7 +137,7 @@ namespace OpenTelemetry.Metrics
                         // which will apply defaults.
                         // Users can turn off this default
                         // by adding a view like below as the last view.
-                        // .AddView(instrumentName: "*", new MetricStreamConfiguration() { Aggregation = Aggregation.Drop })
+                        // .AddView(instrumentName: "*", MetricStreamConfiguration.Drop)
                         metricStreamConfigs.Add(null);
                     }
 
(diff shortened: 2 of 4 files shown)


docs/metrics/customizing-the-sdk/Program.cs
```diff
@@ -60,7 +60,7 @@ public static void Main(string[] args)
             // turn off the above default. i.e any
             // instrument which does not match any views
             // gets dropped.
-            // .AddView(instrumentName: "*", new MetricStreamConfiguration() { Aggregation = Aggregation.Drop })
+            // .AddView(instrumentName: "*", MetricStreamConfiguration.Drop)
```
cijothomas, line 63: unrelated minor fix

src/OpenTelemetry/Metrics/MeterProviderSdk.cs
```diff
@@ -135,7 +137,7 @@ internal sealed class MeterProviderSdk : MeterProvider
                         // which will apply defaults.
                         // Users can turn off this default
                         // by adding a view like below as the last view.
-                        // .AddView(instrumentName: "*", new MetricStreamConfiguration() { Aggregation = Aggregation.Drop })
+                        // .AddView(instrumentName: "*", MetricStreamConfiguration.Drop)
```
cijothomas, line 140: unrelated minor fix

src/OpenTelemetry/Metrics/MeterProviderSdk.cs
```diff
@@ -412,12 +439,27 @@ internal Batch<Metric> Collect()
...
+                    int metricCountCurrentBatch = 0;
                     for (int i = 0; i < target; i++)
                     {
-                        this.metrics[i].SnapShot();
+                        var metric = this.metrics[i];
+                        if (metric != null)
+                        {
+                            if (metric.InstrumentDisposed)
+                            {
+                                metric.SnapShot();
```
cijothomas, line 450: if the instrument was disposed *before* Collect, then only clear it in current Collect. Otherwise, it'll be cleaned up in the *next* collect only.

reyang:
> This also does not re-allow the metric stream name even after the original instrument is disposed. There are some spec updates likely coming in this area, and this restriction might be going away. Hence not touching it in _this_ PR.

That's a good callout. It also has memory implications if someone keeps creating/disposing instruments.

src/OpenTelemetry/Metrics/MeterProviderSdk.cs
reyang: What if the next Collect failed (e.g. timeout)?

reyang (approved):

src/OpenTelemetry/Metrics/MeterProviderSdk.cs
utpilla, line 450: nit: `metric.Snapshot()` is always called. Simplify code by only having `this.metrics[i] = null` in the `if` block.

utpilla (approved):

src/OpenTelemetry/Metrics/MeterProviderSdk.cs
cijothomas: The check `metric.InstrumentDisposed` must be done before snapshot all, so not sure how to rearrange the code here..

cijothomas: good point. The provider currently has no way of knowing that (only readers know that).

src/OpenTelemetry/Metrics/MeterProviderSdk.cs
```diff
@@ -300,9 +305,31 @@ static Regex GetWildcardRegex(IEnumerable<string> collection)
...
+            metric.InstrumentDisposed = true;
+        }
+
         internal void MeasurementsCompleted(Instrument instrument, object state)
         {
-            Console.WriteLine($"Instrument {instrument.Meter.Name}:{instrument.Name} completed.");
+            var metrics = state as List<Metric>;
+            if (metrics == null)
+            {
+                // TODO: log
```
cijothomas, line 325: test comment

PR #2592, opened 2021-11-10
Ignore view when invalid histogram bounds are given

### LINKED ISSUE #2566: Histogram bounds must be validated
Similar to instrument/view name, we need to validate the custom bounds (no duplicates, increasing order etc.), if provided.
If we can fail-fast, do that. - done in #2573 
Else log and move on.

mic-max:
Fixes #2566

## Changes
- Logs error when adding a histogram with invalid bounds at runtime and throws an exception if adding when building the meter provider


src/OpenTelemetry/Metrics/MeterProviderSdk.cs
```diff
@@ -125,7 +125,17 @@ internal sealed class MeterProviderSdk : MeterProvider
                         var metricStreamConfig = viewConfig(instrument);
                         if (metricStreamConfig != null)
                         {
-                            metricStreamConfigs.Add(metricStreamConfig);
+                            // Validate histogram bounds
+                            if (metricStreamConfig is HistogramConfiguration histogramConfiguration &&
+                                histogramConfiguration?.BucketBounds != null &&
```
cijothomas, line 130:
```suggestion
                                histogramConfiguration.BucketBounds != null &&
```

mic-max: Good call

> The is operator returns true when an expression result is non-null...
[Source](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/type-testing-and-cast#is-operator)

alanwest (approved):
LGTM

src/OpenTelemetry/Metrics/MeterProviderBuilderExtensions.cs
```diff
@@ -89,7 +90,8 @@ public static MeterProviderBuilder AddView(this MeterProviderBuilder meterProvid
             if (metricStreamConfiguration is HistogramConfiguration histogramConfiguration)
             {
                 // Validate histogram bounds
-                if (histogramConfiguration.BucketBounds != null && !IsSortedAndDistinct(histogramConfiguration.BucketBounds))
+                if (histogramConfiguration.BucketBounds != null &&
+                    !Utils.IsSortedAndDistinct(histogramConfiguration.BucketBounds))
```
reyang, line 94: I think we need to handle the following cases:
1. if `Double.NaN` is provided in the `BucketBounds`, throw exception.
2. if `Double.NegativeInfinity` or `Double.PositiveInfinity` exist in the `BucketBounds` (and the array is ordered), skip the value.
3. if the resulting `BucketBounds` has no bounds, use `(-Inf, +Inf)` as the single bucket.

reyang (changes requested):
https://github.com/open-telemetry/opentelemetry-dotnet/pull/2592/files#r749866257

src/OpenTelemetry.Api/Internal/Utils.cs
```diff
@@ -0,0 +1,34 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+
+namespace OpenTelemetry.Internal
+{
+    internal static class Utils
+    {
+        public static bool IsSortedAndDistinct(double[] values)
```
reyang, line 21: Minor: I think this method is less likely to be used by other places, perhaps:
1. move it to the `MeterProviderBuilderExtensions.cs` as a private static method.
2. change the name to indicate the specific intention (e.g. `ValidateHistogramBucketBounds`).

mic-max:
The pr https://github.com/open-telemetry/opentelemetry-dotnet/pull/2735 includes some related changes to histogram, so will wait for that one to complete before finishing this one.

src/OpenTelemetry/Metrics/MeterProviderBuilderExtensions.cs
```diff
@@ -107,7 +108,20 @@ public static MeterProviderBuilder AddView(this MeterProviderBuilder meterProvid
             if (metricStreamConfiguration is ExplicitBucketHistogramConfiguration histogramConfiguration)
             {
                 // Validate histogram boundaries
-                if (histogramConfiguration.Boundaries != null && !IsSortedAndDistinct(histogramConfiguration.Boundaries))
+                if (histogramConfiguration.BucketBounds == null ||
```
utpilla, line 111: The user can provide `Boundaries` as `null`. In this case, we use the default bucket distribution for the Histogram: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#explicit-bucket-histogram-aggregation

src/OpenTelemetry/Metrics/MeterProviderBuilderExtensions.cs
```diff
@@ -107,7 +108,20 @@ public static MeterProviderBuilder AddView(this MeterProviderBuilder meterProvid
...
+                    !HasValidHistogramBucketBounds(histogramConfiguration.BucketBounds))
+                {
+                    throw new ArgumentException("Histogram bounds must be in ascending order with distinct values. double.NaN is not allowed.", nameof(histogramConfiguration.BucketBounds));
+                }
+
+                // If double.NegativeInfinity or double.PositiveInfinity exist in the BucketBounds, skip the value.
+                if (histogramConfiguration.BucketBounds != null)
```
utpilla, line 118: A user can provide just `double.NegativeInfinity` and `double.PositiveInfinity` as the `Boundaries` and it should be accepted. The SDK should treat that as one bucket `(-inf, +inf)`.

src/OpenTelemetry/Metrics/MeterProviderBuilderExtensions.cs
```diff
@@ -107,7 +108,20 @@ public static MeterProviderBuilder AddView(this MeterProviderBuilder meterProvid
...
+                {
+                    histogramConfiguration.BucketBounds = histogramConfiguration.BucketBounds.Where(x => x != double.NegativeInfinity || x != double.PositiveInfinity).ToArray();
+                }
+
+                // If the resulting BucketBounds has no bounds, use (-Inf, +Inf) as the single bucket
+                if (histogramConfiguration.BucketBounds == null || !Enumerable.Any(histogramConfiguration.BucketBounds))
```
utpilla, line 124: A user can provide an empty `double[]` as the `Boundaries`. In this case, the SDK should use `AggregationType.HistogramSumCount`.

This should be allowed to satisfy this requirement in the spec: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#view

> - Customize the aggregation - if the default aggregation associated with the Instrument does not meet the needs of the user. For example, an HTTP client library might expose HTTP client request duration as [Histogram](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#histogram) by default, but the application developer might only want the total count of outgoing requests.

src/OpenTelemetry/Metrics/MeterProviderBuilderExtensions.cs
mic-max: Correct, will replace the `||` on line 120 with `&&`

src/OpenTelemetry/Metrics/MeterProviderBuilderExtensions.cs
mic-max: Yes, I am doing that here: https://github.com/open-telemetry/opentelemetry-dotnet/pull/2735/files#diff-a149c9e64379e14c17737caf055686e54be18c08fb57d4f129576a141f61f740R96

src/OpenTelemetry/Metrics/MeterProviderBuilderExtensions.cs
```diff
@@ -107,9 +108,26 @@ public static MeterProviderBuilder AddView(this MeterProviderBuilder meterProvid
...
+                        throw new ArgumentException("Histogram bounds must be in ascending order with distinct values. double.NaN is not allowed.");
+                    }
+
+                    // Remove any infinity values from the histogram boundaries
+                    histogramConfiguration.Boundaries = histogramConfiguration.Boundaries.Where(x => !double.IsInfinity(x)).ToArray();
+
+                    // If empty bounds, use (-inf, +inf) as the single bucket
+                    if (!Enumerable.Any(histogramConfiguration.Boundaries))
+                    {
+                        histogramConfiguration.Boundaries = new double[] { double.NegativeInfinity, double.PositiveInfinity };
```
mic-max, line 124: Make the histogram of type HistogramSumCount?

src/OpenTelemetry/Metrics/MeterProviderSdk.cs
```diff
@@ -119,6 +119,36 @@ internal sealed class MeterProviderSdk : MeterProvider
...
+                                            continue;
+                                        }
+
+                                        // Remove any infinity values from the histogram boundaries
+                                        histogramConfiguration.Boundaries = histogramConfiguration.Boundaries.Where(x => !double.IsInfinity(x)).ToArray();
+
+                                        // If empty bounds, use (-inf, +inf) as the single bucket
+                                        if (!Enumerable.Any(histogramConfiguration.Boundaries))
+                                        {
+                                            histogramConfiguration.Boundaries = new double[] { double.NegativeInfinity, double.PositiveInfinity };
```
mic-max, line 142: Make the histogram of type HistogramSumCount?

src/OpenTelemetry/Metrics/MeterProviderBuilderExtensions.cs
```diff
@@ -223,11 +237,11 @@ public static MeterProvider Build(this MeterProviderBuilder meterProviderBuilder
...
         }
 
-        private static bool IsSortedAndDistinct(double[] values)
+        internal static bool HasValidHistogramBucketBounds(double[] values)
         {
-            for (int i = 1; i < values.Length; i++)
+            for (int i = 0; i < values.Length; i++)
             {
-                if (values[i] <= values[i - 1])
+                if (values[i] == double.NaN || (i > 0 && values[i] <= values[i - 1]))
```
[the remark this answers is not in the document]
utpilla: Ahh my bad! I got a little confused😀 It returns `false` when the condition it `true`.

src/OpenTelemetry/Metrics/MeterProviderBuilderExtensions.cs
utpilla: We should use `(-inf, +inf)` as the bucket only when the user explicitly provides the `Boundaries` as `new double[] { double.NegativeInfinity, double.PositiveInfinity }` in the view configuration.

src/OpenTelemetry/Metrics/MeterProviderBuilderExtensions.cs
```diff
@@ -223,11 +219,45 @@ public static MeterProvider Build(this MeterProviderBuilder meterProviderBuilder
...
+                        if (!Enumerable.Any(histogramConfiguration.Boundaries))
+                        {
+                            histogramConfiguration.Boundaries = new double[] { double.NegativeInfinity, double.PositiveInfinity };
+                        }
+                    }
+                }
+                else
+                {
+                    // Use default value when bounds are null
+                    histogramConfiguration.Boundaries = Metric.DefaultHistogramBounds;
```
utpilla, line 249: Do we need this? This should be taken care of in the `Metric` ctor: https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/Metrics/Metric.cs#L113

mic-max: Tests seem to work without it :)

src/OpenTelemetry/Metrics/MeterProviderBuilderExtensions.cs
```diff
@@ -223,11 +219,40 @@ public static MeterProvider Build(this MeterProviderBuilder meterProviderBuilder
...
+                        {
+                            onInvalidFunction();
+                            return false;
+                        }
+
+                        // Remove any infinity values from the histogram boundaries
+                        histogramConfiguration.Boundaries = histogramConfiguration.Boundaries.Where(x => !double.IsInfinity(x)).ToArray();
+
+                        // If empty bounds, use (-inf, +inf) as the single bucket
+                        if (!Enumerable.Any(histogramConfiguration.Boundaries))
```
utpilla, line 240: Use `histogramConfiguration.Boundaries.Length == 0` for the `if` check.

src/OpenTelemetry/Metrics/MetricPoint.cs
```diff
@@ -313,13 +313,19 @@ internal void Update(double number)
 
                 case AggregationType.Histogram:
                     {
-                        int i;
-                        for (i = 0; i < this.histogramBuckets.ExplicitBounds.Length; i++)
+                        // If -inf is the first bound, +inf will be the second of two bounds
+                        // in which case there is only a single bucket, so set the bucket
+                        // index to increment should be zero.
+                        int i = 0;
+                        if (!double.IsNegativeInfinity(this.histogramBuckets.ExplicitBounds[0]))
```
utpilla, line 320: This is updating the hot path. Could you run the Histogram benchmarks for before and after this change?

utpilla, line 320: Instead of checking for `IsNegativeInfinity`, just check if `this.histogramBuckets.RunningBucketCounts == 1`.

mic-max: Ya, that's a more sensible check to make

mic-max: Updated benchmark file with results in the comments. It is faster than the existing comment but the benchmark has changed since I first branched from main. I'll retest this again when I'm back at my desktop PC

src/OpenTelemetry/Metrics/HistogramBuckets.cs
```diff
@@ -37,8 +39,17 @@ public class HistogramBuckets
         internal HistogramBuckets(double[] explicitBounds)
         {
             this.ExplicitBounds = explicitBounds;
-            this.RunningBucketCounts = explicitBounds != null ? new long[explicitBounds.Length + 1] : null;
-            this.SnapshotBucketCounts = explicitBounds != null ? new long[explicitBounds.Length + 1] : new long[0];
+            if (explicitBounds != null)
+            {
+                var numBuckets = explicitBounds.Where(x => !double.IsInfinity(x)).Count() + 1;
```
TimothyMothra, line 44:
```suggestion
                var numBuckets = explicitBounds.Count(x => !double.IsInfinity(x)) + 1;
```

mic-max: Oo nice, I didn't know Count could take a predicate :)

src/OpenTelemetry/Metrics/MetricPoint.cs
```diff
@@ -325,13 +325,16 @@ internal void Update(double number)
...
+                        int i = 0;
+                        if (this.histogramBuckets.RunningBucketCounts.Length != 1)
                         {
-                            // Upper bound is inclusive
-                            if (number <= this.histogramBuckets.ExplicitBounds[i])
+                            for (; i < this.histogramBuckets.ExplicitBounds.Length; i++)
                             {
-                                break;
+                                // Upper bound is inclusive
+                                if (number <= this.histogramBuckets.ExplicitBounds[i++])
```
utpilla, line 334: Why increment `i` here again?

mic-max: Oops, this was unintentional. I was contemplating changing it to a while loop

mic-max:
Reopen, I added a couple new commits just to show the difference in benchmark results from main and this branch which looks negligible.

cijothomas:
Is this irrelevant now, considering https://github.com/open-telemetry/opentelemetry-dotnet/issues/2566#issuecomment-1085270736 ? If we are missing something, we can add it to the single method  `IsSortedAndDistinct`

PR #2601, opened 2021-11-12
Improve Prometheus example

reyang:
Removed the timeout (it's buggy, and doesn't have obvious benefit).
Used a real world async counter example.
Simplified the code.

### DIFF
--- a/examples/Console/TestPrometheusExporter.cs
+++ b/examples/Console/TestPrometheusExporter.cs
@@ -16,6 +16,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Diagnostics.Metrics;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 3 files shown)


examples/Console/TestPrometheusExporter.cs
```diff
@@ -16,80 +16,63 @@
...
-    internal class TestPrometheusExporter
-    {
-        private static readonly Meter MyMeter = new Meter("TestMeter");
-        private static readonly Counter<double> Counter = MyMeter.CreateCounter<double>("myCounter", description: "A counter for demonstration purpose.");
-        private static readonly Histogram<long> MyHistogram = MyMeter.CreateHistogram<long>("myHistogram");
-        private static readonly ThreadLocal<Random> ThreadLocalRandom = new ThreadLocal<Random>(() => new Random());
-
-        internal static object Run(int port, int totalDurationInMins)
-        {
-            /*
```
cijothomas, line 37: i'd still keep a known prometheus.yml for quickly testing.

We include it in almost all examples in this folder
https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/examples/Console/TestOtlpExporter.cs#L30

examples/Console/TestPrometheusExporter.cs
```diff
@@ -16,80 +16,63 @@
...
-            Following is sample prometheus.yml config. Adjust port,interval as needed.
+namespace Examples.Console;
 
-            scrape_configs:
-              # The job name is added as a label `job=<job_name>` to any timeseries scraped from this config.
-              - job_name: 'OpenTelemetryTest'
+internal class TestPrometheusExporter
+{
+    private static readonly Meter MyMeter = new Meter("MyMeter");
+    private static readonly Counter<double> Counter = MyMeter.CreateCounter<double>("myCounter", description: "A counter for demonstration purpose.");
```
cijothomas, line 32: please put `unit` as well, so as to demo how it'll appear in prometheus.

cijothomas (approved):
LGTM.
Suggest to retain a valid prometheus.yml example in the code comment.

examples/Console/TestPrometheusExporter.cs
reyang: I've added a simple yml file.

Curious, do we treat this project as an "example", a "test" or both?

examples/Console/TestPrometheusExporter.cs
reyang: the thread runaway time has unit so I intend to not put unit here

examples/Console/TestPrometheusExporter.cs
cijothomas: yes its a mix of both now. Agree we need a readme with instructions on how to run etc.

(I use them a lot, (and I believe other contributors as well), to quickly test things out.)

PR #2634, opened 2021-11-17
Allow duplicate metric name when meter is different

### LINKED ISSUE #2627: Instrument name collisions for different Meters
Opening this as a tracking issue for the bug where `Instrument`s that have the same name are dropped, even if they come from separate `Meter`s. From the [API spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument): 

> Different Meters MUST be treated as separate namespaces. The names of the Instruments under one Meter SHOULD NOT interfere with Instruments under another Meter.

This issue does **only** cover the fact that duplicate instrument elimination in the .NET SDK implementation does not take into account the `Meter` name, but only the `Instrument` name. It does **not** cover downstream implications (such as how to deal with Prometheus export in such a case, which has extensively been discussed and resolved (?) [elsewhere](https://github.com/open-telemetry/opentelemetry-specification/pull/2035#issuecomment-965540896)).

While this should occur only rarely, I think its important that this is addressed. As more and more libraries adopt OpenTelemetry, such situations are likely to appear more often. OTLP is able to transport these data streams, so it seems counter-intuitive to drop them.

## Additional Context:

https://github.com/open-telemetry/opentelemetry-specification/issues/2064
[Code where instruments are dropped when using Views](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/Metrics/MeterProviderSdk.cs#L169-L175)
[Code where instruments are dropped without Views](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/Metrics/MeterProviderSdk.cs#L252)

alanwest:
Fixes #2627.

## Changes

### DIFF
--- a/src/OpenTelemetry/CHANGELOG.md
+++ b/src/OpenTelemetry/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-* Metrics with the same name but different from different meters is allowed.
+* Metrics with the same name but from different meters is allowed.
   ([#2634](https://github.com/open-telemetry/opentelemetry-dotnet/pull/2634))
 
 * Metrics SDK will not provide inactive Metrics to delta exporter.
--- a/src/OpenTelemetry/CHANGELOG.md
+++ b/src/OpenTelemetry/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Metrics with the same name but different from different meters is allowed.
+  ([#2634](https://github.com/open-telemetry/opentelemetry-dotnet/pull/2634))
+
 * Metrics SDK will not provide inactive Metrics to delta exporter.
   ([#2629](https://github.com/open-telemetry/opentelemetry-dotnet/pull/2629))
 
--- a/src/OpenTelemetry/CHANGELOG.md
+++ b/src/OpenTelemetry/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-* Metrics with the same name but from different meters is allowed.
+* Metrics with the same name but from different meters are allowed.
   ([#2634](https://github.com/open-telemetry/opentelemetry-dotnet/pull/2634))
 
 * Metrics SDK will not provide inactive Metrics to delta exporter.
--- a/src/OpenTelemetry/Metrics/MeterProviderSdk.cs
+++ b/src/OpenTelemetry/Metrics/MeterProviderSdk.cs
@@ -437,8 +437,8 @@ namespace OpenTelemetry.Metrics
                         OpenTelemetrySdkEventSource.Log.MetricObserverCallbackException(exception);
                     }
 
-                    var indexSnapShot = Math.Min(this.metricIndex, MaxMetrics - 1);
-                    var target = indexSnapShot + 1;
+                    var indexSnapshot = Math.Min(this.metricIndex, MaxMetrics - 1);
+                    var target = indexSnapshot + 1;
                     int metricCountCurrentBatch = 0;
                     for (int i = 0; i < target; i++)
                     {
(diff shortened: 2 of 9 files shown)


src/OpenTelemetry/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Metrics with the same name but from different meters is allowed.
```
reyang, line 5: nit:

```suggestion
* Metrics with the same name but from different meters are allowed.
```

reyang (approved):
LGTM.

test/OpenTelemetry.Tests/Metrics/MetricAPITest.cs
```diff
@@ -102,9 +102,11 @@ public void ObserverCallbackExceptionTest()
...
 
         [Theory]
-        [InlineData(AggregationTemporality.Cumulative)]
-        [InlineData(AggregationTemporality.Delta)]
-        public void StreamNamesDuplicatesAreNotAllowedTest(AggregationTemporality temporality)
+        [InlineData(AggregationTemporality.Cumulative, true)]
+        [InlineData(AggregationTemporality.Cumulative, false)]
+        [InlineData(AggregationTemporality.Delta, true)]
+        [InlineData(AggregationTemporality.Delta, false)]
+        public void StreamNamesDuplicatesAreNotAllowedTest(AggregationTemporality temporality, bool hasView)
```
utpilla, line 109: Rename this to `StreamNameDuplicatesAreAllowedWithDifferentMetersTest`?

We should add another unit test to confirm that we cannot add duplicate metric stream names when the instruments created from the same meter.

utpilla (approved):
Left the suggestions for unit tests.

test/OpenTelemetry.Tests/Metrics/MetricAPITest.cs
alanwest: This unit test currently tests both

1) that duplicate names are not allowed in when the instruments are created from the same meter, and
2) that duplicate names are allowed when instruments are from different meters

Are you suggesting breaking this into two tests for clarity? If so, I'm 👍 on that idea.

utpilla: > We should also add another unit test to confirm that we cannot add duplicate metric stream names when the instruments created from the same meter.

Ohh I just realized that this was already covered before this PR. I guess we could have two different unit tests so that their names match what they are testing.

alanwest: Split em up. ✂️

cijothomas (approved):

src/OpenTelemetry/Metrics/MeterProviderSdk.cs
```diff
@@ -153,12 +153,14 @@ internal sealed class MeterProviderSdk : MeterProvider
                         for (int i = 0; i < maxCountMetricsToBeCreated; i++)
                         {
                             var metricStreamConfig = metricStreamConfigs[i];
-                            var metricStreamName = metricStreamConfig?.Name ?? instrument.Name;
+                            var meterName = instrument.Meter.Name;
+                            var metricName = metricStreamConfig?.Name ?? instrument.Name;
+                            var metricStreamName = $"{meterName}.{metricName}";
 
-                            if (!MeterProviderBuilderSdk.IsValidInstrumentName(metricStreamName))
+                            if (!MeterProviderBuilderSdk.IsValidInstrumentName(metricName))
```
pirgeo, line 160: Shouldn't this be `metricStreamName` still? As it is now, it again uses the instrument name, right?

utpilla: We do want to use the instrument name as the validation check is only for the instrument name. `metricStreamName` is now used to ensure the uniqueness of the combination of meter name and instrument name.

pirgeo: :+1:, thanks for tackling this so quickly!

PR #2636, opened 2021-11-17
Fix Prometheus serialization bug

reyang:
I introduced the bug in #2553 based on a false assumption that `metricPoint.Keys` will be an empty array rather than `null` if there is zero dimension.

@CodeBlanch noticed the bug in https://github.com/open-telemetry/opentelemetry-dotnet/pull/2610#discussion_r749620234.

This PR has provided a complete fix + test coverage.


CodeBlanch (approved):
LGTM

PR #2640, opened 2021-11-18
Leverage PrometheusExporter in stress test

reyang:


test/OpenTelemetry.Tests.Stress.Metrics/Program.cs
```diff
@@ -41,8 +42,15 @@ public static void Main()
 
         using var meterProvider = Sdk.CreateMeterProviderBuilder()
             .AddMeter(TestMeter.Name)
+            .AddPrometheusExporter(options =>
```
cijothomas, line 45: I really like the stress test exposing memory stats via Prometheus, but this (^^ line), could be reconsidered in a future PR, to make it configurable. We surely want it to trigger collection, by some human action, but want to have flexibility to pick Delta vs Cumulative. This PR hard codes Prometheus, and hence Cumulative gets enabled.
May be add a comment about it, or add a Delta reader in comment, so its easy to uncomment and enable for anyone trying this.

cijothomas (approved):

test/OpenTelemetry.Tests.Stress.Metrics/Program.cs
reyang: Maybe OTLP exporter is a better option.

PR #2641, opened 2021-11-19
Fix metric export interval to match spec

### LINKED ISSUE #2575: Is TimeSpan.Infinite the best default aggregation default for console exporter?
# Question

**Describe your environment.**

Local development

**What are you trying to achieve?**

I was creating a POC to use OpenTelemetry metrics and couldn't get it to work. After trying a lot of things, it turns out that the default aggregation interval is `TimeSpan.Infinite`.

This means that, by default, the metrics will only be exported when the application shuts down (if the `MetricProvider` is disposed properly)

Wouldn't it make more sense to use 5 min as a default? What was the reasoning behind this?

svetlanabrennan:
The spec says export interval should be 60000 milliseconds and this pr fixes this for the otlp metric exporter. 

## Changes

### DIFF
--- a/src/OpenTelemetry.Exporter.Console/ConsoleExporterOptions.cs
+++ b/src/OpenTelemetry.Exporter.Console/ConsoleExporterOptions.cs
@@ -14,7 +14,6 @@
 // limitations under the License.
 // </copyright>
 
-using System.Threading;
 using OpenTelemetry.Metrics;
 
 namespace OpenTelemetry.Exporter
--- a/src/OpenTelemetry.Exporter.Console/ConsoleExporterOptions.cs
+++ b/src/OpenTelemetry.Exporter.Console/ConsoleExporterOptions.cs
@@ -14,6 +14,7 @@
 // limitations under the License.
 // </copyright>
 
+using System.Threading;
 using OpenTelemetry.Metrics;
 
 namespace OpenTelemetry.Exporter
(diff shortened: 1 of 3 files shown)


src/OpenTelemetry.Exporter.Console/ConsoleExporterOptions.cs
```diff
@@ -27,9 +26,9 @@ public class ConsoleExporterOptions
         public ConsoleExporterOutputTargets Targets { get; set; } = ConsoleExporterOutputTargets.Console;
 
         /// <summary>
-        /// Gets or sets the metric export interval in milliseconds. The default value is <c>Timeout.Infinite</c>.
+        /// Gets or sets the metric export interval in milliseconds. The default value is 60000.
         /// </summary>
-        public int MetricExportIntervalMilliseconds { get; set; } = Timeout.Infinite;
+        public int MetricExportIntervalMilliseconds { get; set; } = 60000;
```
reyang, line 31: Why would a console exporter use the periodic exporting strategy?

alanwest: Ah, yes that makes sense now. The side-effect of this change causing the console exporter to default to periodic exporting was not immediately obvious.

https://github.com/open-telemetry/opentelemetry-dotnet/blob/6e6afdbc3614cfc52d20825a25f06bd80b5ba64e/src/OpenTelemetry.Exporter.Console/ConsoleExporterMetricsExtensions.cs#L42-L44

My gut says it would make sense to decouple the idea of periodic vs. non-periodic in a separate config setting. Kinda similar to how tracing has the batch exporter options.

alanwest: Maybe break this into two PRs... because I think the change to the OTLP exporter is correct.

reyang (approved):
LGTM.

cijothomas:
@svetlanabrennan Could you update the description as it is outdated now? The linked issue require separate handling, so that may be removed.

src/OpenTelemetry.Exporter.Console/ConsoleExporterOptions.cs
svetlanabrennan: Ok. I broke it up.

alanwest (approved):

PR #2656, opened 2021-11-21
Update docker-compose.yml

mehrdad-goudarzi:


reyang:
@mehrdad-goudarzi thanks for the contribution, would you clear the CLA?

image

ktmitton:
Just noting, after testing it doesn't look like the new environment variables are doing what it _looks_ like they are doing. Those settings are for configuration files, not environment variables:
https://www.rabbitmq.com/configure.html#config-items

While the container would still spin up successfully, it looks like it wasn't because of those env settings. The default user for rabbitmq is already guest/guest per https://www.rabbitmq.com/access-control.html#default-state, so the change this PR presents is really removing the RABBITMQ_* settings that broke from 3.9.0-3.9-3, and allowing the default behavior to trigger.

Testing I did to confirm this was using basically what's in the PR, but changing the user and/or password to something like guest2/securepassowrd. I wasn't able to observe a change in the default account, I still had to log in using guest/guest.

reyang:
Superseded by #2682.

PR #2661, opened 2021-11-23
Rename batch metric point

cijothomas:
Fixes #.

## Changes

### DIFF
--- a/src/OpenTelemetry/Metrics/Metric.cs
+++ b/src/OpenTelemetry/Metrics/Metric.cs
@@ -124,7 +124,7 @@ namespace OpenTelemetry.Metrics
 
         internal bool InstrumentDisposed { get; set; }
 
-        public BatchMetricPoint GetMetricPoints()
+        public MetricPointsAccessor GetMetricPoints()
         {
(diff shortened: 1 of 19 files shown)


src/OpenTelemetry/Metrics/Metric.cs
```diff
@@ -124,9 +124,9 @@ public sealed class Metric
 
         internal bool InstrumentDisposed { get; set; }
 
-        public BatchMetricPoint GetMetricPoints()
+        public MetricPointsAccessor GetMetricPointsAccessor()
```
cijothomas, line 127: unsure if the methodname can still be `GetMetricPoints()` as before...

reyang: IMHO `MetricPointsAccessor GetMetricPoints()` is better.

cijothomas: keeping GetMetricPoints() now.
Its unlikely that anyone would see the name, as exporters typically just do a foreach (var metricpoint in metric.GetMetricPoints())

reyang (approved):

PR #2673, opened 2021-11-24
Have separate Metrictype for each histogram

cijothomas:
Adding specific MetricType for the type of histograms.


src/OpenTelemetry/Metrics/MetricType.cs
```diff
@@ -23,10 +23,13 @@ public enum MetricType : byte
...
-            0x10: Sum
-            0x20: Gauge
-            0x30: Histogram
-            0x40: Summary (reserved)
+            0x10 0b00010000: Sum
+            0x20 0b00100000: Gauge
+            0x40 0b01000000: Histogram
+            0x50 0b01010000: HistogramWithBuckets
+            0x60 0b01100000: HistogramWithMinMax
+            0x70 0b01110000: HistogramWithMinMaxAndBuckets
```
cijothomas, line 31: @reyang when exponentialhistogram arrives, we won't have any bits left for it.. unless we take Summary or make this 16 bits instead of current 8 bytes

reyang: I think int is fine.

src/OpenTelemetry/Metrics/MetricType.cs
```diff
@@ -19,16 +19,29 @@
...
-    public enum MetricType : byte
+    public enum MetricType : int
     {
         /*
-        Type:
-            0x10: Sum
-            0x20: Gauge
-            0x30: Histogram
-            0x40: Summary (reserved)
+        Type (bits 4,5,6,7,8,9,10,11):
```
reyang, line 25: We need to have some consistency here, the spacing, whether binary/hex are used, etc.

src/OpenTelemetry/Metrics/MetricTypeExtensions.cs
```diff
@@ -59,7 +59,7 @@ public static bool IsGauge(this MetricType self)
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static bool IsHistogram(this MetricType self)
         {
-            return (self & METRIC_TYPE_MASK) == METRIC_TYPE_HISTOGRAM;
+            return (self & METRIC_TYPE_HISTOGRAM) == METRIC_TYPE_HISTOGRAM;
```
reyang, line 62: `(self & METRIC_TYPE_HISTOGRAM) == METRIC_TYPE_HISTOGRAM` is misleading, use `self.HasFlag(METRIC_TYPE_HISTOGRAM)`.

src/OpenTelemetry/Metrics/MetricTypeExtensions.cs
```diff
@@ -22,12 +22,12 @@ public static class MetricTypeExtensions
...
 
-        internal const MetricType METRIC_TYPE_MASK = (MetricType)0xf0;
+        internal const MetricType METRIC_TYPE_MASK = (MetricType)0xff0;
 
         internal const MetricType METRIC_TYPE_SUM = (MetricType)0x10;
         internal const MetricType METRIC_TYPE_GAUGE = (MetricType)0x20;
-        internal const MetricType METRIC_TYPE_HISTOGRAM = (MetricType)0x30;
-        /* internal const byte METRIC_TYPE_SUMMARY = 0x40; // not used */
+        internal const MetricType METRIC_TYPE_HISTOGRAM = (MetricType)0x40;
+        /* internal const byte METRIC_TYPE_SUMMARY = 0x100; // not used */
```
reyang, line 30: We can probably remove this (and the revert the 0xff0 part) since it's not used.

cijothomas:
closing in favor of https://github.com/open-telemetry/opentelemetry-dotnet/pull/2693

PR #2682, opened 2021-11-25
Removed explicit setting of default guest user

### LINKED ISSUE #2655: rabbitmq default user and pass needs to be updated in the MicroserviceExample's docker-compose
I tried to run the sample to check the output but something needed to be fixed in the docker-compose, rabbitmq section.

this is the container's log for rabbtmq:

```
error: RABBITMQ_DEFAULT_PASS is set but deprecated

error: RABBITMQ_DEFAULT_USER is set but deprecated

error: deprecated environment variables detected


Please use a configuration file instead; visit https://www.rabbitmq.com/configure.html to learn more
```

ktmitton:
Fixes #2655.

## Changes

RabbitMq 3.9.0 through 3.9.3 had a breaking change that disabled the environment variables for setting the default user/password. This is discussed on https://github.com/docker-library/rabbitmq/issues/508, and as mentioned in the comments, explicitly setting these for guest/guest isn't necessary.

As documented https://www.rabbitmq.com/access-control.html#default-state, the default user is already set to guest/guest, and these links to the docker build process show how that user is configured to allow traffic beyond localhost:
https://github.com/docker-library/rabbitmq/blob/5ec7771d821917f6839d682ab04101f4d87a8daf/3.9/alpine/Dockerfile#L262
https://github.com/docker-library/rabbitmq/blob/master/3.9/alpine/10-default-guest-user.conf

Since this is already default behavior in RabbitMq, and I got the impression that some of the maintainers there may still have their eye on removing the environment variables in the future, it seemed like a good option to remove them from the docker-compose file and avoid this issue cropping up again later.


cijothomas (approved):

reyang:
Related to #2656.

reyang (approved):

ktmitton:
> Related to #2656.

Thanks for linking that @reyang, didn't catch that before! I'll leave a comment on that, I looked into something like that while testing and found it wasn't doing what it _looks_ like it's doing.

Issue #2589, opened 2021-11-10
Provide capability to have more advanced ASP.NET Core telemetry for metrics


tomkerkhove:
# Feature Request

**Is your feature request related to a problem?**

Provide capability to have more advanced ASP.NET Core telemetry allowing us to add more information such as routes which are not supported today.

**Describe the solution you'd like:**

Configure this through the options of the extension

**Describe alternatives you've considered.**

We are currently working around this by writing our own middleware to provide more information

## Additional Context

N/A

cijothomas:
We have Enrich in ASP.NET Core today.
https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/src/OpenTelemetry.Instrumentation.AspNetCore#enrich

If there are scenarios not covered by this, please describe them for tracking.

tomkerkhove:
Sorry, my feature request was not thorough enough as I'm asking for the metrics aspect of it.

In the extension there are still a few to-dos that need to be tackled.

cijothomas:
Yes the instrumentation for metrics are very early stage. It'll be prioritized after the core SDK is done. We are happy to get contributions, if you are interested in!

tomkerkhove:
I will try to have a look when I have time, but no promises

ktmitton:
@tomkerkhove I think I'm ready to take a stab at a _real_ contribution, what kind of metrics are you looking for?

tomkerkhove:
Sure, from top of mind:

- # of requests (tags: status code)
- # of failed requests
- # of successful requests
- Request duration (already there I think)

ktmitton:
Not sure if I'm being too conservative here, but if we are tagging number of requests with status codes, so we need separate metrics for failed/successful, or can that be handled downstream by filtering data?

tomkerkhove:
No problem at all. From my side I'd say it can be beneficial because not all systems support this; for example Prometheus allows you to sum those up but not all of them.

But I'm not a metrics expert.

tomkerkhove:
What I would do though, is make sure people can opt-out and/or extend the metrics that are being emitted fairly easily

ktmitton:
There's still something I'm trying to figure out, but...it looks like this might be something low-cost where you filtering isn't strictly necessary. I've looked at the prometheus implementation in the past, and that relied on middleware, which I think made sense to make sure you have filters so you can limit overhead.

[...]

Issue #2556, opened 2021-11-02
Metric instrumentation for Asp.Net


cijothomas:
Similar to: https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInMetricsListener.cs

alirizaadiyahsi:
We are exporting metrics to New Relic and we can handle request counts, durations, failed request counts, successful request counts. But we can group them by url. For example we want to group by `user/login` and `user/register`. My question is: Is there any estimation about this issue? 

At least may be you can add http attributes like request url or path?

Thanks.

PR #2687, opened 2021-11-25
[Do Not Merge] Added new metric for completed requests

ktmitton:
## Changes

Added a new metric to aspnetcore to track completed requests.

Related Issues:
* #2589
* #2556

### DIFF
--- a/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInMetricsListener.cs
+++ b/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInMetricsListener.cs
@@ -36,7 +36,7 @@ namespace OpenTelemetry.Instrumentation.AspNetCore.Implementation
         {
             this.meter = meter;
             this.httpServerDuration = meter.CreateHistogram<double>("http.server.duration", "ms", "measures the duration of the inbound HTTP request");
-            this.httpServerCompletedRequests = meter.CreateCounter<long>("http.server.completed_requests", "requests", "measures the number of concurrent HTTP requests that are completed");
+            this.httpServerCompletedRequests = meter.CreateCounter<long>("http.server.completed_requests", "requests", "counts the HTTP requests that are completed");
         }
 
         public override void OnStopActivity(Activity activity, object payload)
--- a/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInMetricsListener.cs
+++ b/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInMetricsListener.cs
@@ -29,11 +29,14 @@ namespace OpenTelemetry.Instrumentation.AspNetCore.Implementation
 
         private Histogram<double> httpServerDuration;
 
+        private Counter<long> httpRequestCount;
+
         public HttpInMetricsListener(string name, Meter meter)
             : base(name)
         {
             this.meter = meter;
             this.httpServerDuration = meter.CreateHistogram<double>("http.server.duration", "ms", "measures the duration of the inbound HTTP request");
+            this.httpRequestCount = meter.CreateCounter<long>("http.server.completed_requests", "requests", "measures the number of concurrent HTTP requests that are completed");
         }
 
         public override void OnStopActivity(Activity activity, object payload)
@@ -64,6 +67,7 @@ namespace OpenTelemetry.Instrumentation.AspNetCore.Implementation
             };
 
             this.httpServerDuration.Record(activity.Duration.TotalMilliseconds, tags);
+            this.httpRequestCount.Add(1, tags);
         }
     }
 }
--- a/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInMetricsListener.cs
+++ b/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInMetricsListener.cs
@@ -29,14 +29,14 @@ namespace OpenTelemetry.Instrumentation.AspNetCore.Implementation
 
         private Histogram<double> httpServerDuration;
 
-        private Counter<long> httpRequestCount;
+        private Counter<long> httpServerCompletedRequests;
 
         public HttpInMetricsListener(string name, Meter meter)
             : base(name)
         {
             this.meter = meter;
             this.httpServerDuration = meter.CreateHistogram<double>("http.server.duration", "ms", "measures the duration of the inbound HTTP request");
-            this.httpRequestCount = meter.CreateCounter<long>("http.server.completed_requests", "requests", "measures the number of concurrent HTTP requests that are completed");
+            this.httpServerCompletedRequests = meter.CreateCounter<long>("http.server.completed_requests", "requests", "measures the number of concurrent HTTP requests that are completed");
         }
 
         public override void OnStopActivity(Activity activity, object payload)
@@ -67,7 +67,7 @@ namespace OpenTelemetry.Instrumentation.AspNetCore.Implementation
             };
 
             this.httpServerDuration.Record(activity.Duration.TotalMilliseconds, tags);
-            this.httpRequestCount.Add(1, tags);
+            this.httpServerCompletedRequests.Add(1, tags);
         }
     }
 }
--- a/test/OpenTelemetry.Instrumentation.AspNetCore.Tests/MetricTests.cs
+++ b/test/OpenTelemetry.Instrumentation.AspNetCore.Tests/MetricTests.cs
@@ -55,7 +55,7 @@ namespace OpenTelemetry.Instrumentation.AspNetCore.Tests
         }
 
         [Fact]
-        public async Task RequestMetricIsCaptured()
+        public async Task RequestDurationMetricIsCaptured()
         {
             var metricItems = new List<Metric>();
             var metricExporter = new InMemoryExporter<Metric>(metricItems);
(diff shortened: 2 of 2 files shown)


ktmitton:
Noticed the AspNetCore example app seems to depend on starting up services outside the test, working on a separate PR to move that into docker like the microservice example.

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInMetricsListener.cs
```diff
@@ -29,11 +29,14 @@ internal class HttpInMetricsListener : ListenerHandler
...
         private Histogram<double> httpServerDuration;
 
+        private Counter<long> httpServerCompletedRequests;
+
         public HttpInMetricsListener(string name, Meter meter)
             : base(name)
         {
             this.meter = meter;
             this.httpServerDuration = meter.CreateHistogram<double>("http.server.duration", "ms", "measures the duration of the inbound HTTP request");
+            this.httpServerCompletedRequests = meter.CreateCounter<long>("http.server.completed_requests", "requests", "measures the number of concurrent HTTP requests that are completed");
```
reyang, line 39: Curious - where is this idea coming from?

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/http-metrics.md#http-server

reyang (changes requested):
If we need to add a new instrument, consider updating the experimental spec first https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/http-metrics.md#http-server.

ktmitton:
@reyang updates to a spec are something I'm new to and not yet comfortable with. Is it as simple as submitting a PR for the markdown you linked?

reyang:
> @reyang updates to a spec are something I'm new to and not yet comfortable with. Is it as simple as submitting a PR for the markdown you linked?

Submitting a PR is the easy part, there will be discussions/comments which is the hard part.
The key is - we don't add something just because it is _useful_, we add things when we have good confidence that they will have long term success.

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInMetricsListener.cs
ktmitton: Yeah, I saw that thanks to the pre-existing comment in the file, but to be honest, I wasn't sure what to do. Like I said in my other PR, updating specs is _very_ new to me, so I'm playing things a little cautious and trying not to rock the boat too much until I get more comfortable with this. Wasn't sure if this metric was justified for an overall standard, or was something more specific to this instrumentation, so I just ran with my own name assuming somebody would call it out if it was an issue and help walk me through this 😃 

Here's the context for my decisions, let me know how you think we should proceed:

It came from issue #2589, assuming I'm interpreting @tomkerkhove correctly. I could be wrong, but I think this is a metric I've seen from standard prometheus .net metrics, too, which is why it seemed reasonable to me.

Additionally, where I work I think we have some of our alerts based on this. I think it is based on the delta over a period of time to make sure we are around our baseline traffic, so we can catch when it fluctuates due to things like network/firewall issues, or a spam of bot traffic.

Talking through it like that is making me think this isn't an aspnetcore specific metric, and something useful for other servers...

ktmitton:
> > @reyang updates to a spec are something I'm new to and not yet comfortable with. Is it as simple as submitting a PR for the markdown you linked?
> 
> Submitting a PR is the easy part, there will be discussions/comments which is the hard part. The key is - we don't add something just because it _could be useful_, we add things when we have good confidence that they will have long term success.

We are kind of having the same conversation in the review comment you left, mind if we just 100% move this discussion to that since it has a clearer (to me) conversation flow?

reyang:
> view comment you left, mind if we just 100% move this discussion to that since it has a clearer (to me) conversation flow?

I'm open to Discussions.

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInMetricsListener.cs
reyang: At minimum, these are the things that we must address:
1. The wording in the description, "measures" is used for histogram, here we need to use "counts" https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/supplementary-guidelines.md#instrument-selection
2. The accuracy of the description, e.g. it is not "concurrent HTTP requests that are completed", it is "the HTTP requests that are completed" ("concurrent" adds no additional meaning here).
3. What about requests that are not completed? e.g. requests that are in-flight (this should be easy because it is clearly not included), and requests that are aborted/timed-out (this is not clear at all). In a nutshell, what does "completed" mean?

ktmitton: 1. Ok, I just straight up missed that nuance, I'll get it updated
1. ...that's a result of me trying to reword my original description to be more inline with the other metric, and somehow I skipped that word while reviewing it, I'll get that changed
1. Yeah, that was something I was thinking was a little vague and was waiting for somebody to chime in. It's also like the reasons I was hesitant to add the "successful" and "failed" metrics asked for in #2589.
   1. To me, completed means "we're done processing, regardless of the result". Does that make sense? I'm open to alternative names or definitions here.
   1. I'm wondering if maybe I should start with the spec-defined metric, instead of a community requested one. Since these two are so close to each other, we might benefit from knocking that out first as a stepping stone to this. It might help to see how a request moves from the in-flight metric to the completed one.

ktmitton: actually, for number 1, I based that on the description in the spec for http.server.active_requests, not the description in our code:
image

Should we get the spec updated to reflect what you pointed out?

ktmitton: nvm, re-read what you linked, this is the key bit that makes the spec correct:
image

ktmitton: Just pushed a change to address 1 & 2, so unless you still see something wrong with those bits, we just need to figure out how to resolve 3!

ktmitton: "This will be quick, I can implement active requests for reference before work"...then I realized what is required by the spec means my original idea for it won't work, I don't think I should be wiring that up through the listener like I am this one.

That being said, we can't use my idea of using that as a reference to help showcase how the metric proposed in this PR fits in. How should we proceed? Is there more we can discuss here first, or should we skip to moving this discussion into the spec to flesh out whether this metric has value?

@reyang when it is time to move the discussion to the spec (assuming we don't just decide here that it's unnecessary), is there anything you think I can prep for this part of that conversation:
"The key is - we don't add something just because it could be useful, we add things when we have good confidence that they will have long term success."

Have I even sold you on the value with my previous statement?

ktmitton: I've been looking at this more and ultimately, even if we decide this metric is worth implementing, I don't think this is the proper way to do it. When I get time later, I'm going to close this pull request and open up a separate issue to discuss the things that came up in regards to this metric and the active_requests one.

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInMetricsListener.cs
```diff
@@ -29,11 +29,14 @@ internal class HttpInMetricsListener : ListenerHandler
...
         private Histogram<double> httpServerDuration;
 
+        private Counter<long> httpServerCompletedRequests;
+
         public HttpInMetricsListener(string name, Meter meter)
             : base(name)
         {
             this.meter = meter;
             this.httpServerDuration = meter.CreateHistogram<double>("http.server.duration", "ms", "measures the duration of the inbound HTTP request");
+            this.httpServerCompletedRequests = meter.CreateCounter<long>("http.server.completed_requests", "requests", "counts the HTTP requests that are completed");
```
cijothomas, line 39: The instrumentation libraries are based on the SemanticConventions (https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/http-metrics.md).
Adding new metric or changing the meaning, would have to go through updating the semantic conventions first. Its okay to have prototypes and gather feedback, but before merging, we need to have it part of spec.

All semantic conventions are currently experimental, and are subject to breaking changes. (that is the main reason the instrumentation libraries are still non-stable).

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInMetricsListener.cs
reyang: Here is my personal thinking:
1. Instead of having an UpDownCounter tracking the active_requests, I would prefer to have two Counters - one tracking the requests that we've received so far, another tracking the requests that we've finished processing (whether they succeeded, failed, timed out or aborted - as long as we know we're done with them).
2. The current metrics semantic convention is still Experimental (not Stable), and the instrumentation library is not Stable, too. We should take the opportunity to drive things towards the good direction.
3. This PR can be marked as "Do Not Merge", similar like what I did in #2593 - to facilitate the spec discussion/design. Or we can merge it knowing that we might change it later (given the state that we're not yet Stable, I think it is probably fine).

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInMetricsListener.cs
ktmitton: yeah, @reyang was pointing me in that direction, too. After talking through it and thinking about this more, I realized early this morning that this PR just won't do. There needs to be a larger conversation on these metrics.

My plan is to get rid of this PR, then open up a new issue on our project to discuss this more, moving the conversation out of this soon-to-be abandoned PR. Once I think we're fleshed this out more on what makes sense, I was planning to open a PR to the spec project.

This definitely turned out more complex than I originally thought...maybe I should stick to updates to the test environments 😆

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInMetricsListener.cs
ktmitton: I think I want to straight up abandon this PR. With what you and @cijothomas, and me getting more comfortable with the spec and thinking through it, I think the end result won't include a single line from this one. You two have made good points on this needing a larger discussion with the spec.

I would like to spin out a new issue in this project to talk it out with you two a little more before I feel comfortable submitting a PR to the spec. Want to make sure I understand it a little better before opening that can of worms 😃

ktmitton: I'm not used to github...not sure how to abandon a PR.

Is it as simple as clicking the Close Pull Request button at the bottom?

reyang: > I'm not used to github...not sure how to abandon a PR.
> Is it as simple as clicking the Close Pull Request button at the bottom?

Yep.

ktmitton: sweet, thanks!

Gonna keep this PR around just until I actually get the new issue entered. I want to do a little more research first, though, to help with that conversation.

PR #2688, opened 2021-11-25
Update AspNetCore Example

ktmitton:
## Changes

Updated the AspNetCore example project to include build instructions and provide a docker-compose approach to spinning up the full environment. Since docker-compose stands up all dependencies, I removed the conditional logic to choose which exporters to use, and just default to all of them. If you run the application outside of docker, this shouldn't be an issue, the application will still run if an exporter dependency wasn't stood up.

Related Issues:
* #2589 
* #2556


ktmitton:
For context, the impetus behind this update was my need to test my changes for #2687. This seemed like the appropriate avenue to test, but I saw it was lacking startup instructions. While writing those, I really liked the approach the microservice example used, so I updated this one to match.

examples/AspNetCore/README.md
```diff
@@ -0,0 +1,68 @@
+# OpenTelemetry Example Application
+
+This project is an example of ASP.NET Core instrumentation for tracing and
+logging.
+
+The following exporters are configured for viewing the traces:
+
+* Console
+* Zipkin
```
reyang, line 9: Minor: consider alphabetical sorting by default, unless there is a reason not to do so.

examples/AspNetCore/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+The following exporters are configured for viewing the metrics:
+
+* Console
+* Otlp
+* Prometheus
+
+## Running the example
+
+Running instances of the following services are required to view exports:
```
reyang, line 21: nit:

```suggestion
Running instances of the following services are required to view the exported data:
```

examples/AspNetCore/README.md
ktmitton: Only reason was the order in which I added them, I'll get this updated!

examples/AspNetCore/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+```shell
+docker-compose up --build
+```
+
+With everything running:
+
+* [Invoke the API](http://localhost:5000/WeatherForecast) to create new data.
+* If you have run Zipkin with default settings:
+  * View your traces with Zipkin [here](http://localhost:9411/zipkin)
```
reyang: Minor: consider replacing `here` with the actual URI (so the readers will see the URI / realize that it is a local endpoint).

```suggestion
  * View your traces with Zipkin by accessing the local endpoint [http://localhost:9411/zipkin](http://localhost:9411/zipkin).
```

examples/AspNetCore/README.md
ktmitton: Updated, along with a couple other lists in this file.

What's general best practice, do I mark the conversation resolved after I apply the change, or will you resolve it after looking at my change?

examples/AspNetCore/prometheus.yml
```diff
@@ -0,0 +1,26 @@
...
+  external_labels:
+      monitor: 'codelab-monitor'
+
+# Load rules once and periodically evaluate them according to the global 'evaluation_interval'.
+rule_files:
+  # - "first.rules"
+  # - "second.rules"
+
+# A scrape configuration containing exactly one endpoint to scrape:
+# Here it's Prometheus itself.
```
reyang: This line is misleading, consider removing it.

General question - do we intend to have a minimal config file? E.g.

https://github.com/open-telemetry/opentelemetry-dotnet/blob/71a6969c1c665642f8fec7afa16fc00b436890a8/examples/Console/TestPrometheusExporter.cs#L39-L49

examples/AspNetCore/README.md
ktmitton: I have no problem with this change, but I was mirroring the convention in the microservice example.

Should I update the documentation for that here, or in a separate PR?

reyang (approved):
LGTM with some suggestions.

examples/AspNetCore/README.md
reyang: I would suggest to keep this PR scoped - if we need to update documentation for another example, cover it in a separate PR.

examples/AspNetCore/README.md
ktmitton: @reyang Updated!

For my own reference, what does `nit` mean?

examples/AspNetCore/README.md
ktmitton: Sounds good. I've updated this per your recommendation, as well as the other instances in this README.

examples/AspNetCore/README.md
reyang: > What's general best practice, do I mark the conversation resolved after I apply the change, or will you resolve it after looking at my change?

Here we're quite flexible, I think people normally resolve the comments without waiting for confirmation (and others can revert it if they have different opinion).
On the opentelemetry-specification repo we've been more careful, normally wait for folks who raised the comment to resolve it.

reyang (approved):
LGTM.

examples/AspNetCore/prometheus.yml
ktmitton: The intent is to provide a very minimal configuration, I was just trying to do the bare-minimum to set up scraping since prometheus is different from other exporters in that way and doesn't "just work". I based my configuration off of https://docs.docker.com/config/daemon/prometheus/, and didn't really look at the comments they put in their example. Once I had linting issues, though, because of some whitespace character, I looked at them and decided the comments weren't necessary, since the point of this project isn't to teach people how to configure prometheus.

Now that I've removed the comments, does this file look good to you?

examples/AspNetCore/README.md
reyang: _nit-picking_ - basically "it's a very tiny/trivial suggestion, you decide if you want to take or ignore it" 😄

examples/AspNetCore/prometheus.yml
reyang: Yep, it looks great! Thanks! 👍

examples/AspNetCore/README.md
ktmitton: Gotcha, thanks!

examples/AspNetCore/README.md
ktmitton: Sounds good, thanks!

examples/AspNetCore/README.md
```diff
@@ -0,0 +1,72 @@
...
+* Otlp
+* Prometheus
+
+## Running the example
+
+Running instances of the following services are required to view the exported
+data:
+
+* Jaeger
+* OTEL Collector
```
pellared, line 25: The official acronym is `OTel`. See: https://github.com/open-telemetry/opentelemetry-specification#acronym

```suggestion
* OTel Collector
```

this comment also applies to other places

pellared (approved):
just one nit comment 😉

examples/AspNetCore/README.md
ktmitton: Thanks for the link, updated!

ktmitton:
Had to step away for a bit, I'll get the merge conflicts resolved this weekend

mic-max:
Awesome!

ktmitton:
Sorry for the delay, this is updated now!

examples/AspNetCore/Startup.cs
```diff
@@ -59,91 +56,54 @@ public void ConfigureServices(IServiceCollection services)
...
-
-                    break;
-            }
+            services.AddOpenTelemetryTracing((builder) => builder
+                .SetResourceBuilder(ResourceBuilder.CreateDefault().AddService(serviceName: "aspnetcore", serviceNamespace: "example"))
+                .AddAspNetCoreInstrumentation()
+                .AddConsoleExporter()
+                .AddZipkinExporter(options =>
+                {
+                    var hostname = Environment.GetEnvironmentVariable("ZIPKIN_HOSTNAME") ?? "localhost";
```
cijothomas, line 65: can we use the official Env variables, and let SDK automatically read it? https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#jaeger-exporter

examples/AspNetCore/README.md
```diff
@@ -0,0 +1,72 @@
+# OpenTelemetry Example Application
+
+This project is an example of ASP.NET Core instrumentation for tracing and
```
cijothomas, line 3: and metrics too.

PR #2691, opened 2021-11-26
feat: Fail fast when insecure channel is not configured for .NET Core 3.1 with gRPC

### LINKED ISSUE #2690: Fail fast when unsecure channel is not configured correctly in .NET 3.1
# Feature Request

Yes, unsecured channels must be configured correctly in .NET 3.1 but is not surfaced at runtime.

Fail fast when creating an OTLP export client when the unsecured channel is not configured correctly.

Add logging.

## Additional Context

Learn more about this requirement in the docs - https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/README.md#special-case-when-using-insecure-channel

tomkerkhove:
Signed-off-by: Tom Kerkhove <kerkhove.tom@gmail.com>

Fixes #2690.

## Changes

Fail fast when unsecure channel is not configured for .NET Core 3.1 which is required for the OTLP exporter to work correctly.

This was not configured correctly and costed me a few hours to remind myself to configure this.

* ~[ ] Design discussion issue #~
* ~[ ] Changes in public API reviewed~


tomkerkhove:
Finalizing running tests locally as well, but feel free to trigger the CI for this as well.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/BaseOtlpExportClient.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        private static void EnsureUnencryptedSupportIsEnabled(OtlpExporterOptions options)
+        {
+            if (options.Endpoint.Scheme.Equals("http", StringComparison.InvariantCultureIgnoreCase))
+            {
+                if (AppContext.TryGetSwitch(
+                        "System.Net.Http.SocketsHttpHandler.Http2UnencryptedSupport", out var unencryptedIsSupported) == false
+                    || unencryptedIsSupported == false)
+                {
+                    throw new NotSupportedException(
```
pellared, line 47: I would not say it is "not supported". How about using `InvalidOperationException` instead?

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/BaseOtlpExportClient.cs
```diff
@@ -0,0 +1,53 @@
...
+            this.Options = options;
+
+#if NETCOREAPP3_1
+            EnsureUnencryptedSupportIsEnabled(options);
+#endif
+        }
+
+        internal OtlpExporterOptions Options { get; }
+
+        private static void EnsureUnencryptedSupportIsEnabled(OtlpExporterOptions options)
```
pellared, line 39: Is it also needed for `BaseOtlpHttpExportClient`? MS doc only says about gRPC.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/BaseOtlpExportClient.cs
tomkerkhove: Sounds good.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/BaseOtlpExportClient.cs
tomkerkhove: As far as I know, it's for both:
https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/IntegrationTests.cs#L35-L36

pellared: This test uses gRPC protocol. Can you double-check it?

tomkerkhove: But the docs say:
> If your application is targeting .NET Core 3.1, and you are using an insecure (HTTP) endpoint, the following switch must be set before adding OtlpExporter.

So that means both scenarios?

tomkerkhove: (from what I've seen, this is required for both scenarios)

pellared: I will double-check it then 😉 ETA 1 hour

pellared: It is indeed needed. I am creating a PR to add `HttpProtobuf` to the integration test 😉

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/BaseOtlpExportClient.cs
```diff
@@ -0,0 +1,53 @@
...
+// limitations under the License.
+// </copyright>
+
+using System;
+using OpenTelemetry.Internal;
+
+namespace OpenTelemetry.Exporter.OpenTelemetryProtocol.Implementation.ExportClient
+{
+    /// <summary>Base class for sending OTLP export requests.</summary>
+    internal abstract class BaseOtlpExportClient
```
pellared, line 23: Personally, I avoid using abstract classes for removing code duplication as much as possible.

I would rather create a `internal static EnsureUnencryptedSupportIsEnabled` in some helper static class and use it in `BaseOtlpGrpcExportClient` and `BaseHttpGrpcExportClient` constructors.

Why?

1. Code executed via inheritance is harder to follow/read/maintain
2. There is no need for polymorphism here (virtual methods etc) so why the use of inheritance?

_You can wait a day or two to check what other approvers/maintainers say. Maybe they disagree with my opinion. Therefore it is not a blocker and I am approving._

pellared (approved):

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/BaseOtlpExportClient.cs
tomkerkhove: Happy to run the tests against both protocols if you want?

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/BaseOtlpExportClient.cs
tomkerkhove: I'm happy to use any of the approaches. How would you call the static class then? ExporterClientValidation.Ensure...?

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/BaseOtlpExportClient.cs
pellared: It occurs that I have a problem making the HTTP protocol to work.
See: https://github.com/open-telemetry/opentelemetry-dotnet/pull/2692

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/BaseOtlpExportClient.cs
pellared: `ExporterClientValidation` works for me

tomkerkhove: Done

tomkerkhove:
This has not been my best PR, sorry @pellared 🤦‍♂️

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/BaseOtlpGrpcExportClient.cs
```diff
@@ -33,6 +33,10 @@ protected BaseOtlpGrpcExportClient(OtlpExporterOptions options)
             Guard.Null(options, nameof(options));
             Guard.InvalidTimeout(options.TimeoutMilliseconds, nameof(options.TimeoutMilliseconds));
 
+#if !NETSTANDARD2_1 && !NET5_0_OR_GREATER
```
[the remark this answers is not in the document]
tomkerkhove: Glad you've asked - I'm wondering the same thing :)
But when I was using that, the condition was never respected, for some reason.

I'll change it back - Maybe the CI here is more lucky.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/BaseOtlpHttpExportClient.cs
```diff
@@ -31,6 +31,10 @@ protected BaseOtlpHttpExportClient(OtlpExporterOptions options, HttpClient httpC
             Guard.Null(options, nameof(options));
             Guard.InvalidTimeout(options.TimeoutMilliseconds, $"{nameof(options)}.{nameof(options.TimeoutMilliseconds)}");
 
+#if !NETSTANDARD2_1 && !NET5_0_OR_GREATER
+            ExporterClientValidation.EnsureUnencryptedSupportIsEnabled(options);
+#endif
```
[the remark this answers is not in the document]
tomkerkhove: Ha that's why the comments were changed. So this is gRPC only?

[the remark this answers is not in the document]
tomkerkhove: From what I have tested it was the case (metrics, not traces).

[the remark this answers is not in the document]
tomkerkhove: I trust your judgement and removed it but let's make sure this is correct.

tomkerkhove:
@pellared As you can see the build directive is not respected in https://github.com/open-telemetry/opentelemetry-dotnet/runs/4350754547?check_suite_focus=true

As per https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives#conditional-compilation you'll notice that `NETCOREAPP3_1` is no longer supported as per .NET 5+ SDK which was why I went with the not standard nor .NET 5 or above approach

pellared:
> @pellared As you can see the build directive is not respected in https://github.com/open-telemetry/opentelemetry-dotnet/runs/4350754547?check_suite_focus=true
> 
> As per https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives#conditional-compilation you'll notice that `NETCOREAPP3_1` is no longer supported as per .NET 5+ SDK which was why I went with the not standard nor .NET 5 or above approach

I see but now it fails on .NET Core 3.1 🤦

tomkerkhove:
Ok so I see the issue here, I think, is that while the Docker Compose file is getting the correct target framework & SDK version (https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/build/docker-compose.netcoreapp3.1.yml#L7-L8), it is using a .NET 6 base image:

> Step 1/12 : ARG SDK_VERSION=6.0
> Step 2/12 : FROM mcr.microsoft.com/dotnet/sdk:6.0-focal AS build

Details: https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/Dockerfile#L6

So that means that the directive that I've switched back to per https://github.com/open-telemetry/opentelemetry-dotnet/pull/2691#discussion_r758076953 is not available and thus not running the new validation. So I'd suggest going back to using this:

`#if !NETSTANDARD2_1 && !NET5_0_OR_GREATER` instead of `#if NETCOREAPP3_1`, or change the image for .NET 3.1 tests (less ideal)

tomkerkhove:
Changed it already but of course the check is now failing for .NET Framework.

Chatted with @pellared offline and he raised a valid concern - Given we use .NET 6 as a base image we might have other .NET Core 3.1 build directives that are being ignored.

What I would propose, @cijothomas, would be to:
1. Use `#if NETCOREAPP3_1` in the code
2. Change the integration test workflow on (https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/.github/workflows/integration.yml#L92) so that there is another Dockerfile being used for .NET Core 3.1 so that it's not using .NET 6 but 3.1 instead as this is required per https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives#conditional-compilation

That way we explicitly target this to only run for .NET Core 3.1, and be able to test this.

alanwest:
> Changed it already but of course the check is now failing for .NET Framework.

This check should not be relevant for .NET Framework. There are two gRPC libraries used by the exporter based on the target it is compiled for. The library used in the context of a .NET Framework application does not require the `Http2UnencryptedSupport` switch.

Can you change the check to `#if !NETFRAMEWORK && !NETSTANDARD2_1 && !NET5_0_OR_GREATER`?

tomkerkhove:
Jup, it is not my intention to throw the exception for .NET Framework. 

Another directive is certainly a possible solution but I'm afraid that all checks with the netcore3 directive are not respected in the integration tests (see https://github.com/open-telemetry/opentelemetry-dotnet/pull/2691#issuecomment-982596037) which is a bigger problem/concern

alanwest:
> Another directive is certainly a possible solution but I'm afraid that all checks with the netcore3 directive are not respected in the integration tests (see [#2691 (comment)](https://github.com/open-telemetry/opentelemetry-dotnet/pull/2691#issuecomment-982596037)) which is a bigger problem/concern

Agreed the integration tests should be modified as you've suggested - that is, when the .NET Core 3.1 app is tested it should be built with 3.1 SDK.

Using `#if NETCOREAPP3_1` would require adding an additional `TargetFramework` here https://github.com/open-telemetry/opentelemetry-dotnet/blob/3aab1244f8797395f12d5ac7b177c26bedd52a0c/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OpenTelemetry.Exporter.OpenTelemetryProtocol.csproj#L3 Maybe I'm still missing something here, but I think adding another target framework can be avoided.

tomkerkhove:
I'm happy to take a stab at this test change in a separate PR!


> Using `#if NETCOREAPP3_1` would require adding an additional `TargetFramework` here
> 
> https://github.com/open-telemetry/opentelemetry-dotnet/blob/3aab1244f8797395f12d5ac7b177c26bedd52a0c/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OpenTelemetry.Exporter.OpenTelemetryProtocol.csproj#L3
> 
> Maybe I'm still missing something here, but I think adding another target framework can be avoided.

Not sure about your opinion here - Are you suggesting to change back and add it as a target framework? Because it sounds like you'd prefer not to add a new target framework.

alanwest:
> it sounds like you'd prefer not to add a new target framework

Yes, I'm not sure adding another target framework is necessary. I'm concerned it might even cause some additional confusion.

I do like what this PR is aiming to solve, but I'm beginning to wonder if conditional compilation is the right way to solve it.

One scenario that may be problematic - let's say I have two components:
1) A `netcoreapp3.1` targeted application, and
2) A `netstandard2.1` library which I've implemented and I've placed all my OTel SDK configuration code in this library.

Say the application references the library which then references the OTLP exporter. What I think will happen here is you'll end up consuming the `netstandard2.1` build of the exporter which would not have your new check in it.

So instead of solving this with conditional compilation, I'm curious if adding a check at runtime for the runtime version being used. I think you could do this with [`System.Environment.Version`](https://docs.microsoft.com/en-us/dotnet/api/system.environment.version?view=net-6.0#remarks).

tomkerkhove:
Great suggestion @alanwest - Thanks! PR updated.

tomkerkhove:
@pellared @alanwest PTAL

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/IntegrationTests.cs
```diff
@@ -68,5 +68,40 @@ public void ExportResultIsSuccess(OtlpExportProtocol protocol, string endpoint)
...
+            {
+                Endpoint = new Uri($"http://{CollectorHostname}:4317"),
+            };
+
+            Assert.Throws<InvalidOperationException>(() => new OtlpTraceExporter(exporterOptions));
+        }
+
+        [Trait("CategoryName", "CollectorIntegrationTests")]
+        [SkipUnlessEnvVarFoundFact(CollectorHostnameEnvVarName)]
+        public void ExportResultIsFailedBecauseEncryptedHttpSupportIsNotConfiguredForGrpc()
```
alanwest, line 93:
```suggestion
        public void ConstructingGrpcExporterFailsWhenHttp2UnencryptedSupportIsNotConfiguredForNetcoreapp31()
```

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/IntegrationTests.cs
```diff
@@ -68,5 +68,40 @@ public void ExportResultIsSuccess(OtlpExportProtocol protocol, string endpoint)
             Assert.Single(delegatingExporter.ExportResults);
             Assert.Equal(ExportResult.Success, delegatingExporter.ExportResults[0]);
         }
+
+#if NETCOREAPP3_1
+        [Trait("CategoryName", "CollectorIntegrationTests")]
+        [SkipUnlessEnvVarFoundFact(CollectorHostnameEnvVarName)]
+        public void ExportResultIsFailedBecauseEncryptedHttpSupportIsDisabledForGrpc()
```
alanwest, line 75:
```suggestion
        public void ConstructingGrpcExporterFailsWhenHttp2UnencryptedSupportIsDisabledForNetcoreapp31()
```

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/IntegrationTests.cs
```diff
@@ -68,5 +68,40 @@ public void ExportResultIsSuccess(OtlpExportProtocol protocol, string endpoint)
...
+            Assert.Throws<InvalidOperationException>(() => new OtlpTraceExporter(exporterOptions));
+        }
+
+        [Trait("CategoryName", "CollectorIntegrationTests")]
+        [SkipUnlessEnvVarFoundFact(CollectorHostnameEnvVarName)]
+        public void ExportResultIsFailedBecauseEncryptedHttpSupportIsNotConfiguredForGrpc()
+        {
+            // Adding the OtlpExporter creates a GrpcChannel.
+            // This switch must be set before creating a GrpcChannel/HttpClient when calling an insecure gRPC service.
+            // We want to fail fast so not setting it to ensure it fails
```
alanwest, line 97:
```suggestion
            // Adding the OtlpExporter creates a GrpcChannel.
            // This test will fail fast because Http2UnencryptedSupport is not configured and is disabled by default.
```

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/ExporterClientValidation.cs
```diff
@@ -0,0 +1,45 @@
...
+            }
+
+            if (options.Endpoint.Scheme.Equals("http", StringComparison.InvariantCultureIgnoreCase))
+            {
+                if (AppContext.TryGetSwitch(
+                        "System.Net.Http.SocketsHttpHandler.Http2UnencryptedSupport", out var unencryptedIsSupported) == false
+                    || unencryptedIsSupported == false)
+                {
+                    throw new InvalidOperationException(
+                        "'System.Net.Http.SocketsHttpHandler.Http2UnencryptedSupport' must be enabled for using HTTP with .NET 3.1");
```
alanwest, line 40:
```suggestion
                        "Calling insecure gRPC services on .NET Core 3.x requires enabling the 'System.Net.Http.SocketsHttpHandler.Http2UnencryptedSupport' switch. See: https://docs.microsoft.com/aspnet/core/grpc/troubleshoot#call-insecure-grpc-services-with-net-core-client");
```

alanwest (approved):
Just some minor suggestions, but looks great!

tomkerkhove:
Ugh here we go again with the integration test while it's exactly the same. Shall I change the base image for .NET 3.1 integration tests @alanwest?

tomkerkhove:
Did a quick test in ce662f4, let's see what it says but should work as far as I see.

cijothomas (dismissed):

pellared:
not sure if you noticed:

```
tests_1           | [xUnit.net 00:00:01.04]     OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests.IntegrationTests.ConstructingGrpcExporterFailsWhenHttp2UnencryptedSupportIsNotConfiguredForNetcoreapp31 [FAIL]
tests_1           |   X OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests.IntegrationTests.ConstructingGrpcExporterFailsWhenHttp2UnencryptedSupportIsNotConfiguredForNetcoreapp31 [4ms]
tests_1           |   Error Message:
tests_1           |    Assert.Throws() Failure
tests_1           | Expected: typeof(System.InvalidOperationException)
tests_1           | Actual:   (No exception was thrown)
```

tomkerkhove:
Jup, it's on my radar but need to check why it's suddenly failing.

alanwest:
> Did a quick test in [ce662f4](https://github.com/open-telemetry/opentelemetry-dotnet/commit/ce662f457197e5ec3621687419ec1b3a1323cdb4), let's see what it says but should work as far as I see.
> 
> Edit: Reverted since it's using language version 10.

Since you're doing a runtime check now in the exporter, how about you do the same in the test code? So, remove the `#IF NETCOREAPP3.1` directives and then only check that an exception is thrown when runtime version is 3.x.

tomkerkhove:
Good idea, I'll use that approach and check with `System.Environment.Version`.

tomkerkhove:
The test is still failing and I will check tomorrow;

alanwest:
> The test is still failing and I will check tomorrow;

If I were to guess it's the enabling of the switch by this test that causes a conflict with the `ConstructingGrpcExporterFailsWhenHttp2UnencryptedSupportIsNotConfiguredForNetcoreapp31` test

https://github.com/open-telemetry/opentelemetry-dotnet/pull/2691/files#diff-ea318d753fea852e284d62944e2e6293f83e44cfb12628b3594685e7f13f1763L36-L44.

It's probably safe to just remove the `ConstructingGrpcExporterFailsWhenHttp2UnencryptedSupportIsNotConfiguredForNetcoreapp31` test. I think the two tests that explicitly enable and disable the switch should cover us.

tomkerkhove:
This must have been my worst PR ever, but hey, it looks like we are ready :)

Thanks for the help!

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/ExporterClientValidationTests.cs
```diff
@@ -0,0 +1,90 @@
...
+
+            try
+            {
+                var options = new OtlpExporterOptions
+                {
+                    Endpoint = new Uri("http://localhost:4173"),
+                };
+
+                AppContext.SetSwitch("System.Net.Http.SocketsHttpHandler.Http2UnencryptedSupport", true);
+                ExporterClientValidation.EnsureUnencryptedSupportIsEnabled(options);
```
pellared, line 39: there is no assertion here so future readers may not understand what this test is for

consider this: https://github.com/xunit/xunit/issues/358#issuecomment-249301674

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/ExporterClientValidationTests.cs
```diff
@@ -0,0 +1,90 @@
...
+        [Fact]
+        public void ExporterClientValidation_FlagIsEnabledForHttpEndpoint()
+        {
+            var initialFlagStatus = this.DetermineInitialFlagStatus();
+
+            try
+            {
+                var options = new OtlpExporterOptions
+                {
+                    Endpoint = new Uri("http://localhost:4173"),
```
pellared, line 35: 1. `http://localhost:4173` how about moving to a contant?
2. however, why it is even needed? it is the default value, right?

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/ExporterClientValidationTests.cs
```diff
@@ -0,0 +1,90 @@
...
+
+namespace OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests
+{
+    public class ExporterClientValidationTests
+    {
+#if NETCOREAPP3_1
+        [Fact]
+        public void ExporterClientValidation_FlagIsEnabledForHttpEndpoint()
+        {
+            var initialFlagStatus = this.DetermineInitialFlagStatus();
```
pellared, line 29: how about doing 

```
var initialFlagStatus = this.DetermineInitialFlagStatus();
```

in constructor and

```
AppContext.SetSwitch("System.Net.Http.SocketsHttpHandler.Http2UnencryptedSupport", initialFlagStatus);
```

in Dispose()?

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/IntegrationTests.cs
```diff
@@ -68,5 +68,31 @@ public void ExportResultIsSuccess(OtlpExportProtocol protocol, string endpoint)
...
+
+            if (Environment.Version.Major == 3)
+            {
+                Assert.Throws<InvalidOperationException>(() => new OtlpTraceExporter(exporterOptions));
+            }
+            else
+            {
+                var exporter = new OtlpTraceExporter(exporterOptions);
+                Assert.NotNull(exporter);
+            }
```
pellared, line 95: how about using `Record.Exception` and have `Assert.IsType<T>` and `Assert.Null` depending on `Environment.Version.Major == 3`  ?

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/OtlpExporterOptionsExtensionsTests.cs
```diff
@@ -96,6 +104,24 @@ public void GetTraceExportClient_SupportedProtocol_ReturnsCorrectExportClient(Ot
...
         }
 
+#if NETCOREAPP3_1
+        [Fact]
+        public void GetTraceExportClient_GetClientForGrpcWithoutUnencryptedFlag_ThrowsException()
+        {
+            // Adding the OtlpExporter creates a GrpcChannel.
+            // This switch must be set before creating a GrpcChannel when calling an insecure HTTP/2 endpoint.
+            // See: https://docs.microsoft.com/aspnet/core/grpc/troubleshoot#call-insecure-grpc-services-with-net-core-client
+            AppContext.SetSwitch("System.Net.Http.SocketsHttpHandler.Http2UnencryptedSupport", false);
```
pellared, line 114: shouldn't the value be reverted as in `ExporterClientValidationTests`

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/OtlpTraceExporterTests.cs
```diff
@@ -372,6 +372,14 @@ public void ToOtlpSpanPeerServiceTest()
         [Fact]
         public void UseOpenTelemetryProtocolActivityExporterWithCustomActivityProcessor()
         {
+            if (Environment.Version.Major == 3)
+            {
+                // Adding the OtlpExporter creates a GrpcChannel.
+                // This switch must be set before creating a GrpcChannel when calling an insecure HTTP/2 endpoint.
+                // See: https://docs.microsoft.com/aspnet/core/grpc/troubleshoot#call-insecure-grpc-services-with-net-core-client
+                AppContext.SetSwitch("System.Net.Http.SocketsHttpHandler.Http2UnencryptedSupport", true);
```
pellared, line 380: shouldn't the value be reverted as in `ExporterClientValidationTests`?

pellared (commented):
Is it not worth creating an issue to remove this feature when .NET Core 3.1 is no longer supported?

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/ExporterClientValidationTests.cs
tomkerkhove: To be explicit as the default value can change but I'll remove this if need be.

tomkerkhove: I'm moving it to a const and keeping it since it's more explicit given we want to test against HTTP and HTTPS endpoints.

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/IntegrationTests.cs
pellared: Yes

tomkerkhove:
PTAL @pellared, @alanwest and @cijothomas

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/ExporterClientValidationTests.cs
```diff
@@ -0,0 +1,67 @@
...
+using Xunit;
+
+namespace OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests
+{
+    public class ExporterClientValidationTests : Http2UnencryptedSupportTests
+    {
+        private const string HttpEndpoint = "http://localhost:4173";
+        private const string HttpsEndpoint = "https://localhost:4173";
+
+#if NETCOREAPP3_1
```
pellared, line 28: https://github.com/open-telemetry/opentelemetry-dotnet/pull/2691#issuecomment-988268390

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/ExporterClientValidationTests.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        [Fact]
+        public void ExporterClientValidation_FlagIsNotEnabledForHttpsEndpoint()
+        {
+            var options = new OtlpExporterOptions
+            {
+                Endpoint = new Uri(HttpsEndpoint),
+            };
+
+            ExporterClientValidation.EnsureUnencryptedSupportIsEnabled(options);
```
pellared, line 63:
```suggestion
            var exception = Record.Exception(() => ExporterClientValidation.EnsureUnencryptedSupportIsEnabled(options));
            Assert.Null(exception);
```

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/OtlpExporterOptionsExtensionsTests.cs
```diff
@@ -96,6 +104,24 @@ public void GetTraceExportClient_SupportedProtocol_ReturnsCorrectExportClient(Ot
             Assert.Equal(expectedExportClientType, exportClient.GetType());
         }
 
+#if NETCOREAPP3_1
```
pellared, line 107: https://github.com/open-telemetry/opentelemetry-dotnet/pull/2691#issuecomment-988268390

pellared (approved):

tomkerkhove:
@pellared are those the final remarks or do you want to do a full review?

I'm happy to keep on changing things but at some point we'll need to just complete the review.

pellared:
I made a full review. That is also why I approved the PR.

tomkerkhove:
> I made a full review. That is also why I approved the PR.

Thanks.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+* Added validation that unsecure channel is configured correctly when using
```
cijothomas, line 5:
```suggestion
* Added validation that insecure channel is configured correctly when using
```

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
...
+  .NET Core 3.1 for gRPC-based exporting.
```
cijothomas, line 6:
```suggestion
  .NET Core 3.x for gRPC-based exporting.
```

alanwest (approved):

tomkerkhove:
Any thoughts on this @cijothomas ?

PR #2692, opened 2021-11-26
Add HTTP protocol test case to OTLP exporter integration tests

pellared:
Add HTTP protocol test case to OTLP exporter integration tests to help with https://github.com/open-telemetry/opentelemetry-dotnet/pull/2691

## Remarks

```
AppContext.SetSwitch("System.Net.Http.SocketsHttpHandler.Http2UnencryptedSupport", true);
```

is only needed for the gRPC OTLP protocol (on .NET 3.1 Core of course).


tomkerkhove:
I don't see it either so it might be the collector :/

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/IntegrationTests.cs
```diff
@@ -24,23 +24,26 @@ namespace OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests
...
+            // Adding the OtlpExporter creates a GrpcChannel/HttpClient.
+            // This switch must be set before creating a GrpcChannel/HttpClient when calling an insecure HTTP/2 endpoint.
             // See: https://docs.microsoft.com/aspnet/core/grpc/troubleshoot#call-insecure-grpc-services-with-net-core-client
             AppContext.SetSwitch("System.Net.Http.SocketsHttpHandler.Http2UnencryptedSupport", true);
 #endif
 
             var exporterOptions = new OtlpExporterOptions
             {
-                Endpoint = new System.Uri($"http://{CollectorEndpoint}"),
+                Endpoint = new System.Uri($"http://{CollectorEndpoint}:{port}"),
```
reyang, line 45: Not introduced by this PR, the term `CollectorEndpoint` seems to be confusing - perhaps endpoint should include the port?

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/IntegrationTests.cs
```diff
@@ -24,23 +24,26 @@ namespace OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests
 {
     public class IntegrationTests
     {
-        private const string CollectorEndpointEnvVarName = "OTEL_EXPORTER_OTLP_ENDPOINT";
-        private static readonly string CollectorEndpoint = SkipUnlessEnvVarFoundFactAttribute.GetEnvironmentVariable(CollectorEndpointEnvVarName);
+        private const string CollectorEndpointEnvVarName = "OTEL_COLLECTOR_HOSTNAME";
```
reyang, line 27: Curious, what's the different between endpoint and hostname? (my understanding is that TCP endpoint contains more information than a hostname - based on experience working with C++ Boost https://www.boost.org/doc/libs/1_47_0/doc/html/boost_asio/reference/ip__tcp/endpoint.html)

reyang (approved):
LGTM in general, perhaps we want to rename `CollectorEndpoint` to `CollectorHostname`/`CollectorHostName`.

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/IntegrationTests.cs
pellared: CollectorHostname will be better

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/IntegrationTests.cs
pellared: That is exactly how I see it

pellared:
> LGTM in general, perhaps we want to rename `CollectorEndpoint` to `CollectorHostname`/`CollectorHostName`.

Agree. The main problem is.... That the test for HttpProtobuf are failing and the reason is unknown. I do not know when I will have time to investigate it further

reyang:
> > LGTM in general, perhaps we want to rename `CollectorEndpoint` to `CollectorHostname`/`CollectorHostName`.
> 
> Agree. The main problem is.... That the test for HttpProtobuf are failing and the reason is unknown. I do not know when I will have time to investigate it further

Is this a priority? I don't know much history about this component, why would people want to use HTTP instead of gRPC? I can imagine the following:
1. Work with HTTP proxies, so it's simpler to penetrate firewalls.
2. Required by some company policies.
3. Less dependency? (I doubt if is this the case today)

pellared:
I heard that it is also more performant and it is the default proposed by spec https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md.

I think the URL for HttpProtobuf is bad in the test code. I will check it on Monday

pellared:
@tomkerkhove There were two issues:
1. a bug in test (missing whole trace endpoint URL)
2. For an unknown reason the collector was not listening on the default HTTP endpoint port at all. I had to explicitly set the endpoint URL in the config. When I have more time I'll check if it is already fixed in https://github.com/open-telemetry/opentelemetry-collector

pellared:
@open-telemetry/dotnet-approvers PTAL

tomkerkhove:
Thanks for sharing and taking a closer look! I've noticed you have removed the HTTP comment remark, is that on purpose?

pellared:
> Thanks for sharing and taking a closer look! I've noticed you have removed the HTTP comment remark, is that on purpose?

Yes, because it is not needed and the integration test confirms it. Take notice:

```csharp
#if NETCOREAPP3_1
            // Adding the OtlpExporter creates a GrpcChannel.
            // This switch must be set before creating a GrpcChannel when calling an insecure HTTP/2 endpoint.
            // See: https://docs.microsoft.com/aspnet/core/grpc/troubleshoot#call-insecure-grpc-services-with-net-core-client
            if (protocol == OtlpExportProtocol.Grpc)
            {
                AppContext.SetSwitch("System.Net.Http.SocketsHttpHandler.Http2UnencryptedSupport", true);
            }
#endif
```

alanwest (approved):
LGTM

cijothomas (approved):

cijothomas:
> > > LGTM in general, perhaps we want to rename `CollectorEndpoint` to `CollectorHostname`/`CollectorHostName`.
> > 
> > 
> > Agree. The main problem is.... That the test for HttpProtobuf are failing and the reason is unknown. I do not know when I will have time to investigate it further
> 
> Is this a priority? I don't know much history about this component, why would people want to use HTTP instead of gRPC? I can imagine the following:
> 
> 1. Work with HTTP proxies, so it's simpler to penetrate firewalls.
> 2. Required by some company policies.
> 3. Less dependency? (I doubt if is this the case today)

^ There are backends which doesn't support Grpc.

PR #2702, opened 2021-11-29
Rename GetCounterSum to GetSum

cijothomas:
Counter is the name of the instrument only and its not required to be part of the `MetricPoint` seen by the Exporters. The data model calls it simply "Sum" without "counter".
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#sums

For Gauge and Histogram instruments, the name of the point also has the instrument name in it. So we can keep current name as is.
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#gauge
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#histogram


reyang (approved):
LGTM.

PR #2704, opened 2021-11-29
Update TraceProviderSdkTest for legacy activities

### LINKED ISSUE #2700: Improve unit and functional tests for propagation in case of legacy Activities
Related to #2659. 

In case of legacy activities (i.e. new Activity()), we run the sampler after the activity is started. When the sampler runs it updates the ActivityTraceFlags property based on the sampling result. However, updating the trace flags on activity does not get reflected on Id if the Id was accessed before sampler runs. This results is incorrect propagation when the propagator uses Id property to set the traceparent header.

vishweshbankwar:
Partially Fixes #2700


test/OpenTelemetry.Tests/Trace/TracerProviderSdkTest.cs
```diff
@@ -648,6 +648,11 @@ public void SdkSamplesLegacyActivityWithAlwaysOnSampler()
             Assert.True(activity.IsAllDataRequested);
             Assert.True(activity.ActivityTraceFlags.HasFlag(ActivityTraceFlags.Recorded));
 
+            // Validating ActivityTraceFlags is not enough as it does not get reflected on
+            // Id, If the Id is accessed before the sampler runs.
+            // https://github.com/open-telemetry/opentelemetry-dotnet/issues/2700
+            Assert.EndsWith("01", activity.Id);
```
reyang, line 654: Maybe adding `-`?

```suggestion
            Assert.EndsWith("-01", activity.Id);
```

reyang (approved):

cijothomas (approved):

PR #2705, opened 2021-11-29
Allow future addition of Histogram MinMax via View

cijothomas:
MinMax is not supported currently, so not exposing any setting for that publicly. But this introduces a base class `HistogramConfiguration` to hold `RecordMinMax` as it'll be applied to `ExplicitBucketHistogram` and `ExponentialHistogram`(when that feature is added). Alternate would be to add `RecordMinMax` to both `ExplicitBucketHistogram` and `ExponentialHistogram`

### DIFF
--- a/src/OpenTelemetry/Metrics/HistogramConfiguration.cs
+++ b/src/OpenTelemetry/Metrics/HistogramConfiguration.cs
@@ -1,4 +1,4 @@
-// <copyright file="ExplicitBucketHistogramConfiguration.cs" company="OpenTelemetry Authors">
+// <copyright file="HistogramConfiguration.cs" company="OpenTelemetry Authors">
 // Copyright The OpenTelemetry Authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


reyang (approved):

alanwest (approved):

src/OpenTelemetry/Metrics/HistogramConfiguration.cs
```diff
@@ -0,0 +1,27 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+
+namespace OpenTelemetry.Metrics
+{
+    public class HistogramConfiguration : MetricStreamConfiguration
```
utpilla, line 19: Do we want to make this an `abstract` class?

cijothomas: We need things like below in the future.
`.AddView(instrumentName: "MyHistogram", new HistogramConfiguration() { Name = "newname", RecordMinMax = false })`

cijothomas:
@utpilla is checking if this can be added after 1.2.0 as well, if yes, will revert to avoid the confusion when one uses this:

```
AddView(instrumentName: "MyCounter", new HistogramConfiguration() { Name = "newname"})  // user may assume this instruments aggregated as Histogram, but we dont support it.
```

PR #2706, opened 2021-11-30
Revert "Allow future addition of Histogram MinMax via View"

utpilla:
Reverts open-telemetry/opentelemetry-dotnet#2705

We want to avoid the situation where a user can add a statement such as:

`.AddView(instrumentName: "MyCounter", new HistogramConfiguration() { Name = "newname"})` 
This will not update the `Counter` to a `Histogram`

Since this will not be a breaking change if added after `1.2.0` release, we can add it later.

Documentation on types of breaking changes: https://docs.microsoft.com/en-us/dotnet/core/compatibility/#types

> ❓ REQUIRES JUDGMENT: Introducing a new base class
>
> A type can be introduced into a hierarchy between two existing types if it doesn't introduce any new abstract members or change the semantics or behavior of existing types. For example, in .NET Framework 2.0, the DbConnection class became a 
> new base class for SqlConnection, which had previously derived directly from Component.

I also manually ran the APICompat tool to ensure that this will not be a breaking change if we decide to add it later on.


reyang (approved):

alanwest (approved):

cijothomas:
@utpilla showed that we still have this issue today with `ExplicitHistogramConfiguration`. So may be okay to leave this in.
And add a validation in AddView wherever feasible to fail-fast.

cijothomas:
> @utpilla showed that we still have this issue today with `ExplicitHistogramConfiguration`. So may be okay to leave this in. And add a validation in AddView wherever feasible to fail-fast.

Merging this PR (i.e reverting the newly added confusion)

PR #2707, opened 2021-11-30
[Do not merge] Refactoring OTLP exporter options

alanwest:
This is experimenting with an idea that makes the OTLP exporter options more hierarchical.

Few goals:

* Make the base `OtlpExporterOptions` class agnostic of whether it's for a trace, metric, or log exporter. This means some properties get marked obsolete.
* Demonstrate how we might handle the signal specific env var configs like `OTLP_EXPORTER_OTLP_METRICS_*` environment variables.


alanwest:
Closing this for now in favor of discussing things on #2716 which is a bit more narrowly scoped.

Issue #2543, opened 2021-10-28
Review and audit of EventSource logging for self-diagnostics


alanwest:
OpenTelemetry .NET uses `EventSource` for internal logging. Each component defines an `EventSource` including the [API](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Api/Internal/OpenTelemetryApiEventSource.cs), [SDK](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/Internal/OpenTelemetrySdkEventSource.cs), and each of the exporters and instrumentation components.

The purpose of this issue is threefold:

1. **Review the usage of these `EventSource`'s for correctness in what gets logged and where it gets logged.**

Example of incorrectness:`BaseExporter.Shutdown` logs a span processor related error when there is a failure. This error should not be about spans or processors:
https://github.com/open-telemetry/opentelemetry-dotnet/blob/635028834c7d435bc64dd64510e4f7b7ec4207a4/src/OpenTelemetry/BaseExporter.cs#L123-L127

2. **Improve coverage of diagnostic logging by identifying gaps.**

Example: Sometimes (but not always) we have TODOs throughout the code highlighting gaps where a diagnostic log may be useful, but has not been implemented yet. Here is an example where an attempt to update a metric has failed:
https://github.com/open-telemetry/opentelemetry-dotnet/blob/db8f0e712555716932e323f5fc7c301b17ec4c11/src/OpenTelemetry/Metrics/AggregatorStore.cs#L292

3. **Aim to make diagnostic logging actionable**

Based on this comment https://github.com/open-telemetry/opentelemetry-dotnet/pull/2525#discussion_r737777141, we should seek to provide actionable guidance when there are errors. For example, if an error is due to misconfiguration, then the log message should give an indication for how to resolve the error.

### Related in-flight work
The following PR was held off until the post-1.0 release. It may be beneficial to land this work prior to the review of self-diagnostics:

https://github.com/open-telemetry/opentelemetry-dotnet/pull/1529

### References
[OpenTelemetry general error handling guidelines/self-diagnostics](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/error-handling.md#self-diagnostics)
[OpenTelemetry .NET self-diagnostics guide](https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/src/OpenTelemetry#troubleshooting)

Issue #2621, opened 2021-11-16
Review Metric TODOs


cijothomas:
Plenty of small TODOs in Metric SDK currently. Opening an issue to track reviewing them and taking action, if needed, before 1st stable release.

cijothomas:
Closing as all TODO's which *MUST* be addressed before 1.2.0 are either addressed, or have separate issues tracking them with 1.2.0 milestone attached to it.

PR #2710, opened 2021-11-30
Add logs when exception occurs in critical paths

cijothomas:
Metrics SDK is blind if an exception occurs today. This PR adds logging on critical `MetricReader` and `MeterProvider` paths (i.e anything which can result in not metrics being exported) and includes exception.ToString() in the log message.

We have this issue https://github.com/open-telemetry/opentelemetry-dotnet/issues/2543 to be addressed. Its a bigger undertaking, so this PR is just a temporary fix to avoid being totally blind if Metric SDK fails.

There'll be couple more follow ups before Metric stable release, to do some cleanups, and add logs in the recording paths as well.

### DIFF
--- a/src/OpenTelemetry/Internal/OpenTelemetrySdkEventSource.cs
+++ b/src/OpenTelemetry/Internal/OpenTelemetrySdkEventSource.cs
@@ -74,6 +74,15 @@ namespace OpenTelemetry.Internal
         }
 
         [NonEvent]
+        public void MetricReaderException(string evnt, Exception ex)
+        {
+            if (this.IsEnabled(EventLevel.Error, EventKeywords.All))
+            {
+                this.MetricReaderException(evnt, ex.ToInvariantString());
+            }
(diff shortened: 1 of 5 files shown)


utpilla (approved):

src/OpenTelemetry/Internal/OpenTelemetrySdkEventSource.cs
```diff
@@ -73,6 +73,15 @@ public void MetricObserverCallbackException(Exception exception)
             }
         }
 
+        [NonEvent]
+        public void MetricReaderException(string evnt, Exception ex)
```
reyang, line 77:
```suggestion
        public void MetricReaderException(string methodName, Exception ex)
```

reyang (approved):

cijothomas:
related https://github.com/open-telemetry/opentelemetry-dotnet/issues/2621

PR #2714, opened 2021-11-30
More error logging in Metric SDK

cijothomas:
More logging, just enough to uncover telemetry loss.

### DIFF
--- a/src/OpenTelemetry/Metrics/AggregatorStore.cs
+++ b/src/OpenTelemetry/Metrics/AggregatorStore.cs
@@ -19,6 +19,7 @@ using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Runtime.CompilerServices;
 using System.Threading;
+using OpenTelemetry.Internal;
 
 namespace OpenTelemetry.Metrics
 {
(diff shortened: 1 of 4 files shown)


cijothomas:
Towards https://github.com/open-telemetry/opentelemetry-dotnet/issues/2621

src/OpenTelemetry/Metrics/AggregatorStore.cs
```diff
@@ -255,15 +262,19 @@ private void UpdateLong(long value, ReadOnlySpan<KeyValuePair<string, object>> t
                 var index = this.FindMetricAggregatorsDefault(tags);
                 if (index < 0)
                 {
-                    // TODO: Measurement dropped due to MemoryPoint cap hit.
+                    if (Interlocked.CompareExchange(ref this.metricCapHitMessageLogged, 1, 0) == 0)
+                    {
+                        OpenTelemetrySdkEventSource.Log.MeasurementDropped(this.name, this.metricPointCapHitMessage, "Modify instrumentation to reduce the number of unique key/value pair combinations. Or use MeterProviderBuilder.SetMaxMetricPointsPerMetricStream to set higher limit.");
```
reyang, line 267: I think there is a general problem - when should we log issue related to the measurement reporting? Considering the logging API is much slower.

reyang: Maybe - instead of reporting these via logs, we can have some built-in metrics (e.g. dropped count).

reyang (approved):

src/OpenTelemetry/Metrics/AggregatorStore.cs
cijothomas: yes agree.
Will address it as future.

PR #2717, opened 2021-12-02
OTLP exporter options separate out MetricReader options

alanwest:
Related to #2552.

This separates out options meant for configuring the MetricReader when using the `AddOltpExporter` extension method.

I think this PR is a better alternative to #2716. However unlike #2716 it does not attempt to add support the for the `OTEL_EXPORTER_OTLP_METRICS_*` environment variables. This could be a follow up PR.

If folks like this approach, I can port it to the Console and In-memory metric exporters.


src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpMetricReaderOptions.cs
```diff
@@ -0,0 +1,42 @@
...
+/// Options for configuring the <see cref="MetricReader"/> used in conjunction with an <see cref="OtlpMetricExporter"/>.
+/// This is meant to be used with the MeterProviderBuilder.AddOtlpExporter() extension method.
+/// </summary>
+public class OtlpMetricReaderOptions
+{
+    /// <summary>
+    /// Gets or sets the AggregationTemporality used for Histogram
+    /// and Sum metrics.
+    /// </summary>
+    public AggregationTemporality AggregationTemporality { get; set; } = AggregationTemporality.Cumulative;
```
alanwest, line 31: Should we rename this to `Temporality` to match `MetricReader.Temporality`?

joaopgrassi (approved):
I think this approach is great!

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpMetricExporterExtensions.cs
```diff
@@ -25,48 +25,90 @@ namespace OpenTelemetry.Metrics
...
+
+        /// <summary>
+        /// Adds <see cref="OtlpMetricExporter"/> to the <see cref="MeterProviderBuilder"/>.
+        /// </summary>
+        /// <param name="builder"><see cref="MeterProviderBuilder"/> builder to use.</param>
+        /// <param name="configure">Exporter and <see cref="MetricReader"/> configuration options.</param>
+        /// <returns>The instance of <see cref="MeterProviderBuilder"/> to chain the calls.</returns>
+        public static MeterProviderBuilder AddOtlpExporter(
+            this MeterProviderBuilder builder,
+            Action<OtlpExporterOptions, OtlpMetricReaderOptions> configure)
```
cijothomas, line 67:
```suggestion
            Action<OtlpExporterOptions, OtlpMetricReaderOptions> configureExporterAndMetricReader)
```

cijothomas (approved):

alanwest:
@CodeBlanch requested I converted this to draft for now. He's working on another approach and we'd like to give people a chance to compare.

examples/Console/TestMetrics.cs
```diff
@@ -66,11 +66,11 @@ internal static object Run(MetricsOptions options)
                 AppContext.SetSwitch("System.Net.Http.SocketsHttpHandler.Http2UnencryptedSupport", true);
 
                 providerBuilder
-                    .AddOtlpExporter(o =>
+                    .AddOtlpExporter((exporterOptions, metricReaderOptions) =>
```
mic-max, line 69: Maybe we could add some use of `exporterOptions` in this example, otherwise the proper code would replace this variable name with an underscore?

mic-max (approved):
I'm liking this option

examples/Console/TestMetrics.cs
alanwest: Added an option to choose gRPC or HTTP

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpMetricExporterExtensions.cs
```diff
@@ -25,48 +25,100 @@ namespace OpenTelemetry.Metrics
...
+            exporterOptions.AppendExportPath(initialEndpoint, OtlpExporterOptions.MetricsExportPath);
 
-            options.TryEnableIHttpClientFactoryIntegration(serviceProvider, "OtlpMetricExporter");
+            var metricExporter = new OtlpMetricExporter(exporterOptions);
 
-            options.AppendExportPath(initialEndpoint, OtlpExporterOptions.MetricsExportPath);
+            if (metricReaderOptions.Temporality == (AggregationTemporality)0)
+            {
+                metricReaderOptions.Temporality = AggregationTemporality.Cumulative;
+            }
```
alanwest, line 110: I'm not thrilled by this way of handling default values. Open to other options.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpMetricExporterExtensions.cs
```diff
@@ -25,46 +25,98 @@ namespace OpenTelemetry.Metrics
...
+
+        /// <summary>
+        /// Adds <see cref="OtlpMetricExporter"/> to the <see cref="MeterProviderBuilder"/>.
+        /// </summary>
+        /// <param name="builder"><see cref="MeterProviderBuilder"/> builder to use.</param>
+        /// <param name="configureExporterAndMetricReader">Exporter and <see cref="MetricReader"/> configuration options.</param>
+        /// <returns>The instance of <see cref="MeterProviderBuilder"/> to chain the calls.</returns>
+        public static MeterProviderBuilder AddOtlpExporter(
+            this MeterProviderBuilder builder,
+            Action<OtlpExporterOptions, MetricReaderOptions> configureExporterAndMetricReader)
```
cijothomas, line 67: `AddOtlpExporter(Action<OtlpExporterOptions> configureExporter, Action <MetricReaderOptions> configureMetricReader)`

^ would this be more easy to use?

src/OpenTelemetry/Metrics/MetricReaderOptions.cs
```diff
@@ -0,0 +1,78 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+
+using OpenTelemetry.Metrics;
+
+namespace OpenTelemetry.Exporter;
```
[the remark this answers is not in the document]
alanwest: Opps, yup!

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpMetricExporterExtensions.cs
alanwest: Maybe? It depends on who you ask 😄. I do not personally have a strong preference either way.

But... I think I may have landed on this solution in part because we're anticipating following this pattern eventually with traces as well. If I recall, the one issue I encountered was that we have

```csharp
public static TracerProviderBuilder AddOtlpExporter(this TracerProviderBuilder builder, Action<OtlpExporterOptions> configure = null)
```

Because of the default parameter for `configure` we can't introduce the following because the methods would be ambiguous.

```csharp
public static TracerProviderBuilder AddOtlpExporter(this TracerProviderBuilder builder, Action<OtlpExporterOptions> configure = null, Action<ProcessorOptions> configureProcessor = null)
```

... though this was a while ago, so I may not have my head on straight about this 😄 - I try it out and refresh my memory.

cijothomas: :) not a blocker.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpMetricExporterExtensions.cs
alanwest: I'm running under the assumption that for the OTLP exporter we're going to want to default to periodic. But for the console exporter we'll want to default to manual, so a uniform default will not work for all exporters. `MetricReaderType == -1` helps distinguish that the user did not explicitly configure this option.

alanwest: Oh, oops, you're commenting on Temporality, yea I guess you're right, Cumulative can be considered the default always right?

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpMetricExporterExtensions.cs
```diff
@@ -25,46 +25,93 @@ namespace OpenTelemetry.Metrics
...
-            options.TryEnableIHttpClientFactoryIntegration(serviceProvider, "OtlpMetricExporter");
+            exporterOptions.TryEnableIHttpClientFactoryIntegration(serviceProvider, "OtlpMetricExporter");
 
-            options.AppendExportPath(OtlpExporterOptions.MetricsExportPath);
+            exporterOptions.AppendExportPath(OtlpExporterOptions.MetricsExportPath);
 
-            var metricExporter = new OtlpMetricExporter(options);
+            var metricExporter = new OtlpMetricExporter(exporterOptions);
+
+            if (metricReaderOptions.MetricReaderType == (MetricReaderType)(-1))
```
cijothomas, line 105: lets take this as a follow up to decide 
1. What should be the MetricReaderType default
2. Would that be same for all exporters (Console , OTLP). Or do we want separate defaults.

Not blocking this PR.

cijothomas (approved):

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpMetricExporterExtensions.cs
alanwest: #2921

PR #2735, opened 2021-12-08
Adding MinMax to Histograms

### LINKED ISSUE #2560: Histogram aggregation should support Min Max
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#explicit-bucket-histogram-aggregation
It should be collected by default, with the ability to turn off.

mic-max:
Fixes #2560

## Changes

Histograms will now record the minimum and maximum measurement values.


mic-max:
To do:
- Remove all TODO comments from PR files
- Review public API addition
- Review the two added `MetricPointValueStorage` instances in `MetricPoint` for perf
- Do I need a test case for a histogram that has recorded no values? Or will that not even be sent to the exporter?
  - If so, what would be good defaults for min and max, -Inf and +Inf ?
- Add a way to configure the recording of Min and Max - true by default

cijothomas:
https://github.com/open-telemetry/opentelemetry-dotnet/pull/2718 changes the layout, so this will be affected.

cijothomas:
https://github.com/open-telemetry/opentelemetry-dotnet/pull/2705/files - need to bring this back as well. (it was reverted)

cijothomas:
https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/Metrics/MetricType.cs also see this class.

mic-max:
Failing tests:
- TestHistogramToOltpMetric
- ViewToProduceCustomHistogramBound

mic-max:
Note: `MetricTypeExtensions.cs` changes are not used by anything

src/OpenTelemetry/Metrics/MetricPoint.cs
```diff
@@ -86,90 +108,185 @@ public DateTimeOffset EndTime
...
-                return this.primaryValue.SnapshotAsLong;
+                this.ThrowNotSupportedMetricTypeException(nameof(this.GetHistogramMin));
             }
-            else
+
+            if (this.runningValue.AsLong == 0)
             {
-                throw new NotSupportedException($"{nameof(this.GetHistogramCount)} is not supported for this metric type.");
+                throw new InvalidOperationException("No values have been recorded yet.");
             }
```
mic-max, line 246: What should the behaviour be when calling `GetHistogramMin()` if no values have been recorded?

src/OpenTelemetry/Metrics/MetricPoint.cs
```diff
@@ -86,90 +108,185 @@ public DateTimeOffset EndTime
...
-                return this.secondaryValue.SnapshotAsDouble;
+                this.ThrowNotSupportedMetricTypeException(nameof(this.GetHistogramMax));
             }
-            else
+
+            if (this.runningValue.AsLong == 0)
             {
-                throw new NotSupportedException($"{nameof(this.GetHistogramSum)} is not supported for this metric type.");
+                throw new InvalidOperationException("No values have been recorded yet.");
             }
```
mic-max, line 269: What should the behaviour be when calling GetHistogramMax() if no values have been recorded?

src/OpenTelemetry/Metrics/MetricPoint.cs
```diff
@@ -258,9 +375,42 @@ internal void Update(double number)
...
-                            this.histogramBuckets.BucketCounts[i]++;
+                            this.runningValue.AsLong++;
+                            this.histogramBuckets.RunningSum += number;
+                            this.histogramBuckets.RunningBucketCounts[i]++;
+                        }
+
+                        break;
+                    }
+
+                case AggregationType.HistogramWithMinMax:
```
mic-max, line 387: This case shares a lot of code with `case AggregationType.Histogram`
Should we try to dedupe it?

src/OpenTelemetry/Metrics/MetricPoint.cs
```diff
@@ -258,9 +375,42 @@ internal void Update(double number)
...
+                            if (this.runningValue.AsLong == 1)
+                            {
+                                this.histogramBuckets.RunningMin = number;
+                                this.histogramBuckets.RunningMax = number;
+                            }
+                            else
+                            {
+                                this.histogramBuckets.RunningMin = Math.Min(this.histogramBuckets.RunningMin, number);
+                                this.histogramBuckets.RunningMax = Math.Max(this.histogramBuckets.RunningMax, number);
+                            }
```
mic-max, line 414: If we had a default value for `this.histogramBuckets.RunningMin` and `this.histogramBuckets.RunningMax` then the `Math.Min` and `Math.Max` respectively might not work as intended, unless the defaults were `-Inf` and `+Inf`

Related to what should happen when calling `GetHistogramMin()` and `GetHistogramMax()` without recording any records.

src/OpenTelemetry/Metrics/MetricPoint.cs
```diff
@@ -395,20 +545,52 @@ internal void TakeSnapshot(bool outputDelta)
...
-                            for (int i = 0; i < this.histogramBuckets.BucketCounts.Length; i++)
+                            for (int i = 0; i < this.histogramBuckets.RunningBucketCounts.Length; i++)
                             {
-                                this.histogramBuckets.AggregatedBucketCounts[i] = this.histogramBuckets.BucketCounts[i];
+                                this.histogramBuckets.SnapshotBucketCounts[i] = this.histogramBuckets.RunningBucketCounts[i];
                                 if (outputDelta)
                                 {
-                                    this.histogramBuckets.BucketCounts[i] = 0;
+                                    this.histogramBuckets.RunningBucketCounts[i] = 0;
+                                }
```
mic-max, line 563: Would there be a performance benefit to doing the following?

```csharp
for (int i = 0; i < this.histogramBuckets.RunningBucketCounts.Length; i++)
{
    this.histogramBuckets.SnapshotBucketCounts[i] = this.histogramBuckets.RunningBucketCounts[i];
}

if (outputDelta)
{
    for (int i = 0; i < this.histogramBuckets.RunningBucketCounts.Length; i++)
    {
        this.histogramBuckets.RunningBucketCounts[i] = 0;
    }
}
```

What about using `Array.Copy` and `Array.Clear`?

src/OpenTelemetry/.publicApi/net461/PublicAPI.Unshipped.txt
```diff
@@ -20,7 +20,11 @@ OpenTelemetry.Metrics.HistogramBuckets.Enumerator.Current.get -> OpenTelemetry.M
 OpenTelemetry.Metrics.HistogramBuckets.Enumerator.Enumerator() -> void
 OpenTelemetry.Metrics.HistogramBuckets.Enumerator.MoveNext() -> bool
 OpenTelemetry.Metrics.HistogramBuckets.GetEnumerator() -> OpenTelemetry.Metrics.HistogramBuckets.Enumerator
+OpenTelemetry.Metrics.HistogramConfiguration
```
mic-max, line 23: Review API additions

src/OpenTelemetry/Metrics/MetricTypeExtensions.cs
```diff
@@ -62,6 +63,12 @@ public static bool IsHistogram(this MetricType self)
             return self.HasFlag(METRIC_TYPE_HISTOGRAM);
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static bool IsHistogramWithMinMax(this MetricType self)
+        {
+            return self.HasFlag(METRIC_TYPE_HISTOGRAM_WITH_MIN_MAX);
+        }
```
mic-max, line 70: This method has zero references.

src/OpenTelemetry/Metrics/MetricPoint.cs
mic-max: |          Method | size |      Mean |     Error |    StdDev |
|---------------- |----- |----------:|----------:|----------:|
| ManualArrayZero |    5 |  5.098 ns | 0.0694 ns | 0.0649 ns |
|      ArrayClear |    5 |  8.213 ns | 0.0422 ns | 0.0395 ns |
| ManualArrayZero |   10 |  8.610 ns | 0.0791 ns | 0.0701 ns |
|      ArrayClear |   10 | 11.941 ns | 0.1635 ns | 0.1530 ns |
| ManualArrayZero |   20 | 15.596 ns | 0.1303 ns | 0.1219 ns |
|      ArrayClear |   20 | 12.075 ns | 0.1288 ns | 0.1205 ns |
| ManualArrayZero |   50 | 41.655 ns | 0.3566 ns | 0.3335 ns |
|      ArrayClear |   50 | 18.182 ns | 0.3551 ns | 0.4861 ns |
| ManualArrayZero |  100 | 82.170 ns | 1.6516 ns | 1.5449 ns |
|      ArrayClear |  100 | 34.746 ns | 0.5619 ns | 0.5256 ns |

Keep existing version since the default size 10 is faster

src/OpenTelemetry/Metrics/HistogramConfiguration.cs
```diff
@@ -0,0 +1,26 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+
+namespace OpenTelemetry.Metrics;
+
+public class HistogramConfiguration : MetricStreamConfiguration
```
utpilla, line 19: Do we need a separate class for this? Why not just add `RecordMinMax` to `ExplicitBucketHistogramConfiguration`?

src/OpenTelemetry/Metrics/MetricPoint.cs
```diff
@@ -61,6 +61,10 @@ public struct MetricPoint
             {
                 this.histogramBuckets = new HistogramBuckets(histogramExplicitBounds);
             }
+            else if (this.aggType == AggregationType.HistogramWithMinMax)
```
utpilla, line 64: Simplify this by adding an `OR` condition in the previous `if` check.

src/OpenTelemetry/Metrics/HistogramConfiguration.cs
cijothomas: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#exponentialhistogram MinMax is possible for ExponentialHistogram as well. So i think a base class is right place to capture MinMax.

src/OpenTelemetry/Metrics/HistogramConfiguration.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    /// <summary>
+    /// Gets or sets a value indicating whether Min, Max
+    /// should be collected.
+    /// </summary>
+    internal bool RecordMinMax { get; set; } = true;
```
cijothomas, line 25:
```suggestion
    public bool RecordMinMax { get; set; } = true;
```

src/OpenTelemetry/Metrics/MetricReaderExt.cs
```diff
@@ -124,9 +124,19 @@ internal List<Metric> AddMetricsListWithViews(Instrument instrument, List<Metric
                         Metric metric;
                         var metricDescription = metricStreamConfig?.Description ?? instrument.Description;
                         string[] tagKeysInteresting = metricStreamConfig?.TagKeys;
-                        double[] histogramBucketBounds = (metricStreamConfig is ExplicitBucketHistogramConfiguration histogramConfig
-                            && histogramConfig.Boundaries != null) ? histogramConfig.Boundaries : null;
-                        metric = new Metric(instrument, this.Temporality, metricName, metricDescription, this.maxMetricPointsPerMetricStream, histogramBucketBounds, tagKeysInteresting);
+                        double[] histogramBucketBounds = null;
+                        bool histogramRecordMinMax = false;
```
cijothomas, line 128:
```suggestion
                        bool histogramRecordMinMax = true;
```

src/OpenTelemetry/Metrics/Metric.cs
```diff
@@ -97,6 +98,11 @@ public sealed class Metric
                 {
                     aggType = AggregationType.HistogramSumCount;
                 }
+                else if (histogramRecordMinMax)
```
cijothomas, line 101: MinMax should be independent of bounds.
You'll need some logic like 
Start with MetricType.Histogram
If RecordMinMax, set it to HistogramWithMinMax

For aggregationtype (its internal usage only):
We'll likely need:
Histogram  -- count,sum,buckets
HistogramSumCount - count,sum
HistogramMinMax - count, sum, minmax, buckets
HistogramSumCountMinMax - count,sum,minmax

(would be good to rename the internal Aggtype to better reflect intent)

src/OpenTelemetry/Metrics/AggregationType.cs
```diff
@@ -62,5 +62,10 @@ internal enum AggregationType
         /// Histogram with sum, count only.
         /// </summary>
         HistogramSumCount = 7,
+
+        /// <summary>
+        /// Histogram with minimum and maximum.
+        /// </summary>
+        HistogramWithMinMax = 8,
```
mic-max, line 69: Rework this, main histogram is only sum and count

- Histogram
- HistogramWithBuckets
- HistogramWithBucketsAndMinMax
- HistogramWithMinMax

something like this

src/OpenTelemetry/Metrics/MetricType.cs
```diff
@@ -69,5 +69,10 @@ public enum MetricType : byte
         /// Histogram.
         /// </summary>
         Histogram = 0x40,
+
+        /// <summary>
+        /// Histogram with minimum and maximum.
+        /// </summary>
+        HistogramWithMinMax = 0x50,
```
mic-max, line 76: TODO

src/OpenTelemetry/Metrics/Metric.cs
```diff
@@ -92,15 +93,10 @@ public sealed class Metric
...
-                {
-                    aggType = AggregationType.HistogramSumCount;
-                }
-                else
-                {
-                    aggType = AggregationType.Histogram;
-                }
+                // Use histogramBounds and histogramRecordMinMax to determine the aggregation type
+                aggType = histogramBounds != null && histogramBounds.Length == 0
+                    ? (histogramRecordMinMax ? AggregationType.HistogramSumCountMinMax : AggregationType.HistogramSumCount)
```
cijothomas, line 98: the MetricType should be updated to reflect minmax, right?

src/OpenTelemetry/Metrics/MetricPoint.cs
```diff
@@ -213,14 +218,65 @@ public long GetHistogramCount()
...
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public double GetHistogramMax()
+        {
+            if (this.aggType != AggregationType.HistogramMinMax &&
+                this.aggType != AggregationType.HistogramSumCountMinMax)
+            {
+                this.ThrowNotSupportedMetricTypeException(nameof(this.GetHistogramMax));
+            }
+
+            if (this.runningValue.AsLong == 0)
```
cijothomas, line 272: would this condition ever occur?

mic-max: Here's the test case for it https://github.com/open-telemetry/opentelemetry-dotnet/pull/2735/files#diff-a3e99c5f40d4f5df6f97f7c944d7c4b560a70b57fc299321c5b656757cf0d76cR651

Are you asking that when we export and there are no items then we do nothing so having this check is redundant? In that case I think so, I could replace with a debug.assert, since the exporters haven't been modified to include the min/max values yet.

src/OpenTelemetry/Metrics/Metric.cs
mic-max: I'm confused as to what the purpose of `MetricType` is compared to using the `AggregationType`

cijothomas: `MetricType` is public interface to allow Exporters to know what type of Metric they are receiving.

AggregationType is just internal to let aggregatos know "what to do with the incoming measurement".

mic-max: Okay, added in the latest commit. Let me know how that looks 👍🏻

mic-max:
Could this be reopened please :)

jdaigle:
Comment to keep this open. I'm excited and eager to see this changed merged and released.

src/OpenTelemetry/Metrics/HistogramBuckets.cs
```diff
@@ -24,21 +26,56 @@ namespace OpenTelemetry.Metrics
...
         internal HistogramBuckets(double[] explicitBounds)
         {
             this.ExplicitBounds = explicitBounds;
+            this.findHistogramBucketIndex = this.FindBucketIndexLinear;
+            if (explicitBounds != null && explicitBounds.Length >= DefaultBoundaryCountForBinarySearch)
+            {
+                this.bucketLookupTreeRoot = ConstructBalancedBST(explicitBounds, 0, explicitBounds.Length);
+                this.findHistogramBucketIndex = this.FindBucketIndexBinary;
+
+                static BucketLookupNode ConstructBalancedBST(double[] values, int min, int max)
```
cijothomas, line 60: this does not seem related to adding MinMax to histograms.

mic-max: Ya, that's from the binary search PR I mentioned in the description that this branch was created from as it touches similar code.

arianmotamedi:
Can this be reopened? Is it being worked on?

utpilla:
@mic-max Please resolve the merge conflicts for getting this PR ready for review.

ejsmith:
Would love to have this. For me, I'm recording queue processing times and having the average is alright, but what I really want to know is what the max time is.

ejsmith:
I guess there wouldn't be any way to get something like the 99th percentile from a histogram even with this change. I was reading docs on OpenTelemetry specs and it seems like doing percentiles isn't really going to be supported.

cijothomas:
> I guess there wouldn't be any way to get something like the 99th percentile from a histogram even with this change. I was reading docs on OpenTelemetry specs and it seems like doing percentiles isn't really going to be supported.

Histogram exports buckets, allowing backends to calculate percentiles.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/MetricItemExtensions.cs
```diff
@@ -256,6 +257,11 @@ internal static OtlpMetrics.Metric ToOtlpMetric(this Metric metric)
                             AddAttributes(metricPoint.Tags, dataPoint.Attributes);
                             dataPoint.Count = (ulong)metricPoint.GetHistogramCount();
                             dataPoint.Sum = metricPoint.GetHistogramSum();
+                            if (metric.MetricType == MetricType.HistogramWithMinMax)
```
utpilla, line 260: Add a CHANGELOG entry for this.

src/OpenTelemetry/Metrics/MetricPoint.cs
```diff
@@ -565,5 +544,95 @@ internal void TakeSnapshot(bool outputDelta)
...
+                    }
+
+                    if (outputDelta)
+                    {
+                        this.runningValue.AsLong = 0;
+                        this.histogramBuckets.RunningSum = 0;
+
+                        if (hasMinMax)
+                        {
+                            this.histogramBuckets.RunningMin = 0;
```
utpilla, line 610: In case of delta temporality, `RunninMin` and `RunningMax` should be set `double.PositiveInfinity` and `double.NegativeInfinity` respectively.

src/OpenTelemetry/Metrics/MetricPoint.cs
```diff
@@ -565,5 +544,95 @@ internal void TakeSnapshot(bool outputDelta)
         {
             throw new NotSupportedException($"{methodName} is not supported for this metric type.");
         }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private void HistogramUpdate(double number, bool hasBuckets, bool hasMinMax)
```
utpilla, line 549: I think we should investigate having a single method that would account for all scenarios vs having a dedicated method for each of them (without these extra if-checks)

The `Update` and `Snapshot` methods for Histogram use a spin-lock mechanism. We want the time taken by these methods during the lock to be as less as possible. Moreover, `Update` method is in the hot path.

We need Stress Test results for before and after this change.

ejsmith:
Do we need help getting this completed and merged? It seems like an important feature for otel metrics.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/MetricItemExtensions.cs
utpilla: This needs the CHANGELOG of OTLP Exporter to be updated.

src/OpenTelemetry/Metrics/MetricPoint.cs
mic-max: Added this perf improvement to the following issue: https://github.com/open-telemetry/opentelemetry-dotnet/issues/3428

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/MetricItemExtensions.cs
mic-max: Ahh, I didn't realise you meant OTLP Exporter CHANGELOG originally. Added.

mic-max:
Let me know when you want to merge this and I will resolve the conflicts, not going to waste my time fixing preemptively just for new conflicts to come up

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Recording `Min` and `Max` for histograms is configurable, enabled by default.
```
cijothomas, line 5: a good changelog for the SDK, but not describing the change for  OTLP Exporter!

src/OpenTelemetry/Metrics/MetricPoint.cs
```diff
@@ -187,7 +189,10 @@ internal MetricPointStatus MetricPointStatus
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public readonly long GetHistogramCount()
         {
-            if (this.aggType != AggregationType.Histogram && this.aggType != AggregationType.HistogramSumCount)
+            if (this.aggType != AggregationType.Histogram &&
```
cijothomas, line 192: we can do a faster checking by checking if the bit for histogram is set.
for future PR :)

cijothomas: nvm, not easy for aggregationtype.

src/OpenTelemetry/Metrics/MetricPoint.cs
cijothomas: this might be a minor regression from existing scenario where configured just SUM and Count. Can you run the main vs this branch with MinMax=false, and share numbers?
I'll leave this comment open, but wont block PR merge as we can still address perf regression, if any, as follow ups

src/OpenTelemetry/Metrics/MetricReaderExt.cs
```diff
@@ -156,8 +156,8 @@ internal List<Metric> AddMetricsListWithViews(Instrument instrument, List<Metric
                     }
                     else
                     {
-                        Metric metric;
-                        metric = new Metric(metricStreamIdentity, this.GetAggregationTemporality(metricStreamIdentity.InstrumentType), this.maxMetricPointsPerMetricStream, metricStreamIdentity.HistogramBucketBounds, metricStreamIdentity.TagKeys);
+                        bool histogramRecordMinMax = (metricStreamConfig as HistogramConfiguration)?.RecordMinMax ?? true;
+                        Metric metric = new(metricStreamIdentity, this.GetAggregationTemporality(metricStreamIdentity.InstrumentType), this.maxMetricPointsPerMetricStream, metricStreamIdentity.HistogramBucketBounds, metricStreamIdentity.TagKeys, histogramRecordMinMax);
```
cijothomas, line 160: @alanwest Could you take a close look at this portion? I think the bool recordMinMax should be part of the MetricIdentity.

test/OpenTelemetry.Instrumentation.Http.Tests/HttpClientTests.cs
```diff
@@ -139,7 +139,7 @@ public async Task HttpOutCallsAreCollectedSuccessfullyAsync(HttpTestData.HttpOut
 
                 var metric = requestMetrics[0];
                 Assert.NotNull(metric);
-                Assert.True(metric.MetricType == MetricType.Histogram);
+                Assert.True(metric.MetricType == MetricType.HistogramWithMinMax);
```
cijothomas, line 142: Since we are changing the MetricType, any exporters which rely on this will be broken. For example: https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/MetricItemExtensions.cs#L241
will stop exporting Histograms.

We need to hotfix the OTLP Exporter to use the IsHistogram() check instead. (prometheus already handles this correctly. So is ConsoleExporter..)

cijothomas (changes requested):
The PR looks good except for 2 things:
1. MinMwx should be part of Identity. Must be fixed in this PR.
2. Other exporters will break with this change. The fix is in other exporters, but blocking PR to prevent accidental merge until exporters are hotfixed.  See https://github.com/open-telemetry/opentelemetry-dotnet/pull/2735/files#r989476122

cijothomas:
> The PR looks good except for 2 things:
> 
> 1. MinMwx should be part of Identity. Must be fixed in this PR.
> 2. Other exporters will break with this change. The fix is in other exporters, but blocking PR to prevent accidental merge until exporters are hotfixed.  See https://github.com/open-telemetry/opentelemetry-dotnet/pull/2735/files#r989476122

if we keep the metric type same, then it wont break existing exporters. and can use Nan/Infinity to indicate lack of min/max

src/OpenTelemetry/Metrics/MetricPoint.cs
mic-max: I'll just unfold these functions 👍

test/OpenTelemetry.Instrumentation.Http.Tests/HttpClientTests.cs
mic-max: We will continue to use `MetricType.Histogram` to not break exporters, and only when min and max are not the default +inf and -inf respectively will they be exported.

src/OpenTelemetry/Metrics/MetricReaderExt.cs
alanwest: Yes, for sure. So something like the following should produce the duplicate metric instrument warning above

```csharp
            var histogram = meter.CreateHistogram<double>("MyHistogram");
                .AddMeter(meter.Name)
                .AddView(histogram.Name, histogramConfigurationWithMinMax)
                .AddView(histogram.Name, histogramConfigurationWithoutMinMax)
                .AddInMemoryExporter(exportedItems)
                .Build();
```

That is each view will produce a metric stream which will be duplicated by name.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/MetricItemExtensions.cs
```diff
@@ -257,6 +257,18 @@ internal static OtlpMetrics.Metric ToOtlpMetric(this Metric metric)
                             dataPoint.Count = (ulong)metricPoint.GetHistogramCount();
                             dataPoint.Sum = metricPoint.GetHistogramSum();
 
+                            var max = metricPoint.GetHistogramMax();
+                            if (!double.IsNegativeInfinity(max))
```
cijothomas, line 261: double.IsNaN instead?

alanwest: Would that work? Reading [this](https://learn.microsoft.com/en-us/dotnet/api/system.double.isnan?view=net-7.0#remarks)

> [IsNaN](https://learn.microsoft.com/en-us/dotnet/api/system.double.isnan?view=net-7.0) returns false if a [Double](https://learn.microsoft.com/en-us/dotnet/api/system.double?view=net-7.0) value is either [PositiveInfinity](https://learn.microsoft.com/en-us/dotnet/api/system.double.positiveinfinity?view=net-7.0) or [NegativeInfinity](https://learn.microsoft.com/en-us/dotnet/api/system.double.negativeinfinity?view=net-7.0). To test for these values, use the [IsInfinity](https://learn.microsoft.com/en-us/dotnet/api/system.double.isinfinity?view=net-7.0), [IsPositiveInfinity](https://learn.microsoft.com/en-us/dotnet/api/system.double.ispositiveinfinity?view=net-7.0), and [IsNegativeInfinity](https://learn.microsoft.com/en-us/dotnet/api/system.double.isnegativeinfinity?view=net-7.0) methods.

alanwest: Though, I will admit that `!double.IsNegativeInfinity(max)` does hurt my 🧠 just a little. Maybe we should push the implementation details of negative/positive infinity meaning "unset" somewhere else.

Maybe something like

```csharp
var maxIsSet = metricPoint.TryGetHistogramMax(out var max)
```

src/OpenTelemetry/Metrics/MetricReaderExt.cs
mic-max: Moved `HistogramRecordMinMax` into `MetricStreamIdentity`, let me know if that looks good

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/MetricItemExtensions.cs
alanwest: Another thought... would it ever be the case where max is set but min is not? Maybe it would be nicer for an exporter author to be able to write something like


```csharp
if (metricPoint.HasMinMax()) // This new method does the negative/positive infinity check
{
    dataPoint.Min = metricPoint.GetHistogramMin();
    dataPoint.Max = metricPoint.GetHistogramMax();
}
```

src/OpenTelemetry/Metrics/MetricStreamIdentity.cs
```diff
@@ -36,6 +36,7 @@ public MetricStreamIdentity(Instrument instrument, MetricStreamConfiguration met
             this.MetricStreamName = $"{this.MeterName}.{this.MeterVersion}.{this.InstrumentName}";
             this.TagKeys = metricStreamConfiguration?.CopiedTagKeys;
             this.HistogramBucketBounds = (metricStreamConfiguration as ExplicitBucketHistogramConfiguration)?.CopiedBoundaries;
+            this.HistogramRecordMinMax = (metricStreamConfiguration as HistogramConfiguration)?.RecordMinMax ?? true;
```
alanwest, line 39: Should include in the hash as well.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/MetricItemExtensions.cs
mic-max: I liked the `TryGetHistogramMax` way so switched it to that.

cijothomas: **would it ever be the case where max is set but min is not**

^ I cannot think of such case. I like the HasMinMax() slightly better. TryGetHistogramMax/Min() is okay as well. Will unblock PR, and we can adjust if needed

src/OpenTelemetry/Metrics/MetricPoint.cs
```diff
@@ -223,14 +231,61 @@ internal MetricPointStatus MetricPointStatus
...
+        /// </summary>
+        /// <param name="min">Output variable for the minimum value.</param>
+        /// <remarks>
+        /// Applies to <see cref="MetricType.Histogram"/> metric type.
+        /// </remarks>
+        /// <returns>A minimum value exists.</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public bool TryGetHistogramMin(out double min)
+        {
+            if (this.aggType != AggregationType.HistogramMinMax &&
```
cijothomas, line 257: we should not throw for any of the histogram type

test/OpenTelemetry.Tests/Metrics/MetricViewTests.cs
```diff
@@ -572,6 +572,77 @@ public void ViewToProduceCustomHistogramBound()
...
+
+            var metricPoints = new List<MetricPoint>();
+            foreach (ref readonly var mp in exportedItems[0].GetMetricPoints())
+            {
+                metricPoints.Add(mp);
+            }
+
+            var histogramPoint = metricPoints[0];
+
+            var ex = Assert.Throws<NotSupportedException>(() => histogramPoint.TryGetHistogramMin(out var min));
```
cijothomas, line 639: this should not throw. Exports has to call TryGetHistogramMin for MetricType=Histogram, irrespective of it has min/max. The way exporters determine if min/max available is from the return value of TryGet.

cijothomas (changes requested):
https://github.com/open-telemetry/opentelemetry-dotnet/pull/2735/files#r993773985 
See the above.

Rest all looks good. We definitely need to add more tests for validation, esp. the delta/cumulate, and across multiple snapshots. But it can be a future PR.

src/OpenTelemetry/Metrics/MetricPoint.cs
mic-max: Okay, I shall return false instead

test/OpenTelemetry.Tests/Metrics/MetricViewTests.cs
mic-max: Yes, thank you!

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/MetricItemExtensions.cs
mic-max: I think I will go with the HasMinMax way now after evaluating more carefully. It also seems like this is what Java did. It is complex to use just the current value to determine if a Min/Max is present :P

src/OpenTelemetry/Metrics/MetricPoint.cs
```diff
@@ -223,14 +231,60 @@ internal MetricPointStatus MetricPointStatus
...
+        /// Should be used before
+        /// <see cref="GetHistogramMin"/> and <see cref="GetHistogramMax"/> to
+        /// ensure a valid value is returned.
+        /// </summary>
+        /// <returns>A minimum and maximum value exist.</returns>
+        public bool HasMinMax()
+        {
+            return (this.aggType == AggregationType.HistogramMinMax ||
+                   this.aggType == AggregationType.HistogramSumCountMinMax) &&
+                   !(this.histogramBuckets.RunningMin == double.PositiveInfinity &&
```
cijothomas, line 284: do we actually need to check for +inf,-inf? I think just aggType check is sufficient.

mic-max: This is making sure we have actually seen a value. Would there be an easier way to check that like histogram count > 0?
Actually, checking this is actually not correct since both running min and max will be reset to +inf and -inf after snapshot is called. Any suggestions?

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/MetricItemExtensions.cs
alanwest: > Is recording an infinity value something we should support?

[The spec address this](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#handle-all-normal-values):

> Implementations SHOULD NOT incorporate non-normal values (i.e., +Inf, -Inf, and NaNs) into the sum, min, and max fields, because these values do not map into a valid bucket.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+* OTLP histogram data points will now include `Min` and `Max` values when
+  they are present.
+  ([#2735](https://github.com/open-telemetry/opentelemetry-dotnet/pull/2735))
```
alanwest, line 7:
```suggestion
* Min-Max brought to you by Mic-Max.
  ([#2735](https://github.com/open-telemetry/opentelemetry-dotnet/pull/2735))
```

src/OpenTelemetry/Metrics/MetricPoint.cs
alanwest: > do we actually need to check for +inf,-inf? I think just aggType check is sufficient.

Ah interesting, yes I think you're right. I was originally thinking +inf/-inf would be the meaning of `HasMinMax`, but now that the logic is pushed down then just `AggregationType` can be used.

> Actually, checking this is actually not correct since both running min and max will be reset to +inf and -inf after snapshot is called. Any suggestions?

I think this is ok. From the standpoint of an exporter, all the histogram metric points it received on export that have min/max _should_ be set by that point. I can't think of a scenario where it would still remain unset.

cijothomas: ^ is correct in my understanding too. @mic-max please remove the +inf and -inf check, and i think we can merge after that.

cijothomas (approved):

cijothomas:
Merging. @alanwest please take a final look and share any issues you see!

src/OpenTelemetry/Metrics/MetricPoint.cs
```diff
@@ -223,14 +231,58 @@ internal MetricPointStatus MetricPointStatus
...
+        }
+
+        /// <summary>
+        /// Gets whether the histogram has a valid Min and Max value.
+        /// Should be used before
+        /// <see cref="GetHistogramMin"/> and <see cref="GetHistogramMax"/> to
+        /// ensure a valid value is returned.
+        /// </summary>
+        /// <returns>A minimum and maximum value exist.</returns>
+        public bool HasMinMax()
```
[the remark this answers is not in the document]
mic-max: Ya that's valid, we could change to `HistogramHasMinMax` maybe?
Please share any other nits I can take care of before inclusion in its initial release :)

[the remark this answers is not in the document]
cijothomas: the initial release train just left the station 🤣 . Lets include it in next beta. (its okay to break between betas)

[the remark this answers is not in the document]
CodeBlanch: `HistogramHasMinMax` works for me!

> Please share any other nits I can take care of before inclusion in its initial release :)

I was going to suggest maybe folding it all into a single method like...

`bool TryGetHistogramMinMaxValues(out double histogramMinimumValue, out double histogramMaximumValue)`

That might be faster (probably needs inlining) with less API surface but not totally sure. Would need to measure it to see feel free to try if you are interested.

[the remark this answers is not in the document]
CodeBlanch: @cijothomas Hey I was just looking at this API as part of some other work. I think we should change it! Reason being, as it is right now...

```csharp
        public double GetHistogramMin()
        {
            return this.histogramBuckets.SnapshotMin;
        }
```

...that is a NullRef just waiting to be thrown 🤣 All the other methods in here validate the type before accessing the value.

This would be much safer and have comparable perf...

```csharp
public bool TryGetHistogramMinMaxValues(out double histogramMinimumValue, out double histogramMaximumValue)
{
   if (this.aggType == AggregationType.HistogramMinMax ||
                   this.aggType == AggregationType.HistogramSumCountMinMax)
   {
      Debug.Assert(this.histogramBuckets != null, "histogramBuckets was null");

      histogramMinimumValue = this.histogramBuckets.SnapshotMin;
      histogramMaximumValue = this.histogramBuckets.SnapshotMax;
      return true;
   }

   histogramMinimumValue = 0;
   histogramMaximumValue = 0;
   return false;
}
```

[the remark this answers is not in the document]
utpilla: Instead of returning `true` or `false`, could we follow the pattern used for other public methods in MetricPoint such as `GetSumLong` etc. of throwing a not supported metric type exception when called for a non-applicable metric type?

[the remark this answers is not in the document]
cijothomas: this is slightly different. Its possible to be histogram, but min/max is not there (as min max is optional).

Issue #2729, opened 2021-12-07
How to build solution


sumit84mca:
I cloned the repository but getting a lot of errors. I spent hours to resolve them. Now I am having these errors and dont know how to resolve these.
![image](https://user-images.githubusercontent.com/18229166/145035587-ee23c6c7-c97b-4d92-be15-bf76c4f06828.png)

**Describe your environment.**
I am using Visual Studio 2019 on Window 10 Professional

**What are you trying to achieve?**

The solution/project file should get build. 
If there is any guide to build the solution/project please let me know.

tomkerkhove:
Did you check if you've met the prerequisites?

https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/CONTRIBUTING.md#development-environment

sumit84mca:
> Did you check if you've met the prerequisites?
> 
> https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/CONTRIBUTING.md#development-environment

Yes I have all the prerequisite.
![image](https://user-images.githubusercontent.com/18229166/145045983-17ded9c2-900a-4f6e-a717-6468d7a329c3.png)
![image](https://user-images.githubusercontent.com/18229166/145049272-82d63856-c7a7-4dd5-82f5-bd2b41abb740.png)

tomkerkhove:
Looks like you don't have .NET 4.6.x installed, can you add that please?

sumit84mca:
I have it
![image](https://user-images.githubusercontent.com/18229166/145076602-764d8316-fa3b-41ef-8b72-ed8557ead759.png)

cijothomas:
The contributing guide may be outdated. If you are using Visual Studio, could you try with Visual Studio 2022? 2019 don't support .net 6.0 I believe.

sumit84mca:
Its failing in VS2022 as well.

PR #2745, opened 2021-12-13
Update CONTRIBUTING.md with Visual Studio 2022

mic-max:
Fixes https://github.com/open-telemetry/opentelemetry-dotnet/issues/2729#issuecomment-988139167

## Changes

State that Visual Studio 2022 is required.
The minimum requirements on all platforms still looks good (git and ,NET Core 3.1+) but maybe we should also callout the other versions of .NET we target `net5.0` and `net6.0` so if they contributor wants to build and test all versions it's easier.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -48,7 +48,7 @@ On all platforms, the minimum requirements are:
 
 ### Linux or MacOS
 
-* Visual Studio for Mac or Visual Studio Code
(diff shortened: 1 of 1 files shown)


cijothomas (approved):

cijothomas:
> but maybe we should also callout the other versions of .NET we target net5.0 and net6.0 so if they contributor wants to build and test all versions it's easier.

+1 to the above.

PR #2754, opened 2021-12-17
Histogram Record Method Benchmark

mic-max:
## Changes

Benchmarking the Histogram.Record function for various number of explicitly provided bounds for both long and double values.

|                 Method | BoundCount |     Mean |    Error |   StdDev | Allocated |
|----------------------- |----------- |---------:|---------:|---------:|----------:|
|   HistogramLongHotPath |         10 | 55.44 ns | 0.211 ns | 0.187 ns |         - |
| HistogramDoubleHotPath |         10 | 55.69 ns | 0.129 ns | 0.107 ns |         - |
|   HistogramLongHotPath |         20 | 57.71 ns | 0.297 ns | 0.278 ns |         - |
| HistogramDoubleHotPath |         20 | 58.10 ns | 0.117 ns | 0.110 ns |         - |
|   HistogramLongHotPath |         50 | 65.21 ns | 0.356 ns | 0.333 ns |         - |
| HistogramDoubleHotPath |         50 | 66.34 ns | 0.381 ns | 0.356 ns |         - |
|   HistogramLongHotPath |        100 | 79.49 ns | 0.804 ns | 0.753 ns |         - |
| HistogramDoubleHotPath |        100 | 85.77 ns | 0.947 ns | 0.840 ns |         - |

Aside: The reason this was added was since I was investigating whether changing the bucket finding algorithm from a linear one to a binary search one would improve the speed. It only did at ~100 boundaries. So opted not to make the change. Here is the results of that benchmark which is not included in this PR.

Current: Linear
|                           Method | BoundCount |     Mean |    Error |   StdDev | Allocated |
|--------------------------------- |----------- |---------:|---------:|---------:|----------:|
| HistogramLongCustomBoundsHotPath |         10 | 42.78 ns | 0.159 ns | 0.141 ns |         - |
| HistogramLongCustomBoundsHotPath |         20 | 45.02 ns | 0.158 ns | 0.148 ns |         - |
| HistogramLongCustomBoundsHotPath |         50 | 53.00 ns | 0.242 ns | 0.215 ns |         - |
| HistogramLongCustomBoundsHotPath |        100 | 69.03 ns | 1.399 ns | 1.555 ns |         - |

Proposed: Binary
|                           Method | BoundCount |     Mean |    Error |   StdDev | Allocated |
|--------------------------------- |----------- |---------:|---------:|---------:|----------:|
| HistogramLongCustomBoundsHotPath |         10 | 51.56 ns | 0.161 ns | 0.151 ns |         - |
| HistogramLongCustomBoundsHotPath |         20 | 57.33 ns | 0.156 ns | 0.146 ns |         - |
| HistogramLongCustomBoundsHotPath |         50 | 66.93 ns | 0.219 ns | 0.205 ns |         - |
| HistogramLongCustomBoundsHotPath |        100 | 72.99 ns | 0.219 ns | 0.183 ns |         - |


src/OpenTelemetry/Metrics/MetricPoint.cs
```diff
@@ -247,12 +247,18 @@ internal void Update(double number)
...
-                        for (i = 0; i < this.histogramBuckets.ExplicitBounds.Length; i++)
+                        if (double.IsNaN(number))
                         {
-                            // Upper bound is inclusive
-                            if (number <= this.histogramBuckets.ExplicitBounds[i])
+                            i = this.histogramBuckets.ExplicitBounds.Length;
+                        }
+                        else
+                        {
+                            // If number is not a histogram bound, then the result will be zero
```
alanwest, line 256: I think you meant "will be negative" right?

src/OpenTelemetry/Metrics/MetricPoint.cs
```diff
@@ -247,12 +247,18 @@ internal void Update(double number)
                 case AggregationType.Histogram:
                     {
                         int i;
-                        for (i = 0; i < this.histogramBuckets.ExplicitBounds.Length; i++)
+                        if (double.IsNaN(number))
```
alanwest, line 250: Interesting... 🤔  makes me wonder if `NaN` should be an error case that we just log instead of recording.

src/OpenTelemetry/Metrics/MetricPoint.cs
mic-max: Yes, I did!

src/OpenTelemetry/Metrics/MetricPoint.cs
mic-max: Ya, I'm not sure. I was trying to fix the Prometheus exporter test for `PrometheusSerializerTests.HistogramNaN` since I was trying to compare by using the `==` operator and also `double.NaN` is greater than all number, I didn't check the result of `double.NaN > double.Infinity` though 😅

alanwest: Yea, doesn't make sense to me that we should increment the largest bucket... why not the smallest? or even a random one?

Interestingly, I checked the Java SDK, and it turns out they too increment the last bucket. Wonder if this is just accidental because I don't see anything in the spec.

Though I looked at that prometheus test and noted that `NaN` is a valid value in the [prometheus format](https://github.com/prometheus/docs/blob/main/content/docs/instrumenting/exposition_formats.md#comments-help-text-and-type-information).

Maybe something @reyang could shed some light on.

reyang: The [SDK Specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#numerical-limits-handling) gives certain flexibility to the SDK implementations for perf consideration.

Prometheus Exporter spec (which is still Experimental) hasn't covered `NaN` histogram yet, further clarifications are needed.

src/OpenTelemetry/Metrics/MetricPoint.cs
```diff
@@ -247,12 +247,18 @@ internal void Update(double number)
...
                         {
-                            // Upper bound is inclusive
-                            if (number <= this.histogramBuckets.ExplicitBounds[i])
+                            i = this.histogramBuckets.ExplicitBounds.Length;
+                        }
+                        else
+                        {
+                            // If `number` is not a histogram bound, the result will be negative
+                            // The bitwise complement of the returned value is the insertion index
+                            i = Array.BinarySearch(this.histogramBuckets.ExplicitBounds, number);
```
reyang, line 258: For array with small number of bounds, this extra invocation could turn out to be slower.

We need to have a perf benchmark to show the numbers.

Refer to the prom-client and see how low/high buckets are handled differently:
* https://github.com/prom-client-net/prom-client/blob/master/src/Prometheus.Client/HistogramImpl/HistogramHighBucketsStore.cs
* https://github.com/prom-client-net/prom-client/blob/master/src/Prometheus.Client/HistogramImpl/HistogramLowBucketsStore.cs

reyang (changes requested):
https://github.com/open-telemetry/opentelemetry-dotnet/pull/2754#discussion_r777629039

src/OpenTelemetry/Metrics/MetricPoint.cs
alanwest: If perf is what drives our decision then it would make sense to try and eliminate the `if (double.IsNaN(number))` check.

If we settle on using `Array.BinarySearch` then `~Array.BinarySearch(buckets, double.NaN)` equals 0. Seems it would be slightly more efficient to increment the first bucket in this case.

alanwest: Though, the bummer of aggregating a `NaN` value is that it not only messes up one of the bucket counts, it also clobbers the sum.

reyang: > If perf is what drives our decision then it would make sense to try and eliminate the `if (double.IsNaN(number))` check.

+1

This is also called out in the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#numerical-limits-handling):

_It is unspecified how the SDK should handle the input limits. The SDK authors MAY **leverage/follow the language runtime behavior for better performance, rather than perform a check on each value coming from the API**._

reyang: > Though, the bummer of aggregating a `NaN` value is that it not only messes up one of the bucket counts, it also clobbers the sum.

In addition to `NaN`, other values could result in `NaN` as well, for example (note that currently histogram does not support negative value):

```csharp
histogram.Record(double.PositiveInfinity);
histogram.Record(double.NegativeInfinity);
```

src/OpenTelemetry/Metrics/MetricPoint.cs
mic-max: I'll write a self-contained benchmark that compares the existing linear search implementation with the binary search one and run it with a range of boundary counts.

I agree that linear is likely faster for a small number of boundaries, but eventually the binary search will outperform it. Then I can a boundary number where anything less uses linear, and anything more uses binary.

I'll generate the random numbers so that each index gets an equal amount of finds.
Ex: If the sorted array is [ 20, 30, 40 ] then the random numbers will be generated like `random.next(10, 50)`
An array of all the results from performing 10,000 checks should looks similar to:

| Value | Count |
|-------|--------|
| 0 | 2500 |
| 1 | 2500 |
| 2 | 2500 |
| 3 | 2500 |

test/Benchmarks/Metrics/HistogramBenchmarks.cs
```diff
@@ -0,0 +1,100 @@
...
+        public void Cleanup()
+        {
+            this.meter?.Dispose();
+            this.provider?.Dispose();
+        }
+
+        [Benchmark]
+        public void HistogramLongHotPath()
+        {
+            this.histogramLong?.Record(this.random.Next(MaxValue));
```
reyang, line 91: This doesn't seem to be thread-safe?

reyang (approved):

utpilla:
We should add benchmarks for recording histogram values with different number of dimensions similar to how we have benchmarks for `Counter`: https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/test/Benchmarks/Metrics/MetricsBenchmarks.cs

If it becomes too complicated, we could either remove the `double` Histogram benchmarks or have them in a separate class.

test/Benchmarks/Metrics/HistogramBenchmarks.cs
```diff
@@ -0,0 +1,104 @@
...
+| HistogramDoubleHotPath |        100 | 86.93 ns | 0.915 ns | 0.856 ns |         - |
+*/
+
+namespace Benchmarks.Metrics
+{
+    [MemoryDiagnoser]
+    public class HistogramBenchmarks
+    {
+        private const int MaxValue = 1000;
+        private static readonly string[] DimensionValues = new string[] { "DimVal1", "DimVal2", "DimVal3", "DimVal4", "DimVal5", "DimVal6", "DimVal7", "DimVal8", "DimVal9", "DimVal10" };
```
cijothomas, line 46: it doesn't look like we use tags at all?

test/Benchmarks/Metrics/HistogramBenchmarks.cs
mic-max: Fixed

test/Benchmarks/Metrics/HistogramBenchmarks.cs
mic-max: Accidentally included this when considering implementing @utpilla 's suggestion, but will defer to a separate later PR

test/Benchmarks/Metrics/HistogramBenchmarks.cs
```diff
@@ -0,0 +1,103 @@
...
+using System.Collections.Generic;
+using System.Diagnostics.Metrics;
+using System.Threading;
+using BenchmarkDotNet.Attributes;
+using OpenTelemetry;
+using OpenTelemetry.Exporter;
+using OpenTelemetry.Metrics;
+using OpenTelemetry.Tests;
+
+/*
```
cijothomas, line 27: consider adding the full text to see which framework/hardware was used.

mic-max: added!

alanwest (approved):

cijothomas (approved):

reyang:
@mic-max I learned from some users of OpenTelemetry .NET that in some extreme cases they might be using more than 1k buckets. I guess we might want to revisit this and see if certain https://github.com/open-telemetry/opentelemetry-dotnet/pull/2754#discussion_r777629039 (e.g. I guess `140` might be a good magical number to trigger binary search).

PR #2756, opened 2021-12-21
HTTP retries and redirects instrumentation

denisivan0v:
## Changes
This PR brings an implementation for HTTP retries and redirects proposed in [opentelemetry-specification #2078](https://github.com/open-telemetry/opentelemetry-specification/pull/2078).

It addresses a scenario which is in the scope for bringing the existing HTTP semantic conventions for tracing to an initial stable state, see related [otep #174](https://github.com/open-telemetry/oteps/pull/174).


test/OpenTelemetry.Instrumentation.Http.Tests/HttpWebRequestActivitySourceTests.netfx.cs
```diff
@@ -653,19 +661,37 @@ public async Task TestSecureTransportRetryFailureRequest(string method)
                 Assert.True(ex is HttpRequestException);
             }
 
-            // We should have one Start event and one Stop event with an exception.
-            Assert.Equal(2, eventRecords.Records.Count());
-            Assert.Equal(1, eventRecords.Records.Count(rec => rec.Key == "Start"));
-            Assert.Equal(1, eventRecords.Records.Count(rec => rec.Key == "Stop"));
+            // We should have two Start events and twos Stop events, both with an exceptions.
```
mic-max, line 664:
```suggestion
            // We should have two Start events and twos Stop events, both with exceptions.
```

src/OpenTelemetry.Instrumentation.Http/HttpRequestMessageContextPropagation.cs
```diff
@@ -34,6 +34,11 @@ internal static class HttpRequestMessageContextPropagation
 
         internal static Action<HttpRequestMessage, string, string> HeaderValueSetter => (request, name, value) =>
         {
+            if (request.Headers.Contains(name))
+            {
+                request.Headers.Remove(name);
+            }
```
mic-max, line 40:
```suggestion
request.Headers.Remove(name);
```

[Remove](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.headers.httpheaders.remove?view=net-6.0#system-net-http-headers-httpheaders-remove(system-string)) already handles it not being present in the `request.Headers`

denisivan0v: Thanks!

src/OpenTelemetry.Instrumentation.Http/Implementation/HttpHandlerDiagnosticListener.cs
```diff
@@ -88,14 +90,32 @@ public override void OnStartActivity(Activity activity, object payload)
...
             {
-                // this request is already instrumented, we should back off
+                // Handling request retry or redirect.
+                var retryCount = 1;
+                if (request.Properties.TryGetValue(SpanAttributeConstants.RetryCountKey, out var previousRetryCount))
+                {
+                    retryCount = (int)previousRetryCount + 1;
+                }
+
+                // Suppressing activity started by HttpClient DiagnosticsHandler.
```
cijothomas, line 103: this section is a bit unclear without some examples and explanation.
some questions:
1. Does httpClient diagnostichandler created a single Activity only, or does it create one for every retry?
2. Which one are we stopping here.
3. We are creating a new activity here - is it supposed to be the child of the original http client activity or is it expected to be its sibling (i.e same parent, likely that of the Asp.Net Core Request)
4. Why do we explicitly set Activity.Current in line 108?

PR #2761, opened 2021-12-24
refactor Resource::SanitizeValue method to be more readable

kiapanahi:
Code clean-up for the `SanitizeValue` method on the `Resource` type.

### DIFF
--- a/src/OpenTelemetry/Resources/Resource.cs
+++ b/src/OpenTelemetry/Resources/Resource.cs
@@ -102,7 +102,7 @@ namespace OpenTelemetry.Resources
                 sanitizedKey = attribute.Key;
             }
 
-            object sanitizedValue = SanitizeValue(attribute.Value, sanitizedKey);
+            var sanitizedValue = SanitizeValue(attribute.Value, sanitizedKey);
             return new KeyValuePair<string, object>(sanitizedKey, sanitizedValue);
         }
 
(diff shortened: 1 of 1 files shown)


mic-max:
LGTM thanks!

src/OpenTelemetry/Resources/Resource.cs
```diff
@@ -102,61 +102,74 @@ public Resource Merge(Resource other)
...
+                string[] => value,
+                bool[] => value,
+                double[] => value,
+                long[] => value,
+                int => Convert.ToInt64(value),
+                short => Convert.ToInt64(value),
+                float => Convert.ToDouble(value, System.Globalization.CultureInfo.InvariantCulture),
+                int[] v => SanitizationUtils.ConvertoToInt64Array(v),
+                short[] v => SanitizationUtils.ConvertoToInt64Array(v),
+                float[] v => SanitizationUtils.ConvertToDoubleArray(v),
```
mic-max, line 128: Instead of these helper methods consider using the built-in [Array.ConvertAll ](https://docs.microsoft.com/en-us/dotnet/api/system.array.convertall?view=net-6.0) method

It might be useful to also check how the performance of this is compared to the current solution

src/OpenTelemetry/Resources/Resource.cs
```diff
@@ -102,61 +102,74 @@ public Resource Merge(Resource other)
...
+                return result;
             }
 
-            if (value is float[])
+            public static double[] ConvertToDoubleArray(float[] value)
             {
-                double[] convertedArr = new double[((float[])value).Length];
-                int i = 0;
-                foreach (float val in (float[])value)
+                var convertedArr = new double[value.Length];
```
mic-max, line 163:
```suggestion
                var result = new double[value.Length];
```

To be consistent with the other similar methods.

mic-max (changes requested):

src/OpenTelemetry/Resources/Resource.cs
kiapanahi: fixed via 83f5baaf

src/OpenTelemetry/Resources/Resource.cs
kiapanahi: good point. thanks :)

from the [docs](https://docs.microsoft.com/en-us/dotnet/api/system.array.convertall?view=net-6.0#remarks):

> This method is an O(n) operation, where n is the Length of array.

but I'll create a benchmark.net test checking allocations on these 3 options. (considering the code prior to this PR as the baseline)

kiapanahi: based on [this gist](https://gist.github.com/kiapanahi/1f74a39a6917474eb5a3e5e17a14bfa1) and the following results from a simple benchmark, I guess its safe to assume that `Array.ConvertAll` is pretty performant

``` ini

Intel Xeon E-2176M CPU 2.70GHz, 1 CPU, 12 logical and 6 physical cores
.NET SDK=6.0.101
  [Host]     : .NET 6.0.1 (6.0.121.56705), X64 RyuJIT
  DefaultJob : .NET 6.0.1 (6.0.121.56705), X64 RyuJIT


```
|                                           Method |      Mean |     Error |    StdDev | Ratio |
|------------------------------------------------- |----------:|----------:|----------:|------:|
| &#39;Convert int[] to long[] using traditional loop&#39; | 54.787 ms | 1.0834 ms | 2.1886 ms |  1.00 |
| &#39;Convert int[] to long[] using Array.ConvertAll&#39; |  5.425 ms | 0.1179 ms | 0.3420 ms |  0.10 |

mic-max: Awesome! Looks great to me 😄 
One small nit comment is you could convert `ConvertToDoubleUsingIvariantCulture` into a lambda function since it is only used in that one place 👍

kiapanahi: You're absolutely right.
Although the static local function will be almost always aggressively inlined by JIT, the one-liner lambda notation is much easier on the eye.

cijothomas (approved):

Issue #2371, opened 2021-09-17
Follow up on AggregationTemporality vs. AggregationModes


reyang:
https://github.com/open-telemetry/opentelemetry-dotnet/pull/2357#discussion_r710704718

reyang:
> In this context, does something like `new SumAggregation { AggregationMode = AggregationModes.DeltaTemporality }` feel better to folks than `new SumAggregation { AggregationTemporality = AggregationTemporality.Delta`?

I was struggling with this, too! And that's why I didn't send out the PR yet 😆

reyang:
> `new SumAggregation { Temporality = AggregationTemporality.Delta`, feels most https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#temporality

How do you think about this https://github.com/open-telemetry/opentelemetry-specification/issues/1200?

reyang:
I guess we can close it since it's not going anywhere, and the current names seem to be fine.

PR #2771, opened 2022-01-07
Added sanity check for the setter of PrometheusExporter's HTTP listeners.

Yun-Ting:
## Changes
Added sanity check for PrometheusExporter's HTTP listener.

### DIFF
--- a/src/OpenTelemetry.Exporter.Prometheus/PrometheusExporterOptions.cs
+++ b/src/OpenTelemetry.Exporter.Prometheus/PrometheusExporterOptions.cs
@@ -29,13 +29,14 @@ namespace OpenTelemetry.Exporter
         internal Func<DateTimeOffset> GetUtcNowDateTimeOffset = () => DateTimeOffset.UtcNow;
 
         private int scrapeResponseCacheDurationMilliseconds = 10 * 1000;
+        private IReadOnlyCollection<string> httpListenerPrefixes = new string[] { "http://*:80/" };
 
 #if NETCOREAPP3_1_OR_GREATER
-        /// <summary>
-        /// Gets or sets a value indicating whether or not an http listener
-        /// should be started. Default value: False.
-        /// </summary>
-        public bool StartHttpListener { get; set; }
+    /// <summary>
+    /// Gets or sets a value indicating whether or not an http listener
+    /// should be started. Default value: False.
+    /// </summary>
+    public bool StartHttpListener { get; set; }
 #else
         /// <summary>
         /// Gets or sets a value indicating whether or not an http listener
@@ -48,7 +49,23 @@ namespace OpenTelemetry.Exporter
         /// Gets or sets the prefixes to use for the http listener. Default
         /// value: http://*:80/.
         /// </summary>
-        public IReadOnlyCollection<string> HttpListenerPrefixes { get; set; } = new string[] { "http://*:80/" };
+        public IReadOnlyCollection<string> HttpListenerPrefixes
+        {
+            get => this.httpListenerPrefixes;
+            set
+            {
+                foreach (string inputUri in value)
+                {
+                    if (!(Uri.TryCreate(inputUri, UriKind.Absolute, out var uri) &&
+                        (uri.Scheme == Uri.UriSchemeHttp || uri.Scheme == Uri.UriSchemeHttps)))
+                    {
+                        throw new ArgumentException("Prometheus server path should be a valid uri with http/https protocol.");
+                    }
+
+                    this.httpListenerPrefixes = value;
+                }
+            }
+        }
 
         /// <summary>
         /// Gets or sets the path to use for the scraping endpoint. Default value: /metrics.
--- a/src/OpenTelemetry.Exporter.Prometheus/PrometheusExporterOptions.cs
+++ b/src/OpenTelemetry.Exporter.Prometheus/PrometheusExporterOptions.cs
@@ -29,7 +29,7 @@ namespace OpenTelemetry.Exporter
         internal Func<DateTimeOffset> GetUtcNowDateTimeOffset = () => DateTimeOffset.UtcNow;
 
         private int scrapeResponseCacheDurationMilliseconds = 10 * 1000;
-        private IReadOnlyCollection<string> httpListenerPrefixes = new string[] { "http://*:9184/" };
+        private IReadOnlyCollection<string> httpListenerPrefixes = new string[] { "http://*:80/" };
 
 #if NETCOREAPP3_1_OR_GREATER
         /// <summary>
--- a/src/OpenTelemetry.Exporter.Prometheus/PrometheusExporterOptions.cs
+++ b/src/OpenTelemetry.Exporter.Prometheus/PrometheusExporterOptions.cs
@@ -54,6 +54,8 @@ namespace OpenTelemetry.Exporter
             get => this.httpListenerPrefixes;
             set
             {
+                _ = value ?? throw new ArgumentNullException(nameof(this.httpListenerPrefixes));
+
                 foreach (string inputUri in value)
                 {
                     if (!(Uri.TryCreate(inputUri, UriKind.Absolute, out var uri) &&
@@ -61,11 +63,11 @@ namespace OpenTelemetry.Exporter
                     {
                         throw new ArgumentException(
                             "Prometheus server path should be a valid URI with http/https scheme.",
-                            nameof(this.HttpListenerPrefixes));
+                            nameof(this.httpListenerPrefixes));
                     }
-
-                    this.httpListenerPrefixes = value;
                 }
+
+                this.httpListenerPrefixes = value;
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/OpenTelemetry.Exporter.Prometheus/PrometheusExporterOptions.cs
```diff
@@ -29,6 +29,7 @@ public class PrometheusExporterOptions
         internal Func<DateTimeOffset> GetUtcNowDateTimeOffset = () => DateTimeOffset.UtcNow;
 
         private int scrapeResponseCacheDurationMilliseconds = 10 * 1000;
+        private IReadOnlyCollection<string> httpListenerPrefixes = new string[] { "http://*:80/" };
```
reyang, line 32: Now I start to wonder, do we want to have `80` as the default or `9090` (or depending on whether we are exposing it via `PrometheusExporterMiddleware` vs. `PrometheusExporterHttpServer`)?

reyang: Doesn't have to be solved in this PR though.

src/OpenTelemetry.Exporter.Prometheus/PrometheusExporterOptions.cs
```diff
@@ -48,7 +49,23 @@ public class PrometheusExporterOptions
...
+        {
+            get => this.httpListenerPrefixes;
+            set
+            {
+                foreach (string inputUri in value)
+                {
+                    if (!(Uri.TryCreate(inputUri, UriKind.Absolute, out var uri) &&
+                        (uri.Scheme == Uri.UriSchemeHttp || uri.Scheme == Uri.UriSchemeHttps)))
+                    {
+                        throw new ArgumentException("Prometheus server path should be a valid uri with http/https protocol.");
```
reyang, line 62:
```suggestion
                        throw new ArgumentException("Prometheus server path should be a valid URI with http/https scheme.", nameof(HttpListenerPrefixes));
```

reyang: I don't know if "Prometheus server path" is a good name, and I cannot think of a better one at this moment.

test/OpenTelemetry.Exporter.Prometheus.Tests/PrometheusExporterHttpServerTests.cs
```diff
@@ -105,5 +105,39 @@ public async Task PrometheusExporterHttpServerIntegration()
                 "^# TYPE counter_double counter\ncounter_double{key1='value1',key2='value2'} 101.17 \\d+\n$".Replace('\'', '"'),
                 await response.Content.ReadAsStringAsync().ConfigureAwait(false));
         }
+
+        [Theory]
+        [InlineData("http://http.example.com")]
+        [InlineData("https://https.example.com")]
```
reyang, line 111: Maybe just "https://example.com"?

reyang (approved):
LGTM.

src/OpenTelemetry.Exporter.Prometheus/PrometheusExporterOptions.cs
Yun-Ting: I know! I struggled with the name as well and apparently this is the best I could come up with so far 😅 .

Yun-Ting: Maybe I should update the http/https to be upper case as well?

reyang: I think http/https is fine, and probably better in this case.

https://datatracker.ietf.org/doc/html/rfc3986#section-3.1

image

reyang: > I know! I struggled with the name as well and apparently this is the best I could come up with so far 😅 .

Finding good names is probably the hardest thing in coding #2371.

src/OpenTelemetry.Exporter.Prometheus/PrometheusExporterOptions.cs
Yun-Ting: I'll investigate this further.

src/OpenTelemetry.Exporter.Prometheus/PrometheusExporterOptions.cs
```diff
@@ -48,7 +49,25 @@ public class PrometheusExporterOptions
...
+                {
+                    if (!(Uri.TryCreate(inputUri, UriKind.Absolute, out var uri) &&
+                        (uri.Scheme == Uri.UriSchemeHttp || uri.Scheme == Uri.UriSchemeHttps)))
+                    {
+                        throw new ArgumentException(
+                            "Prometheus server path should be a valid URI with http/https scheme.",
+                            nameof(this.HttpListenerPrefixes));
+                    }
+
+                    this.httpListenerPrefixes = value;
```
mic-max, line 67: Should this be outside the loop

test/OpenTelemetry.Exporter.Prometheus.Tests/PrometheusExporterHttpServerTests.cs
```diff
@@ -105,5 +105,44 @@ public async Task PrometheusExporterHttpServerIntegration()
                 "^# TYPE counter_double counter\ncounter_double{key1='value1',key2='value2'} 101.17 \\d+\n$".Replace('\'', '"'),
                 await response.Content.ReadAsStringAsync().ConfigureAwait(false));
         }
+
+        [Theory]
+        [InlineData("http://example.com")]
+        [InlineData("https://example.com")]
```
mic-max, line 111: add string[] tests with multiple elements

test/OpenTelemetry.Exporter.Prometheus.Tests/PrometheusExporterHttpServerTests.cs
```diff
@@ -105,5 +105,44 @@ public async Task PrometheusExporterHttpServerIntegration()
...
+                using MeterProvider meterProvider = Sdk.CreateMeterProviderBuilder()
+                    .AddPrometheusExporter(opt =>
+                    {
+                        opt.HttpListenerPrefixes = new string[] { uri };
+                    })
+                    .Build();
+            }
+            catch (Exception ex)
+            {
+                Assert.Equal("System.ArgumentException", ex.GetType().ToString());
```
mic-max, line 139: Would something like this work?
```suggestion
                Assert.Equal(typeof(System.ArgumentException), ex.GetType());
```

test/OpenTelemetry.Exporter.Prometheus.Tests/PrometheusExporterHttpServerTests.cs
```diff
@@ -105,5 +105,44 @@ public async Task PrometheusExporterHttpServerIntegration()
...
+                .Build();
+        }
+
+        [Theory]
+        [InlineData("")]
+        [InlineData(null)]
+        [InlineData("ftp://example.com")]
+        public void ServerEndpointSanityCheckNegativeTest(string uri)
+        {
+            try
```
mic-max, line 128: Instead of using `try-catch` in tests you can use Assert.Throws, search in this repo for examples

Yun-Ting: Right, the arrange/act/assert is a more standard pattern.
I'll leave the current try-catch so that I can assert the exception message easily.

src/OpenTelemetry.Exporter.Prometheus/PrometheusExporterOptions.cs
Yun-Ting: I guess we can take 9090 to align with the default prometheus yaml file?

Yun-Ting: 9090 is the port of Prometheus application itself; so we should use something else as the port for the metrics we want Prometheus to collect data from. I'm curious about how did we come up with port 80 in the beginning. Thanks!

reyang: #2780

Yun-Ting: Updated the port to be 9184 as is in Noah's [example](https://github.com/dotnet/docs/blob/main/docs/core/diagnostics/metrics-collection.md#:~:text=opt.HttpListenerPrefixes%20%3D%20new%20string%5B%5D%20%7B%20%24%22http%3A//localhost%3A9184/%22%20%7D%3B ).

reyang (changes requested):
The scope of PR has changed, now it is beyond sanity check. Need to either revert the change (scope down) or address the new issues (e.g. update the PR title, update relevant docs to reflect the change, update changelog, etc.)

reyang (approved):

test/OpenTelemetry.Exporter.Prometheus.Tests/PrometheusExporterHttpServerTests.cs
mic-max: You can still check the exception message using this way.
See [here](https://github.com/open-telemetry/opentelemetry-dotnet/blob/33f3596b4b3b4703daa536e6b5901c554f90a14a/test/OpenTelemetry.Tests/Metrics/MetricViewTests.cs#L64)

cijothomas (approved):

dmcgiv:
@Yun-Ting this has caused previous valid configuration to fail, e.g. http://+:9184/ now fails this validation when previously worked for when your service needs to respond to multiple hosts or hostname is know known at time of config.

Issue #2780, opened 2022-01-12
The default listening port for Prometheus Exporter


reyang:
# Bug Report

List of [NuGet packages](https://www.nuget.org/profiles/OpenTelemetry) and
version that you are using (e.g. `OpenTelemetry 1.0.2`):

* OpenTelemetry.Exporter.Prometheus 1.2.0-rc1

Runtime version (e.g. `net461`, `net48`, `netcoreapp3.1`, `net5.0` etc. You can
find this information from the `*.csproj` file):

* net6.0

## Symptom

PrometheusExporter by default listens on `*:80` which gives two concerns - security / attack surface, 80 is a well known port and it requires privilege on certain operating systems.

**What is the expected behavior?**

1. the exporter should listen on localhost by default
2. the exporter should use a port that is greater than 1024 by default

**What is the actual behavior?**

The exporter is listening on `*:80` by default

## Additional Context

It seems some existing Prometheus components are using `9100` and `9091` by default https://prometheus.io/docs/guides/node-exporter/.

https://github.com/open-telemetry/opentelemetry-dotnet/blob/18d73c59f021b1c7a9bd433ea703e6f2c22a0b90/src/OpenTelemetry.Exporter.Prometheus/PrometheusExporterOptions.cs#L51

Yun-Ting:
There's a issue opened on adding authentication on the Prometheus middleware endpoint lately: #2776

Yun-Ting:
Note: The PR is merged and this issue can be closed. (I'm not authorized to do that.)

mic-max:
Should we automate the removal of the `Stale` tag on merged pull requests?

PR #2282, opened 2021-08-25
Add histogram support for Prometheus exporter

cijothomas:
Sample output for prometheus to scrape, when running the sample in examples.

```
# TYPE myCounter counter
myCounter{tag1="value1",tag2="value2"} 570 1629860968455
# TYPE myCounter counter
myCounter{tag1="anothervalue",tag2="somethingelse"} 5700 1629860968458
# TYPE myHistogram histogram
myHistogram_sum{tag1="value1",tag2="value2"} 44620 1629860968462
myHistogram_count{tag1="value1",tag2="value2"} 57 1629860968462
myHistogram_bucket{tag1="value1",tag2="value2",le="0"} 0 1629860968462
myHistogram_bucket{tag1="value1",tag2="value2",le="5"} 0 1629860968462
myHistogram_bucket{tag1="value1",tag2="value2",le="10"} 0 1629860968462
myHistogram_bucket{tag1="value1",tag2="value2",le="25"} 0 1629860968462
myHistogram_bucket{tag1="value1",tag2="value2",le="50"} 0 1629860968462
myHistogram_bucket{tag1="value1",tag2="value2",le="75"} 0 1629860968462
myHistogram_bucket{tag1="value1",tag2="value2",le="100"} 1 1629860968462
myHistogram_bucket{tag1="value1",tag2="value2",le="250"} 4 1629860968462
myHistogram_bucket{tag1="value1",tag2="value2",le="500"} 15 1629860968462
myHistogram_bucket{tag1="value1",tag2="value2",le="1000"} 38 1629860968462
myHistogram_bucket{tag1="value1",tag2="value2",le="+Inf"} 57 1629860968462
# TYPE Gauge gauge
Gauge{tag1="value1",tag2="value2"} 998 1629860968473
```

There are issues with Prometheus exporter as it emits multiple rows with "Type" info, which is against prometheus rule. Okay to do this for the coming alpha2 release. Will address this along with rest of fixes in next release. (the fix requires change in export data structure which is anyway changing)

PR #2783, opened 2022-01-13
Update default host and port for PrometheusExporter to align with the OTel spec.

Yun-Ting:
Fixes #.
https://github.com/open-telemetry/opentelemetry-dotnet/issues/2780

## Changes
Update default httpListenerPrefixes for PrometheusExporter to be http://localhost:9464/ to align with the spec: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#prometheus-exporter


reyang:
Couple things to discuss:
1. Do we want the binding host/port to be optional or required? (e.g. it seems Java implementation is making it required https://github.com/open-telemetry/opentelemetry-java/tree/main/exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus)
2. If the answer to 1) is optional, what should be the default value and why?
3. Given the .NET PrometheusExporter can be used either as a standalone HttpListener or integrated into ASP.NET Core pipeline, do we want to have different default value depending on the usage case, or the same value across the board? (@CodeBlanch you might have more insights here).

alanwest (approved):

alanwest:
For sake of comparison, looks like [opentelemetry-java choose 9464](https://github.com/open-telemetry/opentelemetry-java/blob/969000363d6dd53e882626b3864aef751a1c9921/exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/PrometheusHttpServerBuilder.java#L16)

alanwest:
> 1. it seems Java implementation is making it required

Looks like [Java's default is 9464](https://github.com/open-telemetry/opentelemetry-java/blob/969000363d6dd53e882626b3864aef751a1c9921/exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/PrometheusHttpServerBuilder.java#L16)

Yun-Ting:
> > it seems Java implementation is making it required
> 
> Looks like [Java's default is 9464](https://github.com/open-telemetry/opentelemetry-java/blob/969000363d6dd53e882626b3864aef751a1c9921/exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/PrometheusHttpServerBuilder.java#L16)
> 
> Appears there is some precedent for 9464 as here's the OpenCensus exporter for node https://github.com/census-instrumentation/opencensus-node/blob/master/packages/opencensus-exporter-prometheus/README.md.
> 
> The spec references it too https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#prometheus-exporter

Interesting! Not sure how did people come up with 9464. BTW, [here](https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers) contains the list of ports that we may need to avoid when choosing default ports.

src/OpenTelemetry.Exporter.Prometheus/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Update default httpListenerPrefixes for PrometheusExporter to be <http://0.0.0.0:9464/>.
```
reyang, line 5:
```suggestion
* Update default `httpListenerPrefixes` for PrometheusExporter to be `http://0.0.0.0:9464/`.
```

src/OpenTelemetry.Exporter.Prometheus/PrometheusExporterOptions.cs
```diff
@@ -29,7 +29,7 @@ public class PrometheusExporterOptions
         internal Func<DateTimeOffset> GetUtcNowDateTimeOffset = () => DateTimeOffset.UtcNow;
 
         private int scrapeResponseCacheDurationMilliseconds = 10 * 1000;
-        private IReadOnlyCollection<string> httpListenerPrefixes = new string[] { "http://*:80/" };
+        private IReadOnlyCollection<string> httpListenerPrefixes = new string[] { "http://0.0.0.0:9464/" };
```
reyang, line 32: Interesting, it seems to be a spec bug to me.

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#prometheus-exporter

I believe `0.0.0.0` stands for "all IPv4 addresses" rather than "all addresses".
If "all addresses" is the desired behavior, `*` is the correct way.
I also doubt if "all addresses" is the right direction because it could be a security concern for some folks.

reyang: https://github.com/open-telemetry/opentelemetry-specification/issues/2275

Yun-Ting: I agree on the potential security concern introduced by having all (IPv4) addresses as the default host. 
I guess we can wait a bit for the comments by folks regarding the above mentioning issue.

reyang: See if this would get merged #2282.

Yun-Ting: I think we've got a good number of approvals for the spec change PR here. Thank you.

Yun-Ting: I think we've got a good number of approvals for the spec change PR here. Thank you.

reyang (approved):

cijothomas (approved):

PR #2785, opened 2022-01-13
Fail-fast when using AddView with guaranteed conflict

### LINKED ISSUE #2751: AddView validations based on new spec
Check if AddView should react to the following spec clarification, by fail-fast on AddView if it can be detected that the view can lead to conflicts:
See https://github.com/open-telemetry/opentelemetry-specification/pull/2168#discussion_r756325622

cijothomas:
Fixes #2751 

## Changes

### DIFF
--- a/src/OpenTelemetry/Metrics/MeterProviderBuilderExtensions.cs
+++ b/src/OpenTelemetry/Metrics/MeterProviderBuilderExtensions.cs
@@ -57,6 +57,14 @@ namespace OpenTelemetry.Metrics
                 throw new ArgumentException($"Custom view name {name} is invalid.", nameof(name));
             }
 
+            if (instrumentName.IndexOf('*') != -1)
+            {
+                throw new ArgumentException(
+                    $"When view is used to provide name of metric stream, the instrument name" +
+                    $"{instrumentName} should not contain wildcard selection, as it'll lead to conflict.",
+                    nameof(instrumentName));
+            }
+
             if (meterProviderBuilder is MeterProviderBuilderBase meterProviderBuilderBase)
             {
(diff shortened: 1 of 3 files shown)


src/OpenTelemetry/Metrics/MeterProviderBuilderExtensions.cs
```diff
@@ -57,6 +57,14 @@ public static MeterProviderBuilder AddView(this MeterProviderBuilder meterProvid
                 throw new ArgumentException($"Custom view name {name} is invalid.", nameof(name));
             }
 
+            if (instrumentName.IndexOf('*') != -1)
+            {
+                throw new ArgumentException(
```
cijothomas, line 62: @alanwest Any better wording suggestions?

src/OpenTelemetry/Metrics/MeterProviderBuilderExtensions.cs
```diff
@@ -57,6 +57,14 @@ public static MeterProviderBuilder AddView(this MeterProviderBuilder meterProvid
...
+                    $"When view is used to provide name of metric stream, the instrument name" +
```
reyang, line 63:
```suggestion
                    $"When view is used to rename metric stream, the instrument name" +
```

src/OpenTelemetry/Metrics/MeterProviderBuilderExtensions.cs
alanwest: maybe something like...

"Instrument selection criteria is invalid. Instrument name '{instrumentName}' contains a wildcard character. This is not allowed when using a view to rename a metric stream as it would lead to conflicting metric stream names."

alanwest (approved):

PR #2792, opened 2022-01-14
Add an example for LoggerMessageAttribute

reyang:
Add an example demonstrating the compile-time logging source generation.


utpilla (approved):

docs/logs/source-generation/Program.cs
```diff
@@ -0,0 +1,42 @@
...
+                .AddConsoleExporter());
+        });
+
+        var logger = loggerFactory.CreateLogger<Program>();
+
+        logger.FoodPriceChanged("artichoke", 9.99);
+
+        logger.FoodRecallNotice(
+            logLevel: LogLevel.Critical,
+            brandName: "Dole",
```
utpilla, line 36: Do we want to use a dummy company? This is not a public repo so it should be fine anyway.

reyang: I don't know. This information is coming from today's FDA official web site https://www.fda.gov/safety/recalls-market-withdrawals-safety-alerts so I think it's already public?

reyang: Maybe we can use Contoso or Northwind? (which seems old school) 😃

utpilla: I'm okay with having this company's name here in the example. I was just wondering if there's any best practice or guideline   for calling out actual companies in our example to avoid any legal ramifications. I'm not sure if there is any.

utpilla: > Maybe we can use Contoso or Northwind? (which seems old school) 😃

Yeah we could go with any of the standard fictional companies that are used in Microsoft docs: https://social.technet.microsoft.com/wiki/contents/articles/1117.list-of-fictional-companies-used-in-microsoft-material.aspx

reyang: Good point, I've changed the name to Contoso.

cijothomas (approved):

PR #2805, opened 2022-01-22
Metric AggregatorStore optimizations for sorting tags- New

utpilla:
Fixes item 1 of #2374

## Changes

- In order to avoid sorting the tags on every update call, we keep track of the tag combinations that we sorted so far.
- This mapping of user provided tag combinations to their respective sorted tag combination is maintained in `ConcurrentDictionary<Tags, Tags> sortedTagsDictionary`.
- On every update, we first check if we have sorted the user provided tags combination before.
  + If yes, we then lookup the MetricPoint index for the sorted combination in `ConcurrentDictionary<Tags, int> tagsToMetricPointIndexDictionary`.
  + If no, then we sort the user provided tag combination and add the mapping to the `sortedTagsDictionary`
- The logic for `tagsToMetricPointIndexDictionary` is comparable to the previously used `value2metrics` dictionary. The only difference is that we now use the composite key of tag keys and tag values as the `Key` for the dictionary.
- `Tags` is a struct that contains `string[]` keys and `object[]` values. It represents a tag key-value combination. This is used as the key for both the new dictionaries.

Lessons learnt from #2777: We have to sort the key and values together and their mapping information has to be stored together as well.


utpilla:
## Benchmarks
#### There is up to ~46% improvement in Perf for higher number of Tags

// * Summary *

.NET SDK=6.0.101
  [Host]     : .NET 6.0.1 (6.0.121.56705), X64 RyuJIT
  DefaultJob : .NET 6.0.1 (6.0.121.56705), X64 RyuJIT

### main

|                    Method | AggregationTemporality |      Mean |     Error |    StdDev | Allocated |
|-------------------------- |----------------------- |----------:|----------:|----------:|----------:|
|            CounterHotPath |             Cumulative |  15.82 ns |  0.078 ns |  0.069 ns |         - |
| CounterWith1LabelsHotPath |             Cumulative |  71.28 ns |  0.270 ns |  0.252 ns |         - |
| CounterWith3LabelsHotPath |             Cumulative | 390.50 ns |  1.550 ns |  1.450 ns |         - |
| CounterWith5LabelsHotPath |             Cumulative | 594.25 ns | 11.624 ns | 13.386 ns |         - |
| CounterWith6LabelsHotPath |             Cumulative | 720.17 ns |  7.979 ns |  6.663 ns |         - |
| CounterWith7LabelsHotPath |             Cumulative | 768.72 ns |  5.820 ns |  4.544 ns |         - |
|            CounterHotPath |                  Delta |  15.68 ns |  0.097 ns |  0.086 ns |         - |
| CounterWith1LabelsHotPath |                  Delta |  68.67 ns |  0.233 ns |  0.195 ns |         - |
| CounterWith3LabelsHotPath |                  Delta | 375.43 ns |  3.357 ns |  3.140 ns |         - |
| CounterWith5LabelsHotPath |                  Delta | 588.71 ns |  5.730 ns |  5.360 ns |         - |
| CounterWith6LabelsHotPath |                  Delta | 683.18 ns |  3.388 ns |  3.169 ns |         - |
| CounterWith7LabelsHotPath |                  Delta | 782.50 ns |  5.883 ns |  5.503 ns |         - |


### With the new changes

|            CounterHotPath |             Cumulative |  15.61 ns | 0.046 ns | 0.039 ns |         - |
| CounterWith1LabelsHotPath |             Cumulative |  87.49 ns | 0.626 ns | 0.585 ns |         - |
| CounterWith3LabelsHotPath |             Cumulative | 219.34 ns | 1.950 ns | 1.824 ns |         - |
| CounterWith5LabelsHotPath |             Cumulative | 344.24 ns | 1.789 ns | 1.673 ns |         - |
| CounterWith6LabelsHotPath |             Cumulative | 388.55 ns | 1.570 ns | 1.392 ns |         - |
| CounterWith7LabelsHotPath |             Cumulative | 434.44 ns | 3.176 ns | 2.971 ns |         - |
|            CounterHotPath |                  Delta |  15.64 ns | 0.078 ns | 0.069 ns |         - |
| CounterWith1LabelsHotPath |                  Delta |  88.44 ns | 0.554 ns | 0.463 ns |         - |
| CounterWith3LabelsHotPath |                  Delta | 220.13 ns | 0.982 ns | 0.871 ns |         - |
| CounterWith5LabelsHotPath |                  Delta | 335.49 ns | 1.135 ns | 1.006 ns |         - |
| CounterWith6LabelsHotPath |                  Delta | 392.99 ns | 2.364 ns | 2.096 ns |         - |
| CounterWith7LabelsHotPath |                  Delta | 430.89 ns | 1.063 ns | 0.995 ns |         - |

utpilla:
## Stress Tests
- There is an **increase of ~7M loops/sec** from 14.7M in the main branch to 21.5M with the new changes
- The various memory size numbers are comparable for both the branches


### main
image


### With the new changes
image

src/OpenTelemetry/Metrics/AggregatorStore.cs
```diff
@@ -178,27 +179,46 @@ private void InitializeZeroTagPointIfNotInitialized()
...
+                }
+                else
                 {
-                    this.keyValue2MetricAggs.TryGetValue(seqKey, out value2metrics);
+                    sortedTagKeys = givenKeys;
+                    sortedTagValues = givenValues;
                 }
+
+                sortedTags = new Tags(sortedTagKeys, sortedTagValues);
+                this.sortedTagsDictionary.TryAdd(givenTags, sortedTags);
```
alanwest, line 218: Similar to comment here https://github.com/open-telemetry/opentelemetry-dotnet/pull/2777#discussion_r783563923

Can we defer this add until after `this.tagsToMetricPointIndexDictionary.TryAdd(sortedTags, aggregatorIndex);` below? I figure at we at least know we've gotten a metric point by then.

My concern is that one poorly instrumented library that suffers from a dimensional explosion could cause this dictionary to grow unbounded.

utpilla: We cannot defer it as we use the value of the `sortedTagsDictionary` as a key to lookup the `tagsToMetricPointIndexDictionary`.

We could add this check right after `if (!this.sortedTagsDictionary.TryGetValue(givenTags, out var sortedTags))`
```
if (!this.sortedTagsDictionary.TryGetValue(givenTags, out var sortedTags))
{
    aggregatorIndex = this.metricPointIndex;
    if (aggregatorIndex >= this.maxMetricPoints)
    {
        // sorry! out of data points.
        // TODO: Once we support cleanup of
        // unused points (typically with delta)
        // we can re-claim them here.
        return -1;
    }
    ...
    ...
    this.sortedTagsDictionary.TryAdd(givenTags, sortedTags);
}
```

This way we could avoid adding new entries to `sortedTagsDictionary` once we run out of MetricPoints. Does that seem good?

utpilla: This would still not completely prevent the dictionary from growing too large. If there is a bad instrumentation that emitsall possible combinations of a set of tags, they would all contribute to just a single `MetricPoint` but we would still have an entry for each of those combinations.

For e.g. If we have a Counter with 8 tags. #possible combinations = 8! = 40320. If the code is instrumented badly such that all these combinations are emitted then we would have 40320 entries in the `sortedTagsDictionary` which would all be contributing to just one `MetricPoint`.

src/OpenTelemetry/Metrics/AggregatorStore.cs
```diff
@@ -178,27 +179,46 @@ private void InitializeZeroTagPointIfNotInitialized()
...
-            string[] seqKey = null;
+            var givenTags = new Tags(tagKeys, tagValues);
 
-            // GetOrAdd by TagKeys at 1st Level of 2-level dictionary structure.
-            // Get back a Dictionary of [ Values x Metrics[] ].
-            if (!this.keyValue2MetricAggs.TryGetValue(tagKeys, out var value2metrics))
+            // We only need to sort if there is more than one Tag Key.
+            if (!this.sortedTagsDictionary.TryGetValue(givenTags, out var sortedTags))
             {
                 // Note: We are using storage from ThreadStatic, so need to make a deep copy for Dictionary storage.
```
Yun-Ting, line 187: Just for my learning, I assume that we are making a deep copy here because this is happening in the main thread (thread A), and we want to achieve parallelism (non-blocking) call in another thread (thread B). 
We want to make sure the if thread A is modifying tags, thread B won't got affected. Is my assumption correct?
If that is the case, could you point me to where is the logic of thread A spawning thread B? Thanks!

src/OpenTelemetry/Metrics/AggregatorStore.cs
```diff
@@ -178,27 +179,46 @@ private void InitializeZeroTagPointIfNotInitialized()
...
+                sortedTags = new Tags(sortedTagKeys, sortedTagValues);
+                this.sortedTagsDictionary.TryAdd(givenTags, sortedTags);
             }
 
-            // GetOrAdd by TagValues at 2st Level of 2-level dictionary structure.
-            // Get back Metrics[].
-            if (!value2metrics.TryGetValue(tagValues, out aggregatorIndex))
+            if (!this.tagsToMetricPointIndexDictionary.TryGetValue(sortedTags, out var aggregatorIndex))
             {
                 aggregatorIndex = this.metricPointIndex;
```
Yun-Ting, line 223: I don't think I understand why do we need to have two checks for the boundary, one outside of the lock and one inside the lock. I assume `aggregatorIndex` indicates the current index. I guess there are other modifiers to `tagsToMetricPointIndexDictionary` other than this function which results in having two checks; thus, we want to make sure the current Index is not out of boundary outside of the lock, and the next writable index is valid?

src/OpenTelemetry/Metrics/AggregatorStore.cs
Yun-Ting: Or maybe, it is because the users can spawn more than 1 thread when doing their instrumentation, so it is not necessarily that main thread is spawning a child thread...

src/OpenTelemetry/Metrics/AggregatorStore.cs
```diff
@@ -225,24 +245,14 @@ private int LookupAggregatorStore(string[] tagKeys, object[] tagValues, int leng
...
                         ref var metricPoint = ref this.metricPoints[aggregatorIndex];
                         var dt = DateTimeOffset.UtcNow;
-                        metricPoint = new MetricPoint(this.aggType, dt, seqKey, seqVal, this.histogramBounds);
+                        metricPoint = new MetricPoint(this.aggType, dt, sortedTags.Keys, sortedTags.Values, this.histogramBounds);
 
                         // Add to dictionary *after* initializing MetricPoint
                         // as other threads can start writing to the
                         // MetricPoint, if dictionary entry found.
-                        value2metrics.TryAdd(seqVal, aggregatorIndex);
+                        this.tagsToMetricPointIndexDictionary.TryAdd(sortedTags, aggregatorIndex);
```
cijothomas, line 255: to confirm the approach here : with this PR, we'll now have to repeatedly store the key and value for every entry, as opposed to the existing code, which stored keys only once.

This would lead to memory wastage storing duplicates entries, but if the benefits outweigh, then we can move forward.

src/OpenTelemetry/Metrics/AggregatorStore.cs
cijothomas: > This would still not completely prevent the dictionary from growing too large. If there is a bad instrumentation that emitsall possible combinations of a set of tags, they would all contribute to just a single MetricPoint but we would still have an entry for each of those combinations

^ We need to address this. We enforce MaxMetrics and MaxMetricPointPerMetric caps to bound the resource usage. this negates them and allows unbound growth. We could store only the unsorted->sorted for the very 1st time, and not after. if user keeps providing different order, we'll just sort it - its slow, but won't lead to unbounded memory

src/OpenTelemetry/Metrics/AggregatorStore.cs
cijothomas: IncomingTags
Lookup IncomingTags->Index
Success: use index
Fail: 
Lookup SortedIncomingTags->Index
             Success: use index
             Failure:
             Insert IncomingTags->Index, SortedIncomingTags->Index to dictionary.


^ Is the above better? didn't think fully through.. but it avoids memory explosion. at max 2 entries. and there is only a single dictionary

src/OpenTelemetry/Metrics/Tags.cs
```diff
@@ -0,0 +1,107 @@
...
+
+        public readonly object[] Values { get; }
+
+        public static bool operator ==(Tags tag1, Tags tag2) => tag1.Equals(tag2);
+
+        public static bool operator !=(Tags tag1, Tags tag2) => !tag1.Equals(tag2);
+
+        public readonly override bool Equals(object obj)
+        {
+            if (!(obj is Tags))
```
CodeBlanch, line 39: nit: I forget what C# version we have but on newer, you can make this a bit simpler:

```csharp
 return obj is not Tags other
   ? false
   : this.Equals(other);
```

Something like that 😄

src/OpenTelemetry/Metrics/AggregatorStore.cs
utpilla: This could work.

src/OpenTelemetry/Metrics/AggregatorStore.cs
```diff
@@ -210,12 +230,12 @@ private int LookupAggregatorStore(string[] tagKeys, object[] tagValues, int leng
...
 
-                lock (value2metrics)
+                lock (this.tagsToMetricPointIndexDictionary)
                 {
                     // check again after acquiring lock.
-                    if (!value2metrics.TryGetValue(tagValues, out aggregatorIndex))
+                    if (!this.tagsToMetricPointIndexDictionary.TryGetValue(sortedTags, out aggregatorIndex))
                     {
-                        aggregatorIndex = Interlocked.Increment(ref this.metricPointIndex);
+                        aggregatorIndex = ++this.metricPointIndex;
```
CodeBlanch, line 238: I don't think ++/-- are atomic operations and we need the `Interlocked.Increment`. The reason is ++/-- compile to a read and then a set. Each are atomic individually, but they are not atomic together. Other threads can jump in between the read and set.

This...

```csharp
metricPointIndex++;
```

...compiles like...

```msil
    IL_0000: ldsfld       int32 class_name::metricPointIndex
    IL_0005: ldc.i4.1
    IL_0006: add
    IL_0007: stsfld       int32 class_name::metricPointIndex
```

Not 100% certain, but dangerous bet to lose 😄

cijothomas: we are inside lock (for whole dictionary..), so ++ should be fine?

CodeBlanch: Was chatting with @utpilla on slack, he said we dropped the `Interlocked.Increment` because we now have `lock (this.tagsToMetricPointIndexDictionary)`. I have different concerns with that. Before we had `lock (value2metrics)` which was locking on the child lookup. The reason for that is it reduced contention for updates with different keys. If we have this global lock we are going to add contention for updates across different keys. If the only reason for this was to remove the `Interlocked`, I think the old way was a better mechanism. Do our benchmarks test the same key/values over and over or is there a mix of uniqueness?

utpilla: The benchmarks always use the same set of keys, but they randomize the values for the respective keys.

cijothomas: this PR removed the 2 level lookup, so we lose some niceness from the original approach. (wasted space to store keys over and over, more contention between new updates with different keys.)

Benchmarks *and* stress test should tell us if the benefits outweigh these concerns

utpilla: The main reason behind doing it was not to get rid of `Interlocked.Increment` but to get rid of sorting the key-value pairs on every update call. Separating the keys from values made it tough to identify the metricPoint index assigned to a tags combination when we tried to avoid sorting.

src/OpenTelemetry/Metrics/AggregatorStore.cs
alanwest: > We could store only the unsorted->sorted for the very 1st time, and not after.

I'm not sure this solves the issue. @utpilla reminded me it's not just about combinations of sort order. It's also about the combinations of values.

Assume a metric with the following dimensions and range of value for each dimension.
```
A = [red, black, blue], B = [sour, tangy, sweet], C = [gross, delicious]
```

We'll need to store 18 different combinations for this in `sortedTagsDictionary` - and this example is pretty reasonable cardinality-wise. Though, in poorly designed instrumentation, one of these dimensions may be unbounded. Imagine an `http.url` attribute where `url` was infinite. I've seen something like this before in pen tests where an application is hit with a barrage of bogus urls. If those urls are used as a value in one of the metric dimensions then... bad news bears 🐻 - the OTel SDK just took your app down.

Maybe we just need to pick another arbitrary cap on the size of the `sortedTagsDictionary` and then adjust it when we learn more.

cijothomas: @alanwest see this -> https://github.com/open-telemetry/opentelemetry-dotnet/pull/2805#discussion_r793795826 It might solve the unbounded memory usage issue..

src/OpenTelemetry/Metrics/Tags.cs
utpilla: Simplified this.

src/OpenTelemetry/Metrics/AggregatorStore.cs
utpilla: I have implemented the above suggestion of using just one dictionary. With this approach, we store at most two entries for any given MetricPoint. (sorted + an optional unsorted combination).

Here are the Benchmark results:

// * Summary *

.NET SDK=6.0.101
  [Host]     : .NET 6.0.1 (6.0.121.56705), X64 RyuJIT
  DefaultJob : .NET 6.0.1 (6.0.121.56705), X64 RyuJIT


|            CounterHotPath |             Cumulative |  15.78 ns | 0.024 ns | 0.022 ns |         - |
| CounterWith1LabelsHotPath |             Cumulative |  56.90 ns | 0.043 ns | 0.036 ns |         - |
| CounterWith3LabelsHotPath |             Cumulative | 144.11 ns | 0.492 ns | 0.461 ns |         - |
| CounterWith5LabelsHotPath |             Cumulative | 227.84 ns | 0.692 ns | 0.578 ns |         - |
| CounterWith6LabelsHotPath |             Cumulative | 260.65 ns | 0.640 ns | 0.599 ns |         - |
| CounterWith7LabelsHotPath |             Cumulative | 289.10 ns | 1.119 ns | 0.934 ns |         - |
|            CounterHotPath |                  Delta |  15.98 ns | 0.219 ns | 0.194 ns |         - |
| CounterWith1LabelsHotPath |                  Delta |  62.86 ns | 0.157 ns | 0.147 ns |         - |
| CounterWith3LabelsHotPath |                  Delta | 141.15 ns | 0.415 ns | 0.368 ns |         - |
| CounterWith5LabelsHotPath |                  Delta | 224.20 ns | 1.083 ns | 0.960 ns |         - |
| CounterWith6LabelsHotPath |                  Delta | 257.57 ns | 0.670 ns | 0.594 ns |         - |
| CounterWith7LabelsHotPath |                  Delta | 286.83 ns | 1.061 ns | 0.941 ns |         - |

utpilla: Here the Stress Test results with the new changes to the PR. Loops per second has increased to **~29M**.

image

src/OpenTelemetry/Metrics/AggregatorStore.cs
```diff
@@ -25,14 +25,12 @@ namespace OpenTelemetry.Metrics
...
     {
-        private static readonly ObjectArrayEqualityComparer ObjectArrayComparer = new ObjectArrayEqualityComparer();
         private readonly object lockZeroTags = new object();
         private readonly HashSet<string> tagKeysInteresting;
         private readonly int tagsKeysInterestingCount;
 
-        // Two-Level lookup. TagKeys x [ TagValues x Metrics ]
-        private readonly ConcurrentDictionary<string[], ConcurrentDictionary<object[], int>> keyValue2MetricAggs =
-            new ConcurrentDictionary<string[], ConcurrentDictionary<object[], int>>(new StringArrayEqualityComparer());
+        private readonly ConcurrentDictionary<Tags, int> tagsToMetricPointIndexDictionary =
```
cijothomas, line 32: any gotchas for using readyonly struct as dictionary key, other than implementing iequatable/

src/OpenTelemetry/Metrics/AggregatorStore.cs
alanwest: Ah ha, yes, didn't see your suggestion earlier.

src/OpenTelemetry/Metrics/Tags.cs
```diff
@@ -0,0 +1,101 @@
...
+            int hash = 17;
+
+            unchecked
+            {
+                for (int i = 0; i < this.Keys.Length; i++)
+                {
+                    hash = (hash * 31) + this.Keys[i].GetHashCode();
+                }
+
+                for (int i = 0; i < this.Values.Length; i++)
```
CodeBlanch, line 92: @utpilla One idea for perf optimization here. Could we do keys + values in the same loop?

src/OpenTelemetry/Metrics/AggregatorStore.cs
CodeBlanch: I can't think of anything personally. Implement `IEquatable` and a nice fast `GetHashCode`.

src/OpenTelemetry/Metrics/AggregatorStore.cs
```diff
@@ -178,71 +176,83 @@ private void InitializeZeroTagPointIfNotInitialized()
...
-                    // sorry! out of data points.
-                    // TODO: Once we support cleanup of
-                    // unused points (typically with delta)
-                    // we can re-claim them here.
-                    return -1;
+                    sortedTags = givenTags;
                 }
 
-                lock (value2metrics)
+                if (!this.tagsToMetricPointIndexDictionary.TryGetValue(sortedTags, out aggregatorIndex))
```
cijothomas, line 204: minor optimization - this is now executed for keys.lengtgh==1 also. we can avoid dual look up with same tags!

src/OpenTelemetry/Metrics/Tags.cs
utpilla: Fixed this.

src/OpenTelemetry/Metrics/AggregatorStore.cs
utpilla: I have updated to code to not do an additional lookup if tags `tagsLength <= 1`.

cijothomas: Could you update the PR description with the final code, along with updated benchmarks/stress test before/after?

utpilla: The Benchmarks/Stress Test numbers are the same as:
https://github.com/open-telemetry/opentelemetry-dotnet/pull/2805#discussion_r794096208
https://github.com/open-telemetry/opentelemetry-dotnet/pull/2805#discussion_r794101246

The numbers didn't change because of the optimizations of the last commit. That's because those optimizations would only show their effect when the tags are not yet added to the dictionary. Once, the tags are added they will not touch the paths that we have optimized. The `GetHashCode` optimization didn't affect the numbers much either.

cijothomas: Consider adding a changelog entry - Performance improvements. (Users are advised to provide tags with same Key order, to achieve maximum performance)..

utpilla: I have updated the Benchmark and Stress Test numbers.

src/OpenTelemetry/Metrics/AggregatorStore.cs
utpilla: We need to store the tagKeys and tagValues in the dictionary. The tagKeys and tagValues provided to the method `LookupAggregatorStore` are local to the thread that is running that code. These values will keep changing as long as different metrics are emitted on that thread. The dictionary entries should not be modified once they are in the dictionary. Hence, we need an allocated memory space to store these. That's why we do a deep copy.

src/OpenTelemetry/Metrics/AggregatorStore.cs
cijothomas: "check... take lock..check again" is a standard pattern for maximum performance, and yet correct from multi-threading point.

src/OpenTelemetry/CHANGELOG.md
```diff
@@ -8,6 +8,10 @@
 * Fail-fast when using AddView with guaranteed conflict.
   ([2751](https://github.com/open-telemetry/opentelemetry-dotnet/issues/2751))
 
+* Performance Improvement: When emitting metrics, users are strongly advised to
+  provide tags with same Key order, to achieve maximum performance.
+  ([2805](https://github.com/open-telemetry/opentelemetry-dotnet/issues/2805))
```
cijothomas, line 13:
```suggestion
  ([2805](https://github.com/open-telemetry/opentelemetry-dotnet/pull/2805/files))
```

utpilla: Copy-past errors 🥲

src/OpenTelemetry/Metrics/AggregatorStore.cs
```diff
@@ -225,24 +216,94 @@ private int LookupAggregatorStore(string[] tagKeys, object[] tagValues, int leng
...
+                }
+                else
+                {
+                    // Note: We are using storage from ThreadStatic, so need to make a deep copy for Dictionary storage.
+                    string[] givenKeys;
+                    object[] givenValues;
 
-                        var seqVal = new object[length];
-                        tagValues.CopyTo(seqVal, 0);
+                    if (length == 0)
```
cijothomas, line 255: does this ever occur, given we special-case 0 keys already before this method?

src/OpenTelemetry/Metrics/AggregatorStore.cs
```diff
@@ -225,24 +216,94 @@ private int LookupAggregatorStore(string[] tagKeys, object[] tagValues, int leng
...
+
+                            ref var metricPoint = ref this.metricPoints[aggregatorIndex];
+                            var dt = DateTimeOffset.UtcNow;
+                            metricPoint = new MetricPoint(this.aggType, dt, givenTags.Keys, givenTags.Values, this.histogramBounds);
+
+                            // Add to dictionary *after* initializing MetricPoint
+                            // as other threads can start writing to the
+                            // MetricPoint, if dictionary entry found.
+
+                            // givenTags will alwayas be sorted when tags length < 1
```
cijothomas, line 304:
```suggestion
                            // givenTags will always be sorted when tags length == 1
```

src/OpenTelemetry/Metrics/AggregatorStore.cs
utpilla: Good point!

src/OpenTelemetry/Metrics/AggregatorStore.cs
```diff
@@ -178,44 +176,37 @@ private void InitializeZeroTagPointIfNotInitialized()
...
-                value2metrics = new ConcurrentDictionary<object[], int>(ObjectArrayComparer);
-                if (!this.keyValue2MetricAggs.TryAdd(seqKey, value2metrics))
+                if (length > 1)
                 {
-                    this.keyValue2MetricAggs.TryGetValue(seqKey, out value2metrics);
-                }
-            }
+                    // Note: We are using storage from ThreadStatic, so need to make a deep copy for Dictionary storage.
+                    // Create a new array for the sorted Tag keys.
+                    var sortedTagKeys = new string[length];
```
cijothomas, line 187: I think we could leverage another ThreadStatic to store the sorted one, and do the new string[],object[] allocs *only* after we determine that we need to insert something to dictionary. For example, if user provides random order of keys every time, the current behavior does not cause allocation. This PR causes allocations. Something to think as future improvement, not for this PR.

src/OpenTelemetry/Metrics/AggregatorStore.cs
utpilla: Fixed this.

TimothyMothra:
Looking at the `LookupAggregatorStore` method, the `if (length > 1)` and `else` bodies are almost identical...
- Lines 185-244 Perform the sort, and then add both `givenTags` and `sortedTags`.
- Lines 251-294 Are only adding the `givenTags`.

If it's important to keep these bodies consistent, consider refactoring this out to a helper method.

alanwest (approved):
Nice work @utpilla! This looks like a solid improvement.

src/OpenTelemetry/Metrics/AggregatorStore.cs
```diff
@@ -225,24 +216,83 @@ private int LookupAggregatorStore(string[] tagKeys, object[] tagValues, int leng
                             return -1;
                         }
 
-                        // Note: We are using storage from ThreadStatic, so need to make a deep copy for Dictionary storage.
-                        if (seqKey == null)
+                        lock (this.tagsToMetricPointIndexDictionary)
```
reyang, line 219: Use a dedicated lock object instead of the default one from the dict (since the dict's own implementation might use it for its internal operation).

reyang: Curious - would `SpinWait` give better behavior under contention (less CSWitch)?

src/OpenTelemetry/Metrics/Tags.cs
```diff
@@ -0,0 +1,96 @@
...
+                if (!this.Values[i].Equals(other.Values[i]))
+                {
+                    return false;
+                }
+            }
+
+            return true;
+        }
+
+        public readonly override int GetHashCode()
```
reyang, line 81: Could this benefit from aggressive inlining?

src/OpenTelemetry/CHANGELOG.md
```diff
@@ -8,6 +8,10 @@
 * Fail-fast when using AddView with guaranteed conflict.
   ([2751](https://github.com/open-telemetry/opentelemetry-dotnet/issues/2751))
 
+* Performance Improvement: When emitting metrics, users are strongly advised to
```
reyang, line 11:
```suggestion
* Performance improvement: when emitting metrics, users are strongly advised to
```

reyang (approved):
LGTM.

src/OpenTelemetry/Metrics/AggregatorStore.cs
cijothomas: Would be good to explore as follow ups.

src/OpenTelemetry/Metrics/AggregatorStore.cs
utpilla: I have created #2838 to track this.

src/OpenTelemetry/Metrics/AggregatorStore.cs
utpilla: I have created the issue #2843 to track this.

src/OpenTelemetry/Metrics/Tags.cs
utpilla: I have created the issue #2843 to track this.

utpilla:
> Looking at the `LookupAggregatorStore` method, the `if (length > 1)` and `else` bodies are almost identical...
> 
> * Lines 185-244 Perform the sort, and then add both `givenTags` and `sortedTags`.
> * Lines 251-294 Are only adding the `givenTags`.
> 
> If it's important to keep these bodies consistent, consider refactoring this out to a helper method.

I have created the issue #2843 to track this.

cijothomas (approved):
Great job in improving the perf significantly! LGTM. 
Please address the non blocking comments as follow ups.

src/OpenTelemetry/Metrics/Tags.cs
```diff
@@ -0,0 +1,96 @@
...
+            // Equality check for Values
+            // Check if the two object[] are equal
+            var valuesLength = this.Values.Length;
+
+            if (valuesLength != other.Values.Length)
+            {
+                return false;
+            }
+
+            for (int i = 0; i < valuesLength; i++)
```
CodeBlanch, line 70: @utpilla Sorry, just noticed this. If we re-order this a bit so we validate that key & value lengths are equal first, then we could use one loop to check the keys & values for equality. Faster that way I think.

utpilla: I did try this it didn't affect the benchmark numbers that much. I would still update it nonetheless.

PR #2814, opened 2022-01-25
test: jaeger doesn't support microsecond level precision

### LINKED ISSUE #2719: Invalid Jaeger span duration
# Bug Report

OpenTelemetry packages:

* OpenTelemetry 1.1.0
* OpenTelemetry.Exporter.Jaeger 1.1.0

Environment:

* .NET 5
* Windows 10 (10.0.19044) / Debian 4.19.194-3
* Jaeger 1.29.0 (all-in-one, memory storage)

## Symptom

Invalid Jaeger span duration, especially when instrumenting something fast like in-memory operations - in Jaeger this spans have 0us duration when real duration could be something like 800us or 1300us. Looks like it happens because of invalid operator order in  `JaegerActivityExtensions.ToJaegerSpan`  :

https://github.com/open-telemetry/opentelemetry-dotnet/blob/76d944130f1d84d3d7b08db19a1fe0926378e85a/src/OpenTelemetry.Exporter.Jaeger/Implementation/JaegerActivityExtensions.cs#L131

explicit cast `double` -> `long` has higher priority when multiplication and 0.8 ms becomes 0.

## Proposed solution

This could be fixed with

```c#
return new JaegerSpan(
    duration: (long)(activity.Duration.TotalMilliseconds * 1000),
   // other args
);
```

or even better (with away from zero rounding):

```c#
return new JaegerSpan(
    duration: (long)Math.Round(activity.Duration.TotalMilliseconds * 1000 * 1000, MidpointRounding.AwayFromZero),
   // other args
);
```

mic-max:
Fixes #2719

## Changes

Change how we round the activity's duration. Before anything below 1 millisecond would be made 0. Now it has microsecond precision.
I.e. Before a 800us jaeger span would have a duration of 0. Now it has a duration of 800.

### DIFF
--- a/test/OpenTelemetry.Exporter.Jaeger.Tests/Implementation/JaegerActivityConversionTest.cs
+++ b/test/OpenTelemetry.Exporter.Jaeger.Tests/Implementation/JaegerActivityConversionTest.cs
@@ -250,7 +250,7 @@ namespace OpenTelemetry.Exporter.Jaeger.Implementation.Tests
         [Fact]
         public void JaegerActivityConverterTest_ConvertActivityToJaegerSpan_NoLinks()
         {
-            var activity = CreateTestActivity(addLinks: false);
+            var activity = CreateTestActivity(addLinks: false, ticksToAdd: 8000);
             var traceIdAsInt = new Int128(activity.Context.TraceId);
             var spanIdAsInt = new Int128(activity.Context.SpanId);
 
@@ -480,10 +480,11 @@ namespace OpenTelemetry.Exporter.Jaeger.Implementation.Tests
             Resource resource = null,
             ActivityKind kind = ActivityKind.Client,
             bool isRootSpan = false,
-            Status? status = null)
+            Status? status = null,
+            long ticksToAdd = 60 * TimeSpan.TicksPerSecond)
         {
             var startTimestamp = DateTime.UtcNow;
-            var endTimestamp = startTimestamp.AddSeconds(60);
+            var endTimestamp = startTimestamp.AddTicks(ticksToAdd);
             var eventTimestamp = DateTime.UtcNow;
             var traceId = ActivityTraceId.CreateFromString("e8ea7e9ac72de94e91fabc613f9686b2".AsSpan());
 


reyang:
I don't know what's the expect/right behavior for Jaeger.

For example, would the following thing make sense:

* Parent: duration 2ms (actually 1.8ms)
  * Child 1: duration 1ms (actually 0.6ms)
  * Child 2: duration 1ms (actually 0.6ms)
  * Child 3: duration 1ms (actually 0.6ms)

src/OpenTelemetry.Exporter.Jaeger/Implementation/JaegerActivityExtensions.cs
```diff
@@ -128,7 +128,7 @@ public static JaegerSpan ToJaegerSpan(this Activity activity)
                 operationName: activity.DisplayName,
                 flags: (activity.Context.TraceFlags & ActivityTraceFlags.Recorded) > 0 ? 0x1 : 0,
                 startTime: ToEpochMicroseconds(activity.StartTimeUtc),
-                duration: (long)activity.Duration.TotalMilliseconds * 1000,
+                duration: (long)Math.Round(activity.Duration.TotalMilliseconds * 1000, MidpointRounding.AwayFromZero),
```
reyang, line 131: Would ticks have better precision than milliseconds?

reyang (changes requested):
https://github.com/open-telemetry/opentelemetry-dotnet/pull/2814/files#r792004332

src/OpenTelemetry.Exporter.Jaeger/Implementation/JaegerActivityExtensions.cs
mic-max: Yes, I will update to use it!

mic-max:
> I don't know what's the expect/right behavior for Jaeger.
> 
> For example, would the following thing make sense:
> 
> * Parent: duration 2ms (actually 1.8ms)
>   
>   * Child 1: duration 1ms (actually 0.6ms)
>   * Child 2: duration 1ms (actually 0.6ms)
>   * Child 3: duration 1ms (actually 0.6ms)
> 
> Or:
> 
> * Parent: duration 1ms (actually 1.8ms)
>   
>   * Child 1: duration 0ms (actually 0.6ms)
>   * Child 2: duration 0ms (actually 0.6ms)
>   * Child 3: duration 0ms (actually 0.6ms)

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk_exporters/jaeger.md#unit-of-time
The spec says either is fine. Rounding or truncating. However, there's nothing about the parent and child relationship in regard to their durations. Note: It does specify that microsecond precision is needed for Thrift and nanosecond precision is needed for protobuf.

reyang (approved):
LGTM.

src/OpenTelemetry.Exporter.Jaeger/Implementation/JaegerActivityExtensions.cs
```diff
@@ -128,7 +128,7 @@ public static JaegerSpan ToJaegerSpan(this Activity activity)
                 operationName: activity.DisplayName,
                 flags: (activity.Context.TraceFlags & ActivityTraceFlags.Recorded) > 0 ? 0x1 : 0,
                 startTime: ToEpochMicroseconds(activity.StartTimeUtc),
-                duration: (long)activity.Duration.TotalMilliseconds * 1000,
+                duration: activity.Duration.Ticks * TicksPerMicrosecond,
```
reyang, line 131: Sorry I wasn't paying close attention to this - is this `*` or `/`?

reyang (changes requested):
https://github.com/open-telemetry/opentelemetry-dotnet/pull/2814/files#r793038266

src/OpenTelemetry.Exporter.Jaeger/Implementation/JaegerActivityExtensions.cs
mic-max: Yes, my bad. That's embarrassing 😓

test/OpenTelemetry.Exporter.Jaeger.Tests/Implementation/JaegerActivityConversionTest.cs
```diff
@@ -573,6 +574,11 @@ public void JaegerActivityConverterTest_Status_ErrorFlagTest(StatusCode expected
             return activity;
         }
 
+        private long TimeSpanToMicroseconds(TimeSpan timeSpan)
+        {
+            return timeSpan.Ticks / (TimeSpan.TicksPerMillisecond / 1000);
```
reyang, line 579: Consider `timeSpan.Ticks / TimeSpan.TicksPerMillisecond * 1000`. (it's uncommon, but in case TicksPerMillisecond is a small number on new hardware).

reyang (approved):

PR #2820, opened 2022-01-27
Add diagnostic logs for Metrics

utpilla:
## Changes
- Add `Verbose` diagnostic logs for Metrics for better troubleshooting

### DIFF
--- a/src/OpenTelemetry/Metrics/BaseExportingMetricReader.cs
+++ b/src/OpenTelemetry/Metrics/BaseExportingMetricReader.cs
@@ -83,7 +83,11 @@ namespace OpenTelemetry.Metrics
             // TODO: Do we need to consider timeout here?
             try
             {
-                return this.exporter.Export(metrics) == ExportResult.Success;
+                OpenTelemetrySdkEventSource.Log.MetricReaderEvent($"{nameof(BaseExportingMetricReader)}.{nameof(this.ProcessMetrics)} is calling {nameof(this.exporter)}.{nameof(this.exporter.Export)} method.");
+                var result = this.exporter.Export(metrics);
+                OpenTelemetrySdkEventSource.Log.MetricReaderEvent($"{nameof(this.exporter)}.{nameof(this.exporter.Export)} returned {result}.");
+
+                return result == ExportResult.Success;
             }
             catch (Exception ex)
             {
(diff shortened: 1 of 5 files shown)


src/OpenTelemetry/Metrics/BaseExportingMetricReader.cs
```diff
@@ -83,7 +83,11 @@ internal override bool ProcessMetrics(in Batch<Metric> metrics, int timeoutMilli
             // TODO: Do we need to consider timeout here?
             try
             {
-                return this.exporter.Export(metrics) == ExportResult.Success;
+                OpenTelemetrySdkEventSource.Log.MetricReaderEvent($"{nameof(BaseExportingMetricReader)}.{nameof(this.ProcessMetrics)} is calling {nameof(this.exporter)}.{nameof(this.exporter.Export)} method.");
```
reyang, line 86: This would alloc/format even if there is no listener.

reyang (changes requested):
Perf / memory allocation concern.

src/OpenTelemetry/Metrics/MeterProviderSdk.cs
```diff
@@ -45,6 +46,11 @@ internal sealed class MeterProviderSdk : MeterProvider
             int maxMetricPointsPerMetricStream,
             IEnumerable<MetricReader> readers)
         {
+            OpenTelemetrySdkEventSource.Log.MeterProviderSdkEvent("Setup started.");
```
cijothomas, line 49: such messages are not necessarily useful. Lets add very specific messages in the logs. MeterProvider initialized successfully/failed with ex(exception message) etc.

src/OpenTelemetry/Metrics/MeterProviderSdk.cs
```diff
@@ -93,9 +119,13 @@ internal sealed class MeterProviderSdk : MeterProvider
                 shouldListenTo = instrument => meterSourcesToSubscribe.Contains(instrument.Meter.Name);
             }
 
+            OpenTelemetrySdkEventSource.Log.MeterProviderSdkEvent($"Listening to following meters = \"{string.Join(";", meterSources)}\".");
```
cijothomas, line 122: make these non-events, else the cost is paid even if none is listening to these logs.

src/OpenTelemetry/Metrics/MeterProviderSdk.cs
utpilla: I was thinking since this verbose logging, we can have some logs to mark the beginning of `ctor` as well instead of otherwise just inferring that from the absence of logs.

PR #2822, opened 2022-01-27
Remove unused parameter timeoutMilliseconds from ProcessMetrics

utpilla:
## Changes
- Remove unused parameter `timeoutMilliseconds` parameter from `ProcessMetrics`

### DIFF
--- a/src/OpenTelemetry/Metrics/BaseExportingMetricReader.cs
+++ b/src/OpenTelemetry/Metrics/BaseExportingMetricReader.cs
@@ -78,7 +78,7 @@ namespace OpenTelemetry.Metrics
         }
 
         /// <inheritdoc/>
-        internal override bool ProcessMetrics(in Batch<Metric> metrics, int timeoutMilliseconds)
+        internal override bool ProcessMetrics(in Batch<Metric> metrics)
         {
             // TODO: Do we need to consider timeout here?
             try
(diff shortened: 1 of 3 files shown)


src/OpenTelemetry/Metrics/BaseExportingMetricReader.cs
```diff
@@ -78,7 +78,7 @@ internal override void SetParentProvider(BaseProvider parentProvider)
         }
 
         /// <inheritdoc/>
-        internal override bool ProcessMetrics(in Batch<Metric> metrics, int timeoutMilliseconds)
+        internal override bool ProcessMetrics(in Batch<Metric> metrics)
```
reyang, line 81: Why do we want to remove it?

utpilla: Just to simplify the method signature because `ProcessMetrics` does not use it.

mic-max: How did you find this?
Is there some stylecop rule that can help us clean up any other similar unused parameters for our private and internal functions?

utpilla: I found this while working on a PR that involved adding logs to Metrics SDK code. There is a code quality rule for this: https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0060

However, we don't want to remove this as we would want to modify `BaseExporter.Export` method to add this parameter.

utpilla:
Spoke to @reyang offline. We intend to update the Export method signature in future to support timeout. So, we don't want to remove this.

PR #2823, opened 2022-01-27
Add test to cover spatial aggregation bug

cijothomas:
For Observablecounters, tests with tags were missing. This adds more coverage. Also added a test which is skipped now, until spatially aggregating the ObservableCounter is supported.

### DIFF
--- a/test/OpenTelemetry.Tests/Metrics/MetricAPITest.cs
+++ b/test/OpenTelemetry.Tests/Metrics/MetricAPITest.cs
@@ -15,7 +15,9 @@
 // </copyright>
 
 using System;
+using System.Buffers;
 using System.Collections.Generic;
+using System.ComponentModel.Design;
 using System.Diagnostics;
 using System.Diagnostics.Metrics;
 using System.Threading;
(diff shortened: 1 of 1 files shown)


utpilla (approved):

test/OpenTelemetry.Tests/Metrics/MetricAPITest.cs
```diff
@@ -382,6 +384,131 @@ public void ObservableCounterAggregationTest(bool exportDelta)
...
+            var metricPoint2 = metricPoints[1];
+            Assert.Equal(10, metricPoint2.GetSumLong());
+            ValidateMetricPointTags(tags2, metricPoint2.Tags);
+
+            var metricPoint3 = metricPoints[2];
+            Assert.Equal(10, metricPoint3.GetSumLong());
+            ValidateMetricPointTags(tags3, metricPoint3.Tags);
+        }
+
+        [Theory(Skip = "Known issue.")]
```
utpilla, line 451: Do we have an issue created for this? We could add that information here.

cijothomas: will create and link.

martincostello: 4 years have passed so the chances might not be good, but do you remember what this issue is/was @cijothomas? I stumbled on the skip while doing some refactoring.

cijothomas: The issue still exists, and the test added in this PR shows the issue. I think I missed creating an issue in this repo.
It originated from spec discussion https://github.com/open-telemetry/opentelemetry-specification/issues/1874

PR #2828, opened 2022-01-28
Add unit test to validate Metric name description

cijothomas:
Minor addition to unit test to validate conformance to:

> If the unit is not provided or the unit is null, the API and SDK MUST make sure that the behavior is the same as an empty unit string.

### DIFF
--- a/src/OpenTelemetry.Exporter.Console/ConsoleMetricExporter.cs
+++ b/src/OpenTelemetry.Exporter.Console/ConsoleMetricExporter.cs
@@ -51,13 +51,13 @@ namespace OpenTelemetry.Exporter
             {
                 var msg = new StringBuilder($"\nExport ");
                 msg.Append(metric.Name);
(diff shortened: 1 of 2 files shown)


src/OpenTelemetry.Exporter.Console/ConsoleMetricExporter.cs
```diff
@@ -51,13 +51,13 @@ public override ExportResult Export(in Batch<Metric> batch)
             {
                 var msg = new StringBuilder($"\nExport ");
                 msg.Append(metric.Name);
-                if (!string.IsNullOrEmpty(metric.Description))
+                if (metric.Description != string.Empty)
```
cijothomas, line 54: similar for all exporters? wondering if there is any perf benefit of doing this, vs `IsNullOrEmpty`

utpilla (approved):

PR #2831, opened 2022-01-29
Remove interfaces used for creating mock gRPC clients

alanwest:
My primary intent with this PR is to enable the move of the generated proto code to its own package. Something like `OpenTelemetry.Proto`. The OTLP exporter package would then depend on `OpenTelemetry.Proto` rather than generating the code itself.

The interfaces we've introduced for testing purposes cause a wrinkle in this effort. The issue is that you can't extend a partial class from another assembly. So, in this case we're extending `[Metric|Log|Trace]ServiceClient` in order to introduce an interface.

```charp
    internal static partial class TraceService
    {
        public interface ITraceServiceClient
        {
            ExportTraceServiceResponse Export(...);
        }

        public partial class TraceServiceClient : ITraceServiceClient
        {
        }
    }
```

As it turns out, we're not actually making very extensive use of these interfaces (I think we were more so before). Furthermore, they're internal. I think we should remove them.

Docs for gRPC .NET essentially just guide folks to use a mocking library https://docs.microsoft.com/en-us/aspnet/core/grpc/test-client?view=aspnetcore-6.0#mock-a-grpc-client.

The only place remaining where we need to mock the client is in the benchmark project, so I've demonstrated things there. Might be nice to centralize this "get me a mock gRPC client" code somewhere if we end up needing to do the same thing in other tests.


cijothomas (approved):

cijothomas:
@alanwest looks like the CI is failing. (I retried and still the same)

alanwest:
Hmm, I ran it once again and it worked... though we may have a bug hiding somewhere (not related to this PR)

Found this in the log of one of the failed runs

```
The active test run was aborted. Reason: Test host process crashed : Unhandled exception. System.ObjectDisposedException: Safe handle has been closed.
Object name: 'SafeHandle'.
   at System.Runtime.InteropServices.SafeHandle.DangerousAddRef(Boolean& success)
   at Interop.Kernel32.ResetEvent(SafeWaitHandle handle)
   at System.Threading.EventWaitHandle.Reset()
   at OpenTelemetry.BatchExportProcessor`1.ExporterProc() in /home/runner/work/opentelemetry-dotnet/opentelemetry-dotnet/src/OpenTelemetry/BatchExportProcessor.cs:line 251
```

reyang:
@alanwest do you have the link to the failed run and test case? I can take a look. I guess it might be related to some race condition that the exporter got disposed before the exporting thread joined.

alanwest:
Failed run: https://github.com/open-telemetry/opentelemetry-dotnet/runs/5028395785?check_suite_focus=true

It's unclear exactly which test case it failed on but the surrounding passing test cases suggests it may be `OpenTelemetry.Trace.Tests.BatchExportActivityProcessorTest.CheckForceFlushExport`. Probably for one of the test cases where the timeout is very short - 0 or 1 ms.

```
  Passed OpenTelemetry.Trace.Tests.BatchExportActivityProcessorTest.CheckExportForRecordingButNotSampledActivity [4 ms]
  Passed OpenTelemetry.Trace.Tests.BatchExportActivityProcessorTest.CheckForceFlushExport(timeout: -1) [1 s]
The active test run was aborted. Reason: Test host process crashed : Unhandled exception. System.ObjectDisposedException: Safe handle has been closed.

...
```

reyang:
> Failed run: https://github.com/open-telemetry/opentelemetry-dotnet/runs/5028395785?check_suite_focus=true
> 
> It's unclear exactly which test case it failed on but the surrounding passing test cases suggests it may be `OpenTelemetry.Trace.Tests.BatchExportActivityProcessorTest.CheckForceFlushExport`. Probably for one of the test cases where the timeout is very short - 0 or 1 ms.

Thanks! That's sufficient for me to look into the problem 💪

PR #2844, opened 2022-02-02
Swallow ObjectDisposedException

reyang:
Refer to the analysis here https://github.com/open-telemetry/opentelemetry-dotnet/pull/2831#issuecomment-1027421862.

I think it is a very rare case that the user might have disposed the `BatchExportingProcessor` while the exporter worker thread `Join()` timed out.

Here goes a self-contained repro simulating this situation:

```csharp

using System;
using System.Text;
using System.Threading;
using OpenTelemetry;
using OpenTelemetry.Logs;

public class Program
{
    public static void Main()
    {
        var processor = new BatchLogRecordExportProcessor(
            exporter: new MyExporter("ExporterX"),
            scheduledDelayMilliseconds: 100,
            maxExportBatchSize: 1);
            builder.AddOpenTelemetry(options => options.AddProcessor(processor)));

        for (int i = 0; i < 100000; i++)
        {
            logger.LogInformation("Hello, World!");
        }

        processor.Dispose();
    }

    internal class MyExporter : BaseExporter<LogRecord>
    {
        private readonly string name;

        public MyExporter(string name = "MyExporter")
        {
            this.name = name;
        }

        public override ExportResult Export(in Batch<LogRecord> batch)
        {
            // SuppressInstrumentationScope should be used to prevent exporter
            // code from generating telemetry and causing live-loop.
            using var scope = SuppressInstrumentationScope.Begin();

            Console.WriteLine($"{this.name}.Export(...)");
            Thread.Sleep(1);
            return ExportResult.Success;
        }

        protected override bool OnShutdown(int timeoutMilliseconds)
        {
            Console.WriteLine($"{this.name}.OnShutdown(timeoutMilliseconds={timeoutMilliseconds})");
            return true;
        }

        {
            Console.WriteLine($"{this.name}.Dispose({disposing})");
        }
    }
}
```

### DIFF
--- a/src/OpenTelemetry/BatchExportProcessor.cs
+++ b/src/OpenTelemetry/BatchExportProcessor.cs
@@ -240,7 +240,7 @@ namespace OpenTelemetry
                     }
                     catch (ObjectDisposedException)
                     {
-                        // the exporter is somehow disposed before the worker thread finishing its job
+                        // the exporter is somehow disposed before the worker thread could finish its job
                         return;
                     }
                 }
@@ -259,7 +259,7 @@ namespace OpenTelemetry
                     }
                     catch (ObjectDisposedException)
                     {
-                        // the exporter is somehow disposed before the worker thread finishing its job
+                        // the exporter is somehow disposed before the worker thread could finish its job
                         return;
                     }
                 }
--- a/src/OpenTelemetry/BatchExportProcessor.cs
+++ b/src/OpenTelemetry/BatchExportProcessor.cs
@@ -251,8 +251,15 @@ namespace OpenTelemetry
                         this.exporter.Export(batch);
                     }
 
-                    this.dataExportedNotification.Set();
-                    this.dataExportedNotification.Reset();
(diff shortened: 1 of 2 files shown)


reyang:
I think we need to discuss about the design principles, in general the telemetry SDK should "be nice to the user, after the SDK is initialized, we should not punish the user even if they did something wrong". And we have [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/error-handling.md#basic-error-handling-principles) covering this.

What's your take on the following scenario:

1. the user wrote a multi-threaded application which calls `exporter.ForceFlush`.
2. the user disposed the exporter explicitly or implicitly
3. the user threads which are calling `exporter.ForceFlush` now could run into exception (depending on how it is implemented)

Next question: do you expect all the telemetry APIs that we exposed from this project to be bullet proof? e.g. even if things got disposed, member functions can still be called without exception.

cijothomas (approved):

PR #2845, opened 2022-02-02
Should histogram sum be filled from negatives

cijothomas:
Trying to get clarity on the expected behavior:
As per this (https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto#L408-L409), SUM should not be filled when negative values are used.

Should we interpret it as
1. (This PR) - If -ve values are reported, then Histogram count, buckets are calculated as before. But the SUM should not include the -ve values.
2. If -ve values are reported, then SUM field *should not* be populated at all.
3. Block -ve values at Histogram.Record API itself. But what to do if user configured View to convert Gauge/UpDownCounter to a Histogram?

### DIFF
--- a/src/OpenTelemetry/Metrics/MetricPoint.cs
+++ b/src/OpenTelemetry/Metrics/MetricPoint.cs
@@ -326,7 +326,11 @@ namespace OpenTelemetry.Metrics
                         lock (this.histogramBuckets.LockObject)
                         {
                             this.runningValue.AsLong++;
-                            this.histogramBuckets.RunningSum += number;
+                            if (number > 0)
+                            {
+                                this.histogramBuckets.RunningSum += number;
+                            }
+
                             this.histogramBuckets.RunningBucketCounts[i]++;
                         }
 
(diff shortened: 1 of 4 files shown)


src/OpenTelemetry/Metrics/MetricPoint.cs
```diff
@@ -326,7 +326,11 @@ internal void Update(double number)
                         lock (this.histogramBuckets.LockObject)
                         {
                             this.runningValue.AsLong++;
-                            this.histogramBuckets.RunningSum += number;
+                            if (number > 0)
```
reyang, line 329: I have different understanding - negative value for histogram should be dropped:

```csharp
if (number < 0)
{
    // do nothing for now, revisit once histogram supports negative
}
else
{
    lock (this.histogramBuckets.LockObject)
    {
        // do the work
    }
}

reyang: Given the current wording is "CAN", I think either way is _okay_, I personally would prefer the conservative approach (do nothing).

https://github.com/open-telemetry/opentelemetry-proto/blob/f8c59fe6b063397714034d9727f08aca1a5b99b3/opentelemetry/proto/metrics/v1/metrics.proto#L406-L410

reyang (approved):

reyang:
> If -ve values are reported, then SUM field should not be populated at all.

We would never know that ahead-of-time.

cijothomas:
Closing in favor of https://github.com/open-telemetry/opentelemetry-dotnet/pull/2846

Issue #2519, opened 2021-10-26
Measurement value validations


cijothomas:
Measurements from Counter are expected to be +ve numbers only. Measurements from ObservableCounter are expected to be non-decreasing.

This issue it to track adding this validations to the Metric SDK.

cijothomas:
Histogram recording should only be +ve numbers?

cijothomas:
> Histogram recording should only be +ve numbers?

Not really validation, but something for the aggregation
https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto#L440-L444

cijothomas:
more notes : validate and add tests for NaN, +Inf, -Inf

cijothomas:
> > Histogram recording should only be +ve numbers?
> 
> Not really validation, but something for the aggregation https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto#L440-L444

Well..after re-reading it seems only +ve allowed :D https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#record

cijothomas:
> Measurements from ObservableCounter are expected to be non-decreasing.

Its not clear if SDK is expected to do this validation. The spec doesn't explicitly say so. 

For histograms, and sync counters spec has the following:

The increment amount, which MUST be a non-negative numeric value.
The amount of the Measurement, which MUST be a non-negative numeric value.

cijothomas:
Moving out of 1.2 milestone.
One attempt was made, but this require a more deep analysis
https://github.com/open-telemetry/opentelemetry-dotnet/pull/2846#issuecomment-1028401221

PR #2846, opened 2022-02-02
Counter and Histogram to ignore negative measurements

cijothomas:
[Counter](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#add) and [Histogram](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#record) API spec says the values must be non-negative.
This PR adds this validation, and ignores the measurement if negative.

Fixes https://github.com/open-telemetry/opentelemetry-dotnet/issues/2519

### DIFF
--- a/src/OpenTelemetry/Metrics/Metric.cs
+++ b/src/OpenTelemetry/Metrics/Metric.cs
@@ -139,7 +139,7 @@ namespace OpenTelemetry.Metrics
         {
             if (!this.allowNegativeMeasurement && value < 0)
             {
-                // do we log?
+                // TODO: do we log?
             }
             else
             {
@@ -151,7 +151,7 @@ namespace OpenTelemetry.Metrics
         {
             if (!this.allowNegativeMeasurement && value < 0)
             {
-                // do we log?
+                // TODO: do we log?
             }
             else
             {
--- a/src/OpenTelemetry/Metrics/Metric.cs
+++ b/src/OpenTelemetry/Metrics/Metric.cs
@@ -25,6 +25,7 @@ namespace OpenTelemetry.Metrics
         internal static readonly double[] DefaultHistogramBounds = new double[] { 0, 5, 10, 25, 50, 75, 100, 250, 500, 1000 };
 
         private readonly AggregatorStore aggStore;
+        private readonly bool allowNegativeMeasurement = true;
 
         internal Metric(
             Instrument instrument,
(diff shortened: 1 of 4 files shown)


utpilla (approved):

src/OpenTelemetry/Metrics/Metric.cs
```diff
@@ -133,12 +137,26 @@ public MetricPointsAccessor GetMetricPoints()
 
         internal void UpdateLong(long value, ReadOnlySpan<KeyValuePair<string, object>> tags)
         {
-            this.aggStore.Update(value, tags);
+            if (!this.allowNegativeMeasurement && value < 0)
+            {
+                // TODO: do we log?
```
reyang, line 142: My answer would be no

cijothomas: For perf reasons, right?

reyang (changes requested):
I guess for integers we should allow negatives as the natural overflow?

cijothomas:
> I guess for integers we should allow negatives as the natural overflow?

Overflow can occur even for long right? (though less likely). Also same for histogram.Record(value), where value is -ve due to overflow.

I think we shouldn't do any special casing - if we receive -ve (due to overflow or for other reasons), we can treat the same.

cijothomas:
> > I guess for integers we should allow negatives as the natural overflow?
> 
> Overflow can occur even for long right? (though less likely). Also same for histogram.Record(value), where value is -ve due to overflow.
> 
> I think we shouldn't do any special casing - if we receive -ve (due to overflow or for other reasons), we can treat the same.

I misinterpreted that you were referring to just Int32 type. Yes this can be done.

Also after digging more, this require more deeper analysis. The aggregator itself can overflow and produce negatives in the outputted metric, so we'll need to carefully document the expected behavior in all cases. Will be closing this PR for now.

test/OpenTelemetry.Tests/Metrics/MetricAPITest.cs
```diff
@@ -379,6 +379,77 @@ public void CounterAggregationTest(bool exportDelta)
...
+            meterProvider.ForceFlush(MaxTimeToAllowForFlush);
+            sumReceived = GetLongSum(exportedItems);
+            if (exportDelta)
+            {
+                Assert.Equal(60, sumReceived);
+            }
+            else
+            {
+                Assert.Equal(100, sumReceived);
+            }
```
mic-max, line 450: nit: This check seems superfluous, or is it trying to cover something other than the export being cumulative or delta?

test/OpenTelemetry.Tests/Metrics/MetricAPITest.cs
```diff
@@ -379,6 +379,77 @@ public void CounterAggregationTest(bool exportDelta)
             }
         }
 
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+        public void CounterAggregationIgnoresNegativeMeasurementsTest(bool exportDelta)
```
mic-max, line 385: Nit: make the parameter of type `AggregationTemporality`

mic-max (approved):

test/OpenTelemetry.Tests/Metrics/MetricAPITest.cs
cijothomas: exported value is different based on Delta vs Cumulative.

mic-max: ya, the previous round of `.Add` methods covered that case though, so is this needed since I think it just checks the same thing again -- is what I meant to say in my comment

PR #2848, opened 2022-02-02
Attempt to reduce provider lifetime issues

cijothomas:
Have seen users not managing the provider lifetime correctly. Partly because of copy-paste from console example: https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/docs/metrics/getting-started/Program.cs#L28, to other workloads like ASP.NET, without changes.

This also adds a log, so self-diag can offer some hints about the issue, in some situations.


docs/metrics/customizing-the-sdk/README.md
```diff
@@ -31,6 +31,16 @@ using OpenTelemetry.Metrics;
 using var meterProvider = Sdk.CreateMeterProviderBuilder().Build();
 ```
 
+In a typical application, a single `MeterProvider` is created at application
+startup and disposed at application shutdown. It is important to ensure that the
+provider is not disposed too early. Actual mechanism depends on the application
+type. For example, in a typical ASP.NET application, `MeterProvider` is created
+in `Application_Start`, and disposed in `Application_End` (both methods part of
+Global.asp.cs file) as shown [here](../../../examples/AspNet/Global.asax.cs). In
```
utpilla, line 39: Global.asax.cs*

docs/metrics/customizing-the-sdk/README.md
```diff
@@ -31,6 +31,16 @@ using OpenTelemetry.Metrics;
...
+a typical ASP.NET Core application, `MeterProvider` lifetime is managed by
+leveraging the built-in Dependency Injection container as shown
+[here](../../../examples/AspNetCore/Startup.cs).
```
utpilla, line 42: These examples don't show `MeterProivder` setup.

reyang (approved):

docs/trace/customizing-the-sdk/README.md
```diff
@@ -34,6 +34,16 @@ using OpenTelemetry.Trace;
 using var tracerProvider = Sdk.CreateTracerProviderBuilder().Build();
 ```
 
+In a typical application, a single `TracerProvider` is created at application
+startup and disposed at application shutdown. It is important to ensure that the
+provider is not disposed too early. Actual mechanism depends on the application
+type. For example, in a typical ASP.NET application, `TracerProvider` is created
+in `Application_Start`, and disposed in `Application_End` (both methods part of
+Global.asp.cs file) as shown [here](../../../examples/AspNet/Global.asax.cs). In
```
utpilla, line 42: Global.asax.cs*

CodeBlanch (approved):
LGTM

docs/metrics/customizing-the-sdk/README.md
cijothomas: 😃 I know. But in future we'll add. It at least shows "tracerprovider". so people might get inspired from its usage.

utpilla (approved):

alanwest (commented):
Late to the 🎉, but LGTM.

An additional thought might be to include some lifetime guidance in some of the XML comments. Maybe on the `Create*ProviderBuilder` methods? https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/Sdk.cs#L57-L74

I could take a stab...

cijothomas:
> Late to the 🎉, but LGTM.
> 
> An additional thought might be to include some lifetime guidance in some of the XML comments. Maybe on the `Create*ProviderBuilder` methods? https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/Sdk.cs#L57-L74
> 
> **I could take a stab...**

Yes please!

PR #2854, opened 2022-02-03
[Do not merge] Unblock LogRecord redaction scenario

Yun-Ting:
Fixes #.

## Changes

### DIFF
--- a/src/OpenTelemetry/Logs/LogRecord.cs
+++ b/src/OpenTelemetry/Logs/LogRecord.cs
@@ -132,7 +132,7 @@ namespace OpenTelemetry.Logs
                 Guard.ThrowIfNull(value, nameof(this._state));
                 try
                 {
-                    _spinlock.Enter(ref _lockTaken);
+                    _spinlock.Enter(ref this._lockTaken);
 
                     var listKvp = value is IReadOnlyList<KeyValuePair<string, object>> ? value as IReadOnlyList<KeyValuePair<string, object>> : value is IEnumerable<KeyValuePair<string, object>> ? value as IReadOnlyList<KeyValuePair<string, object>> : null;
 
@@ -155,7 +155,7 @@ namespace OpenTelemetry.Logs
                 }
                 finally
                 {
-                    if (_lockTaken)
+                    if (this._lockTaken)
                     {
                         _spinlock.Exit();
                     }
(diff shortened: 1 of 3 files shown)


src/OpenTelemetry/Logs/LogRecord.cs
```diff
@@ -81,21 +107,97 @@ public sealed class LogRecord
...
         /// <summary>
-        /// Gets the raw state attached to the log. Set to <see
+        /// Gets or sets the raw state attached to the log. Set to <see
         /// langword="null"/> when <see
         /// cref="OpenTelemetryLoggerOptions.ParseStateValues"/> is enabled.
         /// </summary>
-        public object State { get; }
+        public object State
+        {
+            get => this._stateList;
```
Yun-Ting, line 140: Did not add a lock here, because get is a hot path and if thread A will only get the value before or after the list assignment, which is atomic.

src/OpenTelemetry/Logs/LogRecord.cs
```diff
@@ -81,21 +106,96 @@ public sealed class LogRecord
...
+                        listKvp = new List<KeyValuePair<string, object>>
+                        {
+                            new KeyValuePair<string, object>(string.Empty, this._state),
+                        };
+                    }
+
+                    var tempList = new List<KeyValuePair<string, object>>(listKvp.Count);
+                    for (int i = 0; i < listKvp.Count; ++i)
+                    {
+                        var updatedEntry = new KeyValuePair<string, object>(listKvp[i].Key, listKvp[i].Value);
```
Yun-Ting, line 150: value

src/OpenTelemetry/Logs/LogRecord.cs
```diff
@@ -81,21 +107,97 @@ public sealed class LogRecord
...
         /// <summary>
-        /// Gets the raw state attached to the log. Set to <see
+        /// Gets or sets the raw state attached to the log. Set to <see
         /// langword="null"/> when <see
         /// cref="OpenTelemetryLoggerOptions.ParseStateValues"/> is enabled.
         /// </summary>
-        public object State { get; }
+        public object State
+        {
+            get => (object)this.state;
```
Yun-Ting, line 130: When upcasting back to object, it is no longer of type `FormattedLogValues` can cannot leverage its `.ToString` anymore...

PR #2858, opened 2022-02-04
Update readme.md

xavierjohn:
Updated readme.md with

Metrics Console exporter does not output to the console immediately so you will end up seeing the metrics on the screen after the program has terminated.

### DIFF
--- a/docs/metrics/getting-started/README.md
+++ b/docs/metrics/getting-started/README.md
@@ -58,6 +58,8 @@ is configured to subscribe to instruments from the Meter
 `MyCompany.MyProduct.MyLibrary`, and aggregate the measurements in-memory. The
 pre-aggregated metrics are exported to a `ConsoleExporter`.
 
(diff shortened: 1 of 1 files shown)


docs/metrics/getting-started/README.md
```diff
@@ -30,6 +30,8 @@ Update the `Program.cs` file with the code from [Program.cs](./Program.cs):
 Run the application again (using `dotnet run`) and you should see the metric
 output from the console, similar to shown below:
```
cijothomas, line 31:
```suggestion
output from the console (metrics will be seen once the program ends), similar to shown below:
```

cijothomas: the program ends instantly, so I always see metrics instantly..

cijothomas (approved):

reyang (approved):

alanwest (approved):

PR #1949, opened 2021-03-31
Do not merge: Move proto files to separate project

alanwest:
Fixes #652.

Putting this out here for discussion. In today's SIG meeting I noted that the Go community generates their proto files to make consuming them convenient.

See:
* Source repository: https://github.com/open-telemetry/opentelemetry-proto-go
* Package repository: https://pkg.go.dev/go.opentelemetry.io/proto/otlp

The idea behind this PR is to ask whether we'd like to do something similar.

The idea is that we would release/version this project in alignment with the [opentelemetry-proto](https://github.com/open-telemetry/opentelemetry-proto) project.

The version of the proto files we currently have in the repository roughly aligns with v0.5.0 as noted in this [readme](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/README.md)

We could retroactively release prior versions of the protos up the current [v0.8.0](https://github.com/open-telemetry/opentelemetry-proto/releases/tag/v0.8.0) release.

Note this PR currently does not build because I have not added the `.publicApi` folder

PR #2863, opened 2022-02-05
Add OpenTelemetry.Proto project

alanwest:
This PR introduces a new `OpenTelemetry.Proto` project that, when built, includes code generated from the OpenTelemetry protos.

The project currently targets version 0.3.0 of the opentelemetry-proto project. My thought is that we could land this initial PR and then add the `proto-0.3.0` tag to the repo. We could do follow up PRs incrementing the version up to the current 0.12.0 protos. This would allow us to retroactively publish prior versions.

I have not referenced OpenTelemetry.Proto by any other projects yet. I think it will be cool to reference OpenTelemetry.Proto from the OpenTelemetry.Exporter.OpenTelemetryProtocol project, but there some challenges to consider before doing so. Namely, when a user references the OTLP exporter package, they'd also pull down and expose the API of the OpenTelemetry.Proto project. OpenTelemetry.Proto has not reached 1.0, so there is a potential for breaking changes. If we do end up referencing OpenTelemetry.Proto from the OTLP exporter then we may want to do it in a way that does not expose the proto API to the consumer - ILRepack maybe?

### DIFF
--- a/build/Common.prod.props
+++ b/build/Common.prod.props
@@ -6,7 +6,7 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
-    <PackageReference Include="Microsoft.CodeAnalysis.PublicApiAnalyzers" Version="$(MicrosoftCodeAnalysisAnalyzersPkgVer)" Condition=" $(OS) == 'Windows_NT'">
+    <PackageReference Include="Microsoft.CodeAnalysis.PublicApiAnalyzers" Version="$(MicrosoftCodeAnalysisAnalyzersPkgVer)" Condition="!($(ExcludeFromPublicApiAnalyzer) == 'true') And $(OS) == 'Windows_NT'">
       <PrivateAssets>All</PrivateAssets>
     </PackageReference>
   </ItemGroup>
@@ -74,7 +74,7 @@
   </PropertyGroup>
 
   <!--PublicApi Analyzer-->
-  <ItemGroup>
+  <ItemGroup Condition="!($(ExcludeFromPublicApiAnalyzer) == 'true')">
     <AdditionalFiles Include=".publicApi\$(TargetFramework)\PublicAPI.Shipped.txt" />
     <AdditionalFiles Include=".publicApi\$(TargetFramework)\PublicAPI.Unshipped.txt" />
     <None Include=".publicApi\*\PublicAPI.Shipped.txt" />
--- a/src/OpenTelemetry.Proto/OpenTelemetry.Proto.csproj
+++ b/src/OpenTelemetry.Proto/OpenTelemetry.Proto.csproj
@@ -8,7 +8,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <OpenTelemetryProtoVersion>0.12.0</OpenTelemetryProtoVersion>
+    <OpenTelemetryProtoVersion>0.3.0</OpenTelemetryProtoVersion>
     <OpenTelemetryProtoUrl>https://github.com/open-telemetry/opentelemetry-proto/archive/refs/tags/v$(OpenTelemetryProtoVersion).zip</OpenTelemetryProtoUrl>
     <OpenTelemetryProtoZipFile>opentelemetry-proto-$(OpenTelemetryProtoVersion).zip</OpenTelemetryProtoZipFile>
     <OpenTelemetryProtoSourceRoot>opentelemetry-proto</OpenTelemetryProtoSourceRoot>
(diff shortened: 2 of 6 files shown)


src/OpenTelemetry.Proto/OpenTelemetry.Proto.csproj
```diff
@@ -0,0 +1,40 @@
...
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
+    <Description>OpenTelemetry protos for OpenTelemetry .NET</Description>
+    <MinVerTagPrefix>proto-</MinVerTagPrefix>
+
+    <ExcludeFromPublicApiAnalyzer>true</ExcludeFromPublicApiAnalyzer>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <OpenTelemetryProtoVersion>0.3.0</OpenTelemetryProtoVersion>
```
alanwest, line 11: A suggestion (https://github.com/open-telemetry/opentelemetry-dotnet/pull/1949#issuecomment-810739808) on a prior iteration of this idea is to have a CI job automate the generation of new versions of the OpenTelemetry.Proto package. My thought is we might be able to introduce this in the future. For now, bumping the version here is how it's done.

build/Common.prod.props
```diff
@@ -74,7 +74,7 @@
   </PropertyGroup>
 
   <!--PublicApi Analyzer-->
-  <ItemGroup>
+  <ItemGroup Condition="!($(ExcludeFromPublicApiAnalyzer) == 'true')">
```
alanwest, line 77: I thought there was already a way to accomplish this, but `RunApiCompat=false` did not accomplish what I needed. It seems `RunApiCompat` still requires the presence of the publicApi files.

The problem is, the first build generates the code then fails because then the generated API is not declared in the public API. If I add things to the publicApi files the build works, but then if I do a clean build - wiping out the generated code - the public API analyzer runs prior to generation and fails because there are items declared in the public API that are not in the code (because it hasn't been generated yet).

reyang: Is this project needed, or the entire code-gen can be done as a pre step in the OTLP exporter project?

E.g.

1. OTLP exporter project build started
2. fetch the proto files based on the version specified in the OTLP exporter project
3. run command to build the generated code (it is possible that we need to generate a temp project here) - since this is not tight to the overall project, many rules won't apply (e.g. style check, api compat, etc.)
4. copy the output proto binary (DLLs, XML docs and PDF files) to the target place
5. build the OTLP exporter
6. pack the outputs if nuget pack step is enabled

reyang: Stepping back question - do we in general prefer keeping generated code in the repo or not?

I guess folks would have very different opinions. I've seen 80% projects keep generated code in the repo while 20% try to be mean-and-lean (and sometimes when issue happened, folks had hard time debugging stuff due to the lack of source/symbol).

alanwest: We do not currently keep generated code in the repo. We do keep a copy of the proto files in the OTLP exporter project that it generates the code from, but it internalizes it so public API is not an issue.

https://github.com/open-telemetry/opentelemetry-dotnet/blob/2c9501fcd54aa52f18b1b98e90ce9004a890a89b/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OpenTelemetry.Exporter.OpenTelemetryProtocol.csproj#L54-L58

This PR demonstrates we don't necessarily even need to keep a copy of the protos in the repo. Technically, we could take this same approach on the OTLP exporter project itself.

reyang: Yeah, Jaeger exporter took a different approach here https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/src/OpenTelemetry.Exporter.Jaeger/ApacheThrift/Protocol.

alanwest: I think the value of publishing the .NET generated proto code as a nuget package is a question that can be considered separately.

I find that it is somewhat useful. I have worked on some internal tooling requiring me to interact directly with the protos. It's not hugely difficult to acquire the protos and generate the code, but it is far easier to just reference a nuget package.

There is a precedent for this from some of the other language SIGs
* https://github.com/open-telemetry/opentelemetry-proto-java
* https://github.com/open-telemetry/opentelemetry-proto-go

They have an entire separate repo. I suppose we could consider this as well.

reyang: Yep, IIRC in Python we were also having a separate repo (so the change history is not entangled with the main project), and things can be rolled back/forward independently.

alanwest: > Yeah, Jaeger exporter took a different approach here https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/src/OpenTelemetry.Exporter.Jaeger/ApacheThrift/Protocol.

Ah, yes, in talking with @CodeBlanch he pointed this to me as well. I think maintaining the generated thrift code for jaeger probably makes sense here. I'm not familiar with the tooling for generating this, so not sure how it differs from the MSBuild tooling for generating code from protos.

alanwest: > Yep, IIRC in Python we were also having a separate repo (so the change history is not entangled with the main project), and things can be rolled back/forward independently.

It looks like Python maintains this in their core repo

reyang: I meant this:

* https://github.com/census-instrumentation/opencensus-proto/tree/master/gen-python
* https://github.com/census-instrumentation/opencensus-python/tree/master/contrib/opencensus-ext-ocagent/opencensus/ext/ocagent

PR #2875, opened 2022-02-08
Support question mark in wildcard

reyang:
Previously we only support `*` in wildcards, not `?`.
This PR added `?` support.

"Indirectly" related to https://github.com/open-telemetry/opentelemetry-specification/pull/2325 (this PR only touches the provider/source part, not the View part).


src/OpenTelemetry/ProviderExtensions.cs
```diff
@@ -69,5 +72,18 @@ internal static Action GetObservableInstrumentCollectCallback(this BaseProvider
 
             return null;
         }
+
+        internal static Regex GetWildcardRegex(this BaseProvider baseProvider, IEnumerable<string> wildcards = default)
```
reyang, line 76: This helper method doesn't have any specific logic to do with Providers. However, the providers are the only places that need it so I guess it's better to put it here than creating another place?

cijothomas (approved):
1. Need changelog entry
2. I suppose we didn't have unit test for tracing's AddSource with wildcard?

src/OpenTelemetry/ProviderExtensions.cs
alanwest: 👍 I agree, seems good for now.

reyang:
> 1. Need changelog entry
> 2. I suppose we didn't have unit test for tracing's AddSource with wildcard?

Changelog is already included in the PR.
I'm working on the unit test for tracing part, got interrupted by other things, will update the PR later today.

test/OpenTelemetry.Tests/Metrics/MetricAPITest.cs
```diff
@@ -249,7 +249,7 @@ public void MeterSourcesWildcardSupportMatchTest(bool hasView)
 
             var exportedItems = new List<Metric>();
             var meterProviderBuilder = Sdk.CreateMeterProviderBuilder()
-                .AddMeter("AbcCompany.XyzProduct.*")
+                .AddMeter("AbcCompany.XyzProduct.Component?")
```
alanwest, line 252: I guess we didn't have a similar test for the trace functionality?

src/OpenTelemetry/Metrics/MeterProviderSdk.cs
```diff
@@ -84,7 +83,7 @@ internal sealed class MeterProviderSdk : MeterProvider
             Func<Instrument, bool> shouldListenTo = instrument => false;
             if (meterSources.Any(s => s.Contains('*')))
             {
-                var regex = GetWildcardRegex(meterSources);
+                var regex = this.GetWildcardRegex(meterSources);
                 shouldListenTo = instrument => regex.IsMatch(instrument.Meter.Name);
             }
```
alanwest, line 88: Do we need a to a contains check for * and ? now?

reyang: YES 👍

test/OpenTelemetry.Tests/Metrics/MetricAPITest.cs
alanwest: Oh didn't see earlier comments about this.

src/OpenTelemetry/Metrics/MeterProviderSdk.cs
reyang: I've fixed this and added test cases.
Not sure if it'll be better to extract the `s.Contains('*') || s.Contains('?')` logic into a common helper, looks like it is currently used in 3 places.

alanwest: > Not sure if it'll be better to extract the `s.Contains('*') || s.Contains('?')` logic into a common helper, looks like it is currently used in 3 places.

I think it'd be a decent minor improvement, so something like:

```csharp
internal static bool ContainsWildcard(string s)
{
    return s.Contains('*') || s.Contains('?');
}

...

if (meterSources.Any(ContainsWildcard))
```

but, I'm ok 👍 without if you think it reads better without the helper

alanwest (approved):

src/OpenTelemetry/Metrics/MeterProviderSdk.cs
reyang: Maybe an extension method on String class? 😆

alanwest: That'd work too!

reyang: I've updated the PR by extracting the wildcard logics to `WildcardHelper`.

I didn't use String extension method because it could be too noisy (e.g. if someone is using Intellisense).

alanwest (approved):
LWTM (Looks wild to me)

src/OpenTelemetry/Metrics/MeterProviderSdk.cs
alanwest: > I didn't use String extension method because it could be too noisy (e.g. if someone is using Intellisense).

Yea I think that's a good call. I was thinking that too, but thought maybe since it'd be internal it'd just affect us and not the whole 🌏

src/OpenTelemetry.Api/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Improved wildcard support to cover `?` (which matches exactly one character).
```
cijothomas, line 5:
```suggestion
* Improved wildcard support for `AddSource`, `AddMeter` to cover `?` (which matches exactly one character).
```

cijothomas (approved):

cijothomas:
Considering we are close to 1.2 stable release, i'll leave it open for a day to give more reviewers an opportunity to spot anything risky in this PR.

mic-max (approved):

Issue #2803, opened 2022-01-20
Start Root Span Collects Current Active Span as Parent


omamoo:
# Bug Report

List of [NuGet packages](https://www.nuget.org/profiles/OpenTelemetry) and
version that you are using (e.g. `OpenTelemetry 1.0.2`):

* NuGet: OpenTelemetry.Api Version 1.2.0-rc1

Runtime version (e.g. `net461`, `net48`, `netcoreapp3.1`, `net5.0` etc. You can
find this information from the `*.csproj` file):

* dotnet 6.0.101

## Symptom

I defined simple Api endpoint with the following code:
```
[HttpGet]
public async Task TestOtel()
{
    using (TracerProvider.Default.
               GetTracer(ActivitySourceProvider.ActivitySourceName, ActivitySourceProvider.Version).
               StartActiveSpan("FirstSpan"))
    {
        await Task.Delay(5.Seconds());
        var rootSpan = 
            TracerProvider.Default.
                GetTracer(ActivitySourceProvider.ActivitySourceName, ActivitySourceProvider.Version).
                StartRootSpan("MyRootSpan"); 
        using (Tracer.WithSpan(rootSpan))
        {
            Task.Delay(5.Seconds());
        }
    }
}
```

**What is the expected behavior?**

I'm expecting to see two different traces:
* First one is for incoming HTTP request (using AspDotNet instrumentation library) and one more nested span called "FirstSpan"
* Second is for the MyRootSpan which created by using `CreateRootSpan` and activated with `WithSpan` method.

**What is the actual behavior?**

One trace with all span see attached image.

## Additional Context
Image with trace at Jaeger:
![image](https://user-images.githubusercontent.com/91830944/150417633-2c309132-d438-4a31-bba3-9796b238f618.png)

Thanks for the help.

omamoo:
Anyone?

alanwest:
Yes, this might be a bug. I created a test #2878 that I think models your situation.

omamoo:
@alanwest Yes this is the exact case, thanks. How can I know if the test passes?

alanwest:
We'll leave this issue open as I'm not yet sure what the best fix will be.

In my opinion, the problem is due in part to the behavior of the `StartActivity` activity method when it is passed an empty parent context (i.e., starting a root span). `StartActivity` is called from `StartRootSpan` and is not part of the OpenTelemetry .NET project - it is part of .NET's System.Diagnostics library.

[...]

PR #2878, opened 2022-02-09
Add failing test for StartRootSpan

alanwest:
Relates to #2803 

Just adding a test for now for discussion and to determine whether there's something we can do at the moment.

It appears the `StartActivity` method starts a new span parented from `Activity.Current` even when you pass in a new `ParentContext`. I imagine this might be by design since `Activity.Current` is tracked via `AsyncLocal` and maybe `StartActivity` doesn't want to clobber that state.

https://github.com/open-telemetry/opentelemetry-dotnet/blob/6b7f2dd77cf9d37260a853fcc95f7b77e296065d/src/OpenTelemetry.Api/Trace/Tracer.cs#L79-L82

https://github.com/open-telemetry/opentelemetry-dotnet/blob/6b7f2dd77cf9d37260a853fcc95f7b77e296065d/src/OpenTelemetry.Api/Trace/Tracer.cs#L167-L195

In a scenario where a thread hop has occurred - as in this test - maybe it would be more intuitive to wipe out `Activity.Current`... though, not sure there's a way for us to detect we're on a new thread.

---

Confirmed that `StartActivity` uses `Activity.Current` when the parent's `ActivityContext == default`

https://github.com/dotnet/runtime/blob/3ae87395f638a533f37b8e3385f6d3f199a72f4f/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/ActivitySource.cs#L252-L253

### DIFF
--- a/test/OpenTelemetry.Tests/Trace/TracerTest.cs
+++ b/test/OpenTelemetry.Tests/Trace/TracerTest.cs
@@ -15,7 +15,9 @@
 // </copyright>
 
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
+using System.Threading.Tasks;
 using Xunit;
 
 namespace OpenTelemetry.Trace.Tests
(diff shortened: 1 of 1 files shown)


reyang (approved):

test/OpenTelemetry.Tests/Trace/TracerTest.cs
```diff
@@ -74,6 +76,35 @@ public void Tracer_StartRootSpan_BadArgs_NullSpanName()
...
+            using var tracer = Sdk.CreateTracerProviderBuilder()
+                .AddSource("tracername")
+                .AddInMemoryExporter(exportedItems)
+                .Build();
+
+            async Task DoSomeAsyncWork()
+            {
+                await Task.Delay(10);
+                var newRootSpan = TracerProvider.Default.GetTracer("tracername").StartRootSpan("RootSpan2");
+                using (Tracer.WithSpan(newRootSpan))
```
pellared, line 93: I think (not tested) that it can be simplified to 

```suggestion
                using (TracerProvider.Default.GetTracer("tracername").StartRootSpan("RootSpan2"))
```

pellared (approved):

test/OpenTelemetry.Tests/Trace/TracerTest.cs
```diff
@@ -74,6 +76,35 @@ public void Tracer_StartRootSpan_BadArgs_NullSpanName()
...
+                }
+            }
+
+            using (tracer.GetTracer("tracername").StartActiveSpan("RootSpan1"))
+            {
+                await DoSomeAsyncWork();
+            }
+
+            Assert.Equal(2, exportedItems.Count);
+            Assert.NotEqual(exportedItems[0].TraceId, exportedItems[1].TraceId);
```
cijothomas, line 105: would be good to make it clear in this line which one is item0 and item1, by doing an assert of the `DisplayName`.

test/OpenTelemetry.Tests/Trace/TracerTest.cs
alanwest: simplified

PR #2880, opened 2022-02-09
code style compatible with dotnet new version

mokarchi:
My changes include code style, I believe new changes related to newer versions of dotnet should be made in the code to make it compatible and more readable by new users

## Changes
Beginning with C# 9.0, constructor invocation expressions are target-typed. That is, if a target type of an expression is known, you can omit a type name
In C# 9 they added several new pattern matching operators that can be combined with the is operator
the null-coalescing assignment operator ??= assigns the value of its right-hand operand to its left-hand operand only if the left-hand operand evaluates to null and also new using declaration

### DIFF
--- a/src/OpenTelemetry.Instrumentation.AspNet/Implementation/HttpInListener.cs
+++ b/src/OpenTelemetry.Instrumentation.AspNet/Implementation/HttpInListener.cs
@@ -26,8 +26,8 @@ namespace OpenTelemetry.Instrumentation.AspNet.Implementation
 {
     internal sealed class HttpInListener : IDisposable
     {
-        private readonly PropertyFetcher<object> routeFetcher = new PropertyFetcher<object>("Route");
-        private readonly PropertyFetcher<string> routeTemplateFetcher = new PropertyFetcher<string>("RouteTemplate");
+        private readonly PropertyFetcher<object> routeFetcher = new("Route");
+        private readonly PropertyFetcher<string> routeTemplateFetcher = new("RouteTemplate");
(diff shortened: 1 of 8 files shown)


src/OpenTelemetry.Instrumentation.AspNet/Implementation/HttpInListener.cs
```diff
@@ -57,12 +57,7 @@ public void Dispose()
         /// <returns>Span uri value.</returns>
         private static string GetUriTagValueFromRequestUri(Uri uri)
         {
-            if (string.IsNullOrEmpty(uri.UserInfo))
-            {
-                return uri.ToString();
-            }
-
-            return string.Concat(uri.Scheme, Uri.SchemeDelimiter, uri.Authority, uri.PathAndQuery, uri.Fragment);
+            return string.IsNullOrEmpty(uri.UserInfo) ? uri.ToString() : string.Concat(uri.Scheme, Uri.SchemeDelimiter, uri.Authority, uri.PathAndQuery, uri.Fragment);
```
reyang, line 60: Would this line be too long? (e.g. the original version makes it easier to put a breakpoint on Ln62).

mic-max (approved):

mokarchi:
I didn't no that why some check were not successful
@mic-max @reyang

mokarchi:
@mic-max I don't know why codecov/project not successful.

mic-max:
> @mic-max I don't know why codecov/project not successful.

That check isn't the best. Nothing to worry about on your end!

mic-max (approved):

Issue #2881, opened 2022-02-09
 PrometheusExporter's HTTP Sanity Check breaks valid configuration 


dmcgiv:
# Bug Report

Impacted NuGet package

- OpenTelemetry.Exporter.Prometheus 1.2.0-rc2

Runtime version 

- net461

## Symptom
#2771 introduced validation in PrometheusExporterOptions.HttpListenerPrefixes setter which does not accept valid wildcard hostname e.g. http://+:9184/ 

see documentation which confirms this is valid https://docs.microsoft.com/en-us/dotnet/api/system.net.httplistenerprefixcollection.add?view=netframework-4.6.1#remarks 

**What is the expected behavior?**

No validation exception and the HTTP Listener server request for any hostname associated with the server on the given port on .

**What is the actual behavior?**

ArgumentException("Prometheus server path should be a valid URI with http/https scheme.")

Yun-Ting:
I'll be working on the fix.

8:
Duplicate of #2840

PR #2882, opened 2022-02-10
Prometheus HttpListenerPrefixes setter to accept wildcard hostname.

Yun-Ting:
Fixes https://github.com/open-telemetry/opentelemetry-dotnet/issues/2881.

## Changes
Dropping the sanity check for PrometheusExporterOptions.HttpListenerPrefixes setter.
Currently, TryCreate [implementation](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.Uri/src/System/UriExt.cs#:~:text=private%20void%20InitializeUri(ParsingError%20err%2C%20UriKind%20uriKind%2C%20out%20UriFormatException%3F%20e)) in .NET runtime does not deem `https://+:8080/` and `http://*:8080/` as [valid](https://docs.microsoft.com/en-us/dotnet/api/system.net.httplistenerprefixcollection.add?view=netframework-4.6.1#:~:text=When%20a%20port,that%20include%20paths.) URI which were marked as valid cases for the item in `HttpListenerPrefixCollection`.

Are there ways to incorporate these cases: `https://+:8080/` and `http://*:8080/` without using ugly regexes?

### DIFF
--- a/src/OpenTelemetry.Exporter.Prometheus/PrometheusExporterOptions.cs
+++ b/src/OpenTelemetry.Exporter.Prometheus/PrometheusExporterOptions.cs
@@ -54,19 +54,7 @@ namespace OpenTelemetry.Exporter
             get => this.httpListenerPrefixes;
             set
             {
-                Guard.ThrowIfNull(value, nameof(this.httpListenerPrefixes));
-
-                foreach (string inputUri in value)
-                {
-                    if (!(Uri.TryCreate(inputUri, UriKind.Absolute, out var uri) &&
-                        (uri.Scheme == Uri.UriSchemeHttp || uri.Scheme == Uri.UriSchemeHttps)))
-                    {
-                        throw new ArgumentException(
-                            "Prometheus server path should be a valid URI with http/https scheme.",
-                            nameof(this.httpListenerPrefixes));
-                    }
-                }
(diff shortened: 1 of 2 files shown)


reyang:
Do we want to support `+` and `*` at all?

Yun-Ting:
> Do we want to support `+` and `*` at all?
Right, there will be security concerns to allow `+` and `*`.
@Damien McGivern, could you share about what is the case that you would use `+` with us?

dmcgiv:
@Yun-Ting @reyang `+` and `*` are valid use cases see
https://docs.microsoft.com/en-us/dotnet/api/system.net.httplistenerprefixcollection.add?view=netframework-4.6.1#remarks

It should be available to users of the package who are in control of their server setup. Its not a security concern as permission must be first granted to listen which requires admin access, e.g. below 

`netsh http add urlacl url=http://+:8080/ user=[username]`


HttpListener already has prefix validation so I'd say this sanity check on this property is not needed, why limiting the capabilities of this package?

Yun-Ting:
Not super relevant to the above discussion, but this is how AddPrefix do the validation - 
https://source.dot.net/#System.Net.HttpListener/System/Net/HttpListener.cs,112

PR #2890, opened 2022-02-11
Minor improvement to console trace exporter output

cijothomas:
Fixes #.

## Changes

### DIFF
--- a/src/OpenTelemetry.Exporter.Console/ConsoleActivityExporter.cs
+++ b/src/OpenTelemetry.Exporter.Console/ConsoleActivityExporter.cs
@@ -32,10 +32,16 @@ namespace OpenTelemetry.Exporter
         {
             foreach (var activity in batch)
             {
-                this.WriteLine($"Activity.Id:          {activity.Id}");
-                if (!string.IsNullOrEmpty(activity.ParentId))
+                this.WriteLine($"Activity.TraceId:          {activity.TraceId}");
+                this.WriteLine($"Activity.SpanId:           {activity.SpanId}");
+                this.WriteLine($"Activity.TraceFlags:           {activity.ActivityTraceFlags}");
+                if (!string.IsNullOrEmpty(activity.TraceStateString))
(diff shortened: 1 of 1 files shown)


src/OpenTelemetry.Exporter.Console/ConsoleActivityExporter.cs
```diff
@@ -32,10 +32,16 @@ public override ExportResult Export(in Batch<Activity> batch)
         {
             foreach (var activity in batch)
             {
-                this.WriteLine($"Activity.Id:          {activity.Id}");
-                if (!string.IsNullOrEmpty(activity.ParentId))
+                this.WriteLine($"Activity.TraceId:          {activity.TraceId}");
```
cijothomas, line 35: Activity.Id is what goes inside "traceparent" header... its a bit cleaner to display explicitly traceid,spanid etc.

src/OpenTelemetry.Exporter.Console/ConsoleActivityExporter.cs
```diff
@@ -45,7 +51,7 @@ public override ExportResult Export(in Batch<Activity> batch)
                 this.WriteLine($"Activity.Duration:    {activity.Duration}");
                 if (activity.TagObjects.Any())
                 {
-                    this.WriteLine("Activity.TagObjects:");
+                    this.WriteLine("Activity.Tags:");
```
reyang, line 54: Do you think folks might get confused here?

reyang (approved):
LGTM.

src/OpenTelemetry.Exporter.Console/ConsoleActivityExporter.cs
cijothomas: TagObjects in an implementation detail only Export authors need to know right? Most end users use SetTag, so seeing "Tags" here as opposed to "TagObjects" looks nice in my view.

Issue #2872, opened 2022-02-07
Review all exporters for unhandled exceptions


cijothomas:
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/error-handling.md#basic-error-handling-principles

"The SDK MUST NOT throw unhandled exceptions for errors in their own operations. For example, an exporter should not throw an exception when it cannot reach the endpoint to which it sends telemetry data."

This was exposed in https://github.com/open-telemetry/opentelemetry-dotnet/issues/2870 where OTLP exporter crashed the whole process, but opening a dedicated issue to track fixing this. This is separate from the [issue](https://github.com/open-telemetry/opentelemetry-dotnet/pull/1627) about when/how should SDK handle unhandled exceptions when user components behave bad. 

This issue is only tracking the core components (as required by the spec) : 

- [x] OTLPExporter - Fixed in https://github.com/open-telemetry/opentelemetry-dotnet/pull/2915
- [x] Jaeger - Fix not required. See https://github.com/open-telemetry/opentelemetry-dotnet/issues/2872#issuecomment-1036616930
- [x] Zipkin - Fixed https://github.com/open-telemetry/opentelemetry-dotnet/pull/2935
- [x] Prometheus - Fix not required. See https://github.com/open-telemetry/opentelemetry-dotnet/issues/2872#issuecomment-1043323512

cijothomas:
@Yun-Ting for OTLPExporter. @alanwest can guide here.

TimothyMothra:
I can help :)

TimothyMothra:
I reviewed the Prometheus exporter.
TLDR; If this is used as designed, there should be no issues.
However, if the exporter is ever invoked while OnExport is set to null, this would throw an exception.

If we want to proceed with caution, we can add a try/catch to the Exporter or add a null check.
I'll follow up with @cijothomas offline to discuss.

---

1. `PrometheusExporter.Export` will call `PrometheusExporter.OnExport` without a null check. 
    https://github.com/open-telemetry/opentelemetry-dotnet/blob/1b5765bf43058c6301e0ff0c099aedb34541d54e/src/OpenTelemetry.Exporter.Prometheus/PrometheusExporter.cs#L76-L79

2. `PrometheusExporter.OnExport` should be set to `PrometheusCollectionManager.OnCollect` which DOES has a try/catch. 
    This eventually calls `BaseExportingMetricReader.ProcessMetrics` which will call `exporter.Export`

[...]

PR #2891, opened 2022-02-11
Consolidating OTLPExporter unit tests.

Yun-Ting:
Fixes part of https://github.com/open-telemetry/opentelemetry-dotnet/issues/2872.

### DIFF
--- a/test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/OtlpLogExporterTests.cs
+++ b/test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/OtlpLogExporterTests.cs
@@ -1,0 +1,75 @@
+// <copyright file="OtlpLogExporterTests.cs" company="OpenTelemetry Authors">
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+
+using System.Collections.Generic;
+using Microsoft.Extensions.Logging;
+using OpenTelemetry.Exporter.OpenTelemetryProtocol.Implementation;
+using OpenTelemetry.Logs;
+using OpenTelemetry.Trace;
+using Xunit;
+
+namespace OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests
+{
+    public class OtlpLogExporterTests : Http2UnencryptedSupportTests
+    {
+        [Fact]
+        public void ToOtlpLogRecordTest()
+        {
+            // Just a basic test to demonstrate an
+            // approach useful for testing.
+            // This needs to be expanded to
+            // actually test the conversion.
+            List<LogRecord> logRecords = new List<LogRecord>();
+
+            using var loggerFactory = LoggerFactory.Create(builder =>
+            {
+                builder.AddOpenTelemetry(options =>
+                {
+                    options.IncludeFormattedMessage = true;
+                    options.ParseStateValues = true;
+                    options.AddInMemoryExporter(logRecords);
+                });
+            });
+
+            // TODO:
+            // Validate attributes, severity, traceid,spanid etc.
+
+            var logger = loggerFactory.CreateLogger("log-category");
+            logger.LogInformation("Hello from {name} {price}.", "tomato", 2.99);
+            Assert.Single(logRecords);
+            var logRecord = logRecords[0];
+            var otlpLogRecord = logRecord.ToOtlpLog();
+            Assert.NotNull(otlpLogRecord);
+            Assert.Equal("Hello from tomato 2.99.", otlpLogRecord.Body.StringValue);
+            logRecords.Clear();
+
+            logger.LogInformation(new EventId(10, null), "Hello from {name} {price}.", "tomato", 2.99);
+            Assert.Single(logRecords);
+            logRecord = logRecords[0];
+            otlpLogRecord = logRecord.ToOtlpLog();
+            Assert.NotNull(otlpLogRecord);
+            Assert.Equal("Hello from tomato 2.99.", otlpLogRecord.Body.StringValue);
+            logRecords.Clear();
+
+            logger.LogInformation(new EventId(10, "MyEvent10"), "Hello from {name} {price}.", "tomato", 2.99);
+            Assert.Single(logRecords);
+            logRecord = logRecords[0];
+            otlpLogRecord = logRecord.ToOtlpLog();
+            Assert.NotNull(otlpLogRecord);
+            Assert.Equal("Hello from tomato 2.99.", otlpLogRecord.Body.StringValue);
+        }
+    }
+}
--- a/test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/OtlpLogExporterTests.cs
+++ b/test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/OtlpLogExporterTests.cs
@@ -28,69 +28,99 @@ namespace OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests
 {
     public class OtlpLogExporterTests : Http2UnencryptedSupportTests
     {
-        private readonly ILogger logger;
-        private readonly List<LogRecord> exportedItems = new();
-        private readonly ILoggerFactory loggerFactory;
-        private OpenTelemetryLoggerOptions options;
-
-        public OtlpLogExporterTests()
-        {
-            this.loggerFactory = LoggerFactory.Create(builder =>
-            {
-                builder.AddOpenTelemetry(options =>
-                {
-                    this.options = options;
-                    this.options.AddInMemoryExporter(this.exportedItems);
-                });
-            });
-
-            this.logger = this.loggerFactory.CreateLogger<OtlpLogExporterTests>();
-        }
-
         [Fact]
         public void CheckToOtlpLogRecordStateValues()
         {
-            this.options.IncludeFormattedMessage = true;
-            this.options.ParseStateValues = true;
+            List<LogRecord> logRecords = new List<LogRecord>();
+            using var loggerFactory = LoggerFactory.Create(builder =>
+            {
+                builder.AddOpenTelemetry(options =>
+                {
+                    options.IncludeFormattedMessage = true;
+                    options.ParseStateValues = true;
+                    options.AddInMemoryExporter(logRecords);
+                });
+            });
 
-            this.logger.LogInformation("Hello from {name} {price}.", "tomato", 2.99);
-            Assert.Single(this.exportedItems);
+            var logger = loggerFactory.CreateLogger("OtlpLogExporterTests");
+            logger.LogInformation("Hello from {name} {price}.", "tomato", 2.99);
 
-            var logRecord = this.exportedItems[0];
+            Assert.Single(logRecords);
+
+            var logRecord = logRecords[0];
             var otlpLogRecord = logRecord.ToOtlpLog();
+
             Assert.NotNull(otlpLogRecord);
             Assert.Equal("Hello from tomato 2.99.", otlpLogRecord.Body.StringValue);
+            Assert.Contains("name", otlpLogRecord.Attributes.ToString());
+            Assert.Contains("tomato", otlpLogRecord.Attributes.ToString());
+            Assert.Contains("{OriginalFormat}", otlpLogRecord.Attributes.ToString());
         }
 
         [Fact]
         public void CheckToOtlpLogRecordEventId()
         {
-            this.options.IncludeFormattedMessage = true;
-            this.options.ParseStateValues = true;
+            List<LogRecord> logRecords = new List<LogRecord>();
+            using var loggerFactory = LoggerFactory.Create(builder =>
+            {
+                builder.AddOpenTelemetry(options =>
+                {
+                    options.IncludeFormattedMessage = true;
+                    options.ParseStateValues = true;
+                    options.AddInMemoryExporter(logRecords);
+                });
+            });
 
-            this.logger.LogInformation(new EventId(10, null), "Hello from {name} {price}.", "tomato", 2.99);
-            Assert.Single(this.exportedItems);
+            var logger = loggerFactory.CreateLogger("OtlpLogExporterTests");
+            logger.LogInformation(new EventId(10, null), "Hello from {name} {price}.", "tomato", 2.99);
+            Assert.Single(logRecords);
 
-            var logRecord = this.exportedItems[0];
+            var logRecord = logRecords[0];
             var otlpLogRecord = logRecord.ToOtlpLog();
+
             Assert.NotNull(otlpLogRecord);
             Assert.Equal("Hello from tomato 2.99.", otlpLogRecord.Body.StringValue);
-            this.exportedItems.Clear();
 
-            this.logger.LogInformation(new EventId(10, "MyEvent10"), "Hello from {name} {price}.", "tomato", 2.99);
-            Assert.Single(this.exportedItems);
+            var otlpLogRecordAttributes = otlpLogRecord.Attributes.ToString();
 
-            logRecord = this.exportedItems[0];
+            // Event
+            Assert.Contains("Id", otlpLogRecordAttributes);
+            Assert.Contains("10", otlpLogRecordAttributes);
+
+            logRecords.Clear();
+
+            logger.LogInformation(new EventId(10, "MyEvent10"), "Hello from {name} {price}.", "tomato", 2.99);
+            Assert.Single(logRecords);
+
+            logRecord = logRecords[0];
             otlpLogRecord = logRecord.ToOtlpLog();
             Assert.NotNull(otlpLogRecord);
             Assert.Equal("Hello from tomato 2.99.", otlpLogRecord.Body.StringValue);
+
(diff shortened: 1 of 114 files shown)


test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/OtlpLogExporterTests.cs
```diff
@@ -25,51 +28,162 @@ namespace OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests
...
 
-            // TODO:
-            // Validate attributes, severity, traceid,spanid etc.
+            this.logger = this.loggerFactory.CreateLogger<OtlpLogExporterTests>();
+        }
+
+        [Fact]
+        public void CheckToOtlpLogRecordStateValues()
+        {
+            this.options.IncludeFormattedMessage = true;
```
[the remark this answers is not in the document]
Yun-Ting: There are cases in `LogRecordTest.cs` demonstrating this: https://github.com/open-telemetry/opentelemetry-dotnet/blob/4323c93e14c1aa56e9f81717321818786d80ccd7/test/OpenTelemetry.Tests/Logs/LogRecordTest.cs#L488
So modifying options after the factory was built is not a recommended usage?

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/OtlpLogExporterTests.cs
```diff
@@ -25,51 +28,162 @@ namespace OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests
...
+            Assert.Single(this.exportedItems);
 
-            var logger = loggerFactory.CreateLogger("log-category");
-            logger.LogInformation("Hello from {name} {price}.", "tomato", 2.99);
-            Assert.Single(logRecords);
-            var logRecord = logRecords[0];
+            var logRecord = this.exportedItems[0];
             var otlpLogRecord = logRecord.ToOtlpLog();
             Assert.NotNull(otlpLogRecord);
             Assert.Equal("Hello from tomato 2.99.", otlpLogRecord.Body.StringValue);
```
alanwest, line 62: I think you should be able to also assert the state values have been added as attributes to the otlpLogRecord here.

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/OtlpLogExporterTests.cs
```diff
@@ -25,51 +28,162 @@ namespace OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests
...
-            logRecords.Clear();
+        }
 
-            logger.LogInformation(new EventId(10, null), "Hello from {name} {price}.", "tomato", 2.99);
-            Assert.Single(logRecords);
-            logRecord = logRecords[0];
-            otlpLogRecord = logRecord.ToOtlpLog();
+        [Fact]
+        public void CheckToOtlpLogRecordEventId()
```
alanwest, line 66: Similar thing here. EventId is added as an attribute.

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/OtlpLogExporterTests.cs
```diff
@@ -25,51 +28,162 @@ namespace OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests
...
+            Assert.NotNull(otlpLogRecord);
+            Assert.Equal(logRecord.LogLevel.ToString(), otlpLogRecord.SeverityText);
+        }
+
+        [Fact]
+        public void CheckToOtlpLogRecordFormattedMessage()
+        {
+            this.options.IncludeFormattedMessage = true;
+
+            this.logger.LogInformation("OpenTelemetry!");
```
alanwest, line 155: I like the enthusiasm.

reyang: Or anger? 😆

alanwest: `OpenTelemetry?!` 😠

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/OtlpLogExporterTests.cs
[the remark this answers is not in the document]
cijothomas: I see . Thanks for sharing the pointers. 
Yes logging config is bit different, and would be good doc additions.

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/OtlpLogExporterTests.cs
cijothomas: I'd prefer to use something like the usual Hello from {food} {price}, for testing formatted message. (i.e which has string template, which gets formatted with actual food and price). Or that could be in addition to this one :)

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/OtlpLogExporterTests.cs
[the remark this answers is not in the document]
reyang: I think by default we want to avoid updates on options after ctor, because this would introduce complexity which makes the code difficult to maintain - and in many cases, we will see things ended up in "undefined behavior".

In certain cases we may explicitly allow dynamic configuration update after ctor, these need to be carefully designed, implemented and documented. For example - https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/src/OpenTelemetry#processor.

For this specific case, I don't recall if there is a scenario that needs the dynamic configuration update.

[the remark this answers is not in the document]
Yun-Ting: https://github.com/open-telemetry/opentelemetry-dotnet/issues/2902

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/OtlpLogExporterTests.cs
```diff
@@ -25,51 +28,162 @@ namespace OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests
...
-            List<LogRecord> logRecords = new List<LogRecord>();
+        private readonly ILogger logger;
+        private readonly List<LogRecord> exportedItems = new();
+        private readonly ILoggerFactory loggerFactory;
+        private OpenTelemetryLoggerOptions options;
 
-            using var loggerFactory = LoggerFactory.Create(builder =>
+        public OtlpLogExporterTests()
+        {
+            this.loggerFactory = LoggerFactory.Create(builder =>
```
cijothomas, line 38: If you keep this option, then need to dispose : https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/test/OpenTelemetry.Tests/Logs/LogRecordTest.cs#L609-L612

OR, alternatively, don't do anything inside ctor. Create factory/logger inside each test (My preference is this way). This is a lot easier to test individual options.

cijothomas: For example. if a particular test does `this.options.IncludeFormattedMessage = true;`, then it should "reverse" that in a finally block. (https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/test/OpenTelemetry.Tests/Logs/LogRecordTest.cs#L603-L606)

In short, I strongly encourage to create factory/logger in each unit tests :)

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/OtlpLogExporterTests.cs
Yun-Ting: I've added the check for whether stateValues had been added as attributes. Thank you.

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/OtlpLogExporterTests.cs
Yun-Ting: Adopted the suggested approach, factor/logger in each unit tests! Thank you, Cijo!

reyang: > In short, I strongly encourage to create factory/logger in each unit tests :)

+1. I personally would also prefer each test case to be very self-contained, rather than inventing some test framework and later turn it into a maintenance hell. Here goes what I did for Prometheus Exporter https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/test/OpenTelemetry.Exporter.Prometheus.Tests/PrometheusSerializerTests.cs.

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/OtlpLogExporterTests.cs
```diff
@@ -44,32 +42,216 @@ public void ToOtlpLogRecordTest()
...
+            Assert.True(otlpLogRecord.TraceId.IsEmpty);
+            Assert.True(otlpLogRecord.SpanId.IsEmpty);
+            Assert.True(otlpLogRecord.Flags == 0);
+        }
+
+        [Fact]
+        public void CheckToOtlpLogRecordSpanIdTraceIdAndFlag()
+        {
+            // preparation
+            var sampler = new AlwaysOnSampler();
```
reyang, line 136: Do we need a local variable here or it can be simplified to:
```csharp
                .SetSampler(new AlwaysOnSampler())
                .Build();
```

Yun-Ting: Yup, I think it could be simplified!

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/OtlpLogExporterTests.cs
```diff
@@ -44,32 +42,215 @@ public void ToOtlpLogRecordTest()
...
+            Assert.True(otlpLogRecord.SpanId.IsEmpty);
+            Assert.True(otlpLogRecord.Flags == 0);
+        }
+
+        [Fact]
+        public void CheckToOtlpLogRecordSpanIdTraceIdAndFlag()
+        {
+            // preparation
+            var exportedActivityList = new List<Activity>();
+            var activitySourceName = "toOtlpLogRecordTest";
```
reyang, line 137: Consider using something like this so we don't have to worry about name conflicts:

https://github.com/open-telemetry/opentelemetry-dotnet/blob/1b5765bf43058c6301e0ff0c099aedb34541d54e/test/OpenTelemetry.Tests/Metrics/MetricAPITest.cs#L46

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/OtlpLogExporterTests.cs
```diff
@@ -44,32 +42,215 @@ public void ToOtlpLogRecordTest()
...
+            // assertion
+            var logRecord = logRecords[0];
+            var otlpLogRecord = logRecord.ToOtlpLog();
+
+            var currentActivity = Activity.Current;
+            Assert.NotNull(Activity.Current);
+
+            var expectedTraceId = currentActivity.TraceId;
+            var expectedSpanId = currentActivity.SpanId;
+            Assert.Equal(expectedTraceId.ToString(), ActivityTraceId.CreateFromBytes(otlpLogRecord.TraceId.ToByteArray()).ToString());
```
reyang, line 169: Curious - would `Assert.Equal(expectedTraceId.ToString(), otlpLogRecord.TraceId.ToString())` work? (or general question - is it possible to simplify this).

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/OtlpLogExporterTests.cs
```diff
@@ -44,32 +42,215 @@ public void ToOtlpLogRecordTest()
...
+        }
+
+        [Fact]
+        public void CheckToOtlpLogRecordSpanIdTraceIdAndFlag()
+        {
+            // preparation
+            var exportedActivityList = new List<Activity>();
+            var activitySourceName = "toOtlpLogRecordTest";
+            var activitySource = new ActivitySource(activitySourceName);
+            using var tracerProvider = Sdk.CreateTracerProviderBuilder()
```
cijothomas, line 139: for the purpose of this unit test, we may simply do the below, without setting up traceprovider.

using (var activity = new Activity("something").Start())
{
 // do logging here, it will contained the traceid,spanid from the above activity

}

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/OtlpLogExporterTests.cs
```diff
@@ -44,32 +42,215 @@ public void ToOtlpLogRecordTest()
...
+            {
+                builder.AddOpenTelemetry(options =>
+                {
+                    options.AddInMemoryExporter(logRecords);
+                });
+            });
+
+            var exceptionMessage = "Exception Message";
+            var exception = new Exception(exceptionMessage);
+            var message = "Exception Occurred";
```
reyang, line 236: Same here, consider removing the local variable.

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/OtlpLogExporterTests.cs
```diff
@@ -44,32 +42,215 @@ public void ToOtlpLogRecordTest()
...
+            List<LogRecord> logRecords = new List<LogRecord>();
+            using var loggerFactory = LoggerFactory.Create(builder =>
+            {
+                builder.AddOpenTelemetry(options =>
+                {
+                    options.AddInMemoryExporter(logRecords);
+                });
+            });
+
+            var exceptionMessage = "Exception Message";
```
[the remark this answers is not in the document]
Yun-Ting: removed

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/OtlpLogExporterTests.cs
Yun-Ting: removed

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/OtlpLogExporterTests.cs
Yun-Ting: Thanks for introducing this usage to me.

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/OtlpLogExporterTests.cs
Yun-Ting: Maybe @alanwest could help to shed some light on this too :grinning: ?

alanwest: It appears there's a `ToStringUtf8` method on `ByteString` that might get you the representation you want https://developers.google.com/protocol-buffers/docs/reference/csharp/class/google/protobuf/byte-string#class_google_1_1_protobuf_1_1_byte_string_1a34312693c1a26482efd2aea46d8d4902.

Yun-Ting: Yes, I've tried that. But ByteString method .ToStringUtf8() does not seem to perform as expected.
image

reyang: Maybe it's ToString(Encoding.UTF16) - just need to find out the actual encoding I guess?

reyang (approved):

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/OtlpLogExporterTests.cs
Yun-Ting: image
Hmm... it's not Unicode either.

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/OtlpLogExporterTests.cs
```diff
@@ -44,32 +43,202 @@ public void ToOtlpLogRecordTest()
...
+            using var loggerFactory = LoggerFactory.Create(builder =>
+            {
+                builder.AddOpenTelemetry(options =>
+                {
+                    options.AddInMemoryExporter(logRecords);
+                });
+            });
+
+            var logger = loggerFactory.CreateLogger("OtlpLogExporterTests");
+            logger.LogInformation("Log within a dropped activity");
```
cijothomas, line 123: nit: this can be confusing.. there is no activity here at all, so there is nothing to be dropped.

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/OtlpLogExporterTests.cs
```diff
@@ -44,32 +43,202 @@ public void ToOtlpLogRecordTest()
...
+            {
+                builder.AddOpenTelemetry(options =>
+                {
+                    options.AddInMemoryExporter(logRecords);
+                });
+            });
+
+            var logger = loggerFactory.CreateLogger("OtlpLogExporterTests");
+
+            var expectedTraceId = ActivityTraceId.CreateRandom();
```
cijothomas, line 147: you dont need to initialize this. As its anyway not used. You can assign to `default`.
ActivityTraceId expected = default;

cijothomas (approved):
okay to merge.. there can be follow ups to further improve.

PR #2892, opened 2022-02-11
Add stress test for strong-typed logging

reyang:
Adding a stress test for the logging path.

Related to https://github.com/dotnet/runtime/issues/61947.


test/OpenTelemetry.Tests.Stress.Logs/Program.cs
```diff
@@ -0,0 +1,53 @@
...
+using System.Runtime.CompilerServices;
+using Microsoft.Extensions.Logging;
+using OpenTelemetry.Logs;
+
+namespace OpenTelemetry.Tests.Stress;
+
+public partial class Program
+{
+    private static ILogger logger;
+    private static Payload payload = new Payload();
```
cijothomas, line 26: same payload intentionally? in practice this would be new new new PayLoads? Or are you doing this to explore the pooling of `LogRecord` to avoid heap alloc?

reyang: The latter. I think creating payload is happening before the logging API so we don't need to cover it for the SDK optimization purpose.

cijothomas (approved):

alanwest (approved):

PR #2901, opened 2022-02-15
Invoke SDK initialization when using OpenTelemetry.Extensions.Hosting

### LINKED ISSUE #2898: Propagator.Extract not fetching the baggage
Hi,

I am trying to fetch baggage from the headers using Propagator.Extract. I do get the traceParent but the parentContext.Baggage is empty. I am using RabbitMQ and I am including the baggage before posting the message. I can see the baggage in the header as well, attached screeshot.

This is the sample that I am refering to https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/examples/MicroserviceExample/Utils/Messaging/MessageReceiver.cs#:~:text=Baggage.Current%20%3D%20parentContext.Baggage%3B

image

I am expecting the baggage to be passed over to the worker service.
This is how I am setting the baggage in the MessageSender:
Baggage.SetBaggage("OtelBaggage", "OTelBaggageValue");

alanwest:
Fixes #2898

Using the microservice example to demonstrate that the SDK does not get initialized when expected.

The the static constructor on `OpenTelemtry.Sdk` does not get invoked until after the first message is sent and received in the microservice example:

https://github.com/open-telemetry/opentelemetry-dotnet/blob/aca7efc0f465786c0c307dae864a703a897a517f/src/OpenTelemetry/Sdk.cs#L30-L36

@utpilla

### DIFF
--- a/examples/MicroserviceExample/Utils/Messaging/MessageReceiver.cs
+++ b/examples/MicroserviceExample/Utils/Messaging/MessageReceiver.cs
@@ -31,7 +31,7 @@ namespace Utils.Messaging
     public class MessageReceiver : IDisposable
     {
         private static readonly ActivitySource ActivitySource = new ActivitySource(nameof(MessageReceiver));
-        private static readonly TextMapPropagator Propagator = new TraceContextPropagator();
+        private static readonly TextMapPropagator Propagator = Propagators.DefaultTextMapPropagator;
 
         private readonly ILogger<MessageReceiver> logger;
         private readonly IConnection connection;
--- a/src/OpenTelemetry.Extensions.Hosting/CHANGELOG.md
+++ b/src/OpenTelemetry.Extensions.Hosting/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+* Fixes an issue where the initialization of some aspects of the SDK can be
+  delayed when using the `AddOpenTelemetryTracing` and
+  `AddOpenTelemetryMetrics` methods. Namely, self-diagnostics and the default
+  context propagator responsible for propagating trace context and baggage.
+  ([#2901](https://github.com/open-telemetry/opentelemetry-dotnet/pull/2901))
+
 ## 1.0.0-rc9
 
 Released 2022-Feb-02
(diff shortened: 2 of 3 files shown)


examples/MicroserviceExample/Utils/Messaging/MessageReceiver.cs
```diff
@@ -90,6 +95,7 @@ public void ReceiveMessage(BasicDeliverEventArgs ea)
 
         private IEnumerable<string> ExtractTraceContextFromBasicProperties(IBasicProperties props, string key)
         {
+            Console.WriteLine($"Extracting '{key}' from carrier");
```
alanwest, line 98: Adding this line just to see when context extraction occurs. This line is hit on the second message received.

examples/MicroserviceExample/Utils/Messaging/MessageReceiver.cs
```diff
@@ -58,7 +58,7 @@ public void StartConsumer()
         public void ReceiveMessage(BasicDeliverEventArgs ea)
         {
             // Extract the PropagationContext of the upstream parent from the message headers.
-            var parentContext = Propagator.Extract(default, ea.BasicProperties, this.ExtractTraceContextFromBasicProperties);
+            var parentContext = Propagators.DefaultTextMapPropagator.Extract(default, ea.BasicProperties, this.ExtractTraceContextFromBasicProperties);
```
alanwest, line 61: On the first message received, the `DefaultTextMapPropagator` is the noop instance. By the second message received it has been initialized to the composite propagator that extracts trace context and baggage.

cijothomas: can we "touch" SDK somewhere super early......

https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/Logs/OpenTelemetryLoggerProvider.cs#L35-L40

alanwest: Resolved this by touching the `Sdk` class in OpenTelemetry.Extensions.Hosting.

reyang (approved):

cijothomas (approved):
would be good to add changelog.md entry too in hosting package?

utpilla (approved):

examples/MicroserviceExample/Utils/Messaging/MessageReceiver.cs
cijothomas: A gentle touch, and SDK behaves!

alanwest:
> would be good to add changelog.md entry too in hosting package?

Yes, I wavered on this, but I think I came up with something that makes sense.

PR #2903, opened 2022-02-16
Minor test namespace adjustment

cijothomas:
Less cluttered Test Explorer :)

image
Could use some more refactors to make it cleaner. Like moving Instrumentation.Tests elsewhere?


reyang:
> Less cluttered Test Explorer :)
> 
> image Could use some more refactors to make it cleaner. Like moving Instrumentation.Tests elsewhere?

If you don't use Visual Studio, the problem would disappear 😄

reyang (approved):

utpilla (approved):

PR #2915, opened 2022-02-18
OTLPExporter unhandled exceptions

glucaci:
#2872 

### Changes
- Include batch transformation into try/catch scope

### DIFF
--- a/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpLogExporter.cs
+++ b/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpLogExporter.cs
@@ -70,10 +70,10 @@ namespace OpenTelemetry.Exporter
 
             var request = new OtlpCollector.ExportLogsServiceRequest();
 
-            request.AddBatch(this.ProcessResource, logRecordBatch);
-
             try
             {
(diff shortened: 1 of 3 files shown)


reyang (approved):
LGTM assume changelog is included.

pellared (approved):

cijothomas (approved):

Yun-Ting (approved):

glucaci:
I didn't touch this part 🙄
image

reyang:
> I didn't touch this part 🙄 image

No worries. The current code coverage result is not stable (each run will have slightly different results) and is not considered as a blocker for the PR.
I guess the codecov issue could either be related to the tool itself (e.g. some bug, although I think it's less likely to be the case) or the how the test cases were designed (e.g. some test cases require RPC interactions or some other operations that are sensitive to timing/race).

Given this is not the first time folks asked about this, I'm thinking - maybe for short term we should change the codecov CI so it only shows the coverage percentage but always return Success? In this way it still serves the purpose of giving us a rough estimation of the test coverage, but won't introduce confusion to the contributors? @open-telemetry/dotnet-approvers @open-telemetry/dotnet-maintainers what do you think?

PR #2916, opened 2022-02-18
Allow duplicate instrument registration

### LINKED ISSUE #2925: Support identical instrument registration and handle conflicts
We currently reject instruments with duplicate names - that is, we log out a warning.

https://github.com/open-telemetry/opentelemetry-dotnet/blob/102ac27ce7c93aa760e2d62db023564b7ee09743/src/OpenTelemetry/Metrics/MetricReaderExt.cs#L44-L48

Identical instrument registrations (name, kind, unit, description, and data type - int, double, etc) need to be allowed.

When instruments are not identical but are duplicated by name a warning should still be produced. Though, a new metric stream should be created.

[Related spec change](https://github.com/open-telemetry/opentelemetry-specification/pull/2317)

Relevant sections of the spec:
* https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument-type-conflict-detection
* https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#opentelemetry-protocol-data-model-producer-recommendations
* https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#resolving-duplicate-instrument-registration-conflicts

alanwest:
Fixes #2925

See https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument

> When more than one Instrument of the same name is created for identical Meters, denoted duplicate instrument registration, the implementation MUST create a valid Instrument in every case.

---

### Identical instruments
Instantiating multiple metric instruments with the same name and also identical in all other respects - same type, description, and unit - result in a single metric stream aggregating measurements from all the identical instruments.

For example, measurements from the following instruments will be aggregated together. No warning is produced by the SDK.

```csharp
var instrument = meter.CreateCounter<long>(name: "my.counter", unit: "seconds", description: "counts seconds");
var duplicateInstrument = meter.CreateCounter<long>("my.counter", "seconds", "counts seconds");
```

### Distinct instruments

Instantiating multiple metric instruments with the same name but differ in some respect - different type, description, or unit - will result in a separate metric stream for each distinct instrument.

For example, measurements from the following instruments will result in three distinct metric streams. The second one differs by unit/description and the third one differs by type. A warning is logged by the SDK in the event of distinct duplicate instrument registrations.

```csharp
var instrument = meter.CreateCounter<long>(name: "my.counter", unit: "seconds", description: "counts seconds");
var duplicateInstrument = meter.CreateCounter<long>("my.counter", "milliseconds", "counts milliseconds");
var anotherDuplicateInstrument = meter.CreateCounter<double>("my.counter", "seconds", "counts seconds");
```


reyang (approved):
LGTM.

src/OpenTelemetry/Metrics/MetricReaderExt.cs
```diff
@@ -102,12 +111,16 @@ internal List<Metric> AddMetricsListWithViews(Instrument instrument, List<Metric
                         continue;
                     }
 
-                    if (this.metricStreamNames.Contains(metricStreamName))
+                    if (this.instrumentIdentityToMetric.TryGetValue(instrumentIdentity, out var existingMetric))
```
cijothomas, line 114: we need to add `existingMetric` to the metrics.Add(existingMetric) as well.

cijothomas: consider example:
instr1 has view, which says produce 2 streams named inst1_a, inst1_b.
instr2 has view, which says produce 2 streams named inst1_a, inst1_c.

instr1 creation will return List{M1,M2}.
instr2 creation *should* return List {M1,M3} --> this PR currently only returns {M3}.

^ we should add unittest to cover this.

cijothomas (changes requested):
See this comment : https://github.com/open-telemetry/opentelemetry-dotnet/pull/2916/files#r818252181

I think we are missing a case, when Views are involved.

src/OpenTelemetry/Metrics/MetricReaderExt.cs
```diff
@@ -39,12 +40,17 @@ internal Metric AddMetricWithNoViews(Instrument instrument)
...
+                if (this.instrumentIdentityToMetric.TryGetValue(instrumentIdentity, out var existingMetric))
+                {
+                    return existingMetric;
+                }
+
                 if (this.metricStreamNames.Contains(metricStreamName))
                 {
-                    OpenTelemetrySdkEventSource.Log.MetricInstrumentIgnored(metricName, instrument.Meter.Name, "Metric name conflicting with existing name.", "Either change the name of the instrument or change name using View.");
-                    return null;
+                    OpenTelemetrySdkEventSource.Log.DuplicateMetricInstrument(metricName, meterName, "Metric instrument has the same name as an existing one but differs by description, unit, or instrument type.", "Either change the name of the instrument or use MeterProviderBuilder.AddView to resolve the conflict.");
```
utpilla, line 53: Should we update the message? The user doesn't really need to resolve any conflict here, right?

src/OpenTelemetry/Metrics/MetricReaderExt.cs
```diff
@@ -27,6 +27,7 @@ namespace OpenTelemetry.Metrics
     public abstract partial class MetricReader
     {
         private readonly HashSet<string> metricStreamNames = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+        private readonly Dictionary<InstrumentIdentity, Metric> instrumentIdentityToMetric = new Dictionary<InstrumentIdentity, Metric>();
```
cijothomas, line 30: We need to clear an entry, upon instrument dispose as well.

src/OpenTelemetry/Metrics/MetricReaderExt.cs
alanwest: You are correct. Views hurt my mind 🤕. Fixed https://github.com/open-telemetry/opentelemetry-dotnet/pull/2916/commits/2dd0058ee90faaea17fcdecd606ca919145328f8

src/OpenTelemetry/Metrics/MetricReaderExt.cs
alanwest: Turns out there's a bit more to just freeing up an entry in this dictionary. I hadn't fully followed the dispose path, so I didn't notice that when an instrument is disposed the metric is freed up. Now that multiple instruments can modify a single metric we need to free up the metric only when no more instruments reference it.

Working on this...

src/OpenTelemetry/Metrics/InstrumentIdentity.cs
```diff
@@ -0,0 +1,74 @@
...
+        public InstrumentIdentity(string meterName, string instrumentName, string unit, string description, Type instrumentType)
+        {
+            this.MeterName = meterName;
+            this.InstrumentName = instrumentName;
+            this.Unit = unit;
+            this.Description = description;
+            this.InstrumentType = instrumentType;
+        }
+
+        public readonly string MeterName { get; }
```
alanwest, line 32: This is not quite right. The entire meter - that is, name, version, and schema url (when we eventually support it) should be a component of the identity. Fixing this...

alanwest: Fixed https://github.com/open-telemetry/opentelemetry-dotnet/pull/2916/commits/03a245e2815adbd5f4efa09827a7b2c98b04f074

src/OpenTelemetry/Metrics/MetricReaderExt.cs
alanwest: Ah ok, my initial understanding was incorrect. Instruments don't get disposed, meters do, so I think is less complicated than I had thought... https://github.com/open-telemetry/opentelemetry-dotnet/pull/2916/commits/0ff1d224122c2fd494826a15d72384a565126cce

src/OpenTelemetry/.publicApi/net461/PublicAPI.Unshipped.txt
```diff
@@ -63,7 +63,8 @@ OpenTelemetry.Metrics.MeterProviderExtensions
 OpenTelemetry.Metrics.Metric
 OpenTelemetry.Metrics.Metric.Description.get -> string
 OpenTelemetry.Metrics.Metric.GetMetricPoints() -> OpenTelemetry.Metrics.MetricPointsAccessor
-OpenTelemetry.Metrics.Metric.Meter.get -> System.Diagnostics.Metrics.Meter
+OpenTelemetry.Metrics.Metric.MeterName.get -> string
+OpenTelemetry.Metrics.Metric.MeterVersion.get -> string
```
alanwest, line 67: What do folks think about this change? In part I was thinking since Meter can be disposed, maybe it's safer to not keep a handle on it from the metric. Another option would be to create a new struct to encapsulate these fields.

```csharp
namespace OpenTelemetry.Metrics;

public struct InstrumentationScope
{
    string Name { get; }
    string Version { get; }
    // Eventually
    string SchemaUrl { get; }
}
```

src/OpenTelemetry/Metrics/MetricReaderExt.cs
cijothomas: I think we can also mention in the log that "SDK will still exporter duplicates". Else it is not clear if SDK is going to drop it or pass it to exporters.

src/OpenTelemetry/.publicApi/net461/PublicAPI.Unshipped.txt
cijothomas: This is a good improvement. Don't know if we need to introduce new Struct for this.

cijothomas: Need to add this to changelog as this is breaking any existing exporters.

cijothomas (approved):
LGTM.
Please modify changelog to warn export authors about breaking change from previous version.
Also a improved log message to indicate SDK is not dropping the metric.

reyang (approved):

src/OpenTelemetry/Metrics/InstrumentIdentity.cs
```diff
@@ -0,0 +1,80 @@
...
+                && this.InstrumentName == other.InstrumentName
+                && this.Unit == other.Unit
+                && this.Description == other.Description;
+        }
+
+        public readonly override int GetHashCode()
+        {
+            unchecked
+            {
+                int hash = 17;
```
reyang, line 69: Minor: would it help if the hash is calculated and stored during ctor?

src/OpenTelemetry/Metrics/Metric.cs
```diff
@@ -121,11 +118,15 @@ public sealed class Metric
...
 
-        public string Unit { get; private set; }
+        public string Unit => this.InstrumentIdentity.Unit;
 
-        public Meter Meter { get; private set; }
+        public string MeterName => this.InstrumentIdentity.MeterName;
+
+        public string MeterVersion => this.InstrumentIdentity.MeterVersion;
+
+        internal InstrumentIdentity InstrumentIdentity { get; set; }
```
utpilla, line 129: Make it `private set`?

src/OpenTelemetry/Metrics/Metric.cs
```diff
@@ -30,68 +30,65 @@ public sealed class Metric
         private readonly AggregatorStore aggStore;
 
         internal Metric(
-            Instrument instrument,
+            InstrumentIdentity instrumentIdentity,
             AggregationTemporality temporality,
             string metricName,
```
utpilla, line 35: We can get `metricName` from `InstrumentIdentity` now.

src/OpenTelemetry/Metrics/InstrumentIdentity.cs
utpilla: > This is not quite right. The entire meter - that is, name, version, and schema url (when we eventually support it) should be a component of the identity. Fixing this...

Wouldn't it be better to just have `Meter` as the only `public` property in that case? We would only need to update the `Equals()` check like below:
```
                ...
                && this.Meter.Name == other.Meter.Name
                && this.Meter.Version == other.Meter.Version
                && this.Meter.SchemaUrl == other.Meter.SchemaUrl 
                ...
```
This way if schema url gets added to `Meter` later on, we don't have to add a new `public` property dedicated just for that.

src/OpenTelemetry/Metrics/InstrumentIdentity.cs
alanwest: Interesting thought. Done.

src/OpenTelemetry/Metrics/InstrumentIdentity.cs
utpilla: Since this is an `internal` `struct`, it wouldn't matter as much if we have to add more properties later on, but I think we could just use `Meter` here as it provides the best encapsulation for everything `Meter` related.

alanwest: Do you mean keep a handle on the `Meter` itself? Same concern here https://github.com/open-telemetry/opentelemetry-dotnet/pull/2916#discussion_r819248034 would apply. `Meter` might be disposed so safer to not keep a handle on it.

alanwest: Maybe it'd be less of a concern to keep a handle on the meter since `InstrumentIdentity` is internal?

utpilla: Yeah, I was suggesting to have a handle to `Meter` itself as all of this is happening in the `InstrumentPublished` callback so the `Meter` would not be disposed in this path. But I overlooked the fact that in case of dictionary lookup collisions, the `Equals` check might fail as it might still have entries to instruments whose `Meter`s are disposed. This would only work if we can ensure that the dictionary would never have any instrument whose `Meter` is disposed.

cijothomas: Okay to park this into an issue and come back to this ? (so that we can merge and do a release.)

utpilla (approved):

src/OpenTelemetry/Metrics/MetricReaderExt.cs
cijothomas: And yet, everyone loves a room with a view! 😄

Issue #2558, opened 2021-11-02
Incorrect namespace issue


cijothomas:
The following classes are in the root OpenTelemetry ns, as opposed to being in .Trace or .Logs namespace. Given this was not identified before 1.0.0 stable, changing this now would be breaking change. This issue is opened for tracking fixing this when we do a major bump.

https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/SimpleActivityExportProcessor.cs#L19
https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/BatchActivityExportProcessor.cs#L19
https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/SimpleLogRecordExportProcessor.cs#L19
https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/BatchLogRecordExportProcessor.cs#L19

PR #2927, opened 2022-02-22
Minor clean up

reyang:
Related to #2558.

## Changes

* Moved the log/trace specific code to the relevant folders (keep the namespace as-is to avoid breaking changes).
* Moved `ExportProcessorType` to `BaseExportProcessor.cs` (is this a good change or not?).


cijothomas:
> * Moved the log/trace specific code to the relevant folders (keep the namespace as-is to avoid breaking changes).

This looks good.

> * Moved `ExportProcessorType` to `BaseExportProcessor.cs` (is this a good change or not?).

Not sure if it adds any value? Did you intend to reduce the # of files? Okay, if that was the intention!

reyang:
> Not sure if it adds any value? Did you intend to reduce the # of files? Okay, if that was the intention!

Yes, less files (especially if a file has less than 10 lines of code).

cijothomas (approved):

PR #2930, opened 2022-02-23
Removing codecov status in PRs

### LINKED ISSUE #2926: Code coverage can be confusing
Originally from here:
https://github.com/open-telemetry/opentelemetry-dotnet/pull/2915#issuecomment-1046083751

Copying the content:

The current code coverage result is not stable (each run will have slightly different results) and is not considered as a blocker for the PR.
I guess the codecov issue could either be related to the tool itself (e.g. some bug, although I think it's less likely to be the case) or the how the test cases were designed (e.g. some test cases require RPC interactions or some other operations that are sensitive to timing/race).

Given this is not the first time folks asked about this, I'm thinking - maybe for short term we should change the codecov CI so it only shows the coverage percentage but always return Success? In this way it still serves the purpose of giving us a rough estimation of the test coverage, but won't introduce confusion to the contributors? https://github.com/orgs/open-telemetry/teams/dotnet-approvers https://github.com/orgs/open-telemetry/teams/dotnet-maintainers what do you think?

eddynaka:
Fixes #2926.

## Changes


.github/codecov.yml
```diff
@@ -6,6 +6,7 @@ coverage:
   round: down
   range: "70...100"
   status:
+    project: off
```
eddynaka, line 9: With this change, we won't create the check for code coverage.

cijothomas (approved):

PR #2931, opened 2022-02-23
MetricReaderOptions for AddInMemoryExporter extension methods

alanwest:
As I was working through this PR I was originally feeling "🤮 I don't like it" and then I started feeling "🤔 I suppose this could work". At first I didn't think there'd ever be the need for a periodic InMemory exporter, but I guess we do have some tests that use it.

I'll let you all be the judge. Do you find it nice to have an extension method for the InMemory exporter that allows configuring the standard options for the underlying metric reader?


reyang:
> I'll let you all be the judge. Do you find it nice to have an extension method for the InMemory exporter that allows configuring the standard options for the underlying metric reader?

Yes +1

reyang (approved):

cijothomas (approved):

cijothomas:
> > I'll let you all be the judge. Do you find it nice to have an extension method for the InMemory exporter that allows configuring the standard options for the underlying metric reader?
> 
> Yes +1

The question is self-answered in this PR itself. The tests looks much nicer :)

PR #2939, opened 2022-02-24
Add comments and fix few todos in Metrics

cijothomas:
Towards #2621

### DIFF
--- a/src/OpenTelemetry/Metrics/ExplicitBucketHistogramConfiguration.cs
+++ b/src/OpenTelemetry/Metrics/ExplicitBucketHistogramConfiguration.cs
@@ -24,8 +24,8 @@ namespace OpenTelemetry.Metrics
         /// </summary>
         /// <remarks>
         /// The array must be in ascending order with distinct values.
-        /// Empty array means no histogram buckets will be generated.
-        /// Null means default bucket boundaries will be used.
+        /// An empty array would result in no histogram buckets being calculated.
+        /// A null value would result in default bucket boundaries being used.
         /// </remarks>
         public double[] Boundaries { get; set; }
     }
--- a/src/OpenTelemetry/Metrics/ExplicitBucketHistogramConfiguration.cs
+++ b/src/OpenTelemetry/Metrics/ExplicitBucketHistogramConfiguration.cs
@@ -24,6 +24,8 @@ namespace OpenTelemetry.Metrics
         /// </summary>
         /// <remarks>
         /// The array must be in ascending order with distinct values.
+        /// Empty array means no histogram buckets will be generated.
+        /// Null means default bucket boundaries will be used.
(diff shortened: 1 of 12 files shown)


test/Benchmarks/Metrics/MetricsBenchmarks.cs
```diff
@@ -20,7 +20,6 @@
 using System.Diagnostics.Metrics;
 using BenchmarkDotNet.Attributes;
 using OpenTelemetry;
-using OpenTelemetry.Exporter;
```
reyang, line 23: Do we know why these "unused using statements" are not discovered by the CI?

reyang (approved):

src/OpenTelemetry/Metrics/ExplicitBucketHistogramConfiguration.cs
```diff
@@ -24,6 +24,8 @@ public class ExplicitBucketHistogramConfiguration : MetricStreamConfiguration
         /// </summary>
         /// <remarks>
         /// The array must be in ascending order with distinct values.
+        /// Empty array means no histogram buckets will be generated.
```
utpilla, line 27: Rephrase "means" -> "would result in"
Something like:

- An empty array would result in no histogram buckets being used. 
- Null value would result in default histogram buckets being used.

src/OpenTelemetry/Metrics/PeriodicExportingMetricReader.cs
```diff
@@ -33,6 +38,12 @@ public class PeriodicExportingMetricReader : BaseExportingMetricReader
         private readonly ManualResetEvent shutdownTrigger = new ManualResetEvent(false);
         private bool disposed;
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="PeriodicExportingMetricReader"/> class.
+        /// </summary>
+        /// <param name="exporter">Exporter instance to export Metrics to.</param>
+        /// <param name="exportIntervalMilliseconds">The interval in milliseconds between two consecutive exports. The default value is 60000.</param>
+        /// <param name="exportTimeoutMilliseconds">How long the export can run before it is cancelled. The default value is 30000.</param>
```
utpilla, line 46: nit: canceled*

test/Benchmarks/Metrics/MetricsBenchmarks.cs
cijothomas: I think /test folder is not covered.

src/OpenTelemetry/Metrics/PeriodicExportingMetricReader.cs
cijothomas: will pass that.

utpilla (approved):

PR #2943, opened 2022-02-24
Update Metrics Benchmarks numbers

utpilla:
## Changes
- Update Metrics Benchmarks numbers
- Minor refactoring

### DIFF
--- a/test/Benchmarks/Metrics/MetricsBenchmarks.cs
+++ b/test/Benchmarks/Metrics/MetricsBenchmarks.cs
@@ -26,27 +26,27 @@ using OpenTelemetry.Tests;
 /*
 // * Summary *
 
-BenchmarkDotNet=v0.13.1, OS=Windows 10.0.19043.1288 (21H1/May2021Update)
-Intel Xeon CPU E5-1650 v4 3.60GHz, 1 CPU, 12 logical and 6 physical cores
-.NET SDK=6.0.100
-  [Host]     : .NET 6.0.0 (6.0.21.52210), X64 RyuJIT
-  DefaultJob : .NET 6.0.0 (6.0.21.52210), X64 RyuJIT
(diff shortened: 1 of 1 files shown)


cijothomas (approved):

test/Benchmarks/Metrics/MetricsBenchmarks.cs
```diff
@@ -26,27 +26,27 @@
...
+BenchmarkDotNet=v0.13.1, OS=Windows 10.0.22000
+Intel Core i7-9700 CPU 3.00GHz, 1 CPU, 8 logical and 8 physical cores
+.NET SDK=6.0.200
+  [Host]     : .NET 6.0.2 (6.0.222.6406), X64 RyuJIT
+  DefaultJob : .NET 6.0.2 (6.0.222.6406), X64 RyuJIT
+
+
+|                    Method | AggregationTemporality |      Mean |    Error |   StdDev | Allocated |
+|-------------------------- |----------------------- |----------:|---------:|---------:|----------:|
+|            CounterHotPath |             Cumulative |  16.60 ns | 0.120 ns | 0.094 ns |         - |
```
reyang, line 38: It seems the "Median" column is gone, any theory?

reyang (approved):

test/Benchmarks/Metrics/MetricsBenchmarks.cs
utpilla: From their documentation, it seems like `Median` is not always shown.
https://benchmarkdotnet.org/articles/configs/columns.html#statistics

> Only Mean will be always shown. If the distribution looks strange, BenchmarkDotNet could also print additional columns like Median or P95 (95th percentile).
>
> If you need specific statistics, you always could add them manually.

PR #2944, opened 2022-02-24
Update Histogram benchmarks

utpilla:
## Changes
- Update Histogram benchmarks to emit dimensions
- Use only `long` Histogram

### DIFF
--- a/test/Benchmarks/Metrics/HistogramBenchmarks.cs
+++ b/test/Benchmarks/Metrics/HistogramBenchmarks.cs
@@ -16,6 +16,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Diagnostics.Metrics;
 using System.Threading;
 using BenchmarkDotNet.Attributes;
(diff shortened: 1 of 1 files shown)


cijothomas (approved):

test/Benchmarks/Metrics/HistogramBenchmarks.cs
```diff
@@ -49,12 +62,12 @@ namespace Benchmarks.Metrics
     public class HistogramBenchmarks
     {
         private const int MaxValue = 1000;
-        private static readonly ThreadLocal<Random> ThreadLocalRandom = new(() => new Random());
-        private Histogram<long> histogramLong;
-        private Histogram<double> histogramDouble;
+        private Random random = new Random();
```
mic-max, line 65: Why make this not `ThreadLocal`? I was recommended to use it by @reyang due to potential very bad behaviour when using the same `Random` instance concurrently in a multi-threaded application.
Although, by default `benchmarkdotnet` uses a single thread.

PR #2945, opened 2022-02-25
Add stress tests for Histogram

utpilla:
## Changes
- Add stress tests for Histogram


test/OpenTelemetry.Tests.Stress/Skeleton.cs
```diff
@@ -41,7 +41,7 @@ static Program()
         StressMeter.CreateObservableGauge("Process.VirtualMemorySize64", () => process.VirtualMemorySize64);
     }
 
-    public static void Stress(int concurrency = 0, int prometheusPort = 0)
+    public static void Stress(Action run, int concurrency = 0, int prometheusPort = 0)
```
reyang, line 44: I guess Action would defeat the purpose of `[MethodImpl(MethodImplOptions.AggressiveInlining)] static void Run()`. What's the perf number (CPU cycles per loop) for [the naive test](https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/test/OpenTelemetry.Tests.Stress)?

reyang (approved):

test/OpenTelemetry.Tests.Stress/Meat.cs
```diff
@@ -22,7 +22,7 @@ public partial class Program
 {
     public static void Main()
     {
-        Stress(concurrency: 1, prometheusPort: 9184);
+        Stress(Run, concurrency: 1, prometheusPort: 9184);
```
reyang, line 25: This https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/test/OpenTelemetry.Tests.Stress#writing-your-own-stress-test needs to be updated accordingly.

test/OpenTelemetry.Tests.Stress/Skeleton.cs
utpilla: You're right! The perf did go down with this change for the naive test. CPU Cycles/Loop went from 3 to 5 with this change.

image

utpilla: I guess we could close this PR and anytime someone wants to run Stress Tests for Histogram they can just update the `Run` method.

reyang: Or add the metrics code but comment it out.

test/OpenTelemetry.Tests.Stress.Metrics/Program.cs
```diff
@@ -25,11 +25,16 @@ namespace OpenTelemetry.Tests.Stress;
 public partial class Program
 {
     private const int ArraySize = 10;
+
+    // private const int MaxHistogramMeasurement = 1000;
```
reyang, line 29: Maybe change this to (the same for other places):

```csharp
// Note: uncomment this if you want to run Histogram stress test
// private const int MaxHistogramMeasurement = 1000;
```

reyang (approved):

PR #2946, opened 2022-02-25
Improve GetUri in HttpInListener

Tornhoof:
## Changes
The current GetUri method looks very similar to the GetDisplayUrl() extension method in ASP.NET Core. For that method an open issue exists https://github.com/dotnet/aspnetcore/issues/28906 to improve performance and allocations.
In ASP.NET Core the method can be called by the developer to get an url for display purposes, in the Open Telemetry Listener it is called for each and every request with ``IsAllDataRequested``. So the allocation impact is higher here than in ASP.NET Core.

This PR replaces the existing method with two code paths, one for .NET Standard 2.0 and one for 2.1. The 2.0 code path is the old method with added length calculation to the StringBuilder. The new 2.1 code path uses string.Create to directly create the Display Url based on the calculated length and the supplied values. It is slightly different from the suggested method in the issue above as not everything is available in .NET Standard 2.1.

I changed the tests for the listener to include a query string.


Tornhoof:
Apparently had a very outdated version of the repo around. will do different PR.

PR #2978, opened 2022-03-07
Enable PackageReadMe

### LINKED ISSUE #2977: Nuget readme
Modify nuget to include readme. I think we can put the readme.md file under each project, or simply point to it.
https://docs.microsoft.com/en-us/nuget/nuget-org/package-readme-on-nuget-org

WeihanLi:
Fixes #2977 

## Changes

Enable `PackageReadMe` for nuget packages

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -4,9 +4,6 @@
     <PackageReadmeFile>README.md</PackageReadmeFile>
   </PropertyGroup>
   <ItemGroup>
-    <None Include="README.md">
-      <Pack>True</Pack>
-      <PackagePath>\</PackagePath>
-    </None>
+    <None Include="README.md" Pack="true" PackagePath="\" />
   </ItemGroup>
 </Project>
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -1,3 +1,12 @@
 <Project>
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildProjectDirectory), 'OpenTelemetry.sln'))\build\Common.prod.props" />
+  <PropertyGroup>
+    <PackageReadmeFile>README.md</PackageReadmeFile>
+  </PropertyGroup>
+  <ItemGroup>
+    <None Include="README.md">
+      <Pack>True</Pack>
+      <PackagePath>\</PackagePath>
+    </None>
+  </ItemGroup>
 </Project>


src/Directory.Build.props
```diff
@@ -1,3 +1,12 @@
 <Project>
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildProjectDirectory), 'OpenTelemetry.sln'))\build\Common.prod.props" />
+  <PropertyGroup>
+    <PackageReadmeFile>README.md</PackageReadmeFile>
+  </PropertyGroup>
+  <ItemGroup>
+    <None Include="README.md">
```
reyang, line 7: Would this introduce broken links if the README.md file points to other files?

WeihanLi: Think we could override it in the related project files, it seems ok for most packages

src/Directory.Build.props
```diff
@@ -1,3 +1,9 @@
 <Project>
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildProjectDirectory), 'OpenTelemetry.sln'))\build\Common.prod.props" />
+  <PropertyGroup>
+    <PackageReadmeFile>README.md</PackageReadmeFile>
```
cijothomas, line 4: if we simply put the readme.md file in the nuget, it won't work as there are many relative links to this readme.md file, which'll be broken.

reyang (changes requested):
I'm blocking this PR since we need to make sure the README.md files are useful/helpful in the NuGet context.

With the links to other docs and examples, I'm worried that users might just get more confused.

On a separate note, I guess the *.md file in NuGet is intended to be short, with necessary information to get folks started + links if they want to read more?

WeihanLi:
From some packages I had seen, the README could be the same as Github project README, for example <https://www.nuget.org/packages/Ben.Demystifier/> 

I think the package readme is a description and get-started document for the package, the user could get more by the project website link, it would be better if we could provide a better README experience

Some thoughts on the relative links:
- A standalone README file for the NuGet package
- Custome MSBuild task to convert the relative links for absolute links
- (More ...

cijothomas:
> Some thoughts on the relative links:
> 
> A standalone README file for the NuGet package
> Custome MSBuild task to convert the relative links for absolute links

Either of the above seems okay. Could we have nuget specific readme have very minimal content, and link to the main readme?

WeihanLi:
> Could we have nuget specific readme have very minimal content, and link to the main readme?

Sure

PR #2981, opened 2022-03-07
Make console examples for metrics simpler

cijothomas:
Run till key press.

### DIFF
--- a/examples/Console/TestMetrics.cs
+++ b/examples/Console/TestMetrics.cs
@@ -18,7 +18,6 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Diagnostics.Metrics;
-using System.Threading;
 using System.Threading.Tasks;
 using OpenTelemetry;
 using OpenTelemetry.Exporter;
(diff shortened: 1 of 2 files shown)


examples/Console/TestMetrics.cs
```diff
@@ -116,80 +115,51 @@ internal static object Run(MetricsOptions options)
                 });
             }
 
-            var cts = new CancellationTokenSource();
+            System.Console.WriteLine("Press any key to exit.");
```
reyang, line 118: Curious, would we want "any key" or "ESC"? https://github.com/open-telemetry/opentelemetry-dotnet/blob/6e31ab66d2610216c3bae2cdbf9fc8c4d6aca0d1/test/OpenTelemetry.Tests.Stress/Skeleton.cs#L124

I noticed that sometimes folks pressed Enter twice and the application started then exited.

reyang (approved):

utpilla (approved):

examples/Console/TestMetrics.cs
cijothomas: will revisit this together with Prometheus example (it also uses Any Key today)

cijothomas: https://github.com/open-telemetry/opentelemetry-dotnet/pull/2983

PR #2982, opened 2022-03-07
Default to periodic metric reader for push based exporters

### LINKED ISSUE #2979: Metrics are no longer flushed automatically
# Bug Report

* `OpenTelemetry.Exporter.Console`: 1.2.0-rc3
* `OpenTelemetry.Exporter.InMemory`: 1.2.0-rc3
* `OpenTelemetry.Exporter.OpenTelemetryProtocol`: 1.2.0-rc3
* `OpenTelemetry.Extensions.Hosting`: 1.0.0-rc10

* `net6.0`

## Symptom

Metrics being flushed automatically every 1 minute

Metrics were not being flushed and diagnostic logs don't provide any entries other than null.

However, as mitigation, I need to call `ForceFlush` to emit the metrics.

## Reproduce

When constructing a meter provider as following, it used to automatically flush metrics:
```csharp
var configuredMetricProvider = Sdk.CreateMeterProviderBuilder()
                .AddMeter("Example.*")
                .AddConsoleExporter()
                .AddOtlpExporter(options =>
                {
                    options.Endpoint = new Uri(collectorUri);
                })
                // We can only register global histogram buckets, not per histogram (for now)
                // See https://github.com/open-telemetry/opentelemetry-dotnet/issues/2755
                .AddView("*", new ExplicitBucketHistogramConfiguration { Boundaries = histogramBuckets })
                .Build();
```

alanwest:
Fixes #2979.

The OTLP exporter was mistakingly defaulted to the manual metric reader.

With this PR:
* The periodic reader is reestablished as the default for the OTLP exporter with a default interval of 60 seconds.
* The Console and InMemory exporters now default to using the periodic reader, but with an infinite interval by default (i.e., effectively the same as manual). There will be a spec PR to follow to define what the default interval should be for Console and InMemory exporters.
* As a consequence, the `PeriodicExportingMetricReader` now allows an infinite export interval.


src/OpenTelemetry.Exporter.Console/ConsoleExporterMetricsExtensions.cs
```diff
@@ -97,14 +97,11 @@ public static MeterProviderBuilder AddConsoleExporter(this MeterProviderBuilder
 
             var metricExporter = new ConsoleMetricExporter(exporterOptions);
 
-            if (metricReaderOptions.MetricReaderType == (MetricReaderType)(-1))
-            {
-                metricReaderOptions.MetricReaderType = MetricReaderType.Manual;
-            }
-
             var metricReader = metricReaderOptions.MetricReaderType == MetricReaderType.Manual
```
alanwest, line 100: I'm considering whether we should just get rid of `MetricReaderType.Manual`? My vote is yes and we use the periodic reader with an infinite interval where appropriate. This idea originated from this suggestion https://github.com/open-telemetry/opentelemetry-specification/pull/2378#discussion_r815327718

reyang: I would vote for getting rid of "Manual" at least for now, because:
1. The word "Manual" might be misleading. E.g. one could build a reader/exporter that exports metrics based on network connectivity (e.g. only exports data when running on Wifi or LAN), battery status (e.g. only exports data when batter is >50% or the device is charging).
2. If later we want "Manual", it can be an additive change.

alanwest: This PR fixes the immediate bug. I'll remove `MetricReaderType` in a follow up PR.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpMetricExporterExtensions.cs
```diff
@@ -102,14 +102,11 @@ public static MeterProviderBuilder AddOtlpExporter(this MeterProviderBuilder bui
...
-            {
-                metricReaderOptions.MetricReaderType = MetricReaderType.Periodic;
-            }
-
             var metricReader = metricReaderOptions.MetricReaderType == MetricReaderType.Manual
                 ? new BaseExportingMetricReader(metricExporter)
-                : new PeriodicExportingMetricReader(metricExporter, metricReaderOptions.PeriodicExportingMetricReaderOptions.ExportIntervalMilliseconds);
+                : new PeriodicExportingMetricReader(
+                    metricExporter,
+                    metricReaderOptions.PeriodicExportingMetricReaderOptions.ExportIntervalMilliseconds ?? 60000);
```
tomkerkhove, line 109: Idea - Might be good to have the default outside of this extension in a constant somewhere; but I believe it is already defined somewhere else so using that is OK as well

tomkerkhove (approved):

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpMetricExporterExtensions.cs
alanwest: Moved to a constant. The constant already defined is part of the SDK project and is internal so not accessible here.

alanwest: Actually come to think of it the right thing to do here is to just do `new PeriodicExportingMetricReader(metricExporter);`. I've got a follow up PR to remove `MetricReaderType`. I'll clean that up in that PR.

cijothomas (approved):

PR #2974, opened 2022-03-04
add filtering instructions to ILogger getting started doc

TimothyMothra:
Fixes #.

## Changes

To configure filtering rules, a user must know either the class name `OpenTelemetryLoggingProvider` or it's alias "OpenTelemetry".
This change adds this to the doc and gives concrete configuration examples.

For significant contributions please make sure you have completed the following items:

* [ ] Appropriate `CHANGELOG.md` updated for non-trivial changes
* [ ] Design discussion issue #
* [ ] Changes in public API reviewed

PR #2995, opened 2022-03-08
add doc for ILogger filtering (OpenTelemetryLoggerProvider)

TimothyMothra:
Supersedes #2974.

## Changes

To configure filtering rules, a user must know either the class name `OpenTelemetryLoggingProvider` or it's alias `OpenTelemetry`.
This change adds a doc and working example.


cijothomas (changes requested):
it is not required to extensively document Log filtering as it is not an OpenTelemetry concept, and is fully provided by the Ms.Ext.Logging itself.

Also we don't want to add a hosted application demo app, just to show filtering.

I'd suggest to have something like customizing-the-sdk, similar to Metrics docs, and it can various customization options available for Logging. (IncludeFormattedMessage etc.), and Filtering could be mentioned there with a simple one liner example.

TimothyMothra:
> we don't want to add a hosted application demo app, just to show filtering.

Why don't we want a hosted application?
appsettings.json is the preferred option for configuring ILogger. I don't see an obvious way to demo this without a hosted app.


> I'd suggest to have something like customizing-the-sdk

I'm happy to work on a broader configuration document. I'm going to open a separate issue to identity and track the scope.

cijothomas:
> > we don't want to add a hosted application demo app, just to show filtering.
> 
> Why don't we want a hosted application? appsettings.json is the preferred option for configuring ILogger. I don't see an obvious way to demo this without a hosted app.

The key here is - this repo doesn't own the filtering. Its fully handled by Ms.Extensions.Logging. We dont need to demo appsettings.json or anything related to it. We sure need to mention the name of the Provider, and its alias, so users can apply filtering.

PR #2015, opened 2021-04-27
Improve Visual Studio Code support

pellared:
## Why

Make it easier (or even possible) to develop [OTel .NET](https://github.com/open-telemetry/opentelemetry-dotnet) using VS Code. Right now if you try to use VS Code you would get tons of errors. Moreover, for me, Intellisense was not even working.

## Changes

- Fixes in `.csproj` to reduce the number of "problems". For me, it reduced the number of problems from ~400 to ~100. It was discussed on Slack [here](https://cloud-native.slack.com/archives/C01N3BC2W7Q/p1617727862056600). Most of the ones that left are ASP.NET related which is not worth fixing - it would probably require a dedicated `.sln` for non-Windows dev environments. Unfortunately, AFAIK the [C# extension](https://github.com/OmniSharp/omnisharp-vscode) does not support `*.proj` files. 
- Add `.vscode` and `.devcontainer` to `.gitignore` in case someone wants to use different settings

## Testing

I was testing everything in Development Container (on Windows) that is created in this PR.

It would be great if someone double-checks if this PR does not break anything for Visual Studio. Low probability, but worth ensuring.

PR #2997, opened 2022-03-09
Improve VS Code and cross-platform dev experience

tillig:
## Changes

- Fixed changes that had reverted #2015: VS Code (OmniSharp) looks at the _first_ framework listed in `TargetFrameworks` and uses that for things like code analysis and IntelliSense. Having them ordered from highest to lowest .NET Core, then the .NET desktop framework (e.g., `net461`) last helps.
- Added `.vscode` folder with tasks for build and test as well as default settings and a couple extension recommendations to bootstrap folks. For example, C# dev tools aren't built into VS Code, they're an extension - so that's recommended.
- Added `global.json` to help with tooling determining which version of the .NET SDK to use during build when `dotnet` is invoked.
- Fixed one redundant cast analysis warning which was the only warning that was really throwing OmniSharp for a loop, at least on Mac. It was saying "System.Enum is not defined" but the underlying problem was the redundant cast.

There are definitely still some cross-platform challenges in VS Code. For example, while the `Microsoft.NETFramework.ReferenceAssemblies` package allows you to _compile_, OmniSharp analysis really just doesn't work so there are ~549 errors that all are basically `The type 'XXXXX' is defined in an assembly that is not referenced. You must add a reference to assembly 'netstandard, Version=2.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'. [OpenTelemetry.Instrumentation.AspNet.TelemetryHttpModule]` Build works great, OmniSharp, not so much.

Regardless, this will help folks with VS Code get up and running a little easier, I think.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,6 @@
 *.user
 *.userosscache
 *.sln.docstates
-.vscode/
 .devcontainer/
 
 # User-specific files (MonoDevelop/Xamarin Studio)
@@ -227,7 +226,7 @@ ClientBin/
 *.publishsettings
 orleans.codegen.cs
 
-# Including strong name files can present a security risk 
+# Including strong name files can present a security risk
 # (https://github.com/github/gitignore/pull/2483#issue-259490424)
 #*.snk
 
@@ -326,7 +325,7 @@ __pycache__/
 # OpenCover UI analysis results
 OpenCover/
 
-# Azure Stream Analytics local run output 
+# Azure Stream Analytics local run output
 ASALocalRun/
 
 # MSBuild Binary and Structured Log
@@ -335,7 +334,7 @@ ASALocalRun/
 # NVidia Nsight GPU debugger configuration file
 *.nvuser
 
-# MFractors (Xamarin productivity tool) working folder 
+# MFractors (Xamarin productivity tool) working folder
 .mfractor/
 
 # SonarQube folder
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,8 +1,7 @@
 {
   "recommendations": [
     "editorconfig.editorconfig",
-    "formulahendry.dotnet-test-explorer",
-    "ms-dotnettools.csharp",
-    "streetsidesoftware.code-spell-checker"
+    "davidanson.vscode-markdownlint",
+    "ms-dotnettools.csharp"
   ]
 }
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,8 @@
+{
+  "recommendations": [
+    "editorconfig.editorconfig",
+    "formulahendry.dotnet-test-explorer",
+    "ms-dotnettools.csharp",
+    "streetsidesoftware.code-spell-checker"
+  ]
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,8 +1,0 @@
-{
-  "cSpell.words": [
-    "cref",
-    "otel",
-    "unvalidated"
-  ],
-  "dotnet-test-explorer.testProjectPath": "test/**/*.Tests.csproj"
-}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,9 @@
+{
+  "cSpell.words": [
+    "cref",
+    "otel",
+    "unvalidated"
+  ],
+  "dotnet-test-explorer.runInParallel": true,
+  "dotnet-test-explorer.testProjectPath": "test/**/*.Tests.csproj"
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -4,6 +4,5 @@
     "otel",
     "unvalidated"
   ],
-  "dotnet-test-explorer.runInParallel": true,
   "dotnet-test-explorer.testProjectPath": "test/**/*.Tests.csproj"
 }
--- a/global.json
+++ b/global.json
@@ -1,0 +1,6 @@
+{
+  "sdk": {
+    "rollForward": "latestFeature",
+    "version": "6.0.200"
+  }
+}
--- a/test/OpenTelemetry.Tests.Stress/OpenTelemetry.Tests.Stress.csproj
+++ b/test/OpenTelemetry.Tests.Stress/OpenTelemetry.Tests.Stress.csproj
@@ -1,6 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <OutputType>Exe</OutputType>
+    <!-- OmniSharp/VS Code requires TargetFrameworks to be in descending order for IntelliSense and analysis. -->
     <TargetFrameworks>net6.0;net5.0;netcoreapp3.1;net462</TargetFrameworks>
     <IsPackable>false</IsPackable>
   </PropertyGroup>
--- a/test/OpenTelemetry.Tests.Stress/OpenTelemetry.Tests.Stress.csproj
+++ b/test/OpenTelemetry.Tests.Stress/OpenTelemetry.Tests.Stress.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFrameworks>netcoreapp3.1;net5.0;net6.0;net462</TargetFrameworks>
+    <TargetFrameworks>net6.0;net5.0;netcoreapp3.1;net462</TargetFrameworks>
     <IsPackable>false</IsPackable>
   </PropertyGroup>
 
--- a/.github/workflows/dotnet-format.yml
+++ b/.github/workflows/dotnet-format.yml
@@ -20,10 +20,10 @@ jobs:
     - name: check out code
       uses: actions/checkout@v2
 
-    - name: Setup .NET Core 3.1
+    - name: Setup .NET Core 6.0
       uses: actions/setup-dotnet@v1
       with:
-        dotnet-version: 3.1.x
+        dotnet-version: 6.0.x
 
     - name: Install format tool
       run: dotnet tool install -g dotnet-format
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,0 +1,40 @@
+{
+  "tasks": [
+    {
+      "args": [
+        "build",
+        "${workspaceFolder}/OpenTelemetry.sln",
+        "/property:GenerateFullPaths=true",
+        "/consoleloggerparameters:NoSummary"
+      ],
+      "command": "dotnet",
+      "group": {
+        "isDefault": true,
+        "kind": "build"
+      },
+      "label": "build",
+      "presentation": {
+        "reveal": "silent"
+      },
+      "problemMatcher": "$msCompile",
+      "type": "shell"
+    },
+    {
+      "args": [
+        "test",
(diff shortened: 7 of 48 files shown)


global.json
```diff
@@ -0,0 +1,6 @@
+{
+  "sdk": {
+    "rollForward": "latestFeature",
+    "version": "6.0.200"
```
reyang, line 4: Is this needed? I think in general we want to have less files sitting in the top folder if possible.

test/OpenTelemetry.Tests.Stress/OpenTelemetry.Tests.Stress.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFrameworks>netcoreapp3.1;net5.0;net6.0;net462</TargetFrameworks>
+    <TargetFrameworks>net6.0;net5.0;netcoreapp3.1;net462</TargetFrameworks>
```
reyang, line 4: Why do we have these changes? Is similar thing guarded by the CI in the future? (if not, I think it could run out of sync across files)

reyang: Based on the PR description, I'm guessing this is to workaround OmniSharp on VS Mac?

global.json
tillig: This helps keep folks working with a more predictable build experience, and you can't put it elsewhere, so... Unfortunately I think so. This is pretty common when in other .NET projects, too, even just CLI with no IDE.

test/OpenTelemetry.Tests.Stress/OpenTelemetry.Tests.Stress.csproj
tillig: That's exactly what it is. OmniSharp. It's not specific to Mac - it's a general OmniSharp problem on all platforms.

global.json
tillig: Just occurred to me - the build failed because `dotnet format` was running with an SDK _lower than_ the actual highest target for the project. (I'll have to fix that in the PR.) If various tools happen to behave differently based on target frameworks, you can get incorrect build results. ("Worked on my machine but not the build server!") That's what this helps with, and it already caught something.

test/OpenTelemetry.Tests.Stress/OpenTelemetry.Tests.Stress.csproj
cijothomas: Maybe add a comment above the targetframeworks tag to state the need to have this specific order, so none accidentally reorders this.
(Eventually we need to move it to common place, but thats definitely a separate PR.)

reyang: Here is my guess - without proper CI in place, it is very likely that other contributors might break it again - requiring the maintainers to pay close attention to such details in every single PR is unlikely to sustain.

Consider adding a simple script to validate these or having some OmniSharp CI job https://github.com/open-telemetry/opentelemetry-dotnet/blob/5bb9f26100a0fff8996c0b861b12e27bcf513f79/build/sanitycheck.py.

cijothomas (approved):
This looks reasonable, left a minor comment about this.
Would be great if we get another approver who is a VSCode user to take a look. @pellared ? @alanwest ?

reyang (approved):
LGTM. I have some concern regarding the [lack of CI](https://github.com/open-telemetry/opentelemetry-dotnet/pull/2997#discussion_r822279043), but I think overall this is a good movement, it is okay to me if want to have the CI later.

test/OpenTelemetry.Tests.Stress/OpenTelemetry.Tests.Stress.csproj
cijothomas: a CI job validating this is definitely safer, but a simple comment would also work - its not common for PRs to touch target frameworks.

global.json
cijothomas: https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/CONTRIBUTING.md#development-environment we need to update contributing guide to install net6.0?

.gitignore
```diff
@@ -8,7 +8,6 @@
 *.user
 *.userosscache
 *.sln.docstates
-.vscode/
```
pellared, line 11: I would consider keeping this line and moving all VSCode config to `.vscode.example` so that the developer could tune the configuration to their own preferences.

But we can also consider removing them. See a previous comment by  @reyang (@bruno-garcia had a similar one): https://github.com/open-telemetry/opentelemetry-dotnet/pull/2015#issuecomment-828624988

.vscode/extensions.json
```diff
@@ -0,0 +1,8 @@
+{
+  "recommendations": [
+    "editorconfig.editorconfig",
+    "formulahendry.dotnet-test-explorer",
```
pellared, line 4: Personally, I do not recommend it as:
1. It slows down the VS Code.
2. Has quite of an [issues](https://github.com/formulahendry/vscode-dotnet-test-explorer/issues)
3. Is not maintained by any known/trusted organization
4. I find the test integration in the C# extension good enough

.vscode/extensions.json
```diff
@@ -0,0 +1,8 @@
...
+    "ms-dotnettools.csharp",
+    "streetsidesoftware.code-spell-checker"
```
pellared, line 6: I think it is more of a personal preference than coupled to this repo. I suggest removing it.

.vscode/extensions.json
```diff
@@ -0,0 +1,8 @@
+{
+  "recommendations": [
```
pellared, line 2: We can consider adding https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint as there is even a GitHub workflow which uses `markdownlint-cli` to check the markdown files.

.vscode/settings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "cSpell.words": [
+    "cref",
+    "otel",
+    "unvalidated"
+  ],
+  "dotnet-test-explorer.testProjectPath": "test/**/*.Tests.csproj"
+}
```
pellared, line 8: Personally, I am not using any of these extensions and I think that these configs would get outdated.

pellared (commented):
> There are definitely still some cross-platform challenges in VS Code.

TBH I think you have done the best you could. Especially if one would like to develop e.g. on Linux or MacOs. Nice job 👍 

In general, I do not see any blocking comments. However, I have concerns about the VS Code config-related changes (`.vscode` dir and `global.json` file). I would prefer to have it as a separate PR. It will be easier to roll back if one would find it harmful.

.gitignore
tillig: The `CONTRIBUTING.md` doc specifically lists VS Code as a supported way to work on this project. The whole point of providing the `.vscode` folder is to allow that to happen. [The opinion on this has changed since that discussion in #2015.](https://github.com/open-telemetry/opentelemetry-dotnet/pull/2969#issuecomment-1062145699)

.vscode/settings.json
tillig: I disagree those particular configs would get outdated - they'd be maintained as the project grows. Being able to catch spelling errors is pretty helpful and used to be something FxCop would do. See, for example, [this `jeager-host` typo](https://github.com/open-telemetry/opentelemetry-dotnet/blob/8b8b9348aafeb4629b6a8261e327dccac33e9a0d/test/OpenTelemetry.Exporter.Jaeger.Tests/JaegerExporterOptionsTests.cs#L57).

And I'm pretty sure the test projects have a pattern that wouldn't change, but I feel much less strongly about whether folks want the .NET test explorer.

Regardless, I don't find value in arguing it. It seems like it'd be valuable to help folks - at least VS Code folks - avoid simply typos, but if it's not interesting I can remove it.

tillig: It does appear [there's a spelling check that runs during the `sanitycheck` build stage](https://github.com/open-telemetry/opentelemetry-dotnet/blob/8b8b9348aafeb4629b6a8261e327dccac33e9a0d/.github/workflows/sanitycheck.yml#L23) but I'm not sure where its dictionary is or why it's not erring out things like `jeager`.

.vscode/extensions.json
tillig: Reduced this list to just EditorConfig, C#, and Markdownlint.

.vscode/extensions.json
tillig: Done.

.vscode/settings.json
tillig: Entirety of `settings.json` removed.

.vscode/extensions.json
tillig: Added.

tillig:
I'm _really, really_ not interested in going back here to split out the .csproj and .vscode stuff into two separate PRs. If folks don't like the .vscode stuff, it's "delete the folder and add back to .gitignore" which is a pretty simple operation.

global.json
tillig: Updated `CONTRIBUTING.md` to list .NET 6.0.

test/OpenTelemetry.Tests.Stress/OpenTelemetry.Tests.Stress.csproj
tillig: I added a comment to every `TargetFrameworks` node:

`<!-- OmniSharp/VS Code requires TargetFrameworks to be in descending order for IntelliSense and analysis. -->`

Admittedly, this is _technically inaccurate_: it picks the first one listed. However, it also seemed like having the frameworks listed in a sensible/predictable order was valuable; and it didn't seem like writing a giant paragraph explaining the technical ins-and-outs of the overlap of OmniSharp and ordering was valuable.

However, if you have preferred alternate verbiage, I can run a giant search and replace on the existing comment and update accordingly.

pellared:
> I'm _really, really_ not interested in going back here to split out the .csproj and .vscode stuff into two separate PRs. If folks don't like the .vscode stuff, it's "delete the folder and add back to .gitignore" which is a pretty simple operation.


I was not aware that the maintainers changed their mind. Approving.

test/OpenTelemetry.Tests.Stress/OpenTelemetry.Tests.Stress.csproj
cijothomas: this looks reasonable.

pellared (approved):

cijothomas:
> > I'm _really, really_ not interested in going back here to split out the .csproj and .vscode stuff into two separate PRs. If folks don't like the .vscode stuff, it's "delete the folder and add back to .gitignore" which is a pretty simple operation.
> 
> I was not aware that the maintainers changed their mind. Approving.

Thanks for reviewing. 
We can revisit the decision, if the VSCode things becomes a headache to maintain.

alanwest (commented):
> This looks reasonable, left a minor comment about this.
> Would be great if we get another approver who is a VSCode user to take a look. @pellared ? @alanwest ?

Sorry I'm coming to this late... but

Holy crap. **Hard approve**. Just pulled latest and opened up VSCode. Things are way more seamless - especially when conditional compilation is involved.

I did have to make sure I had the latest .NET SDK installed, but that seems reasonable to me.

Thank you, @tillig!

PR #3003, opened 2022-03-10
[Part1] Support Activity Status and status description in Zipkin Exporter.

Yun-Ting:
Fixes parts of https://github.com/open-telemetry/opentelemetry-dotnet/issues/2569.

Added support for Activity.Status/Description in ZipkinExporter. 
System.Diagnostic.DiagnosticSource version 6.0.0 introduced native support for storing status/description in the Activity itself. 

This PR modified the exporter to retrieve status from the newly added fields if it was set.
To maintain backward compatibility, the control flow would fell back to retrieve status from the activityTags if the native status/description were not set.

Background : Please review Console.Exporter where the same logic were exercised: https://github.com/open-telemetry/opentelemetry-dotnet/pull/3061.

Next items : Add similar support for OTLP/Jaeger Exporters.


src/OpenTelemetry.Exporter.Zipkin/Implementation/ZipkinActivityConversionExtensions.cs
```diff
@@ -45,7 +45,20 @@ internal static ZipkinSpan ToZipkinSpan(this Activity activity, ZipkinEndpoint l
...
+                    ref tagState.Tags,
+                    new KeyValuePair<string, object>(
+                        SpanAttributeConstants.StatusCodeKey,
+                        StatusHelper.GetTagValueForActivityStatusCode(activity.Status)));
+            }
+            else
+            {
+                // In the rare cases when both ActivityStatus and Status Tags were set,
+                // ActivityStatus takes precedence over status tags.
+                activity.EnumerateTags(ref tagState);
```
vishweshbankwar, line 60: This is still needed for other tags. There can be additional tags on activity.

Yun-Ting: Addressed the issue by marking
``
tagState.StatusCode = StatusCode.Unset;
tagState.StatusDescription = string.Empty;
``
in the case when both activity status and status tag were set.
And updated the testcases to cover this scenario.
Thank you.

test/OpenTelemetry.Exporter.Zipkin.Tests/Implementation/ZipkinActivityConversionTest.cs
```diff
@@ -125,5 +126,110 @@ public void ToZipkinSpan_Status_ErrorFlagTest(StatusCode expectedStatusCode, str
...
+            }
+        }
+
+        [Fact]
+        public void ActivityStatus_Takes_precedence_Over_Status_Tags_ActivityStatusCodeIsOk()
+        {
+            // Arrange.
+            var activity = ZipkinExporterTests.CreateTestActivity();
+            activity.SetStatus(ActivityStatusCode.Ok);
+            activity.SetTag(
```
vishweshbankwar, line 177: I think there is an issue with this test. Setting the status code tag value to something outside of pre-defined status values will result in value not being added to exported tags.

https://github.com/open-telemetry/opentelemetry-dotnet/blob/58f13ba7e2a7f31c82db67c8c4999d7f90a78ce7/src/OpenTelemetry.Api/Internal/StatusHelper.cs#L47

vishweshbankwar, line 177: I think there is an issue with this test. Setting the status code tag value to something outside of pre-defined status values will result in value not being added to exported tags.

https://github.com/open-telemetry/opentelemetry-dotnet/blob/58f13ba7e2a7f31c82db67c8c4999d7f90a78ce7/src/OpenTelemetry.Api/Internal/StatusHelper.cs#L47

Yun-Ting: Fixed.

Yun-Ting: Fixed.

src/OpenTelemetry.Api/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Support Activity Status and status description in Zipkin Exporter.
```
vishweshbankwar, line 5:
```suggestion
* Support Activity `Status` and `StatusDescription` in Zipkin Exporter.
```
```suggestion
* Support Activity Status and status description in Zipkin Exporter.
```

vishweshbankwar, line 5:
```suggestion
* Support Activity `Status` and `StatusDescription` in Zipkin Exporter.
```
```suggestion
* Support Activity Status and status description in Zipkin Exporter.
```

src/OpenTelemetry.Exporter.Zipkin/Implementation/ZipkinActivityConversionExtensions.cs
```diff
@@ -72,14 +72,43 @@ internal static ZipkinSpan ToZipkinSpan(this Activity activity, ZipkinEndpoint l
...
+                    PooledList<KeyValuePair<string, object>>.Add(
+                        ref tagState.Tags,
+                        new KeyValuePair<string, object>(
+                            ZipkinErrorFlagTagName,
+                            activity.StatusDescription ?? string.Empty));
+                }
+            }
+            else if (tagState.StatusCode.HasValue && tagState.StatusCode != StatusCode.Unset)
+            {
+                // In the case when both activity status and status tag were set,
```
vishweshbankwar, line 95: nit: Move this comment on the top before checking activity status.

vishweshbankwar, line 95: nit: Move this comment on the top before checking activity status.

src/OpenTelemetry.Exporter.Zipkin/Implementation/ZipkinActivityConversionExtensions.cs
```diff
@@ -207,7 +228,8 @@ public bool ForEach(KeyValuePair<string, object> activityTag)
                         return true;
                     }
                 }
-                else if (activityTag.Value is int intVal && activityTag.Key == SemanticConventions.AttributeNetPeerPort)
+
+                if (activityTag.Value is int intVal && activityTag.Key == SemanticConventions.AttributeNetPeerPort)
```
vishweshbankwar, line 232: Is this change intentional?

src/OpenTelemetry.Exporter.Zipkin/Implementation/ZipkinActivityConversionExtensions.cs
```diff
@@ -72,14 +72,43 @@ internal static ZipkinSpan ToZipkinSpan(this Activity activity, ZipkinEndpoint l
                 }
             }
 
-            if (tagState.StatusCode == StatusCode.Error)
+            if (activity.Status == ActivityStatusCode.Ok || activity.Status == ActivityStatusCode.Error)
```
vishweshbankwar, line 75: activity.Status != ActivityStatusCode.Unset?

Yun-Ting: Logically the same, but I'm thinking maybe the current shape would have better readability?

src/OpenTelemetry.Exporter.Zipkin/Implementation/ZipkinActivityConversionExtensions.cs
Yun-Ting: Nope and fixed. Thanks!

test/OpenTelemetry.Exporter.Zipkin.Tests/Implementation/ZipkinActivityConversionTest.cs
```diff
@@ -125,5 +126,108 @@ public void ToZipkinSpan_Status_ErrorFlagTest(StatusCode expectedStatusCode, str
...
+            // Assert.
+            Assert.Equal(
+                StatusHelper.GetTagValueForActivityStatusCode(ActivityStatusCode.Ok),
+                zipkinSpan.Tags.FirstOrDefault(t => t.Key == SpanAttributeConstants.StatusCodeKey).Value);
+
+            Assert.Contains(zipkinSpan.Tags, t => t.Key == "otel.status_code" && (string)t.Value == "OK");
+            Assert.DoesNotContain(zipkinSpan.Tags, t => t.Key == "otel.status_code" && (string)t.Value == "ERROR");
+
+            // Ensure additional Activity tags were being converted.
+            Assert.Contains(zipkinSpan.Tags, t => t.Key == "myCustomTag" && (string)t.Value == "myCustomTagValue");
```
vishweshbankwar, line 194: (Not for this PR) - can add this as a separate test also. I think we do not have that

vishweshbankwar, line 194: (Not for this PR) - can add this as a separate test also. I think we do not have that

test/OpenTelemetry.Exporter.Zipkin.Tests/Implementation/ZipkinActivityConversionTest.cs
```diff
@@ -125,5 +126,108 @@ public void ToZipkinSpan_Status_ErrorFlagTest(StatusCode expectedStatusCode, str
...
+
+            // ActivityStatusDescription takes higher precedence.
+            Assert.Contains(
+                zipkinSpan.Tags, t =>
+                t.Key == ZipkinActivityConversionExtensions.ZipkinErrorFlagTagName &&
+                (string)t.Value == description);
+            Assert.DoesNotContain(
+                zipkinSpan.Tags, t =>
+                t.Key == ZipkinActivityConversionExtensions.ZipkinErrorFlagTagName &&
+                (string)t.Value == "OK");
```
vishweshbankwar, line 227: Need to set description tag as well. 

https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/src/OpenTelemetry.Api#setting-status

vishweshbankwar, line 227: Need to set description tag as well. 

https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/src/OpenTelemetry.Api#setting-status

src/OpenTelemetry.Exporter.Zipkin/Implementation/ZipkinActivityConversionExtensions.cs
```diff
@@ -47,6 +47,46 @@ internal static ZipkinSpan ToZipkinSpan(this Activity activity, ZipkinEndpoint l
 
             activity.EnumerateTags(ref tagState);
 
+            if (activity.Status != ActivityStatusCode.Unset)
+            {
+                PooledList<KeyValuePair<string, object>>.Add(
+                    ref tagState.Tags,
+                    new KeyValuePair<string, object>(
+                        SpanAttributeConstants.StatusCodeKey,
+                        StatusHelper.GetTagValueForActivityStatusCode(activity.Status)));
```
cijothomas, line 56: I don't think the `GetTagValueForActivityStatusCode` is needed here. There is only 2 possible outcomes - OK or ERROR. A simple if..else should do the job, and it'd be faster than switch..case.

src/OpenTelemetry.Exporter.Zipkin/Implementation/ZipkinActivityConversionExtensions.cs
```diff
@@ -47,6 +47,46 @@ internal static ZipkinSpan ToZipkinSpan(this Activity activity, ZipkinEndpoint l
 
             activity.EnumerateTags(ref tagState);
 
+            if (activity.Status != ActivityStatusCode.Unset)
```
cijothomas, line 50: add a comment above the `If` saying this code is executed when status is set on Activity using the native status field in activity, and not using activity tags.

test/OpenTelemetry.Exporter.Zipkin.Tests/Implementation/ZipkinActivityConversionTest.cs
```diff
@@ -125,5 +126,110 @@ public void ToZipkinSpan_Status_ErrorFlagTest(StatusCode expectedStatusCode, str
...
+                    t.Key == ZipkinActivityConversionExtensions.ZipkinErrorFlagTagName);
+            }
+        }
+
+        [Fact]
+        public void ActivityStatus_Takes_precedence_Over_Status_Tags_ActivityStatusCodeIsOk()
+        {
+            // Arrange.
+            var activity = ZipkinExporterTests.CreateTestActivity();
+            activity.SetStatus(ActivityStatusCode.Ok);
```
cijothomas, line 176: add a test which does the same thing, but in reverse order. We want to make sure `activity.SetStatus(ActivityStatusCode.Ok);` wins no matter what order they were used.

src/OpenTelemetry.Exporter.Zipkin/Implementation/ZipkinActivityConversionExtensions.cs
Yun-Ting: I understand your point; but wouldn't the current implementation have better consistency with `GetTagValueForStatusCode` and `GetStatusCodeForTagValue`?

cijothomas: Let speed be our guide :)

src/OpenTelemetry.Exporter.Zipkin/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Support Activity `Status` and `StatusDescription` in Zipkin Exporter.
```
cijothomas, line 5: Added support for Activity `Status` and `StatusDescription` which were added to Activity from version 6.0. To maintain backward compatibility, the exporter falls back to checking status inside the special tags "otel.status_code", "otel.status_description", if Status is not set using the new Status field

test/OpenTelemetry.Exporter.Zipkin.Tests/ZipkinExporterTests.cs
```diff
@@ -351,14 +351,14 @@ public void ServiceProviderHttpClientFactoryInvoked()
                     break;
                 case StatusCode.Error:
                     statusTag = $@"""{SpanAttributeConstants.StatusCodeKey}"":""ERROR"",";
-                    errorTag = $@",""{ZipkinActivityConversionExtensions.ZipkinErrorFlagTagName}"":""{statusDescription}""";
+                    errorTag = $@"""{ZipkinActivityConversionExtensions.ZipkinErrorFlagTagName}"":""{statusDescription}"",";
```
cijothomas, line 354: given we are doing this in back-compat way, the existing tests should have required no changes, right?

Yun-Ting: The testcase changes were only reflecting the serializing order of Activity tags which is an *unordered-collection*. 
By adding support for Activity Status, updating the tags order to be - Activity Status or Tag Status is immediately followed by the Activity Error description will have better performance (less unnecessary if/else branching checks.)

I've synced offline with @cijothomas, and we agreed to not modify the testcase in this PR for easier code reviewing experience. And use a follow-up PR to do the performance improvement along with the updated testcase.

src/OpenTelemetry.Exporter.Zipkin/Implementation/ZipkinActivityConversionExtensions.cs
```diff
@@ -47,6 +47,41 @@ internal static ZipkinSpan ToZipkinSpan(this Activity activity, ZipkinEndpoint l
...
+                    ref tagState.Tags,
+                    new KeyValuePair<string, object>(
+                        SpanAttributeConstants.StatusCodeKey,
+                        "OK"));
+                }
+
+                // activity.Status is Error
+                else
+                {
+                    PooledList<KeyValuePair<string, object>>.Add(
```
cijothomas, line 66: populate error, "description" also right here, so you don't have to check if Status is error again.

src/OpenTelemetry.Exporter.Zipkin/Implementation/ZipkinActivityConversionExtensions.cs
```diff
@@ -72,9 +107,18 @@ internal static ZipkinSpan ToZipkinSpan(this Activity activity, ZipkinEndpoint l
                 }
             }
 
-            if (tagState.StatusCode == StatusCode.Error)
+            if (activity.Status == ActivityStatusCode.Error)
             {
                 // Error flag rule from https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk_exporters/zipkin.md#status
+                PooledList<KeyValuePair<string, object>>.Add(
```
cijothomas, line 113: move this to the above line ~ 70.

cijothomas (approved):

test/OpenTelemetry.Exporter.Zipkin.Tests/ZipkinExporterTests.cs
Yun-Ting: https://github.com/open-telemetry/opentelemetry-dotnet/pull/3084

Issue #3001, opened 2022-03-09
Update EditorConfig for new language rules


tillig:
# Feature Request

When working in VS Code, OmniSharp will show a lot of info messages based on the defaults for the latest language features.

<img width="725" alt="Example of OmniSharp analysis messages for style" src="https://user-images.githubusercontent.com/1156571/157487410-a88b31b7-0e0f-40e2-9438-aafcbbf1f5e0.png">

It is unclear whether the style for the project matches these or whether a PR would be accepted if they were cleaned up.

The primary missing rules appear to be:

- [IDE0017 `dotnet_style_object_initializer`](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0017) - use object initializers like `var x = new X { Foo = true }` instead of `var x = new X(); x.Foo = true;`
- [IDE0063 `csharp_prefer_simple_using_statement`](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0063) - prefer inline `using var x` over `using(...){...}`
- [IDE0071 `dotnet_style_prefer_simplified_interpolation`](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0071) - simplify/optimize string interpolation expressions
- [IDE0090 `csharp_style_implicit_object_creation_when_type_is_apparent`](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0090) - using target-typed `new()` style expressions

Just those four constitute about 250 or so info messages about style. Having them defined in `.editorconfig` would not only help folks adhere to the desired style of the project, it would make it easy to do a quick "fix all" and submit a PR to clean up all the style at once.

Note there are actually a lot of other things that might be valuable to define, too, like a [pattern matching preferences](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/pattern-matching-preferences). I have used the [`.editorconfig` in the `dotnet/runtime` repo](https://github.com/dotnet/runtime/blob/main/.editorconfig) in the past as a guide to the bajillion preferences you can supply. It's easier than [combing the documentation](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/categories).

**Is your feature request related to a problem?**

[...]

PR #3012, opened 2022-03-10
Fix IDE0090: 'new' expression can be simplified.

tillig:
Continues work on #3001.

## Changes

Fixes all IDE0090 issues such that obvious object construction uses the `new()` syntax instead of `new T()`. Analyzer rule `csharp_style_implicit_object_creation_when_type_is_apparent` explicitly set to `true:suggestion` which is the default behavior, but calling it out helps indicate to folks that it's also the _desired_ value for the project and it's cool to fix those if you see them.

```c#
// var uses remain the same
var x = new Thing();

// explicit types like in class fields switch to new()
Thing x = new();
```

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -65,6 +65,7 @@ csharp_prefer_braces = true:silent
 csharp_preserve_single_line_blocks = true
 csharp_preserve_single_line_statements = true
 dotnet_style_readonly_field = true:suggestion
+csharp_style_implicit_object_creation_when_type_is_apparent = true:suggestion
 
 # Expression-level preferences
 dotnet_style_object_initializer = true:suggestion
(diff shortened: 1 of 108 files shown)


cijothomas (approved):

cijothomas:
Merging without waiting for more reviews. Safe change. Touches a large number of files, so waiting longer could result in merge conflicts.

tillig:
I made an unfortunate error in the title for the PR (just the title) - this was IDE0063, not IDE0090. IDE0090 is simplifying object initialization and that's still to come.

```c#
/// from this
var x = new Foo();
x.Thing = 2;

/// to this
var x = new Foo
{
  Thing = 2
};
```

Sorry about the mix-up. 😦

tillig:
I fixed the title and initial comment but that won't change the [merge commit message here](https://github.com/open-telemetry/opentelemetry-dotnet/commit/d55de344f19875ee09faddf6e4420f56be7b3958). Someone with owner abilities will have to do something like `git rebase` or `git --amend` to fix it. It's probably not that important in the long run, but... again, sorry.

cijothomas:
> It's probably not that important in the long run

I think we can live with that. Appreciate your thoroughness in every PRs :)

PR #3015, opened 2022-03-10
Fix IDE0071: Interpolation can be simplified.

tillig:
Continues work on #3001.

## Changes

Fixes all IDE0071 issues such that string interpolation expressions are simplified.

```c#
// Don't call .ToString() on things, interpolation already does that.
var x = $"{myThing.ToString()}";
// becomes
var x = $"{myThing}";

// Use the padding expressions instead of extension methods.
var x = $"{str.PadRight(Amount)}";
//becomes
var x = $"{str,-Amount}";
```

On original [non-scientific] glance it looked like there were a lot more of these than there actually were, but figured I'd fix them anyway.

.editorconfig setting is based on the default behavior (at least in VS Code) without any explicit setting.


cijothomas (approved):

CodeBlanch (approved):
LGTM

PR #3022, opened 2022-03-11
Fix IDE0017: Use object initializers

### LINKED ISSUE #3001: Update EditorConfig for new language rules

tillig:
Fixes #3001. The last in the series, though I may throw one more PR in with a few one-off analyzer warning fixes since there are still about 300 and they're pretty simple fixes. (And I triple-checked the title this time - it really is IDE0017!)

## Changes

- Set `dotnet_style_object_initializer` to `true:suggestion` in `.editorconfig` since that's the default behavior for the IDE if nothing is specified.
- Fixed all the info issues related to that rule.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -68,6 +68,7 @@ csharp_prefer_simple_using_statement = true:suggestion
 dotnet_style_readonly_field = true:suggestion
 csharp_style_implicit_object_creation_when_type_is_apparent = true:suggestion
 dotnet_style_prefer_simplified_interpolation = true:suggestion
+dotnet_style_object_initializer = true:suggestion
 
 # Expression-level preferences
 dotnet_style_object_initializer = true:suggestion
--- a/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/MetricItemExtensions.cs
+++ b/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/MetricItemExtensions.cs
@@ -144,9 +144,11 @@ namespace OpenTelemetry.Exporter.OpenTelemetryProtocol.Implementation
             {
                 case MetricType.LongSum:
                     {
-                        var sum = new OtlpMetrics.Sum();
-                        sum.IsMonotonic = true;
-                        sum.AggregationTemporality = temporality;
+                        var sum = new OtlpMetrics.Sum
+                        {
+                            IsMonotonic = true,
+                            AggregationTemporality = temporality,
+                        };
 
                         foreach (ref readonly var metricPoint in metric.GetMetricPoints())
                         {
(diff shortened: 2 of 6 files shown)


cijothomas (approved):

cijothomas:
> The last in the series, though I may throw one more PR in with a few one-off analyzer warning fixes since there are still about 300 and they're pretty simple fixes.

Thanks a lot for your efforts in improving this repo!

Issue #3024, opened 2022-03-11
Review the conditional compilations in ASPNET Core and other instrumentations.


cijothomas:
Due to the additional of more targets like net5.0 and net6.0 to the instrumention projects, the conditional checks inside the code seems broken now.

example:
https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs#L313 check for 
`#if NETSTANDARD2_1`, means that code path is *not* triggered for net5.0 or net6.0

Opening an issue to do a round of reviews to make sure the conditional compilation flags are still correct, after the addition of new project targets.

TimothyMothra:
@cijothomas  In your PR, what was the reason for using `#if NETSTANDARD2_1_OR_GREATER || NETCOREAPP3_1_OR_GREATER` instead of just `#if NETSTANDARD2_1_OR_GREATER` ? 

I would have expected the single to work, I'm curious if you found an issue with that.

cijothomas:
Because the project had NETCOREAPP3.1/NET5.0/NET6.0 targets (along with netstandard), and I don't think NETSTANDARD2_1 covers that. Happy to correct myself, if this is not the case.

PR #3025, opened 2022-03-11
Fix conditional compilation in aspnetcore instrumentation

cijothomas:
Towards #3024 
I think this will truly make the benefits from https://github.com/open-telemetry/opentelemetry-dotnet/pull/2947/files available to newer frameworks like NET5.0, NET6.0, which was the intent as well.


src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs
```diff
@@ -310,7 +310,7 @@ private static string GetUri(HttpRequest request)
             var queryString = request.QueryString.Value ?? string.Empty;
             var length = scheme.Length + Uri.SchemeDelimiter.Length + host.Length + pathBase.Length
                          + path.Length + queryString.Length;
-#if NETSTANDARD2_1
+#if !NETSTANDARD2_0
```
reyang, line 313: Would it be possible to describe this in a way that reflects the actual intention?

e.g. `#if NET5_0_OR_GREATER || NETSTANDARD2_1_OR_GREATER`

pellared (approved):

PR #3026, opened 2022-03-11
Fix analyzer issues in docs

tillig:
Related to #3001.

## Changes

Updated items in the `docs/*` tree to resolve analyzer issues.

- Put all classes in file-scoped namespaces, which keeps the readability but handles the analyzer issue that classes should always be in namespaces. I used namespaces based on the name of the project it's in, so `getting-started` got the `GettingStarted` namespace.
- Simplified null checks.
- Removed unused parameters (`string[] args`) isn't required in `Main` and wasn't used).
- Removed one logger line causing CA2254. The line was showing the use of string interpolation in a log message template, which is not really that good. It seems like the docs shouldn't necessarily show "everything works" but maybe do both "show how it works" and "here's what you _should do_" so rather than suppress the issue I removed the line. If folks feel strongly, I can put it back with a suppression on it, but it'd be nice to not have the issue showing up in the VS Code "problems" window.

### DIFF
--- a/docs/logs/extending-the-sdk/Program.cs
+++ b/docs/logs/extending-the-sdk/Program.cs
@@ -39,9 +39,6 @@ public class Program
         // unstructured log
         logger.LogInformation("Hello, World!");
 
-        // unstructured log with string interpolation
-        logger.LogInformation($"Hello from potato {0.99}.");
-
         // structured log with template
         logger.LogInformation("Hello from {name} {price}.", "tomato", 2.99);
 
--- a/docs/logs/extending-the-sdk/Program.cs
+++ b/docs/logs/extending-the-sdk/Program.cs
@@ -18,6 +18,8 @@ using System.Collections.Generic;
 using Microsoft.Extensions.Logging;
 using OpenTelemetry;
 
+namespace ExtendingTheSdk;
+
 public class Program
 {
     public static void Main()
--- a/docs/logs/extending-the-sdk/Program.cs
+++ b/docs/logs/extending-the-sdk/Program.cs
@@ -39,6 +39,9 @@ public class Program
         // unstructured log
         logger.LogInformation("Hello, World!");
 
+        // String interpolation, as in the below line, results in unstructured logging, and is not recommended
+        // logger.LogInformation($"Hello from potato {0.99}.");
+
         // structured log with template
         logger.LogInformation("Hello from {name} {price}.", "tomato", 2.99);
 
--- a/docs/logs/getting-started/Program.cs
+++ b/docs/logs/getting-started/Program.cs
@@ -17,6 +17,8 @@
 using Microsoft.Extensions.Logging;
 using OpenTelemetry.Logs;
 
+namespace GettingStarted;
+
 public class Program
 {
     public static void Main()
--- a/docs/trace/extending-the-sdk/README.md
+++ b/docs/trace/extending-the-sdk/README.md
@@ -197,7 +197,7 @@ client .NET Core](../../../src/OpenTelemetry.Instrumentation.Http/README.md) .
 Instrumentation libraries for these are already provided in this repo. The
 [OpenTelemetry .NET
 Contrib](https://github.com/open-telemetry/opentelemetry-dotnet-contrib)
-repostory also has instrumentations for libraries like
+repository also has instrumentations for libraries like
 [ElasticSearchClient](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.Contrib.Instrumentation.ElasticsearchClient)
 etc. which fall in this category.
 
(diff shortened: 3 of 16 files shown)


docs/logs/extending-the-sdk/Program.cs
```diff
@@ -37,9 +39,6 @@ public static void Main()
         // unstructured log
         logger.LogInformation("Hello, World!");
```
cijothomas, line 40: maybe leave both the unstructure logs lines as-is, but add a comment above saying "unstructured logging is not recommended, but shown here for comparison"..?

cijothomas: or have the whole unstructured log statements as commented out, so CA2254 won't complain, but it serves the purpose of showing the difference between structured and unstructured.

docs/logs/extending-the-sdk/Program.cs
```diff
@@ -18,6 +18,8 @@
 using Microsoft.Extensions.Logging;
 using OpenTelemetry;
 
+namespace ExtendingTheSdk;
```
cijothomas, line 21: if we are adding ns, would `OpenTelemetry.Docs.Logs.ExtendingTheSDK` look better?

docs/trace/extending-the-sdk/README.md
```diff
@@ -197,7 +197,7 @@ client .NET Core](../../../src/OpenTelemetry.Instrumentation.Http/README.md) .
 Instrumentation libraries for these are already provided in this repo. The
 [OpenTelemetry .NET
 Contrib](https://github.com/open-telemetry/opentelemetry-dotnet-contrib)
-repostory also has instrumentations for libraries like
+repository also has instrumentations for libraries like
```
cijothomas, line 200: did you catch it manually or some tools flagged it? (wonder why the CI checks didn't catch it)

docs/logs/extending-the-sdk/Program.cs
cijothomas: Tagging @reyang as he had some concerns about ns. (I am not sure actually...some vague memory about an old comment about namespaces in docs..)

tillig: Maybe, but I sort of guessed that the point of not putting these in namespaces in the first place was to minimize boilerplate, so I erred on the side of "keep it super short" rather than having a larger fully-qualified namespace. Happy to change these if folks think that'd be better.

docs/logs/extending-the-sdk/Program.cs
tillig: If we uncomment it, it'd be good to either add a `[SuppressMessage]` to turn that off for that line or `#pragma warning disable` around it. Or, as you mentioned, I could put it back and leave it commented. It did seem odd to show "you can do this!" for things that you really shouldn't be doing so it felt like removing it was the better option for a documentation project. If it was unit tests then the suppression would be more accurate, I think, to retain the test case.

cijothomas: > I could put it back and leave it commented.

I'd vote for this approach. 
```
// String interpolation, as in the below line, results in unstructured logging, and is not recommended
logger.LogInformation($"Hello from potato {0.99}.");
```

This way, if someone is doing this, its easy to point them to this doc. Having the incorrect way in a comment block will be useful for them to know what they are doing incorrectly.

docs/logs/extending-the-sdk/Program.cs
cijothomas: This seems okay. Will resolve this.

If there are additional comments, we can address later.

docs/trace/extending-the-sdk/README.md
cijothomas: https://github.com/open-telemetry/opentelemetry-dotnet/issues/3031

cijothomas (approved):
LGTM.
A small comment about leaving unstructured logging statement as commented out to demonstrate a wrong approach is pending. Okay to merge without it also.

docs/logs/extending-the-sdk/Program.cs
tillig: OK, I put this back in as a comment.

cijothomas (approved):

docs/logs/extending-the-sdk/Program.cs
reyang: I don't have strong opinion, either way looks fine to me.

docs/logs/getting-started/Program.cs
```diff
@@ -17,6 +17,8 @@
 using Microsoft.Extensions.Logging;
 using OpenTelemetry.Logs;
 
+namespace GettingStarted;
```
reyang, line 20: I'm guessing - maybe we just use `namespace Example` across all these examples?

reyang (approved):

PR #3033, opened 2022-03-11
Add aspnet metric example

vishweshbankwar:

PR #3037, opened 2022-03-12
Add Prometheus Exporter to ASPNET Example

cijothomas:
Continue from https://github.com/open-telemetry/opentelemetry-dotnet/pull/3033
Just default Prometheus (http://localhost:9464/metrics), without any customization. Good enough for demonstrating recently added metrics from ASP.NET instrumentation.

Also removed AddHttpClientInstrumentation() as it does not produce any metrics in .NET Framework.

### DIFF
--- a/examples/AspNet/Global.asax.cs
+++ b/examples/AspNet/Global.asax.cs
@@ -73,8 +73,7 @@ namespace Examples.AspNet
             // https://github.com/open-telemetry/opentelemetry-dotnet/issues/2994
 
             var meterBuilder = Sdk.CreateMeterProviderBuilder()
-                 .AddAspNetInstrumentation()
-                 .AddHttpClientInstrumentation();
+                 .AddAspNetInstrumentation();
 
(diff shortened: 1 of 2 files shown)


cijothomas:
@vishweshbankwar @utpilla

examples/AspNet/Global.asax.cs
```diff
@@ -73,8 +73,7 @@ protected void Application_Start()
             // https://github.com/open-telemetry/opentelemetry-dotnet/issues/2994
 
             var meterBuilder = Sdk.CreateMeterProviderBuilder()
-                 .AddAspNetInstrumentation()
-                 .AddHttpClientInstrumentation();
```
reyang, line 77: Is this intentional?

reyang (approved):

examples/AspNet/Global.asax.cs
cijothomas: yes. HttpClient is not instrumented for .NET FW.

PR #3038, opened 2022-03-12
Clean up metric reader options

alanwest:
* Removes `MetricReaderType` enumeration.
* Adds `PeriodicExportingMetricReaderOptions.ExportTimeoutMilliseconds` with a default of `30000` for all push exporters.
* `MetricReaderOptions.PeriodicExportingMetricReaderOptions` is no longer settable.


src/OpenTelemetry.Exporter.Console/ConsoleExporterMetricsExtensions.cs
```diff
@@ -25,6 +26,9 @@ namespace OpenTelemetry.Metrics
     /// </summary>
     public static class ConsoleExporterMetricsExtensions
     {
+        private const int DefaultExportIntervalMilliseconds = Timeout.Infinite;
+        private const int DefaultExportTimeoutMilliseconds = 30000;
```
reyang, line 30: Do we need timeout 30000 or Infinite?

src/OpenTelemetry.Exporter.InMemory/InMemoryExporterMetricsExtensions.cs
```diff
@@ -26,6 +27,9 @@ namespace OpenTelemetry.Metrics
     /// </summary>
     public static class InMemoryExporterMetricsExtensions
     {
+        private const int DefaultExportIntervalMilliseconds = Timeout.Infinite;
+        private const int DefaultExportTimeoutMilliseconds = 30000;
```
reyang, line 31: Do we need timeout 30000 or Infinite?

reyang (approved):

cijothomas (approved):
LGTM except the same qn as Reiley about default for ExportTimeout to be infinite vs 30

utpilla:
We could get rid of the default exportInterval and timeout in `PeriodicExportingMetricReader` now as we would always be providing the parameter values for the `PeriodicExportingMetricReader` ctor:
https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/Metrics/PeriodicExportingMetricReader.cs#L31-L32

utpilla (approved):

src/OpenTelemetry.Exporter.Console/ConsoleExporterMetricsExtensions.cs
alanwest: I just submitted https://github.com/open-telemetry/opentelemetry-specification/pull/2415 for further clarification.

I'll set to Infinite for now and change depending on how things shake out.

alanwest:
> We could get rid of the default exportInterval and timeout in `PeriodicExportingMetricReader` now as we would always be providing the parameter values for the `PeriodicExportingMetricReader` ctor:

It is still possible to instantiate the `PeriodicExportingMetricReader` by itself, so the defaults are still relevant. Something like

```csharp
var metricExporter = new ConsoleExporter();
var metricReader = new PeriodicExportingMetricReader(exporter);
var meterProviderBuilder = Sdk.CreateMeterProviderBuilder().AddReader(metricReader);
```

Issue #2999, opened 2022-03-09
How to configure ILogger & OpenTelemetryLoggerProvider (Doc Proposal)


TimothyMothra:
I'm seeing a documentation gap around configuration options for using OpenTelemetry with ILogger.
I'm volunteering to work on this (it's a great way to learn more about the SDK!) and I'm using this issue to identify the scope of what should be included in a "customizing-the-sdk" document and example project.

Document: https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/docs/logs/customizing-the-sdk

## Scope

- Intro #3035
- Filtering #3034
   The class name or alias must be known to configure filtering.
    - via appinsights.json: uses the alias `OpenTelemetry`.
    - via code: AddFilter() uses the class name `OpenTelemetryLoggerProvider`
 - [OpenTelemetryLoggerOptions](https://github.com/open-telemetry/opentelemetry-dotnet/blob/d7afc1bcb0092a56ac1ea55bfbcf5a9430c72cf9/src/OpenTelemetry/Logs/OpenTelemetryLoggerOptions.cs)
    - IncludeScopes #3155
    - IncludeFormattedMessage #3158
    - ParseStateValues #3169
    - AddProcessor #3042
    - SetResourceBuilder #3043

### Todo
- [ ] First PR introduced several TODOs. I expect each of these to be resolved before closing this issue.
- [x] Address additional comments from #3034

cijothomas:
IncludeScopes
IncludeFormattedMessage
ParseStateValues
AddProcessor
SetResourceBuilder

^ all of above sounds good to be documented in the repo. You can look at tracing and metrics docs to get some idea on how to organize this.
https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/src/OpenTelemetry#getting-started-with-tracing

https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/docs/metrics/customizing-the-sdk

PR #3043, opened 2022-03-14
Logs customization doc: SetResourceBuilder

TimothyMothra:
#2999.

## Changes
- add content for SetResourceBuilder
- add code sample


docs/logs/customizing-the-sdk/README.md
```diff
@@ -30,7 +30,53 @@ TODO
...
+logger.LogInformation("Hello Information");
+```
+
+This example generates the following output which includes the customized
+service metadata:
+
+```text
+LogRecord.TraceId:            00000000000000000000000000000000
+LogRecord.SpanId:             0000000000000000
+LogRecord.Timestamp:          0001-01-01T00:00:00.0000000Z
```
reyang, line 66: Where is timestamp coming from? It seems wrong.

reyang (approved):

docs/logs/customizing-the-sdk/README.md
TimothyMothra: That was the format for the output, I manually zeroed out the values. 
I contemplated removing most of the example output and only showing the last three lines which show the resource.

reyang: You can put a valid timestamp like the 1st second of 1970.

TimothyMothra: done. changed to `1970-01-01T00:00:01.0000000Z`

docs/logs/customizing-the-sdk/README.md
```diff
@@ -30,7 +30,53 @@ TODO
 
 ### SetResourceBuilder
 
-TODO
+A `ResourceBuilder` provides a `Resource` to the `OpenTelemetryLoggerProvider`.
```
cijothomas, line 33: please borrow as much from https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/src/OpenTelemetry#resource.

(need to test to confirm if the ENV variable is applicable to logs as well, and document accordingly)

TimothyMothra: Do you want the full content copied here or should I link to the doc you shared?

cijothomas: copy here.
(We need to later do some refactoring to extract common concepts like Resource to a common place, and have all signals refer to it.)

Also noticed that, Metrics docs doesn't talk about Resource! Good thing to add it there too.

docs/logs/customizing-the-sdk/README.md
```diff
@@ -30,7 +30,53 @@ TODO
...
+            options.AddConsoleExporter();
+        });
+});
+
+var logger = loggerFactory.CreateLogger<Program>();
+
+logger.LogInformation("Hello Information");
+```
+
+This example generates the following output which includes the customized
```
cijothomas, line 60:
```suggestion
This example generates the following output which includes the Resource information
```

docs/logs/customizing-the-sdk/README.md
TimothyMothra: yeah, i noticed the other docs had TODOs for this content

TimothyMothra: Done. I had to take out references to TracerProvider. Let me know if it needs additional edits

docs/logs/customizing-the-sdk/README.md
```diff
@@ -30,7 +30,71 @@ TODO
 
 ### SetResourceBuilder
 
-TODO
+[Resource](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md)
```
cijothomas, line 33: this has references to "sampler" which is a tracing only concept, so need to tweak this accordingly.
Also the user is not building the provider here, instead they build the logger factory, so we need to modify wording accordingly.

TimothyMothra: I cleaned up this section:

[Resource](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md) is the immutable representation of the entity producing the telemetry.
If no `Resource` is explicitly configured, the default is to use a resource
indicating this [Telemetry SDK](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/resource/semantic_conventions#telemetry-sdk).
The `SetResourceBuilder` method can be used to set a single `ResourceBuilder`.
If `SetResourceBuilder` is called multiple times, only the last is kept..
It is not possible to change the resource builder *after* creating the LoggerFactory.

docs/logs/customizing-the-sdk/README.md
```diff
@@ -30,7 +30,67 @@ TODO
...
+is the immutable representation of the entity producing the telemetry.
+If no `Resource` is explicitly configured, the default is to use a resource
+indicating this [Telemetry
+SDK](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/resource/semantic_conventions#telemetry-sdk).
+The `SetResourceBuilder` method can be used to set a single `ResourceBuilder`.
+If `SetResourceBuilder` is called multiple times, only the last is kept..
+It is not possible to change the resource builder *after* creating the
+LoggerFactory.
```
cijothomas, line 41:
```suggestion
`LoggerFactory`.
```

docs/logs/customizing-the-sdk/README.md
```diff
@@ -30,7 +30,67 @@ TODO
 
 ### SetResourceBuilder
 
-TODO
+[Resource](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md)
+is the immutable representation of the entity producing the telemetry.
+If no `Resource` is explicitly configured, the default is to use a resource
+indicating this [Telemetry
+SDK](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/resource/semantic_conventions#telemetry-sdk).
+The `SetResourceBuilder` method can be used to set a single `ResourceBuilder`.
```
cijothomas, line 38:
```suggestion
The `SetResourceBuilder` method on `OpenTelemetryLoggerOptions` can be used to set a single `ResourceBuilder`.
```

cijothomas, line 38:
```suggestion
The `SetResourceBuilder` method can be used to set a `ResourceBuilder`.
```

docs/logs/customizing-the-sdk/README.md
```diff
@@ -30,7 +30,67 @@ TODO
...
+            options.AddConsoleExporter();
+        });
+});
+
+var logger = loggerFactory.CreateLogger<Program>();
+
+logger.LogInformation("Hello Information");
+```
+
+This example generates the following output which includes the Resource
```
cijothomas, line 64: perhaps not needed to show the actual output here. By this time, users already got started with getting-started, so they should be able to run it and see output themselves.

docs/logs/customizing-the-sdk/README.md
```diff
@@ -30,7 +30,67 @@ TODO
...
+
+See [Program.cs](Program.cs) for complete example.
+
+It is also possible to configure the `Resource` by using following
+environmental variables:
+
+<!-- markdownlint-disable MD013 -->
+| Environment variable       | Description                                        |
+| -------------------------- | -------------------------------------------------- |
+| `OTEL_RESOURCE_ATTRIBUTES` | Key-value pairs to be used as resource attributes. See the [Resource SDK specification](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.5.0/specification/resource/sdk.md#specifying-resource-information-via-an-environment-variable) for more details. |
```
cijothomas, line 91: could you confirm if the ENV variable are respected by LoggingProvider? It should work, but good to add unit tests to it...

cijothomas (approved):
LGTM. Left couple of non blocking comments.
Please do a manual verification that ENV variables are respected.

docs/logs/customizing-the-sdk/README.md
TimothyMothra: added new tests!

TimothyMothra: I wasn't sure the best place for these, so I created a new file. Please let me know if there's a more appropriate file.

docs/logs/customizing-the-sdk/README.md
TimothyMothra: I'm a fan of including the output. 
I think demonstrating what these settings affect is valuable.

What if we remove the top half of the example and only include the bottom section?
```
Resource associated with LogRecord:
    service.name: MyService
    service.version: 1.0.0
    service.instance.id: 00000000-0000-0000-0000-000000000000
```

TimothyMothra: I also agree that these are very simple examples and someone should be able to follow along with ease.

cijothomas: I'd suggest to remove actual output. We don't do that anywhere else (except in the getting-started guide). By completing getting-started, we expect user should already know where to look for output.

test/OpenTelemetry.Tests/Logs/LoggingProviderTests.cs
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        [Fact]
+        public void VerifyResourceBuilderAddService()
+        {
+            var options = new OpenTelemetryLoggerOptions
+            {
+                ResourceBuilder = ResourceBuilder.CreateDefault().AddService(serviceName: "MyService", serviceVersion: "1.2.3"),
+            };
+            var provider = new OpenTelemetryLoggerProvider(options);
```
cijothomas, line 43: this is not how users would be setting resource, so I'd suggest to keep the unit test doing what user would be doing. (there is no public setter `ResourceBuilder`, tests are able to access internals)

cijothomas: its okay to do the test additions separately.

TimothyMothra: if it's okay to contribute tests with this PR I would like to do that to avoid creating MORE PRs :)

PR #3051, opened 2022-03-15
AddHttpClientInstrumentation uses Options<T>.

### LINKED ISSUE #3050: `HttpClientInstrumentationOptions` doesn't support `IDeferredTracerProviderBuilder`
# Bug Report

The `AddAspNetCoreInstrumentation` extension [shows what it looks like to support `IDeferredTracerProviderBuilder`](https://github.com/open-telemetry/opentelemetry-dotnet/blob/a709cfd2a993fef5dedf8beb73b128ff2d03b896/src/OpenTelemetry.Instrumentation.AspNetCore/TracerProviderBuilderExtensions.cs#L41) so it can take advantage of the DI container and registered `Options<AspNetCoreInstrumentationOptions>`.

The corresponding [`AddHttpClientInstrumentation` extension](https://github.com/open-telemetry/opentelemetry-dotnet/blob/a709cfd2a993fef5dedf8beb73b128ff2d03b896/src/OpenTelemetry.Instrumentation.Http/TracerProviderBuilderExtensions.cs#L29) does not currently support `IDeferredTracerProviderBuilder` so there's no way to use `Options<T>` to configure things like enrichers.

(Unclear if this is a bug or an enhancement request so I just picked bug. I'm going to throw a PR in to fix this in a minute anyway.)

tillig:
Fixes #3050.

## Changes

Copied the pattern from [`AddAspNetCoreInstrumentation`](https://github.com/open-telemetry/opentelemetry-dotnet/blob/a709cfd2a993fef5dedf8beb73b128ff2d03b896/src/OpenTelemetry.Instrumentation.AspNetCore/TracerProviderBuilderExtensions.cs#L41) to support `IDeferredTracerProviderBuilder` when in `netstandard`. I did notice there's a different options structure for `net461` but I didn't modify that code path. I guessed that the full framework folks weren't using `Options<T>` and, even if they are, I'm on a Mac in VS Code so I can't really exercise that code path myself.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.Http/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.Http/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+* Updated `TracerProviderBuilderExtensions.AddHttpClientInstrumentation` to support
+  `IDeferredTracerProviderBuilder` and `IOptions<HttpClientInstrumentationOptions>`
+  ([#3051](https://github.com/open-telemetry/opentelemetry-dotnet/pull/3051))
+
 ## 1.0.0-rc10
 
 Released 2022-Mar-04
@@ -92,7 +96,7 @@ Released 2020-Nov-5
   `HttpWebRequest` in Activity.CustomProperty. To enrich activity, use the
   Enrich action on the instrumentation.
   ([#1407](https://github.com/open-telemetry/opentelemetry-dotnet/pull/1407))
-* Renamed TextMapPropagator to TraceContextPropagator, CompositePropapagor to
+* Renamed TextMapPropagator to TraceContextPropagator, CompositePropagator to
   CompositeTextMapPropagator. IPropagator is renamed to TextMapPropagator and
   changed from interface to abstract class.
   ([#1427](https://github.com/open-telemetry/opentelemetry-dotnet/pull/1427))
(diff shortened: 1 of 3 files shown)


tillig:
Sure, I can do those, too, np.

tillig:
So... possibly slight correction. You don't have an established pattern for how to deal with `IDeferredMeterProviderBuilder` in any of the `*.Instrumentation.*` libraries. I see it in the `*.Exporter.*` libraries, but in instrumentation there's a lot of...

```c#
            // TODO: Implement an IDeferredMeterProviderBuilder

            // TODO: Handle HttpClientInstrumentationOptions
            //   SetHttpFlavor - seems like this would be handled by views
            //   Filter - makes sense for metric instrumentation
            //   Enrich - do we want a similar kind of functionality for metrics?
            //   RecordException - probably doesn't make sense for metric instrumentation
```

Which is to say, there are some design things to figure out. If possible, I'd like to punt that to a different PR so I can at least get the tracing side of things out the door. I will work on the changelog, though.

tillig:
Sure. Would it be good to start that maybe as a discussion [over here](https://github.com/open-telemetry/opentelemetry-dotnet/discussions)? I have some ideas based on things like how various .NET "builder" things work as well as `IOptions<T>` handling and more automatic registrations like I talk about in #2971 and #3029. I'll see if I can drop something in there shortly.

tillig:
OK, got `CHANGELOG` updated.

cijothomas (approved):

PR #3062, opened 2022-03-17
Add netstandard2.0 support for GrpcNetClient instrumentation.

dpbevin:
## Changes

Adds support for `netstandard2.0` target framework to the GrpcNetClient instrumentation package.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.GrpcNetClient/OpenTelemetry.Instrumentation.GrpcNetClient.csproj
+++ b/src/OpenTelemetry.Instrumentation.GrpcNetClient/OpenTelemetry.Instrumentation.GrpcNetClient.csproj
@@ -1,7 +1,7 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <!-- OmniSharp/VS Code requires TargetFrameworks to be in descending order for IntelliSense and analysis. -->
-    <TargetFrameworks>netstandard2.1</TargetFrameworks>
+    <TargetFrameworks>netstandard2.1;netstandard2.0</TargetFrameworks>
     <Description>gRPC for .NET client instrumentation for OpenTelemetry .NET</Description>
     <PackageTags>$(PackageTags);distributed-tracing</PackageTags>
     <IncludeDiagnosticSourceInstrumentationHelpers>true</IncludeDiagnosticSourceInstrumentationHelpers>
--- a/src/OpenTelemetry.Instrumentation.GrpcNetClient/.publicApi/netstandard2.0/PublicAPI.Unshipped.txt
+++ b/src/OpenTelemetry.Instrumentation.GrpcNetClient/.publicApi/netstandard2.0/PublicAPI.Unshipped.txt
@@ -1,0 +1,8 @@
+OpenTelemetry.Instrumentation.GrpcNetClient.GrpcClientInstrumentationOptions
+OpenTelemetry.Instrumentation.GrpcNetClient.GrpcClientInstrumentationOptions.Enrich.get -> System.Action<System.Diagnostics.Activity, string, object>
+OpenTelemetry.Instrumentation.GrpcNetClient.GrpcClientInstrumentationOptions.Enrich.set -> void
+OpenTelemetry.Instrumentation.GrpcNetClient.GrpcClientInstrumentationOptions.GrpcClientInstrumentationOptions() -> void
+OpenTelemetry.Instrumentation.GrpcNetClient.GrpcClientInstrumentationOptions.SuppressDownstreamInstrumentation.get -> bool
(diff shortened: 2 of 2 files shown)


src/OpenTelemetry.Instrumentation.GrpcNetClient/OpenTelemetry.Instrumentation.GrpcNetClient.csproj
```diff
@@ -1,7 +1,7 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <!-- OmniSharp/VS Code requires TargetFrameworks to be in descending order for IntelliSense and analysis. -->
-    <TargetFrameworks>netstandard2.1</TargetFrameworks>
+    <TargetFrameworks>netstandard2.1;netstandard2.0</TargetFrameworks>
```
pellared, line 4: Could we target `netstandard2.0` instead of `netstandard2.1`? (I am not sure)

cijothomas: why do we need netstandard2.0 target? @dpbevin can you share some context/reasoning for this.

alanwest: I suspect because Grpc.Net.Client now has some support on netstandard2.0 https://docs.microsoft.com/aspnet/core/grpc/netstandard. Though, having not tried it myself, this documentation leaves me wondering exactly what is required to get it to work on the platforms and runtimes listed.

At a minimum, we need to expand the unit tests for this instrumentation to see if it functions as expected.

https://github.com/open-telemetry/opentelemetry-dotnet/blob/00af7817fdba22b2a5948e574c6769098003f682/test/OpenTelemetry.Instrumentation.Grpc.Tests/OpenTelemetry.Instrumentation.Grpc.Tests.csproj#L5

dpbevin: We use the Grpc.Net.Client in a .NET Framework (4.8) application - we use GRPC-Web comms.
That application is a bridge/proxy between WCF and GRPC. Gotta love the old world of WCF ;-)

dpbevin: Sadly, I don't think we can expand the test project because it's reliant on the `Grpc.AspNetCore.Server` which is not available for net4x.

alanwest: Ah, right. We have an [integration test](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/docker-compose.yml) that runs via docker-compose. It does not currently run against windows/.net framework. I'll see if it's easy to get it ported. If I get it to work, I'll open a separate PR.

dpbevin: @cijothomas @alanwest could we not just drop `netstandard2.1` since this is just a superset of 2.0? (as per @pellared suggestion). That way, tests would prove the `netstandard2.0` library works and we don't have to try and get GRPC servers working on .NET Framework, which apparently is not being considered - see https://github.com/grpc/grpc-dotnet/issues/1368#issuecomment-897121114.

alanwest: It's true we may just be able to drop netstandard2.1. Though, that doesn't ensure we're testing against the netstandard2.0 build of the Grpc.Net.Client library. That is, as it stands, the test suite would use the netstandard2.1 build of the Grpc.Net.Client library irrespective of the instrumentation library's target framework - because currently the test suite only targets net6.0, net5.0 and netcoreapp3.1.

That said, good news, I proved out a simple test against .NET Framework in #3105 that does not require spinning up a gRPC server.

dpbevin: @alanwest nice work on #3105. I'm happy to drop this PR in favour of yours.

alanwest: Ok, sounds good. I'll clean #3105 up soon and move it out of draft. We're focused on getting the stable release of the metrics SDK out, so might be just a bit before we can get this landed.

PR #3066, opened 2022-03-18
Do not allow changes to OTLP exporter configuration after instantiation

alanwest:
Prior to this PR, the underlying HTTP and gRPC export clients kept a handle on the original `OtlpExporterOptions` instance. This allowed for the options to change thereby allowing the behavior of an OTLP exporter instance to be altered after it's been instantiated.

### DIFF
--- a/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/BaseOtlpGrpcExportClient.cs
+++ b/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/BaseOtlpGrpcExportClient.cs
@@ -14,6 +14,7 @@
 // limitations under the License.
 // </copyright>
 
+using System;
 using System.Threading;
 using System.Threading.Tasks;
 using Grpc.Core;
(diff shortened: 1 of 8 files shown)


cijothomas (approved):

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/BaseOtlpGrpcExportClient.cs
```diff
@@ -35,20 +36,23 @@ protected BaseOtlpGrpcExportClient(OtlpExporterOptions options)
...
 
-        internal OtlpExporterOptions Options { get; }
-
 #if NETSTANDARD2_1 || NET5_0_OR_GREATER
         internal GrpcChannel Channel { get; set; }
 #else
         internal Channel Channel { get; set; }
 #endif
 
+        internal Uri Endpoint { get; }
```
reyang, line 50: minor: internal readonly?

reyang: never mind, I see that it only "get"

reyang (approved):

PR #3072, opened 2022-03-23
Create spans for HTTP retries and redirects

denisivan0v:
## Summary

This PR brings an implementation for HTTP retries and redirects instrumentation defined in [opentelemetry-specification #2078](https://github.com/open-telemetry/opentelemetry-specification/pull/2078). 
This is the second attempt to add the implementation (#2756 was the first one) with the reduced scope to reflect the current state of [HTTP semantic conventions specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md).

It addresses a scenario which is in the scope for bringing the existing HTTP semantic conventions for tracing to an initial stable state, see related [otep #174](https://github.com/open-telemetry/oteps/pull/174).

## Changes

1. Now spans will be created for all retry and redirect attempts (previously a span was created for the first attempt only).
   1. A span will be created even though `traceparent` header is already present.
   2. `traceparent` header will be overwritten based on current ambient context (`Activity.Current`) values.
2. We now follow the new spec for retries ad redirects (https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#http-request-retries-and-redirects)


denisivan0v:
@open-telemetry/dotnet-approvers  can you please reopen the PR?

denisivan0v:
Changes from this PR were benchmarked with the model application submitted here https://github.com/denisivan0v/http-semantic-conventions-experiments/pull/2. The main idea of the setup is to compare the current ("original") implementation (the copy of it is [here](https://github.com/denisivan0v/http-semantic-conventions-experiments/blob/no-links-benchmarks/OpenTelemetry.Instrumentation.Http/Implementation/OriginalHttpHandlerDiagnosticListener.cs)) vs the modified ("improved") one (the copy of it is [here](https://github.com/denisivan0v/http-semantic-conventions-experiments/blob/no-links-benchmarks/OpenTelemetry.Instrumentation.Http/Implementation/ImprovedHttpHandlerDiagnosticListener.cs)).

Here are results on my machine:

1. Retries are implemented with a naive `DelegatingHandler` (see [`SimpleRetryHandler`](https://github.com/denisivan0v/http-semantic-conventions-experiments/blob/no-links-benchmarks/HttpRetriesApp/SimpleRetryHandler.cs)).

BenchmarkDotNet=v0.13.1, OS=macOS Monterey 12.3.1 (21E258) [Darwin 21.4.0]
Intel Core i7-1068NG7 CPU 2.30GHz, 1 CPU, 8 logical and 4 physical cores
.NET SDK=6.0.101
  [Host]     : .NET 5.0.13 (5.0.1321.56516), X64 RyuJIT
  DefaultJob : .NET 5.0.13 (5.0.1321.56516), X64 RyuJIT


|          Method |     Mean |    Error |   StdDev |   Median | Ratio | RatioSD |   Gen 0 |  Gen 1 | Allocated |
|---------------- |---------:|---------:|---------:|---------:|------:|--------:|--------:|-------:|----------:|
| HandlerOriginal | 739.3 us | 14.46 us | 21.65 us | 729.2 us |  1.00 |    0.00 | 16.6016 | 0.9766 |     70 KB |
| HandlerImproved | 687.8 us | 13.73 us | 33.69 us | 673.5 us |  0.94 |    0.06 | 17.5781 |      - |     72 KB |

2. Retries are implemented with Polly (see https://github.com/denisivan0v/http-semantic-conventions-experiments/blob/no-links-benchmarks/HttpRetriesApp/Program.cs#L136).


BenchmarkDotNet=v0.13.1, OS=macOS Monterey 12.3.1 (21E258) [Darwin 21.4.0]
Intel Core i7-1068NG7 CPU 2.30GHz, 1 CPU, 8 logical and 4 physical cores
.NET SDK=6.0.101
  [Host]     : .NET 5.0.13 (5.0.1321.56516), X64 RyuJIT
  DefaultJob : .NET 5.0.13 (5.0.1321.56516), X64 RyuJIT


|        Method |     Mean |   Error |  StdDev | Ratio |   Gen 0 |  Gen 1 | Allocated |
|-------------- |---------:|--------:|--------:|------:|--------:|-------:|----------:|
| PollyOriginal | 744.8 us | 4.26 us | 3.56 us |  1.00 | 18.5547 | 0.9766 |     77 KB |
| PollyImproved | 740.4 us | 4.48 us | 3.97 us |  0.99 | 18.5547 |      - |     77 KB |

test/OpenTelemetry.Instrumentation.Http.Tests/HttpWebRequestActivitySourceTests.netfx.cs
```diff
@@ -653,19 +661,28 @@ public async Task TestSecureTransportRetryFailureRequest(string method)
                 Assert.True(ex is HttpRequestException);
             }
 
-            // We should have one Start event and one Stop event with an exception.
-            Assert.Equal(2, eventRecords.Records.Count());
-            Assert.Equal(1, eventRecords.Records.Count(rec => rec.Key == "Start"));
-            Assert.Equal(1, eventRecords.Records.Count(rec => rec.Key == "Stop"));
+            // We should have two Start events and twos Stop events, both with exceptions.
```
reyang, line 664:
```suggestion
            // We should have two Start events and two Stop events, both with exceptions.
```

reyang (approved):
LGTM with a changelog entry.

src/OpenTelemetry.Instrumentation.Http/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+* Implementation of HTTP retries and redirects instrumentation defined in
```
cijothomas, line 5: Did this spec get implemented in this PR? are we adding "http.retry_count" ? Not able to see any change related to that in this PR...

cijothomas (changes requested):
Please see : https://github.com/open-telemetry/opentelemetry-dotnet/pull/3072/files#r887147628

src/OpenTelemetry.Instrumentation.Http/CHANGELOG.md
denisivan0v: Correct. In order to add `http.retry_count` we should propagate a context between tries (which is also required to enable retry/redirect spans correlation with links, see https://github.com/open-telemetry/opentelemetry-specification/pull/2393). 

Since span links concept and its usage is not yet clear for the community, and context propagation topic was postponed, we cannot add `http.retry_count` at this time, but it's going be added in the future.

cijothomas: Thanks!
Could you update the PR description as well? Its not clear what the user can expect once they upgrade to a version with this change.

src/OpenTelemetry.Instrumentation.Http/Implementation/HttpHandlerDiagnosticListener.cs
```diff
@@ -70,14 +70,6 @@ public override void OnStartActivity(Activity activity, object payload)
                 return;
             }
```
cijothomas, line 72: glad to see one TODO removed!

test/OpenTelemetry.Instrumentation.Http.Tests/HttpWebRequestActivitySourceTests.netfx.cs
```diff
@@ -413,8 +413,10 @@ public async Task TestTraceStateAndBaggage()
         [Theory]
         [InlineData("GET")]
         [InlineData("POST")]
-        public async Task DoNotInjectTraceParentWhenPresent(string method)
+        public async Task OverwriteTraceParentWhenPresent(string method)
```
cijothomas, line 416: name is misleading.. as there is no validation done inside this test to validate if the instrumentation actually overwrites the header

src/OpenTelemetry.Instrumentation.Http/Implementation/HttpWebRequestActivitySource.netfx.cs
```diff
@@ -216,13 +212,6 @@ private static void ProcessRequest(HttpWebRequest request)
                 return;
             }
 
-            if (IsRequestInstrumented(request))
```
CodeBlanch, line 219: So in the event of automatic retry (due to ssl things, or redirect, or whatever) the same `HttpWebRequest` is reused. Below we patch the state we need onto the `HttpWebRequest` by taking over the the `IAsyncResult` used by the framework. If you remove this check, on retries, we will patch over what we already patched. This makes me nervous. Did you verify everything is cleaned up and we aren't leaking anything? Is the first activity correctly stopped before `StartActivity` is called again?

cijothomas: I think to unblock progress, @denisivan0v could you remote the HttpClient (.NETFramework) from this PR? Overall this is good, but since we want to address .NETFW HttpClient concerns, it'd be good to split that our to a separate PR.

denisivan0v: Sure, I can do it. But would it be okay the instrumentation works differently for .NET/.NET Core and .NETFW HttpClient?

test/OpenTelemetry.Instrumentation.Http.Tests/HttpClientTests.Basic.netcore31.cs
```diff
@@ -292,7 +293,14 @@ public async Task HttpClientInstrumentationBacksOffIfAlreadyInstrumented()
...
 
-            Assert.Equal(4, processor.Invocations.Count); // SetParentProvider/OnShutdown/Dispose/OnStart called.
+            Assert.Equal(5, processor.Invocations.Count); // SetParentProvider/OnStart/OnEnd/OnShutdown/Dispose called.
+            var activity = (Activity)processor.Invocations[2].Arguments[0];
+
+            Assert.Equal(ActivityKind.Client, activity.Kind);
+            Assert.NotEqual(default, activity.Context.SpanId);
+            Assert.NotEqual(traceId, activity.TraceId.ToString());
+            Assert.NotEqual(parentSpanId, activity.SpanId.ToString());
+            Assert.NotEqual(parentSpanId, activity.ParentSpanId.ToString());
```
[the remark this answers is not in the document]
denisivan0v: I think the asserts I added do exactly that. Am I missing something?

```csharp
// ...            
Assert.NotEqual(traceId, activity.TraceId.ToString());
Assert.NotEqual(parentSpanId, activity.ParentSpanId.ToString());
```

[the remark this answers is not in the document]
cijothomas: The asserts are only checking that the collected span is *not* having the same traceid/spanid from the original traceparent header.

The span will create a new spanid/traceid, and there is no assert to verify that the new traceparent is overwritten to the Request.

src/OpenTelemetry.Instrumentation.Http/Implementation/HttpWebRequestActivitySource.netfx.cs
denisivan0v: @CodeBlanch I've made a quick experiment and it seems the instrumentation doesn't modify `HttpRequestMessage` instance passed to `HttpClient` at all (see https://github.com/open-telemetry/opentelemetry-dotnet/runs/6697287031?check_suite_focus=true#step:5:506). 
Also, the way I understand how `HttpWebRequest` works, it is not really possible to reuse the same `HttpWebRequest` object since these are created per request (and being put to `Connection.m_WriteList`) and reuse the underlying [Connection](https://referencesource.microsoft.com/#System/net/System/Net/_Connection.cs,831) object. 

That said, I don't think removing this check we will patch a previously patched `HttpWebRequest` object we intercept here. Does this make sense?

denisivan0v: Also, verified that in case automatic redirects (`HttpClientHandler.AllowAutoRedirect = true`) those will be handled by `HttpWebRequest`, so we will be re-instrumenting it and starting new activities. Once requests processed, responses will be handled correctly (`LazyAsyncResult.m_AsyncCallback` will be called), so all corresponding activities will be called. It seems to me everything will be properly cleaned up.

image

test/OpenTelemetry.Instrumentation.Http.Tests/HttpWebRequestActivitySourceTests.netfx.cs
denisivan0v: Renamed the test

test/OpenTelemetry.Instrumentation.Http.Tests/HttpClientTests.Basic.netcore31.cs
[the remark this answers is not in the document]
denisivan0v: Got it now, thanks! Done in [7879d01](https://github.com/open-telemetry/opentelemetry-dotnet/pull/3072/commits/7879d01fe1cb798ea39334c6519594995549c083).

test/OpenTelemetry.Instrumentation.Http.Tests/HttpClientTests.Basic.netcore31.cs
```diff
@@ -270,9 +270,10 @@ public async Task HttpClientInstrumentation_AddViaFactory_HttpInstrumentation_Co
         }
 
         [Fact]
-        public async Task HttpClientInstrumentationBacksOffIfAlreadyInstrumented()
+        public async Task HttpClientInstrumentationWorksIfAlreadyInstrumented()
```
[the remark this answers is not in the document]
denisivan0v: Renamed in [7879d01](https://github.com/open-telemetry/opentelemetry-dotnet/pull/3072/commits/7879d01fe1cb798ea39334c6519594995549c083).

src/OpenTelemetry.Instrumentation.Http/Implementation/HttpWebRequestActivitySource.netfx.cs
denisivan0v: @CodeBlanch, @cijothomas does the information above clarifies the initial concern or you still want me move this part to a separate PR?

test/OpenTelemetry.Instrumentation.Http.Tests/HttpClientTests.Basic.netcore31.cs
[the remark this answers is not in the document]
cijothomas: missing the validation that the new traceparent header injected is the activity.Id.

test/OpenTelemetry.Instrumentation.Http.Tests/HttpClientTests.Basic.netcore31.cs
```diff
@@ -281,7 +282,8 @@ public async Task HttpClientInstrumentationBacksOffIfAlreadyInstrumented()
                 Method = new HttpMethod("GET"),
             };
 
-            request.Headers.Add("traceparent", "00-0123456789abcdef0123456789abcdef-0123456789abcdef-01");
+            var traceparent = $"00-{traceId}-{parentSpanId}-01";
```
cijothomas, line 285: nit: please name it either parentId, or spanId. 
https://www.w3.org/TR/trace-context/#parent-id

test/OpenTelemetry.Instrumentation.Http.Tests/HttpClientTests.Basic.netcore31.cs
```diff
@@ -292,7 +294,15 @@ public async Task HttpClientInstrumentationBacksOffIfAlreadyInstrumented()
...
 
-            Assert.Equal(4, processor.Invocations.Count); // SetParentProvider/OnShutdown/Dispose/OnStart called.
+            Assert.Equal(5, processor.Invocations.Count); // SetParentProvider/OnStart/OnEnd/OnShutdown/Dispose called.
+            var activity = (Activity)processor.Invocations[2].Arguments[0];
+
+            Assert.Equal(ActivityKind.Client, activity.Kind);
+            Assert.NotEqual(default, activity.Context.SpanId);
+            Assert.NotEqual(traceId, activity.TraceId.ToString());
+            Assert.NotEqual(parentSpanId, activity.SpanId.ToString());
+            Assert.NotEqual(parentSpanId, activity.ParentSpanId.ToString());
```
cijothomas, line 304: activity.ParentSpanId here is expected to be default, as there is not parent span. Why not simply validate it is default?

test/OpenTelemetry.Instrumentation.Http.Tests/HttpWebRequestActivitySourceTests.netfx.cs
```diff
@@ -413,8 +413,10 @@ public async Task TestTraceStateAndBaggage()
         [Theory]
         [InlineData("GET")]
         [InlineData("POST")]
-        public async Task DoNotInjectTraceParentWhenPresent(string method)
+        public async Task CreateSpansIfTraceParentIsPresent(string method)
         {
+            const string traceId = "abcdef0123456789abcdef0123456789";
+            const string parentSpanId = "abcdef0123456789";
```
cijothomas, line 419: same comment as .net core test.. please name it either parentId or SpanId.

cijothomas (approved):
Approving the overall direction. Please see below:

1. Left couple of nit comments in test. Would highly recommend addressing them.
2. Will wait for @CodeBlanch to approve the .NET Framework part, as he has the most expertise of the complexity involved in the working on .NET Framework HttpClient reflection.
3. The changelog or pr description is not good enough for an end user to understand whats going to happen now onwards. (like we not create span even though traceparent was already present.. we overwrite the traceparent.. we now follow the new spec for redirect/retry.)

test/OpenTelemetry.Instrumentation.Http.Tests/HttpClientTests.Basic.netcore31.cs
denisivan0v: Done in [dfe3520](https://github.com/open-telemetry/opentelemetry-dotnet/pull/3072/commits/dfe352066a7383b4c679ee8bcb72cd34d472b93b)

test/OpenTelemetry.Instrumentation.Http.Tests/HttpWebRequestActivitySourceTests.netfx.cs
denisivan0v: Done in [dfe3520](https://github.com/open-telemetry/opentelemetry-dotnet/pull/3072/commits/dfe352066a7383b4c679ee8bcb72cd34d472b93b)

denisivan0v:
> The changelog or pr description is not good enough for an end user to understand whats going to happen now onwards. (like we now create span even though traceparent was already present.. we overwrite the traceparent.. we now follow the new spec for redirect/retry.)

I've updated the PR description to make it more clear for an end user how the instrumentation's behavior will change and why it has been done.

denisivan0v:
> Will wait for @CodeBlanch to approve the .NET Framework part, as he has the most expertise of the complexity involved in the working on .NET Framework HttpClient reflection.

@CodeBlanch can you please have another look?

src/OpenTelemetry.Instrumentation.Http/Implementation/HttpWebRequestActivitySource.netfx.cs
CodeBlanch: Here is an example of HttpWebRequest re-submitting itself: https://referencesource.microsoft.com/#System/net/System/Net/HttpWebRequest.cs,3885

It is a thing! 😄 [TestSecureTransportRetryFailureRequest](https://github.com/open-telemetry/opentelemetry-dotnet/blob/b680ec624c1ad31ebcf4ec0bc45b94679e4e0a4a/test/OpenTelemetry.Instrumentation.Http.Tests/HttpWebRequestActivitySourceTests.netfx.cs#L635) is the test I added to verify one of the cases that I could figure out how to reproduce on demand. It has been a while since I read that reference code, but there are automatic re-submit cases beyond redirects. Most of them are SSL. Upgrades, downgrades, or during negotiation IIRC.

Anyway sorry I didn't realize I was blocking this. Let me take the code on the PR for a spin tomorrow just to sanity check it and then I'll give feedback or approve.

CodeBlanch: OK I took the HttpWebRequest stuff for a spin.

If we remove this check...

https://github.com/open-telemetry/opentelemetry-dotnet/blob/a58c7a3f62dc1e0e308792db3d7eef17faed9e63/src/OpenTelemetry.Instrumentation.Http/Implementation/HttpWebRequestActivitySource.netfx.cs#L219-L224

...what happens is...

* Activity1 is started. Request is hooked. Original state is wrapped with our instrumentation state.
* Some internal exception is thrown. A resubmit is triggered.
* Activity2 is started. Request is hooked again. Our instrumentation state wraps itself.
* Request finishes. Could be successful, could be another exception.
* Response is processed.
* Activity2 is stopped and tagged with the response. Wrapped state is called (this is our first instrumentation state).
* Activity1 is stopped and tagged with the same response as Activity2. Wrapped state is called (the original internal state).
* Original state fires to finish the Request.

This _appears_ to work. But there are issues. The trace looks like this:

image

Activity2 is a child of Activity1. I don't think that is what the spec is describing. Instead of...

```
Root-------------------------
   POST      --------------
        POST          ---
```

It should be...

```
Root--------------
  POST ---
  POST       ---
```

No?

Now we could stop Activity1 before we start Activity2. But there is a problem with that. We don't have a response or an exception to properly clean up Activity1. There is nothing triggered by HttpWebRequest when one of these internal resubmits happens we could listen to.

Basically we only ever get the final response. So if Activity1 failed and Activity2 succeeded, both would end up being decorated as if both succeeded. It would just look like a dupe to end users. Really what we would want to see on Activity1 is the original exception that triggered the resubmit. Then Activity2 we would want to see the final result. But I don't think that is possible looking at the reference source.

I'm thinking... we might be best just leaving this alone ☹️

src/OpenTelemetry.Instrumentation.Http/Implementation/HttpHandlerDiagnosticListener.cs
```diff
@@ -70,14 +70,6 @@ public override void OnStartActivity(Activity activity, object payload)
...
 
-            // TODO: Investigate why this check is needed.
-            if (Propagators.DefaultTextMapPropagator.Extract(default, request, HttpRequestMessageContextPropagation.HeaderValuesGetter) != default)
```
CodeBlanch, line 74: @denisivan0v @cijothomas 

FYI I took this for a spin too. HttpClient looks much better!

image

@denisivan0v Would you do me a favor and add this test to your PR?

* Modify the server listener like this...

```csharp
        public HttpClientTests()
        {
            this.serverLifeTime = TestHttpServer.RunServer(
                (ctx) =>
                {
                    if (ctx.Request.Url.PathAndQuery.Contains("redirect"))
                    {
                        ctx.Response.RedirectLocation = "/";
                        ctx.Response.StatusCode = 302;
                    }
                    else
                    {
                        ctx.Response.StatusCode = 200;
                    }

                    ctx.Response.OutputStream.Close();
                },
                out var host,
                out var port);

            this.url = $"http://{host}:{port}/";
        }
```

* Add this test...

```csharp
        [Fact]
        public async Task HttpClientRedirectTest()
        {
            var processor = new Mock<BaseProcessor<Activity>>();
            using (Sdk.CreateTracerProviderBuilder()
                .AddProcessor(processor.Object)
                .Build())
            {
                using var c = new HttpClient();
                await c.GetAsync($"{this.url}redirect");
            }

            Assert.Equal(7, processor.Invocations.Count); // SetParentProvider/OnShutdown/Dispose/OnStart/OnEnd/OnStart/OnEnd called.

            var firstActivity = (Activity)processor.Invocations[2].Arguments[0]; // First OnEnd

            Assert.Contains(firstActivity.TagObjects, t => t.Key == "http.status_code" && (int)t.Value == 302);

            var secondActivity = (Activity)processor.Invocations[4].Arguments[0]; // Second OnEnd

            Assert.Contains(secondActivity.TagObjects, t => t.Key == "http.status_code" && (int)t.Value == 200);
        }
```

That verifies for automatic redirect 2 spans are created and the status codes match.

src/OpenTelemetry.Instrumentation.Http/Implementation/HttpWebRequestActivitySource.netfx.cs
denisivan0v: Rolled back HttpWebRequest instrumentation from this PR. Saved changes in https://github.com/denisivan0v/opentelemetry-dotnet/tree/denisivan0v/http-retries-redirects-httpwebrequest branch in case we still need apply them in the future.

denisivan0v:
@open-telemetry/dotnet-maintainers can you please reopen this PR and have another look?

denisivan0v:
@cijothomas / @CodeBlanch / @vishweshbankwar can you please merge this PR in case it looks good for you?

src/OpenTelemetry.Instrumentation.Http/Implementation/HttpHandlerDiagnosticListener.cs
cijothomas: @denisivan0v Could you add this test for redirect specifically? OR is this already covered by existing tests?

denisivan0v: @cijothomas, @CodeBlanch added requested changes and the test for redirect, thanks! And sorry for missing this earlier :)

test/OpenTelemetry.Instrumentation.Http.Tests/HttpClientTests.Basic.netcore31.cs
```diff
@@ -402,6 +421,48 @@ public async Task HttpClientInstrumentationContextPropagation()
...
+                    {
+                        requestUri = $"{this.url}{response.Headers.Location}";
+                    }
+                    else
+                    {
+                        break;
+                    }
+                }
+#else
+                using var c = new HttpClient(new SocketsHttpHandler());
```
vishweshbankwar, line 452: Is this needed? `new SocketsHttpHandler()`

vishweshbankwar (approved):

test/OpenTelemetry.Instrumentation.Http.Tests/HttpClientTests.Basic.netcore31.cs
denisivan0v: Not really - removed.

src/OpenTelemetry.Instrumentation.Http/CHANGELOG.md
```diff
@@ -14,6 +14,10 @@
   occurs.
   ([#3407](https://github.com/open-telemetry/opentelemetry-dotnet/issues/3407))
 
+* Create spans for HTTP retries and redirects according to
```
cijothomas, line 17: could you describe the change here itself for users benefit? 
Having to read spec and PR to understand what changed is too complicated for end users.

cijothomas, line 17: could you describe the change here itself for users benefit? 
Having to read spec and PR to understand what changed is too complicated for end users.

src/OpenTelemetry.Instrumentation.Http/CHANGELOG.md
```diff
@@ -14,6 +14,10 @@
...
+  [opentelemetry-specification #2078](https://github.com/open-telemetry/opentelemetry-specification/pull/2078).
+  ([#3072](https://github.com/open-telemetry/opentelemetry-dotnet/pull/3072))
```
cijothomas, line 19: also be specific that this is only done to the HttpClient .NET Core one, and .NET Framework.

vishweshbankwar:
@cijothomas - Please reopen and merge this PR. I can address the comments in a separate PR. Thanks!
@denisivan0v FYI

vishweshbankwar:
> @cijothomas - Please reopen and merge this PR. I can address the comments in a separate PR. Thanks! @denisivan0v FYI

Looks like we have merge conflicts. @denisivan0v - I will create a separate PR using this one and tag you there.

mirekkukla:
We're very much looking to this being merged to finally resolve https://github.com/microsoft/ApplicationInsights-dotnet/issues/2556

cijothomas:
> We're very much looking to this being merged to finally resolve [microsoft/ApplicationInsights-dotnet#2556](https://github.com/microsoft/ApplicationInsights-dotnet/issues/2556)

This will not fix anything in ApplicationInsights, unless you are using OpenTelemetry based application insights.

@vishweshbankwar this PR can be now closed as this functionality is already merged now?

vishweshbankwar:
> > We're very much looking to this being merged to finally resolve [microsoft/ApplicationInsights-dotnet#2556](https://github.com/microsoft/ApplicationInsights-dotnet/issues/2556)
> 
> This will not fix anything in ApplicationInsights, unless you are using OpenTelemetry based application insights.
> 
> @vishweshbankwar this PR can be now closed as this functionality is already merged now?

Yes - This can be closed now.

cijothomas:
Closing as this is covered by alternative PR. Thanks @denisivan0v and @vishweshbankwar

prateekprshr-nith:
> Closing as this is covered by alternative PR. Thanks @denisivan0v and @vishweshbankwar

@cijothomas, can you please provide a link to alternative PR?

alanwest:
@prateekprshr-nith This was addressed in #3732

prateekprshr-nith:
Thanks @cijothomas, I see that this was released in https://github.com/open-telemetry/opentelemetry-dotnet/releases/tag/core-1.4.0-beta.2.

Is this available in a stable release as well?

Kielek:
Not yet, the latest release is 1.4.0-rc.1 for core components/1.0.0-rc9.10 for non-core.
ETA for 1.4.0 is end of January.

pvsraviteja:
@Kielek  Is this SDK released and available to consume ?

Kielek:
There is some delay:
https://www.nuget.org/packages/OpenTelemetry/1.4.0-rc.3 was released Feb 2, 2023.
From SIG notes: "After it simmers for a bit, we’ll ship 1.4 stable in ~2 weeks."

PR #3061, opened 2022-03-16
[Part0] Support Activity Status and status description in Console Exporter.

Yun-Ting:
Related to https://github.com/open-telemetry/opentelemetry-dotnet/issues/2569.

## Changes
Added support to to set status with Activity.Status, which was introduced in `System.Diagnostic.DiagnosticSource 6.0.0` to ConsoleExporter.
The same idea will be applied to other exporters.

**Case 1:** 
Setting activity status to be OK.
`activity.SetStatus(ActivityStatusCode.Ok);`
Output:
`StatusCode : OK`

**Case 2:** 
Setting activity status to be ERROR with description.
`activity.SetStatus(ActivityStatusCode.Error, "desc");`
Output:
```
StatusCode : ERROR
Error : desc
```

**Case 3:** 
Setting activity tag to be OK.
`activity.SetTag("otel.status_code", "OK");`
Output:
```
Activity.Tags:
   StatusCode : OK
```

**Case 4:** 
Setting activity tag to be ERROR with description.
```
activity.SetTag("otel.status_code", "ERROR");
activity.SetTag("otel.status_description", "desc");
```
Output:
```
Activity.Tags:
   StatusCode : ERROR
   Error : desc
```

**Case 5:** 
Activity status take precedence over activity tag when both were set.
```
activity.SetStatus(ActivityStatusCode.Ok);
activity.SetTag("otel.status_code", "ERROR");
activity.SetTag("otel.status_description", "desc");
```
Output:
```
Activity.Tags:
StatusCode : OK
```
**Case 6:** 
Activity status take precedence over activity tag when both were set.
```
activity.SetStatus(ActivityStatusCode.Error, "desc");
activity.SetTag("otel.status_code", "OK");
```
Output:
```
Activity.Tags:
StatusCode : ERROR
Error : desc
```

For significant contributions please make sure you have completed the following items:

* [x] Appropriate `CHANGELOG.md` updated for non-trivial changes
* [ ] Design discussion issue #
* [ ] Changes in public API reviewed

PR #3073, opened 2022-03-23
[Part2] Support Activity Status and status description in Jaeger Exporter.

Yun-Ting:
Fixes parts of https://github.com/open-telemetry/opentelemetry-dotnet/issues/2569.
Related to: https://github.com/open-telemetry/opentelemetry-dotnet/pull/3003
Background : Please review Console.Exporter where the same logic were exercised: https://github.com/open-telemetry/opentelemetry-dotnet/pull/3061.

Added support for Activity.Status/Description in Jaeger.Exporter.
System.Diagnostic.DiagnosticSource version 6.0.0 introduced native support for storing status/description in the Activity itself.

This PR modified the exporter to retrieve status from the newly added fields if it was set.
To maintain backward compatibility, the control flow would fell back to retrieve status from the activity.Tags if the native status/description were not set.

In order to support Activity Status takes precedence over Activity Tags, the logic that were in `ProcessJaegerTag` were extracted to be inside `ForEach()` method of `TagEnumerationState`.

## Changes

### DIFF
--- a/src/OpenTelemetry.Exporter.Jaeger/Implementation/JaegerActivityExtensions.cs
+++ b/src/OpenTelemetry.Exporter.Jaeger/Implementation/JaegerActivityExtensions.cs
@@ -54,6 +54,42 @@ namespace OpenTelemetry.Exporter.Jaeger.Implementation
 
             activity.EnumerateTags(ref jaegerTags);
 
+            if (activity.Status != ActivityStatusCode.Unset)
+            {
+                if (activity.Status == ActivityStatusCode.Ok)
+                {
+                    PooledList<JaegerTag>.Add(
+                        ref jaegerTags.Tags,
+                        new JaegerTag(SpanAttributeConstants.StatusCodeKey, JaegerTagType.STRING, vStr: "OK"));
+                }
+                else
+                {
+                    PooledList<JaegerTag>.Add(
+                        ref jaegerTags.Tags,
+                        new JaegerTag(SpanAttributeConstants.StatusCodeKey, JaegerTagType.STRING, vStr: "ERROR"));
+
+                    PooledList<JaegerTag>.Add(
+                        ref jaegerTags.Tags,
+                        new JaegerTag(JaegerErrorFlagTagName, JaegerTagType.BOOL, vBool: true));
+                }
+            }
+            else if (jaegerTags.StatusCode.HasValue && jaegerTags.StatusCode != StatusCode.Unset)
+            {
+                PooledList<JaegerTag>.Add(
+                        ref jaegerTags.Tags,
+                        new JaegerTag(
+                            SpanAttributeConstants.StatusCodeKey,
+                            JaegerTagType.STRING,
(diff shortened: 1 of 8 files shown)


test/OpenTelemetry.Exporter.Jaeger.Tests/Implementation/JaegerActivityConversionTest.cs
```diff
@@ -472,6 +472,106 @@ public void JaegerActivityConverterTest_Status_ErrorFlagTest(StatusCode expected
...
+        {
+            // Arrange
+            var activity = CreateTestActivity();
+            activity.SetStatus(expectedStatusCode);
+
+            // Act
+            var jaegerSpan = activity.ToJaegerSpan();
+
+            // Assert
+
```
reyang, line 489: nit, remove this empty line

reyang (approved):
LGTM.

cijothomas (changes requested):
Blocking to address the issue : https://github.com/open-telemetry/opentelemetry-dotnet/pull/3073/files#r833442567

src/OpenTelemetry.Exporter.Jaeger/Implementation/JaegerActivityExtensions.cs
```diff
@@ -312,7 +322,47 @@ public bool ForEach(KeyValuePair<string, object> activityTag)
...
+                                this.StatusCode = statusCode;
+                                return true;
+                            }
+                            else if (!statusCode.HasValue || statusCode == Trace.StatusCode.Unset)
+                            {
+                                // Unset Status is not sent: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk_exporters/jaeger.md#status
+                                return true;
+                            }
+
+                            this.StatusCode = statusCode;
```
cijothomas, line 346: https://github.com/open-telemetry/opentelemetry-dotnet/pull/3073/files#diff-6d1626406ea3dabab7357969eb1e8401d78114acaf5e916f6a85f300b23afe96R331-R348 - This whole block is doing unnecessary checks. It should just check if this tag is storing statuscode, and if yes, store it to this.StatusCode. We dont need of the other checks, as it is handled elsewhere.

Yun-Ting: It will provide early termination for the tags and perform normalization. I can merge to normalization logic onto the  ToJaegerSpan if we think that is a better way to go.

cijothomas: I am not sure what is the early termination benefit here. Its doing some checks, which are repeated when we set jaeger tags again.

The following should be enough
```
if (key == SpanAttributeConstants.StatusCodeKey)
{
StatusCode? statusCode = StatusHelper.GetStatusCodeForTagValue(jaegerTag.VStr);
this.StatusCode = statusCode;
return true;
}
```

cijothomas (approved):

PR #3074, opened 2022-03-23
View tests

alanwest:
Adding a number of broken tests

### DIFF
--- a/test/OpenTelemetry.Tests/Metrics/MetricAPITest.cs
+++ b/test/OpenTelemetry.Tests/Metrics/MetricAPITest.cs
@@ -489,11 +489,15 @@ namespace OpenTelemetry.Metrics.Tests
 
             meterProvider.ForceFlush(MaxTimeToAllowForFlush);
 
-            Assert.Single(exportedItems);
+            Assert.Equal(2, exportedItems.Count);
             var metric1 = new List<Metric>() { exportedItems[0] };
+            var metric2 = new List<Metric>() { exportedItems[1] };
 
             Assert.Equal("newname", exportedItems[0].Name);
+            Assert.Equal("newname", exportedItems[1].Name);
+
             Assert.Equal(20, GetLongSum(metric1));
+            Assert.Equal(20, GetLongSum(metric2));
         }
 
         [Fact]
@@ -578,13 +582,20 @@ namespace OpenTelemetry.Metrics.Tests
 
             meterProvider.ForceFlush(MaxTimeToAllowForFlush);
 
-            Assert.Single(exportedItems);
+            Assert.Equal(2, exportedItems.Count);
             var metric1 = new List<Metric>() { exportedItems[0] };
+            var metric2 = new List<Metric>() { exportedItems[1] };
+
             var tag1 = new List<KeyValuePair<string, object>> { tags[0] };
 
             Assert.Equal("name", exportedItems[0].Name);
+            Assert.Equal("name", exportedItems[1].Name);
+
             Assert.Equal(20, GetLongSum(metric1));
+            Assert.Equal(20, GetLongSum(metric2));
+
             CheckTagsForNthMetricPoint(metric1, tag1, 1);
+            CheckTagsForNthMetricPoint(metric2, tag1, 1);
         }
 
         [Fact]
--- a/test/OpenTelemetry.Tests/Metrics/MetricAPITest.cs
+++ b/test/OpenTelemetry.Tests/Metrics/MetricAPITest.cs
@@ -468,6 +468,204 @@ namespace OpenTelemetry.Metrics.Tests
         }
 
         [Fact]
+        public void DuplicateInstrumentRegistration_WithViews_TwoIdenticalInstruments_TwoViews_SameRename()
+        {
+            var exportedItems = new List<Metric>();
+
+            using var meter = new Meter($"{Utils.GetCurrentMethodName()}");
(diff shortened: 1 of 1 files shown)


test/OpenTelemetry.Tests/Metrics/MetricAPITest.cs
```diff
@@ -467,6 +467,242 @@ long GetAggregatedValue(Metric metric)
             }
         }
 
+        [Fact]
+        public void DuplicateInstrumentRegistration_WithViews_TwoIdenticalInstruments_TwoViews_SameRename()
```
cijothomas, line 471: this test matches my understanding.

test/OpenTelemetry.Tests/Metrics/MetricAPITest.cs
```diff
@@ -467,6 +467,242 @@ long GetAggregatedValue(Metric metric)
...
+            Assert.Single(exportedItems);
+            var metric1 = new List<Metric>() { exportedItems[0] };
+            Assert.Equal("newname", exportedItems[0].Name);
+
+            // Both instruments record
+            Assert.Equal(20, GetLongSum(metric1));
+        }
+
+        [Fact]
+        public void DuplicateInstrumentRegistration_WithViews_TwoIdenticalInstruments_TwoViews_DifferentTags()
```
cijothomas, line 502: this one matches my understanding

test/OpenTelemetry.Tests/Metrics/MetricAPITest.cs
```diff
@@ -467,6 +467,242 @@ long GetAggregatedValue(Metric metric)
...
+
+            // Both instruments record
+            Assert.Equal(20, GetLongSum(metric1));
+
+            // Only key1 is kept not key2
+            CheckTagsForNthMetricPoint(metric1, tag1, 1);
+        }
+
+        [Fact]
+        public void DuplicateInstrumentRegistration_WithViews_TwoIdenticalInstruments_TwoViews_SameTags()
```
cijothomas, line 550: this one matches my understanding

test/OpenTelemetry.Tests/Metrics/MetricAPITest.cs
```diff
@@ -467,6 +467,242 @@ long GetAggregatedValue(Metric metric)
...
+
+            Assert.Equal("name", exportedItems[0].Name);
+
+            // Both instruments record
+            Assert.Equal(20, GetLongSum(metric1));
+            CheckTagsForNthMetricPoint(metric1, tag1, 1);
+        }
+
+        [Fact]
+        public void DuplicateInstrumentRegistration_WithViews_TwoIdenticalInstruments_TwoViews_DifferentHistogramBounds()
```
cijothomas, line 597: this one matches my understanding

test/OpenTelemetry.Tests/Metrics/MetricAPITest.cs
```diff
@@ -467,6 +467,242 @@ long GetAggregatedValue(Metric metric)
...
+                    {
+                        return null;
+                    }
+                })
+                .AddInMemoryExporter(exportedItems);
+
+            using var meterProvider = meterProviderBuilder.Build();
+
+            var instrument1 = meter.CreateCounter<long>("name");
+            var instrument2 = meter.CreateCounter<long>("othername");
```
cijothomas, line 678: mm..this part is tricky.
The 2nd instrument, when following the View spec, should result in conflict, and hence must be ignored.
We should not merge it with the instrument1.

^ My read of spec.

alanwest: I know this isn't quite "duplicate instrument registration" since technically the instruments are not duplicates - and it is only until the view is used to create a duplicate stream, but this is the part of the spec that makes me think it was intended for every instrument registration to result in a recording instrument.

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument-type-conflict-detection

> When more than one Instrument of the same name is created for identical Meters, denoted duplicate instrument registration, the implementation MUST create a valid Instrument in every case. Here, "valid" means an instrument that is functional and can be expected to export data, despite potentially creating a [semantic error in the data model](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#opentelemetry-protocol-data-model-producer-recommendations).

cijothomas: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#resolving-duplicate-instrument-registration-conflicts

"The implementation MUST aggregate data from identical Instruments together in its export pipeline". -- This part of the spec says merging instruments together, but this is only for identical instruments. In this case, the instruments are not identical. User made a mistake in their View config, and we should write an error log about it, and ignore instrument2.

alanwest: And since the second instrument does not match a view then I think this part of the view spec applies

> If the Instrument could not match with any of the registered View(s), the SDK SHOULD enable the instrument using the default aggregation and temporality.

cijothomas: Okay I can see this can be interpreted differently. based on this, then 2nd instrument should be allowed to proceed with default view. So we should get 2 streams. The 2nd instrument should do defaults. 1st instrument should aggregate based on view config.
Ofcourse this results in duplicate stream in the exporters.

PR #3080, opened 2022-03-23
Improve internal log message about exporter to include reader information

cijothomas:
Sample output
OpenTelemetry-Sdk - EventId: [39], EventName: [MeterProviderSdkEvent], Message: [MeterProviderSdk event: 'Exporters added = 
"OpenTelemetry.Exporter.ConsoleMetricExporter (Paired with PeriodicExportingMetricReader exporting at 10000 milliseconds intervals.);OpenTelemetry.Exporter.PrometheusExporter (Paired with a MetricReader requiring manual trigger to export.)".']

### DIFF
--- a/src/OpenTelemetry/Metrics/MeterProviderSdk.cs
+++ b/src/OpenTelemetry/Metrics/MeterProviderSdk.cs
@@ -76,15 +76,15 @@ namespace OpenTelemetry.Metrics
 
                 if (reader is PeriodicExportingMetricReader periodicExportingMetricReader)
                 {
-                    exportersAdded.Append(((PeriodicExportingMetricReader)reader).Exporter);
+                    exportersAdded.Append(periodicExportingMetricReader.Exporter);
                     exportersAdded.Append(" (Paired with PeriodicExportingMetricReader exporting at ");
-                    exportersAdded.Append(((PeriodicExportingMetricReader)reader).ExportIntervalMilliseconds);
+                    exportersAdded.Append(periodicExportingMetricReader.ExportIntervalMilliseconds);
                     exportersAdded.Append(" milliseconds intervals.)");
                     exportersAdded.Append(';');
                 }
                 else if (reader is BaseExportingMetricReader baseExportingMetricReader)
                 {
-                    exportersAdded.Append(((BaseExportingMetricReader)reader).Exporter);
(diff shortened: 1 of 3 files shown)


src/OpenTelemetry/Metrics/MeterProviderSdk.cs
```diff
@@ -74,9 +74,18 @@ internal sealed class MeterProviderSdk : MeterProvider
                     this.reader = new CompositeMetricReader(new[] { this.reader, reader });
                 }
 
-                if (reader is BaseExportingMetricReader baseExportingMetricReader)
+                if (reader is PeriodicExportingMetricReader periodicExportingMetricReader)
+                {
+                    exportersAdded.Append(((PeriodicExportingMetricReader)reader).Exporter);
+                    exportersAdded.Append(" (Paired with PeriodicExportingMetricReader exporting at ");
+                    exportersAdded.Append(((PeriodicExportingMetricReader)reader).ExportIntervalMilliseconds);
```
cijothomas, line 81: no special casing for Timeout.Infinite, but i think even without that, this is still a useful log message.

alanwest (approved):
Very helpful! LGTM

src/OpenTelemetry/Metrics/MeterProviderSdk.cs
```diff
@@ -74,9 +74,18 @@ internal sealed class MeterProviderSdk : MeterProvider
                     this.reader = new CompositeMetricReader(new[] { this.reader, reader });
                 }
 
-                if (reader is BaseExportingMetricReader baseExportingMetricReader)
+                if (reader is PeriodicExportingMetricReader periodicExportingMetricReader)
+                {
+                    exportersAdded.Append(((PeriodicExportingMetricReader)reader).Exporter);
```
TimothyMothra, line 79: I think this can be simplified.
Line 77, the if statement uses the `is` operator to cast.
Lines 79 and 81 both perform the explicit conversion.

Seems like a preexisting issue, the same pattern is on lines 85 and 88.

reyang:
```suggestion
                    exportersAdded.Append(periodicExportingMetricReader.Exporter);
```

reyang (approved):

PR #3085, opened 2022-03-24
Control `MemoryDiagnoser` benchmark option via CLI instead of code

mic-max:
## Changes

The idea is to remove all the `MemoryDiagnoser` class attributes that we include in all the benchmark classes and instead just add the console argument -m to always enable it.


mic-max:
If people think this change is worthwhile I can do a final commit that removes all `[MemoryDiagnoser]` lines from benchmark classes like done here https://github.com/open-telemetry/opentelemetry-dotnet/pull/3085/commits/a57afff61448084329cc9183e9ff68ead900dcdc

test/Benchmarks/EventSourceBenchmarks.cs
```diff
@@ -20,7 +20,6 @@
 
 namespace OpenTelemetry.Benchmarks
 {
-    [MemoryDiagnoser]
```
cijothomas, line 23: what is the benefit of providing this via command line? 
The cmd line is straightforward right now, as one can run it just like any dotnet run. I don't see any advantage of making the command line user needing to remember one more param.

reyang (approved):

test/Benchmarks/EventSourceBenchmarks.cs
reyang: E.g. if I made a change that I know won't change any memory alloc, I could run the test in a much faster way 😃

cijothomas: okay that seems to be a good reason.

test/Benchmarks/README.md
```diff
@@ -1,10 +1,7 @@
 # OpenTelemetry Benchmarks
 
-Use the following example to run Benchmarks from command line:
-(change parameters as necessary)
-
 Navigate to `./test/Benchmarks` directory and run the following command:
 
 ```sh
-dotnet run --framework net6.0 --configuration Release --filter *TraceBenchmarks*
+dotnet run -c Release -f net6.0 -- -m
```
cijothomas, line 6: if we remove the filter , please add example with the filter as well below (to make it easy to copy paste to cmd).
Also - add a one line explaining why would folks want to use/not-use -m parameter.

reyang (approved):

cijothomas (approved):

PR #2395, opened 2021-09-21
Throw an exception when parsing an env var fails

pellared:
## Why

By @reyang from [Slack thread](https://cloud-native.slack.com/archives/C01N3BC2W7Q/p1631125173035600):

Environmental variables handling:

1. If we don't have access to the environment variable (due to security reasons), we'll proceed as if nothing is set in the env vars (just log)
2. When we process the environment variables, anything that wouldn't make logical sense (e.g. we expect int, but we got a string that doesn't look like int) would result in exceptions
3. We only process environment variables during the SDK initialize (e.g. by taking a snapshot), any env vars change later will not be used or picked up.

Reference: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/error-handling.md#basic-error-handling-principles

## Remarks

This PR is so small just to make sure we agree that this is the way want to handle invalid values of env vars.

The only problem with this approach is that if an env var is invalid e.g. `OTEL_BSP_EXPORT_TIMEOUT=invalid` then there is no other way of fixing it than changing or clearing the env var. Code like:

```csharp
 var options = new BatchExportActivityProcessorOptions
{
    ExporterTimeoutMilliseconds = 89000,
};
``` 

won't help as the env var parsing is in the constructor.

TBH I do not think it is bad but I prefer to double-check it with you.

PR #3098, opened 2022-03-28
Fix default URL for OTLP Trace/Metrics for http/protobuf

pjanotti:
## Changes

Setting only the protocol to `http/protobuf` wasn't enough to get the OTLP exporters, trace and metrics, sending to the correct endpoint. Setting the env var or the property directly works, but, it should just work by setting the protocol.


src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/BaseOtlpHttpExportClient.cs
```diff
@@ -46,17 +46,17 @@ protected BaseOtlpHttpExportClient(OtlpExporterOptions options, HttpClient httpC
...
-                using var httpRequest = this.CreateHttpRequest(request);
-
                 using var httpResponse = this.SendHttpRequest(httpRequest, cancellationToken);
 
                 httpResponse?.EnsureSuccessStatusCode();
             }
             catch (HttpRequestException ex)
             {
-                OpenTelemetryProtocolExporterEventSource.Log.FailedToReachCollector(this.Endpoint, ex);
+                OpenTelemetryProtocolExporterEventSource.Log.FailedToReachCollector(httpRequest.RequestUri, ex);
```
pjanotti, line 59: This change is to ensure that the log has the actual path that it attempted to use.

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/IntegrationTests.cs
```diff
@@ -59,11 +61,16 @@ public void TraceExportResultIsSuccess(OtlpExportProtocol protocol, string endpo
             }
 #endif
 
-            var exporterOptions = new OtlpExporterOptions
+            OtlpExporterOptions exporterOptions = new();
```
pjanotti, line 64: Not sure what style is preferred for this.

pjanotti:
🤔 the tests passed locally, need to look...

pjanotti:
I had a conversation with @pellared and will be updating the PR: I will be trying to remove the extension method that already does some checks regarding the path for the exporters.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/OtlpHttpMetricsExportClient.cs
```diff
@@ -33,25 +32,15 @@ namespace OpenTelemetry.Exporter.OpenTelemetryProtocol.Implementation.ExportClie
     internal sealed class OtlpHttpMetricsExportClient : BaseOtlpHttpExportClient<OtlpCollector.ExportMetricsServiceRequest>
     {
         internal const string MediaContentType = "application/x-protobuf";
-        private readonly Uri exportMetricsUri;
 
         public OtlpHttpMetricsExportClient(OtlpExporterOptions options, HttpClient httpClient)
-            : base(options, httpClient)
+            : base(options, httpClient, "v1/metrics")
```
pjanotti, line 37: Not sure of what is the preference for literals used in a single place, happy to add the corresponding constant if that is the preference.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/OtlpHttpTraceExportClient.cs
```diff
@@ -33,25 +32,15 @@ namespace OpenTelemetry.Exporter.OpenTelemetryProtocol.Implementation.ExportClie
     internal sealed class OtlpHttpTraceExportClient : BaseOtlpHttpExportClient<OtlpCollector.ExportTraceServiceRequest>
     {
         internal const string MediaContentType = "application/x-protobuf";
-        private readonly Uri exportTracesUri;
 
         public OtlpHttpTraceExportClient(OtlpExporterOptions options, HttpClient httpClient)
-            : base(options, httpClient)
+            : base(options, httpClient, "v1/traces")
```
pjanotti, line 37: Not sure of what is the preference for literals used in a single place, happy to add the corresponding constant if that is the preference.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpExporterOptionsExtensions.cs
```diff
@@ -155,25 +154,6 @@ public static void TryEnableIHttpClientFactoryIntegration(this OtlpExporterOptio
...
-        internal static void AppendExportPath(this OtlpExporterOptions options, string exportRelativePath)
-        {
-            // The exportRelativePath is only appended when the options.Endpoint property wasn't set by the user,
-            // the protocol is HttpProtobuf and the OTEL_EXPORTER_OTLP_ENDPOINT environment variable
-            // is present. If the user provides a custom value for options.Endpoint that value is taken as is.
-            if (!options.ProgrammaticallyModifiedEndpoint)
-            {
-                if (options.Protocol == OtlpExportProtocol.HttpProtobuf)
-                {
-                    if (EnvironmentVariableHelper.LoadUri(OtlpExporterOptions.EndpointEnvVarName, out Uri endpoint))
```
pjanotti, line 167: This env var is already read on the options constructor.

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/BaseOtlpHttpExportClientTests.cs
```diff
@@ -0,0 +1,67 @@
...
+                {
+                    options.Endpoint = new Uri(optionEndpoint);
+                }
+
+                var exporterClient = new TestOtlpHttpExportClient(options, new HttpClient());
+                Assert.Equal(new Uri(expectedExporterEndpoint), exporterClient.Endpoint);
+            }
+            finally
+            {
+                Environment.SetEnvironmentVariable(OtlpExporterOptions.EndpointEnvVarName, null);
```
Kielek, line 50: Consider pattern used in https://github.com/Kielek/opentelemetry-dotnet/blob/563236b51d4d1312677d749fface18b6422ebfb5/test/OpenTelemetry.Exporter.Jaeger.Tests/JaegerExporterOptionsTests.cs#L24-L32 - clear env vars before end after tests (very defensive approach)

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/BaseOtlpHttpExportClient.cs
```diff
@@ -26,13 +26,16 @@ namespace OpenTelemetry.Exporter.OpenTelemetryProtocol.Implementation.ExportClie
...
     {
-        protected BaseOtlpHttpExportClient(OtlpExporterOptions options, HttpClient httpClient)
+        protected BaseOtlpHttpExportClient(OtlpExporterOptions options, HttpClient httpClient, string signalPath)
         {
             Guard.ThrowIfNull(options);
             Guard.ThrowIfNull(httpClient);
+            Guard.ThrowIfNull(signalPath);
             Guard.ThrowIfInvalidTimeout(options.TimeoutMilliseconds);
 
-            this.Endpoint = new UriBuilder(options.Endpoint).Uri;
```
alanwest, line 35: This line was intentionally making a copy of `options.Endpoint` to prevent later changes to the options from affecting the exporter after it's been constructed. Though, it might have not been necessary since I think Uri may be immutable in practice. Maybe safer to just make a copy?

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/OtlpHttpMetricsExportClient.cs
alanwest: +1 to making it a const.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/BaseOtlpHttpExportClient.cs
pjanotti: 🤔 I always thought that `Uri` is immutable, see [`UriBuilder` remarks](https://docs.microsoft.com/en-us/dotnet/api/system.uribuilder?view=net-6.0#remarks). Let me see if there is anything to the contrary.

pjanotti: Looking at its source code some fields are filled on demand, but the calling code cannot change what the `Uri` points to. I think it is safe to remove the copy, that said, for everyone peace of mind near a release, for something that shouldn't be in a hot path, I will use the old code.

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/BaseOtlpHttpExportClientTests.cs
pjanotti: In practice this is what this code is doing: it sets the value for the test on line 36 and then cleans it here.

Kielek (approved):

pellared (approved):

cijothomas:
@pjanotti one conflict came another PR merge.. could you resolve that please?

pjanotti:
Done @cijothomas

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/BaseOtlpHttpExportClientTests.cs
```diff
@@ -0,0 +1,67 @@
...
+namespace OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests
+{
+    public class BaseOtlpHttpExportClientTests
+    {
+        [Theory]
+        [InlineData(null, null, "http://localhost:4318/signal/path")]
+        [InlineData(null, "http://from.otel.exporter.env.var", "http://from.otel.exporter.env.var/signal/path")]
+        [InlineData("https://custom.host", null, "https://custom.host")]
+        [InlineData("http://custom.host:44318/custom/path", null, "http://custom.host:44318/custom/path")]
+        [InlineData("https://custom.host", "http://from.otel.exporter.env.var", "https://custom.host")]
```
alanwest, line 31: I don't think this is anything new you've introduced, so I'm not looking to block your PR on this, but my preference is that environment variables take precedence over code, however, I'm not sure we ever came to a clear agreement on this.

This was discussed a bit here https://github.com/open-telemetry/opentelemetry-dotnet/pull/2395#issuecomment-924091917.

Also, I've spot checked a few other places and we are preferring code over env vars.

@cijothomas and @CodeBlanch, I just want to confirm that your preference is that config via code should take precedence over environment variables.

alanwest (approved):
LGTM. Just a question about config precedence not really related specifically to this PR.

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/BaseOtlpHttpExportClientTests.cs
pellared: > my preference is that environment variables take precedence over code

1. This would change the existing behavior (IMO breaking change)
2. This would be not consistent with other SDKs. At least the ones that I am aware of (Go, Python).

pjanotti: I agree with @pellared: env vars should only override the defaults, then any other value set in code should be respected. It seems the typical pattern for configuration in general: attempt to read from some source(s) while "creating" the config to perhaps override the defaults and then respect what the developer coded.

pjanotti:
Thanks for the review @alanwest 

@cijothomas @CodeBlanch ptal @ https://github.com/open-telemetry/opentelemetry-dotnet/pull/3098#discussion_r840975028

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/BaseOtlpHttpExportClientTests.cs
alanwest: > The current behavior is consistent with other SDKs. At least the ones that I am aware of (Go, Python).

Good to know, I was wondering about this. I also checked with Java and this seems to be the behavior there too. I'm content with these findings - it's a plus that the SIGs have maintained some consistency here.

cijothomas:
@pjanotti As as follow up, would you help modifying the examples to show how to use the http option?
For example: I think the collector needs to enable http support 1st...
https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/examples/Console/otlp-collector-example/config.yaml

Anything else?

pjanotti:
Sure thing @cijothomas - I will run the examples locally and create a PR if needed. IIRC the OTLP receiver on the collector receives both grpc and http by default, but, as I said I will double-check if the examples are looking good for OTLP.

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/BaseOtlpHttpExportClientTests.cs
cijothomas: https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/src/OpenTelemetry.Exporter.OpenTelemetryProtocol#environment-variables This may not be very obvious from the doc here, would be nice to have a follow up to this doc to avoid any confusion for end users.

PR #3110, opened 2022-03-30
Minor tweak to getting started tracing doc

cijothomas:
Modified example output to match current output.
Remove sampler from getting started. Its not really a concept required to get started, so removing it to keep things simple (under 5 minutes)

### DIFF
--- a/docs/trace/getting-started/README.md
+++ b/docs/trace/getting-started/README.md
@@ -31,12 +31,15 @@ Run the application again (using `dotnet run`) and you should see the trace
 output from the console.
 
 ```text
-Activity.Id:          00-8389584945550f40820b96ce1ceb9299-745239d26e408342-01
+Activity.TraceId:          d4a7d499698d62f0e2317a67abc559b6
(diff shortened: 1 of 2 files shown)


docs/trace/getting-started/README.md
```diff
@@ -74,12 +79,10 @@ An OpenTelemetry
 [TracerProvider](#tracerprovider)
 is configured to subscribe to the activities from the source
 `MyCompany.MyProduct.MyLibrary`, and export it to `ConsoleExporter`.
-`ConsoleExporter` simply displays it on the console. The `AlwaysOnSampler` is
-used by default which does not sample any traces.
```
reyang, line 78: +1, btw I feel that "does not sample any traces" might be interpreted in very different ways depending on the background of the reader, the spec in general uses "create" and "record" as the verb rather than "sample".

reyang (approved):

docs/trace/getting-started/README.md
cijothomas: yup.

PR #3111, opened 2022-03-30
Add SetStatus to getting started

cijothomas:
Now that there is proper support for `Status`, I think this concept is worth documenting in getting started. Open to other suggestions as well.

### DIFF
--- a/docs/trace/getting-started/Program.cs
+++ b/docs/trace/getting-started/Program.cs
@@ -36,5 +36,6 @@ public class Program
         activity?.SetTag("foo", 1);
         activity?.SetTag("bar", "Hello, World!");
         activity?.SetTag("baz", new int[] { 1, 2, 3 });
+        activity?.SetStatus(ActivityStatusCode.Ok);
     }
 }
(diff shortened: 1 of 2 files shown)


docs/trace/getting-started/Program.cs
```diff
@@ -36,5 +36,6 @@ public static void Main()
         activity?.SetTag("foo", 1);
         activity?.SetTag("bar", "Hello, World!");
         activity?.SetTag("baz", new int[] { 1, 2, 3 });
+        activity?.SetStatus(ActivityStatusCode.Ok);
```
cijothomas, line 39: depending on other's opinion on this, I will explore documenting this somewhere else...

reyang: maybe Error is better for demo purpose?

reyang (approved):

docs/trace/getting-started/Program.cs
cijothomas: I thought about it.. But it'd be more apt to show Error, with an Exception.. that would then bring all the complexities associated with Exception into a getting-started doc..... (we already have a section on Exception-reporting, which I'll tweak to reflect new Status api)

reyang: Makes sense, thanks!

Yun-Ting (approved):

Issue #3106, opened 2022-03-29
Markdown Only PRs blocked by required builds


TimothyMothra:
I've got some PRs updating docs and I've noticed that not all the builds run.

<img width="676" alt="image" src="https://user-images.githubusercontent.com/28785781/160723433-98649bdd-2523-4e82-a408-2795132c34a6.png">

I think this is due to this line:
https://github.com/open-telemetry/opentelemetry-dotnet/blob/add7eee59cb86fafec1896dd5ce54714348b5ae4/.github/workflows/windows-ci.yml#L10-L11

GitHub suggests having alternate workflows with the same name to work around this.
https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks

> If someone submits a pull request that changes a markdown file in the root of the repository, then the workflow above won't run at all because of the path filtering. As a result you won't be able to merge the pull request. You would see the following status on the pull request:
You can fix this by creating a generic workflow, with the same name, that will return true in any case similar to the workflow below :

--

## Scope
- [ ] apicompatibility (not required)
- [ ] code-coverage (not required)
- [x] integration #3132
- [x] linux-ci #3128
- [x] windows-ci #3116 
- [x] dotnet-format #3133

TimothyMothra:
FYI @cijothomas 

I'll start with the windows-ci and if it works as expected, will expand to the others.

TimothyMothra:
I tested a few iterations of this and my first PR is what I think is the best version.

For normal PRs (not documentation), I tried excluding this workflow but wasn't successful.
I verified it's not detrimental to have this run in addition to required workflows.
However it does make the list of Checks a little bit confusing with duplicate names.

After this first PR merges, I proposing that we wait a few days before adding the other workflows.
I'll check new PRs and verify that nothing is blocked/broken by adding this workflow.
I also want to confirm that having the duplicate checks doesn't irritate our maintainers before adding additional checks.

TimothyMothra:
tested with #3125.
I think we can close this issue :)

PR #3116, opened 2022-03-30
alternate workflow for "windows-ci" to accomodate documentation-only PRs.

TimothyMothra:
#3106.

## Changes
- new workflow "windows-ci-md" for documentation-only PRs. same name as "windows-ci"

I tested a few iterations of this on my forked repo.
The `name` and matrix values are required to match.

For normal PRs (not documentation),
I tried excluding this workflow but wasn't successful.
I verified it's not detrimental to have this run in addition to required workflows.
However it does make the list of Checks a little bit confusing with duplicate names.

https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks

> If someone submits a pull request that changes a markdown file in the root of the repository, then the workflow above won't run at all because of the path filtering. As a result you won't be able to merge the pull request. You would see the following status on the pull request:
You can fix this by creating a generic workflow, with the same name, that will return true in any case similar to the workflow below :


TimothyMothra:
I tested this in my forked repo and confirmed that the negative filter `!**.md`

TimothyMothra:
> Thanks. Will try this and let you know

This may need something extra to work with a version matrix. 
I'm testing in my fork.

I don't seem to have permission to change this PR to a DRAFT

.github/workflows/windows-ci-md.yml
```diff
@@ -0,0 +1,25 @@
...
+    paths-ignore:
+    - '!**.md'
+
+jobs:
+  build-test:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        version: [net461,netcoreapp3.1,net5.0,net6.0]
```
reyang, line 22: Is this line needed?

reyang (approved):

.github/workflows/windows-ci-md.yml
TimothyMothra: It appears to be required, yes.

The Required check includes the value from the matrix.
When testing without, that check isn't run.

reyang: Thanks for confirmation!

PR #3119, opened 2022-03-31
Improve View error messaging

cijothomas:
Follow up from https://github.com/open-telemetry/opentelemetry-dotnet/pull/3117 by @CodeBlanch 

If a View configuration was invalid, and we were able to detect it at runtime only, this PR improves the messaging.

Consider the following View configuration, which provides invalid Histogram boundaries.
```
.AddView((instrument) =>
                {
                    return instrument.Name == "myname"
                        ? new ExplicitBucketHistogramConfiguration() { Boundaries = invalidones }
                        : null;
                })
```

Without this PR:

OpenTelemetry-Sdk - EventId: [33], EventName: [MetricInstrumentIgnored], Message: [Measurements from Instrument 'counter1', Meter 'AddViewWithInvalidHistogramBoundsIgnored' will be ignored. Reason: 'SDK internal error occurred.'. Suggested action: 'Contact SDK owners..']

With this PR:

OpenTelemetry-Sdk - EventId: [33], EventName: [MetricInstrumentIgnored], Message: [Measurements from Instrument 'counter1', Meter 'AddViewWithInvalidHistogramBoundsIgnored' will be ignored. Reason: 'Applying View Configuration failed with error:Histogram boundaries are invalid. Histogram boundaries must be in ascending order with distinct values. (Parameter 'value')'. Suggested action: 'Fix the view configuration.']

The message could be improved, but this is better than blaming SDK Internal error.

### DIFF
--- a/src/OpenTelemetry/Metrics/MeterProviderSdk.cs
+++ b/src/OpenTelemetry/Metrics/MeterProviderSdk.cs
@@ -163,8 +163,10 @@ namespace OpenTelemetry.Metrics
                             {
                                 metricStreamConfig = viewConfig(instrument);
                             }
-                            catch (ArgumentException ex)
+                            catch (Exception ex)
                             {
+                                // TODO: This allocates the string even if none listening,
+                                // could be improved with a separate dedicated method.
                                 OpenTelemetrySdkEventSource.Log.MetricInstrumentIgnored(instrument.Name, instrument.Meter.Name, "Applying View Configuration failed with error:" + ex.Message, "Fix the view configuration.");
                                 return;
                             }
--- a/src/OpenTelemetry/Metrics/MeterProviderSdk.cs
+++ b/src/OpenTelemetry/Metrics/MeterProviderSdk.cs
@@ -157,7 +157,18 @@ namespace OpenTelemetry.Metrics
                         var metricStreamConfigs = new List<MetricStreamConfiguration>(viewConfigCount);
                         foreach (var viewConfig in this.viewConfigs)
                         {
-                            var metricStreamConfig = viewConfig(instrument);
+                            MetricStreamConfiguration metricStreamConfig = null;
+
+                            try
+                            {
+                                metricStreamConfig = viewConfig(instrument);
+                            }
+                            catch (ArgumentException ex)
+                            {
+                                OpenTelemetrySdkEventSource.Log.MetricInstrumentIgnored(instrument.Name, instrument.Meter.Name, "Applying View Configuration failed with error:" + ex.Message, "Fix the view configuration.");
+                                return;
+                            }
+
                             if (metricStreamConfig != null)
                             {
                                 metricStreamConfigs.Add(metricStreamConfig);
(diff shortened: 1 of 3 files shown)


src/OpenTelemetry/Metrics/MeterProviderSdk.cs
```diff
@@ -157,7 +157,18 @@ internal sealed class MeterProviderSdk : MeterProvider
...
-                            var metricStreamConfig = viewConfig(instrument);
+                            MetricStreamConfiguration metricStreamConfig = null;
+
+                            try
+                            {
+                                metricStreamConfig = viewConfig(instrument);
+                            }
+                            catch (ArgumentException ex)
+                            {
+                                OpenTelemetrySdkEventSource.Log.MetricInstrumentIgnored(instrument.Name, instrument.Meter.Name, "Applying View Configuration failed with error:" + ex.Message, "Fix the view configuration.");
```
cijothomas, line 168: yes this allocates string, even if none listening. Could have added a different Event to avoid this, but I think this is acceptable, given that this is not hot path.

src/OpenTelemetry/Metrics/MeterProviderSdk.cs
```diff
@@ -157,7 +157,18 @@ internal sealed class MeterProviderSdk : MeterProvider
...
+                                return;
```
reyang, line 169: Just to confirm - we want to proceed (with a warning/error msg) rather than rethrow?

cijothomas: Yes. This is instrument creation path, and it can occur anytime, so we are following the general principle of not punishing user after SDK is initialized successfully.

src/OpenTelemetry/Metrics/MeterProviderSdk.cs
```diff
@@ -157,7 +157,18 @@ internal sealed class MeterProviderSdk : MeterProvider
...
                         {
-                            var metricStreamConfig = viewConfig(instrument);
+                            MetricStreamConfiguration metricStreamConfig = null;
+
+                            try
+                            {
+                                metricStreamConfig = viewConfig(instrument);
+                            }
+                            catch (ArgumentException ex)
+                            {
```
cijothomas, line 167: `ArgumentException` only captures the ones thrown by the SDK while validating the arguments.. If the user code throws something else, it'd still fallback to "SDK Error Occurred".... Need to comeback to this to see if we should club that into "hey user, your View config is bad, please fix it...."

src/OpenTelemetry/Metrics/MeterProviderSdk.cs
reyang: I see, thanks!

src/OpenTelemetry/Metrics/MeterProviderSdk.cs
reyang: Consider adding a TODO comment

reyang (approved):

src/OpenTelemetry/Metrics/MeterProviderSdk.cs
cijothomas: Modified to catch `Exception`

PR #3120, opened 2022-03-31
Bad View result in ignoring the View but not the instrument itself

cijothomas:
Following from https://github.com/open-telemetry/opentelemetry-dotnet/pull/3119 and https://github.com/open-telemetry/opentelemetry-dotnet/pull/3117

This PR makes the following change:
If a View Config is invalid, and it was detected after the `MeterProvider` was built, we simply Ignore that particular View Config, and continue with the Instrument. If there are other Views, which matches the instrument, then it'll be applied. If there are none, a default configuration is applied (as required by the spec) In short, a "bad view results in *that* view being ignored. SDK proceeds as if this View did not even exist".

Added test to make the point clear. 

*NOTE* : This is different from a View which is invalid and is detected statically. In those cases, we simply throw, and fail to build the `MeterProvider` itself.

General idea with Views:
We validate and fail-fast whenever feasible.
If we missed the window to do fail-fast, then we would ignore the View (log that it is ignored), but continue with instrument as if **that** View did not exist.

### DIFF
--- a/src/OpenTelemetry/Internal/OpenTelemetrySdkEventSource.cs
+++ b/src/OpenTelemetry/Internal/OpenTelemetrySdkEventSource.cs
@@ -378,6 +378,12 @@ namespace OpenTelemetry.Internal
             this.WriteEvent(40, message);
         }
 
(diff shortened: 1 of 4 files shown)


cijothomas:
@alanwest merging to make progress. Please comment as you see issues/different interpretation.

alanwest (commented):
👍

Issue #3031, opened 2022-03-11
Spell check in CI seems to not useful


cijothomas:
https://github.com/open-telemetry/opentelemetry-dotnet/pull/3026/files#r824913711 Check this comment. The CI was supposed to have caught these, but apparently it is not. Opening an issue to see if the CI needs some fix, or we simply need a alternate tool.

cijothomas:
More https://github.com/open-telemetry/opentelemetry-dotnet/pull/3127

cijothomas:
![image](https://user-images.githubusercontent.com/5232798/163417350-c552a9cb-cbf1-45ce-a4f2-51ff3d73cb77.png)
Its not entirely useless.. Maybe need to send spell-check to a better English school 🙄

PR #3127, opened 2022-04-01
Add VS Code recommended spell check extension

tillig:
## Changes

- Added a spell checker extension as a recommended VS Code extension. (People aren't _required_ to install it, but it's _suggested_.)
- Added some custom words including names to the local workspace dictionary.
- Cleaned up some of the typos and misspellings found using the spell checker. No public APIs affected.

There are definitely a few words I didn't add to the dictionary, like when URLs have anchors (`http://server.com/#anchor-with-weirdlonglowercaseword`). Unclear if those should be in the dictionary or not. I figured probably not, but could be added later.


.vscode/settings.json
```diff
@@ -0,0 +1,48 @@
+{
+  "cSpell.words": [
+    "asax",
+    "cijo",
```
cijothomas, line 4: 😆 this is interesting.. I think the only place where people's name are relevant are the main readme which lists maintainer/approver/etc. 
Could we exclude a section from spell check? Or anything to avoid listing all names in dictionary..

tillig: This is like how I mentioned I just sort of... _ignored_ the fact that `weirdlongwordinaurl` isn't really a word. The spell checker see's it's not a word, it reports it's not a word, but I just didn't do anything about it. We could always just ignore that the spell checker reports people's names. However, I thought it would be more pleasant for the folks listed to see that the checker "knows their name" rather than reporting it as a misspelled word. I don't think _literally every contributor_ shows up in here so the list shouldn't change that often, right?

You can set up an `ignorePaths` configuration for the spell checker to ignore certain full documents, but it seemed like it was valuable to actually spell check the main README since it's... the main README. There is not a way to say "ignore just this one section" that I'm aware of. Perhaps the listing of all the contributors, etc., should be on a separate page that's linked from the main README; that'd allow for exclusion of a whole document.

Should the [GitHub pipeline action that checks spelling](https://github.com/open-telemetry/opentelemetry-dotnet/blob/7ed97be90841221ec2451d8d14a622768cb214b5/.github/workflows/sanitycheck.yml#L17) get fixed and start working, I don't think that has the ability to ignore sections of docs, either.

cijothomas: Want to hear what other people think of having people's name in dictionary.
@open-telemetry/dotnet-approvers

reyang: It seems weird to me but what @tillig explained makes sense. I'm fine either way.

My only question - is this guarded by CI? If not, it's likely that things will start to rot over time. (my personal take is that if something is not controlled by CI, then it's likely to be unsuccessful in the long run)

cijothomas: *this*, as of now, is not CI guarded. Once https://github.com/open-telemetry/opentelemetry-dotnet/issues/3031 is fixed, this would be guarded by CI, as adding a new maintainer/approver etc. without modifying the dictionary will break CI. Though there can be extra names in the dictionary, if approvers/maintainers leave a lot.

I'll approve this change now. Once https://github.com/open-telemetry/opentelemetry-dotnet/issues/3031 is addressed, we'll no longer have to worry about rotting.

cijothomas (approved):

reyang (approved):

PR #3134, opened 2022-04-01
new classes ExportedMetric and Exporter

TimothyMothra:
contributing to #2361 

## Changes
- new class `ExportedMetric`. This class is used to deep-copy `Metric`. 
- new exporter. This will be RECOMMENDED for Metrics unit tests.
  - Option 1: new class `InMemoryMetricExporter`. This is a one-off Exporter that directly creates `ExporterMetric`.
  - Option 2: new class `InMemoryTranslationExporter<Tinput, Toutput>`. This is a generic Exporter, and the caller MUST provide a Func to translate input/output types.
- new class `InMemoryMetricExporter`. This will be RECOMMENDED for unit tests.

- Modify `OpenTelemetry` project to expose internals to the `InMemoryExporter` project. (aka ["cheating"](https://github.com/open-telemetry/opentelemetry-dotnet/pull/3071#discussion_r833602498))
- Modify `InMemoryExporter` removing classes now exposed from `OpenTelemetry` project.

## Next Steps
- After this merges, I'll create a final PR to update the extension method in `InMemoryExporterMetricsExtensions` and specific unit tests.

## Discussion
This design is based on examples shared by @CodeBlanch (https://github.com/open-telemetry/opentelemetry-dotnet/compare/main...CodeBlanch:inmemory-metrics-copy?expand=1).

The end goal is to enable the `InMemoryExporter` to deep-copy `Metric`s.
The `Metric` class itself cannot be copied. (For more details see https://github.com/open-telemetry/opentelemetry-dotnet/issues/2361#issuecomment-1042421246).
Instead an instance of `Metric` can be provided to the new `ExportedMetric` class which copies the minimum necessary members.
This builds on the changes from #3113 to make copies of `MetricPoint` and `HistogramBuckets`.

## Alternative
Currently the `InMemoryExporter<T>` will export the same Type that it receives in the exported `Batch<T>`
Alternatively, we could introduce a new generic class that allows the caller to specify both the `Tinput, Toutput` and a provide translation `Func<Tinput, Toutput>`.

- Pro: Much more flexible, not tied to any specific use case.
- Con: As of today, we don't need additional exporters
- Pro: Future proofs us from a scenario of several one-off exporters.

<details>
    <summary>alternative implementation example</summary>

```
public class InMemoryTransformationExporter<Tinput, Toutput> : BaseExporter<Tinput>
{
    public InMemoryTransformationExporter(ICollection<Toutput> exportedItems, Func<Tinput, Toutput> transformation)
    {
        this.translation = transformation;
    }

    public override ExportResult Export(in Batch<Tinput> batch)
    {
        foreach (var data in batch)
        {
            this.exportedItems.Add(this.transformation(data));
        }
    }
}
```

</details>


src/OpenTelemetry.Exporter.InMemory/InMemoryMetricExporter.cs
```diff
@@ -0,0 +1,47 @@
...
+// limitations under the License.
+// </copyright>
+
+using System.Collections.Generic;
+
+using OpenTelemetry.Metrics;
+
+namespace OpenTelemetry.Exporter
+{
+    public class InMemoryMetricExporter : BaseExporter<Metric>
```
reyang, line 23: Why do we need a new exporter class?

TimothyMothra: We need an exporter that can perform the translation from `Metric` to `ExportedMetric`, and this new exporter does that.

The current `InMemoryExporter` is generic and the exported Type must match the Type in the Batch.
https://github.com/open-telemetry/opentelemetry-dotnet/blob/d3c0c6ed614b1496d47e58a184f462be71e22558/src/OpenTelemetry.Exporter.InMemory/InMemoryExporter.cs#L21-L31

reyang: Sorry, what's the real problem?

TimothyMothra: The real problem is that users cannot depend on a normally exported `Metric` to be representative of the point in time it was exported.
Because the MetricApi reuses `Metric` instances (by design), an exported `Metric` can continue to be updated.

The `Metric` class itself cannot be deep-copied, therefore we're introducing a new class which can be representative of the `Metric` and a new Exporter to perform the translation.

TimothyMothra: Something I don't like about this implementation is that it's a one-off class for a specific use case.
We could attempt to future the Exporter library by defining an InMemoryTranslationExporter<Tinput, Toutput>.
It's _still_ a new class, but this would be a single new class to hopefully prevent dozens of one-off classes.

TimothyMothra: Hi @reyang, circling back to this.

The problem I'm trying to address is that `InMemoryExporter<T>` receives and exports the same type, in this case `Metric`.
I want to introduce a new type `ExportedMetric` which is our copy of the `Metric`.

I don't see a way to accomplish this copy without introducing a new Exporter.

- Option 1: the simplest way to do this is a one-off exporter which performs this copy; `InMemoryMetricExporter`
- Option 2: to prevent this and future one-off exporters, I could solve this with an `InMemoryTranslationExporter<Tinput, Toutput>`.
  In this case, the caller must provide a `Func<Tinput, Toutput>`. 
  
In either option, the caller would be the extension method `AddInMemoryExporter` which I will contribute in a follow up PR along with test changes.

reyang: Can InMemoryExporter be improved to cover the case, so you don't have to introduce a new class?

TimothyMothra: Yes, I could introduce a means to provide an different behavior to the Export method.
I'm struggling with the best way to add this to the PublicApi and also explain customers would use it.

On this same thought, I'm investigating a different approach. I want to open a draft PR to get more feedback.

TimothyMothra:
abandoning in favor of #3198

PR #3148, opened 2022-04-07
Always generate a new metric stream for each view an instrument matches

alanwest:
This is a follow up from https://github.com/open-telemetry/opentelemetry-dotnet/pull/3006#issuecomment-1090875850

Followups included in this PR:
* Each view an instrument matches now results separate stream.
* View related tests have been moved to MetricViewTests.cs. (Some refactoring happened to move some helper methods from MetricApiTests.cs)
* Tests have been expanded.

Followups not included in this PR:
* Using `IStructuralEquatable` https://github.com/open-telemetry/opentelemetry-dotnet/pull/3006#discussion_r832545920


src/OpenTelemetry/Metrics/MetricReaderExt.cs
```diff
@@ -137,45 +125,39 @@ internal List<Metric> AddMetricsListWithViews(Instrument instrument, List<Metric
...
                     if (this.instrumentIdentityToMetric.TryGetValue(metricStreamIdentity, out var existingMetric))
                     {
-                        // The list of metrics may already contain a matching metric with the same
-                        // identity when a single instrument is selected by multiple views.
-                        if (!metrics.Contains(existingMetric))
-                        {
-                            metrics.Add(existingMetric);
-                        }
-
+                        metrics.Add(existingMetric);
```
alanwest, line 128: The conditional here is no longer required. Its purpose was to "merge" together identical streams resulting from different views. This primary purpose of this PR is to _not_ do this merging.

src/OpenTelemetry/Metrics/MetricStreamIdentity.cs
```diff
@@ -24,16 +24,19 @@ namespace OpenTelemetry.Metrics
...
+            this.Description = metricStreamConfiguration?.Description ?? instrument.Description ?? string.Empty;
+            this.InstrumentType = instrument.GetType();
+            this.ViewId = metricStreamConfiguration?.ViewId;
+            this.MetricStreamName = $"{this.MeterName}.{this.MeterVersion}.{this.InstrumentName}";
+
+            var tagKeys = metricStreamConfiguration?.CopiedTagKeys;
+            if (tagKeys != null)
             {
                 this.TagKeys = new string[tagKeys.Length];
                 tagKeys.CopyTo(this.TagKeys, 0);
```
alanwest, line 42: Hmm. I guess this copying may no longer be required since we're making a copy of stuff further upstream. Likewise for histogram bounds.

cijothomas: agree

PR #3161, opened 2022-04-12
Remove TryFindExporter from MeterProvider public API

cijothomas:
https://github.com/open-telemetry/opentelemetry-dotnet/issues/2620#issuecomment-1097122573

^ As per this, only use case is for Prometheus Exporter. We'll explore how to make Prometheus work without this. Given Prometheus is not going stable, but the SDK is, proposing to make this internal, and give special access to Prometheus temporarily to unblock stable release.


CodeBlanch (approved):
LGTM

Just FYI, the reason this is needed is so the middleware can call the exporter: 

https://github.com/open-telemetry/opentelemetry-dotnet/blob/c4abf0b330f1aac7c4f0e319957659d9f7730078/src/OpenTelemetry.Exporter.Prometheus/Implementation/PrometheusExporterMiddleware.cs#L43

PR #3162, opened 2022-04-13
Spell monotonic correctly

alanwest:


src/OpenTelemetry/Metrics/MetricReaderTemporalityPreference.cs
```diff
@@ -28,7 +28,7 @@ public enum MetricReaderTemporalityPreference
     Cumulative = 1,
 
     /// <summary>
-    /// All measurements that are monotnic in nature are aggregated using delta temporality.
+    /// All measurements that are monotonic in nature are aggregated using delta temporality.
```
cijothomas, line 31: Be careful! else you'll lose the reputation as the go-to person for any English wording questions 🤣

cijothomas (approved):

Issue #917, opened 2020-07-24
Should all packages have the same version?


reyang:
Currently we're taking the easy approach - all the packages have the same version.

We were using this approach initially in OpenCensus Python, until we hit beta - there were new packages being added which are not ready (based on beta quality) so we ended up allowing each package to have its independent version. And this is what we carried over to OpenTelemetry Python.

Some of the obvious pros/cons to consider:
1. Having independent version gives us extra work.
2. Not having independent version means if someone added a cool component (e.g. logging), it will immediately inherit whatever version used across the project (e.g. it could be GA 1.0).

pcwiese:
In my opinion, true semantic versioning is totally lost when everything shares the same package version. From a consumer perspective, it is better to be able to pull in targeted fixes rather than be forced to upgrade all packages. But doing this is hard and introduces developer friction for sure.

beef3333:
I have had good success using this to help automate generating versions per package, https://github.com/dotnet/Nerdbank.GitVersioning

alanwest:
I think it's nearing time for us to come back to this discussion.

Up to now, we have maintained a distinction between core and non-core packages.

## Core packages

Per our current [versioning guidelines](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/VERSIONING.md#core-components), we have indicated that core packages will always have the same version.

> Core components refer to the set of components which are required as per the spec. This includes API, SDK, and exporters which are required by the specification. These exporters are OTLP, Jaeger, Zipkin, Console and InMemory.
>
> The core components are always versioned and released together. For example, if Zipkin exporter has a bug fix and is released as 1.0.1, then all other core components are also released as 1.0.1, even if there is no code change in other components.

### Option 1: Maintain the current versioning policy long-term

There are a number of packages which should be considered core but do not yet have a stable release. I believe our process should be that as these packages reach stability, we should bump their versions to whatever the latest stable version is for the rest of the core packages. This includes the following:

* Future core packages
  * Prometheus exporter
  * Prometheus framework dependent bits - e.g. Prometheus.Exporter.AspNetCore (or whatever we end up calling it)
  * Jaeger propagator (#1881)
  * OpenTracing shim
  * ZPages exporter
  * Any other future component that becomes specified

[...]

PR #3197, opened 2022-04-15
Update changelog for next release

cijothomas:
1.3.0.beta.1 for Core
  (A bit odd for Prometheus which goes 1.2.0-rc5 to 1.3.0-beta. After the initial stable release, we expect Prometheus to be versioned release together with the core packages)
1.0.0-rc9.3 for others.


reyang (approved):

reyang (approved):
LGTM.

src/OpenTelemetry.Exporter.Prometheus/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+## 1.3.0-alpha.1
```
reyang: I'm kind of battling with myself.

Do we think we can/should do the same for Instrumentation Libraries? (e.g. OpenTelemetry.Instrumentation.Http)

Also it seems a bit weird to have the following release sequence:

* 1.2.0-rc1
* 1.2.0-rc2
* 1.2.0-rc3
* 1.2.0-rc4
* 1.2.0-rc5
* BOOM there is no 1.2.0 Stable because we're waiting for the spec
* 1.3.0-alpha.1

alanwest: I think what we need to decide in the long run is what our versioning scheme should communicate to end users. Looks like you started this discussion with #917. I'll add some thoughts on that issue to continue the conversation.

Even though it's odd, I think it's ok for now that Prometheus will jump past a 1.2.0 stable release.

>  if OpenTelemetry.Instrumentation.Http is Stable today, do we release it as 1.0.0 or 1.2.0? (I feel it should be 1.2.0)

I also agree that it should be 1.x.0 where x = current core stable version at the time instrumentation goes stable. Again, while odd, I think it'll be ok to be odd up to the point where we've gotten a stable release out of all components.

alanwest (approved):

mic-max (approved):

PR #3198, opened 2022-04-15
[DRAFT] fixing InMemoryExporter & Metrics bug. new class: ExportableMetricCopy. new ctor: InMemoryExporter(Func)

### LINKED ISSUE #2361: InMemoryExporter for Metric is buggy

TimothyMothra:
Fixes #2361

## Changes
- new class `ExportableMetricCopy`. This is used for a strategic selective copy of the minimum members to facilitate unit testing.
- new constructor added to `InMemoryExporter` to accept `Func<Batch<T>, ExportResult`.
  This allows the export behavior to be fully replaced and provides greater flexibility for unit tests.
- new extension methods `AddInMemoryExporter(ICollection<ExportableMetricCopy>)` added to `InMemoryExporterMetricsExtensions`.
- all affected tests have been updated

### Please provide a brief description of the changes here.

`Metric` instances are reused by design. 
As a side effect, the `InMemoryExporter` will repeatedly export duplicates of `Metric` instances.
Because of this, `Metric`s can be modified after export. See also: https://github.com/open-telemetry/opentelemetry-dotnet/issues/2361#issuecomment-1049175566

This change introduces a new class `ExportableMetricCopy` to be used in unit tests. 
This class is a collection of minimum members needed for existing unit tests.
An end user has the option to use either this copy or the original `Metric`.

This change also introduces a new ctor `public InMemoryExporter(Func<Batch<T>, ExportResult> exportFunc)`.
This is needed to export a type that differs from the `Batch` type, such as converting `Metric` to `ExportableMetricCopy`. 
This is also used by other tests: 
- `PrometheusSerializerTests` needs `Metric` to test `PrometheusSerializer.WriteMetric()`
- `OtlpMetricsExporterTests` needs `Batch<Metric>` to test `ExportMetricsServiceRequest.AddMetrics()`
  

### For significant contributions please make sure you have completed the following items:

## Todo
- waiting for general acceptance before updating documentation.
  - [ ] Update InMemoryExporter changelog
  - [ ] Update InMemoryExporter readme

## Alternate approaches investigated.
- Deep-clone of `Metric` has many obstacles. (Not `Serializeable`, missing parameterless constructors, significant MetricPoint buffer could cause OOM).
- Re-Using `Metric` class for a selective copy would give a false impression of a deep-clone and introduces a LARGE maintenance tax.


src/OpenTelemetry.Exporter.InMemory/InMemoryExporterMetricsExtensions.cs
```diff
@@ -36,6 +36,10 @@ public static class InMemoryExporterMetricsExtensions
         /// <param name="builder"><see cref="MeterProviderBuilder"/> builder to use.</param>
         /// <param name="exportedItems">Collection which will be populated with the exported MetricItem.</param>
         /// <returns>The instance of <see cref="MeterProviderBuilder"/> to chain the calls.</returns>
+        /// <remarks>Note that when working with <see cref="Metric"/> these objects are not deep-cloned,
+        /// and repeated flushes may introduce duplicates.
+        /// If you need deep-cloned it is recommended to use <see cref="ExportableMetricCopy"/>.
+        /// </remarks>
         public static MeterProviderBuilder AddInMemoryExporter(this MeterProviderBuilder builder, ICollection<Metric> exportedItems)
```
reyang, line 43: Can this be removed?

src/OpenTelemetry/AssemblyInfo.cs
```diff
@@ -20,5 +20,6 @@
 [assembly: InternalsVisibleTo("OpenTelemetry.Exporter.Prometheus" + AssemblyInfo.PublicKey)]
 [assembly: InternalsVisibleTo("OpenTelemetry.Exporter.Prometheus.Tests" + AssemblyInfo.PublicKey)]
 [assembly: InternalsVisibleTo("OpenTelemetry.Extensions.Hosting.Tests" + AssemblyInfo.PublicKey)]
+[assembly: InternalsVisibleTo("OpenTelemetry.Exporter.InMemory" + AssemblyInfo.PublicKey)]
```
reyang, line 23: nit, sort alphabetically

src/OpenTelemetry.Exporter.InMemory/InMemoryExporterMetricsExtensions.cs
TimothyMothra: I agree offering both extension methods can cause confusion.

I can't remove it entirely because the unit tests for Prometheus and OTLP are using this. 
`PrometheusSerializer.WriteMetric()` and `OtlpCollector.ExportMetricsServiceRequest.AddMetrics()` both need the actual `Metric` object.
These tests could possibly be re-written, but I think that should be addressed in a separate PR.

Regarding the PublicAPI, yes! 
This hasn't shipped yet and we can at least mark these as `internal`.

However, _should_ we remove this from the public api?
I don't know this well enough to say if there are legitimate use cases for a public user to have access to the `Metric`.

TimothyMothra: Thinking out loud,
We can remove the extension method and users can still manually create the `InMemoryExporter<Metric>`

reyang: I would think users should not be able to create an instance of `InMemoryExporter<Metric>` at all.

reyang: > Thinking out loud, We can remove the extension method and users can still manually create the `InMemoryExporter<Metric>`

This sounds like a bug we should fix.

TimothyMothra: I made a change that allows `InMemoryExporter` to be constructed with a `Func<Batch<T>, ExportResult>` which fully replaces the `Export()` method.
This unblocks Prometheus and OTLP tests which were previously dependent on `InMemoryExporter<Metric>` and allows us to remove those extension methods.

src/OpenTelemetry.Exporter.InMemory/OpenTelemetry.Exporter.InMemory.csproj
```diff
@@ -27,8 +27,9 @@
   </ItemGroup>
 
   <ItemGroup>
-    <Compile Include="$(RepoRoot)\src\OpenTelemetry\Internal\PeriodicExportingMetricReaderHelper.cs" Link="Includes\PeriodicExportingMetricReaderHelper.cs" />
-    <Compile Include="$(RepoRoot)\src\OpenTelemetry\Internal\ServiceProviderExtensions.cs" Link="Includes\ServiceProviderExtensions.cs" />
+    <!--<Compile Include="$(RepoRoot)\src\OpenTelemetry\Internal\PeriodicExportingMetricReaderHelper.cs" Link="Includes\PeriodicExportingMetricReaderHelper.cs" />
+    <Compile Remove="C:\REPOS\opentelemetry-dotnet-fork\src\OpenTelemetry\Internal\ServiceProviderExtensions.cs" />
```
reyang, line 31: nit: needs cleanup

src/OpenTelemetry.Exporter.InMemory/InMemoryExporterMetricsExtensions.cs
```diff
@@ -75,15 +93,69 @@ public static MeterProviderBuilder AddInMemoryExporter(this MeterProviderBuilder
...
+        /// <summary>
+        /// Adds InMemory metric exporter to the <see cref="MeterProviderBuilder"/> using default options.
+        /// </summary>
+        /// <param name="builder"><see cref="MeterProviderBuilder"/> builder to use.</param>
+        /// <param name="exportFunc">A function which will replace <see cref="InMemoryExporter{T}.Export(in Batch{T})"/>.</param>
+        /// <param name="configureMetricReader"><see cref="MetricReader"/> configuration options.</param>
+        /// <returns>The instance of <see cref="MeterProviderBuilder"/> to chain the calls.</returns>
+        public static MeterProviderBuilder AddInMemoryExporter(
+            this MeterProviderBuilder builder,
+            Func<Batch<Metric>, ExportResult> exportFunc,
```
reyang, line 105: Why Func rather than Action?

test/OpenTelemetry.Exporter.Prometheus.Tests/PrometheusSerializerTests.cs
```diff
@@ -34,7 +34,15 @@ public void GaugeZeroDimension()
             using var meter = new Meter(Utils.GetCurrentMethodName());
             using var provider = Sdk.CreateMeterProviderBuilder()
                 .AddMeter(meter.Name)
-                .AddInMemoryExporter(metrics)
+                .AddInMemoryExporter(exportFunc: batch =>
+                {
+                    foreach (var metric in batch)
```
reyang, line 39: Is there a way (e.g. list.AddRange) to reduce the boilerplate code (ideally keeping it a one liner)?

TimothyMothra: I had that same thought while i was writing this. i'll spend some time on this today

src/OpenTelemetry.Exporter.InMemory/InMemoryExporterMetricsExtensions.cs
TimothyMothra: I went back and forth on this detail a few times.

1. I think it's a reasonable ask, if a user is replacing the functionality of Export, making them responsible for the ExportResult as well.

2. Being able to control the return type offers more flexibility to unit test authors.
  I found one test that was dependent on `ExportResult.Failure`. 
    - **Before**: This test was taking advantage of a side effect of the `InMemoryExporter`, setting the collection to `null` causes the Export to return `Failure`.
https://github.com/open-telemetry/opentelemetry-dotnet/blob/a64128f896168e18b1d66fb8229130fd9ff8dbeb/test/OpenTelemetry.Tests/Trace/BatchExportActivityProcessorTest.cs#L190-L193
https://github.com/open-telemetry/opentelemetry-dotnet/blob/a64128f896168e18b1d66fb8229130fd9ff8dbeb/src/OpenTelemetry.Exporter.InMemory/InMemoryExporter.cs#L31-L36
    - **After**: Here the test can clearly define the behavior that it's validating. 
https://github.com/open-telemetry/opentelemetry-dotnet/blob/26d7e1fc30332a0aa17c32a287e156d0091bc050/test/OpenTelemetry.Tests/Trace/BatchExportActivityProcessorTest.cs#L193

reyang: Makes sense to me, thanks!

reyang (approved):
LGTM with some suggestions (clean up, reduce boilerplate, etc.).

TimothyMothra:
This PR got very large. 
I'm going to pull out some smaller changes into separate PRs.

test/OpenTelemetry.Exporter.Prometheus.Tests/PrometheusSerializerTests.cs
TimothyMothra: @reyang 
AddRange doesn't work here because there's not a built in way to convert Batch to IEnumerable.
If we need to convert Batch to something, I would prefer to convert it to a List.

My initial idea was to add a helper method to the Batch class itself.
Instead, my most recent change adds a helper method to only this test class. 
See: `private static List<Metric> BatchToList(Batch<Metric> batch)`

reyang: Looks better, still have room for simplification. Could you make it even simpler?

TimothyMothra: The simplest would be to add an additional Extension method that accepts `Action<T>`. I didn't think I could get away with adding both Func and Action to the PublicApi and between the two I chose Func.

reyang: One simple way would be having an extension method just in this test project - `AddInMemoryExporter(some list type)` which handles the boilerplate job.

TimothyMothra: I put a new extension method `AddInMemoryExporter(List<Metric>)` in `Tests.Shared` so it can be reused by multiple projects.
This helps reduce the number of changes in this PR, the individual Prometheus and OTLP unit tests don't need any changes.

src/OpenTelemetry/AssemblyInfo.cs
```diff
@@ -17,6 +17,7 @@
 using System.Runtime.CompilerServices;
 
 [assembly: InternalsVisibleTo("OpenTelemetry.Tests" + AssemblyInfo.PublicKey)]
+[assembly: InternalsVisibleTo("OpenTelemetry.Exporter.InMemory" + AssemblyInfo.PublicKey)]
```
reyang, line 20: Why is this needed?

test/OpenTelemetry.Tests/Trace/BatchExportActivityProcessorTest.cs
```diff
@@ -190,7 +190,8 @@ public void CheckExportForRecordingButNotSampledActivity()
         [Fact]
         public void CheckExportDrainsBatchOnFailure()
         {
-            using var exporter = new InMemoryExporter<Activity>(null);
+            using var exporter = new InMemoryExporter<Activity>(exportFunc: x => ExportResult.Failure);
```
reyang, line 193: Is it possible to just put `exportAction: null` (and avoid using Func)?

test/OpenTelemetry.Exporter.Prometheus.Tests/PrometheusSerializerTests.cs
reyang: I don't think we want to encourage more use of this copy approach given the fact that Metric can be changed after the export.

src/OpenTelemetry/AssemblyInfo.cs
TimothyMothra: When making a copy of `Metric`, `InstrumentIdentity` is internal.

https://github.com/open-telemetry/opentelemetry-dotnet/blob/d1ec45ce15cfe3062f828706bdd883f6ebcc5ded/src/OpenTelemetry/Metrics/Metric.cs#L127

`ExportableMetricCopy` needs access to this. It's referenced 80+ times in tests.
https://github.com/open-telemetry/opentelemetry-dotnet/blob/afb1d863c740aed3d326bf16df45ed093bb2736c/src/OpenTelemetry.Exporter.InMemory/ExportableMetricCopy.cs#L30-L42

test/OpenTelemetry.Tests/Trace/BatchExportActivityProcessorTest.cs
TimothyMothra: I wouldn't use a null here, because I think then this test would be depending on side-effects to test the Failure behavior.
I like this one as it is because it clearly describes the behavior being tested here.
If you think this needs to be cleaned up, I could add an Extension method that just accepts the `ExportResult` enum.

reyang: Or maybe we shouldn't use InMemoryExporter in this test case at all.

TimothyMothra: I thought of that. :)
I thought it was a plus that this exporter was flexible enough to handle cases like this. 
But I have no problem putting a private custom exporter here.

TimothyMothra:
@cijothomas @CodeBlanch, this is ready for review :)

src/OpenTelemetry.Exporter.InMemory/InMemoryExporter.cs
```diff
@@ -22,13 +23,29 @@ public class InMemoryExporter<T> : BaseExporter<T>
         where T : class
     {
         private readonly ICollection<T> exportedItems;
+        private readonly ExportDelegate onExport;
 
         public InMemoryExporter(ICollection<T> exportedItems)
         {
+            if (typeof(T) == typeof(Metrics.Metric))
+            {
+                throw new NotSupportedException("Exported Metrics are not trustworthy because they can continue to be updated after export. Recommend use ExportableMetricCopy.");
```
cijothomas, line 32: I am not sure I follow the reason behind this throw...

test/Benchmarks/Metrics/HistogramBenchmarks.cs
```diff
@@ -83,7 +83,7 @@ public void Setup()
                 this.bounds[i] = i * MaxValue / this.bounds.Length;
             }
 
-            var exportedItems = new List<Metric>();
+            var exportedItems = new List<ExportableMetricCopy>();
```
cijothomas, line 86: is it required to force every test to use the new MetricCopy unless they need to rely on it? 
Or is it by design that InMemoryExporter will force one to use the MetricCopy option only? What I don't need/care about the fact that Metric could be update after export() call is done?

src/OpenTelemetry.Exporter.InMemory/InMemoryExporter.cs
TimothyMothra: It is by design to force the use of the Copy. 
The thought behind this was to prevent users from creating the `InMemoryExporter<Metric>` because the `Metric` is known to be faulty. https://github.com/open-telemetry/opentelemetry-dotnet/pull/3198#discussion_r851547483

I've documented some scenarios where the `Metric` can and cannot be trusted: https://github.com/open-telemetry/opentelemetry-dotnet/issues/2361#issuecomment-1049175566

That being said, Prometheus and OTLP have some units tests where they need the actual `Metric` to provide to their own internals. (documented in the Summary above).

I'm eager to hear your perspective on this!
I'm trying to find the happy medium to support all these use cases while protecting users.
If we decide to allow users to create `InMemoryExporter<Metric>` we also need to inform/educate users on the risks.

test/Benchmarks/Metrics/HistogramBenchmarks.cs
TimothyMothra: Yes, current approach was to force the use of the Copy.
See the comment above for more details https://github.com/open-telemetry/opentelemetry-dotnet/pull/3198#discussion_r867199870

src/OpenTelemetry.Exporter.InMemory/InMemoryExporter.cs
cijothomas: We don't need to prevent users from using the <Metric> one. A lot of tests are already using it, and working fine. (they understand the documented limitation of Metric that it may be updated after Export() call, but they are orchestrating things so as to not be affected)

This can be additional option, for InMemory users, who wants to test metrics, without being affected by the limitation that the memory could be updated.

cijothomas: To unblock progress - lets split PR into sub Prs.
1. Add the additional option in InMemoryExporter to use MetricCopy, leave existing functionality untouched. i.e don't change unit tests, except the one breaking due to the metric-being-updated-after-export issue.
2. We can use the SIG meeting to see if there are strong reasons favoring/against keeping the current <Metric> exporter, and based on that keep it or remove it.

PR #3211, opened 2022-04-20
ConsoleExporter Log minor cosmetic improvement.

cijothomas:
Better padding.
Print TraceContext only if non-default, same with Event.
Group TraceContext fields (traceid,spanid,traceflags) together.

### DIFF
--- a/docs/logs/getting-started/Program.cs
+++ b/docs/logs/getting-started/Program.cs
@@ -14,7 +14,6 @@
 // limitations under the License.
 // </copyright>
 
-using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using OpenTelemetry.Logs;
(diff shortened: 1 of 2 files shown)


TimothyMothra:
could we also consider improving the padding for the Scope Key-Value pairs?

When printing the Resource attributes, we indent by 4 spaces (Line 90).
I think it would make it easier to read if we did the same at for Scope KVP (Line 80).

cijothomas:
> could we also consider improving the padding for the Scope Key-Value pairs?
> 
> When printing the Resource attributes, we indent by 4 spaces (Line 90). I think it would make it easier to read if we did the same at for Scope KVP (Line 80).

Fixed Resource.
Scopes can be a diff. PR.

reyang (approved):

PR #3212, opened 2022-04-21
ConsoleLogExporter extensions to enable ParseStateValues

cijothomas:
Similar change as https://github.com/open-telemetry/opentelemetry-dotnet/pull/3186 to ConsoleExporter


src/OpenTelemetry.Exporter.Console/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+`AddConsoleExporter` extension method on `OpenTelemetryLoggerOptions`
```
reyang, line 5: I have reservations on this. I guess Console Exporter is likely to be used by Exporter developers to understand the behavior, hence it should be neutral about the loggerOptions (e.g. the dev who owns the provider/logger configuration should have the saying).

I guess it might be surprising if the developer set ParseStateValues to false, and then realized that AddConsoleExporter is setting to true.

cijothomas: Does it look any better with the additional remarks on `AddConsoleExporter` extension method?

The default output from getting-started guide would look better, and that would be a good justification for this change. Exporter developers/others are informed about this behavior with the remark in the `AddConsoleExporter` method.

Default getting started output:
With this PR
image

Without this PR
image

I'll be tweaking the ConsoleExporter separately (to better display "Body", etc.),. We can wait for that to complete, to see if it makes this PR more stronger...

reyang: Maybe I'm missing some context.

It seems what we are trying to achieve here for Console Exporter is something that can be done in the console output logic, doesn't have to touch the ParseStateValues options at all?

cijothomas: Yes ConsoleExporter *can* take the `State` and parse it by itself. Same with any other exporter.
We are trying to make SDK deal with it, and not exporter. (https://github.com/open-telemetry/opentelemetry-dotnet/pull/3183#issuecomment-1099728854)

PR #3213, opened 2022-04-21
Console LogExporter to display Body

cijothomas:
Specially treats {OriginalFormat}, to hint that it is also the "body".
If `ParseStateValue` is enabled, the getting started doc would show the below output with this PR:
image


src/OpenTelemetry.Exporter.Console/ConsoleLogRecordExporter.cs
```diff
@@ -58,10 +60,25 @@ public override ExportResult Export(in Batch<LogRecord> batch)
...
+                        // Special casing {OriginalFormat}
+                        // See https://github.com/open-telemetry/opentelemetry-dotnet/pull/3182
+                        // for explanation.
+                        if (logRecord.StateValues[i].Key.Equals("{OriginalFormat}") && string.IsNullOrEmpty(body))
+                        {
+                            body = logRecord.StateValues[i].Value as string;
+                        }
+                        else
+                        {
+                            this.WriteLine($"{logRecord.StateValues[i].Key,-RightPaddingLength}{logRecord.StateValues[i].Value}");
```
reyang, line 72: Given this "belongs" to the `StateValues`, consider adding indentation.

src/OpenTelemetry.Exporter.Console/ConsoleLogRecordExporter.cs
```diff
@@ -58,10 +60,25 @@ public override ExportResult Export(in Batch<LogRecord> batch)
...
+                        }
                     }
                 }
 
+                if (!string.IsNullOrEmpty(body))
+                {
+                    this.WriteLine($"{"Body:",-RightPaddingLength}{body}");
```
reyang, line 79: This might be confusing since `Body` is not a first-class citizen in `LogRecord`.

reyang: Maybe consider something like:

```
"{OriginalFormat}" (a.k.a. Body): "Hello from {name} {price}."
```

cijothomas: Upon reading again, yes, I agree.

reyang (approved):

Issue #2877, opened 2022-02-08
Make some fields in `LogRecord` mutable for consistency with `Activity`.


Yun-Ting:
# Feature Request
Currently, in `Activity`, there are methods that allowed users to modify the fields. For example: `public Activity SetTag(string key, object? value)`.

I'm proposing to introduce methods that allow some fields in `LogRecord` to be mutable - for better consistency with `Activity`. Also, with this, use cases like redaction or fault-injection in Processor or Exporter could be achieved.

**Describe the solution you'd like:**

Introduce setter methods to `FormattedMessage`, `State`, and `StateValues`.
```
OpenTelemetry.Logs.LogRecord.SetFormattedMessage(string formattedMessage) -> void
OpenTelemetry.Logs.LogRecord.SetState(object state) -> void
OpenTelemetry.Logs.LogRecord.SetStateValues(System.Collections.Generic.IReadOnlyList<System.Collections.Generic.KeyValuePair<string, object>> stateValues) -> void
```

**Describe alternatives you've considered.**

Making the constructor public to provide users the ability to copy/update fields of their liking. 
I think the public setters solution above has better encapsulation based on the OOP principles.
And for use cases like redaction or fault-injection, initialization is not mandatory for modifying a `LogRecord`.

dpk83:
+1 to this ask. We have a similar requirement where we are able to perform redaction for traces because Activity supports mutability but same isn't possible with logs.

noblejacob-kubra:
**+1** to this ask. 
Similar to this we have a requirement to do custom scrubbing in logs to mask or remove sensitive information.
But was not able to implement because of having no provision to update the State/ StateValues/ FormattedMessage.
Can we add this feature so that it will be consistent with Activity.

PR #3217, opened 2022-04-21
Exposed public setters for LogRecord.State, LogRecord.StateValues,   and LogRecord.FormattedMessage.

Yun-Ting:
Fixes https://github.com/open-telemetry/opentelemetry-dotnet/issues/2877.


docs/logs/extending-the-sdk/MyClassWithRedactionEnumerator.cs
```diff
@@ -0,0 +1,72 @@
...
+
+using System.Collections;
+using System.Collections.Generic;
+using System.Text;
+using System.Text.RegularExpressions;
+
+internal class MyClassWithRedactionEnumerator : IReadOnlyList<KeyValuePair<string, object>>
+{
+    private static readonly Regex Rgx = new("\\(?\\d{3}[\\.|\\)]?\\d{3}[\\.|\\-]\\d{4}");
+    private readonly IReadOnlyList<KeyValuePair<string, object>> myList;
```
reyang, line 25:
```suggestion
    private readonly IReadOnlyList<KeyValuePair<string, object>> traits;
```

docs/logs/extending-the-sdk/MyClassWithRedactionEnumerator.cs
```diff
@@ -0,0 +1,72 @@
...
+// </copyright>
+
+using System.Collections;
+using System.Collections.Generic;
+using System.Text;
+using System.Text.RegularExpressions;
+
+internal class MyClassWithRedactionEnumerator : IReadOnlyList<KeyValuePair<string, object>>
+{
+    private static readonly Regex Rgx = new("\\(?\\d{3}[\\.|\\)]?\\d{3}[\\.|\\-]\\d{4}");
```
reyang, line 24:
```suggestion
    private static readonly Regex Rule = new("\\(?\\d{3}[\\.|\\)]?\\d{3}[\\.|\\-]\\d{4}").Compile();
```

docs/logs/extending-the-sdk/MyClassWithRedactionEnumerator.cs
```diff
@@ -0,0 +1,72 @@
...
+    private readonly IReadOnlyList<KeyValuePair<string, object>> myList;
+
+    public MyClassWithRedactionEnumerator(IReadOnlyList<KeyValuePair<string, object>> traits)
```
cijothomas, line 27:
```suggestion
    public MyClassWithRedactionEnumerator(IReadOnlyList<KeyValuePair<string, object>> state)
```

docs/logs/extending-the-sdk/MyClassWithRedactionEnumerator.cs
```diff
@@ -0,0 +1,72 @@
...
+            }
+        }
+    }
+
+    IEnumerator IEnumerable.GetEnumerator()
+    {
+        return this.GetEnumerator();
+    }
+
+    public override string ToString()
```
cijothomas, line 57: do we need this?

reyang, line 57: I feel that this part is not needed and might actually be misleading.

I guess @cijothomas' change would actually help to eliminate the need https://github.com/open-telemetry/opentelemetry-dotnet/pull/3213.

docs/logs/extending-the-sdk/Program.cs
```diff
@@ -31,6 +31,7 @@ public static void Main()
                 options.AddProcessor(new MyProcessor("ProcessorA"))
                        .AddProcessor(new MyProcessor("ProcessorB"))
                        .AddProcessor(new SimpleLogRecordExportProcessor(new MyExporter("ExporterX")))
+                       .AddProcessor(new MyRedactionProcessor())
```
reyang, line 34: Consider putting this before all the exporters?

docs/logs/extending-the-sdk/MyClassWithRedactionEnumerator.cs
cijothomas, line 24: for this tutorial, i dont think we need to show actual regex! its upto end user to make their own.. we can do simple match like value.Contains("secret"), as it is sufficient to demonstrate the approach.

reyang: +1

reyang: Or it the class is renamed to "PhoneNumberRedaction(something)" it sounds reasonable.

docs/logs/extending-the-sdk/MyClassWithRedactionEnumerator.cs
Yun-Ting: We'll need to do serialization.
Alternatives I have in mind:
1. Update MyExporter, which is currently the Exporter used in extending-the-sdk to contain the logic as is in https://github.com/open-telemetry/opentelemetry-dotnet/pull/3213/files, which would add some unnecessary complexity to MyExporter.
2. Instead of using MyExporter as the Exporter for extending-the-sdk, use `ConsoleLogRecordExporter.cs` which already takes care of serialization. But with this, we probably don't need MyExporter anymore.

Yun-Ting: https://github.com/open-telemetry/opentelemetry-dotnet/pull/3217#discussion_r855522959

Yun-Ting: We'll need this for the compiler:

```
IEnumerator IEnumerable.GetEnumerator()
{
    return this.GetEnumerator();
}
```
Screenshot (4)

Yun-Ting: https://github.com/open-telemetry/opentelemetry-dotnet/pull/3217#discussion_r855526612

cijothomas: > . But with this, we probably don't need MyExporter anymore.

Its there to demonstrate how to write own exporter part.

cijothomas: The comment is about `ToString()` override. Enumeration is needed.

cijothomas: > We'll need to do serialization.

No need. This doc (the parts from this PR) is just how to write a custom processor, which can do redaction of properties.

Yun-Ting: Interesting - My browser is showing me code highlight as this:
image

docs/logs/extending-the-sdk/MyClassWithRedactionEnumerator.cs
Yun-Ting: I've updated name of the private field to be state.
Not sure which one is better, trait or state; but both of them definitely are better than myList :nerd_face:.

docs/logs/extending-the-sdk/MyClassWithRedactionEnumerator.cs
Yun-Ting: I've removed the regex entirely as suggested in https://github.com/open-telemetry/opentelemetry-dotnet/pull/3217#discussion_r855507969.

reyang (approved):

docs/logs/extending-the-sdk/MyClassWithRedactionEnumerator.cs
Yun-Ting: I've removed the ToString() override of [MyClassWithRedactionEnumerator.cs](https://github.com/open-telemetry/opentelemetry-dotnet/pull/3217/files/9b54a621507e30eabb67e05b298a2bf05ec744bb#diff-2cd84068fc8e2ab3d7a411c5a7eb10c71ab09b1f0c9d4aa224c9083738029af1). Thanks!

docs/logs/extending-the-sdk/MyRedactionProcessor.cs
```diff
@@ -0,0 +1,30 @@
...
+
+using System.Collections.Generic;
+using OpenTelemetry;
+using OpenTelemetry.Logs;
+
+internal class MyRedactionProcessor : BaseProcessor<LogRecord>
+{
+    public override void OnEnd(LogRecord logRecord)
+    {
+        if (logRecord.State is IReadOnlyList<KeyValuePair<string, object>> listOfKvp)
```
cijothomas, line 25: just for completion, we could show how to use same approach for logRecord.StateValues as well. (not a blocker for this PR)

docs/logs/extending-the-sdk/Program.cs
```diff
@@ -64,6 +65,9 @@ public static void Main()
         {
             logger.LogError("{name} is broken.", "refrigerator");
         }
+
+        // message will be redacted by MyRedactionProcessor
+        logger.LogInformation("OpenTelemetry {sensitiveString}.", "<secret>");
```
cijothomas, line 70: also it'd be good to modify the readme.md for the extending-the-sdk to include the redaction part. Similar to this https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/docs/trace/extending-the-sdk#filtering-processor

cijothomas (approved):

cijothomas:
Merging. Couple of non-blocking doc improvements suggested.

PR #3219, opened 2022-04-21
add internal method ToCollection() to Batch

TimothyMothra:
related to #2361.
pulling a minor change out of #3198.

## Changes
- Add internal method `ToCollection` to `Batch`

## Please provide a brief description of the changes here.

This change is to eliminate the `foreach` boilerplate code: 
  ```
  foreach (var data in batch)
  {
      someCollection.Add(data);
  }
  ```

## For significant contributions please make sure you have completed the following items:

## Other options considered

1. do nothing
    this will cause unit tests to have a lot of `foreach` statements.

2. `ToCollection(ICollection<T> someCollection)`
    I considered this but decided to imitate [Enumerable.ToList](https://docs.microsoft.com/dotnet/api/system.linq.enumerable.tolist?view=net-6.0) which returns the collection.

3. `ToList()`
    I don't have a strong opinion about the return type. I chose to return `ICollection` because I see this used more often in this repo.

4. add extension method/class to OpenTelemetry.Tests.Shared
    this would work, but is an extra file that would need to be imported into test projects.
    having this method in the Batch class allows it to access private fields.

### DIFF
--- a/src/OpenTelemetry/Batch.cs
+++ b/src/OpenTelemetry/Batch.cs
@@ -110,16 +110,7 @@ namespace OpenTelemetry
         /// Creates a <see cref="ICollection{T}"/> containing the times of the <see cref="Batch{T}"/>.
         /// </summary>
         /// <returns>A collection containing the items of the <see cref="Batch{T}"/>.</returns>
-        public ICollection<T> ToCollection()
-        {
-            var collection = new List<T>();
-            foreach (var item in this.items)
-            {
-                collection.Add(item);
-            }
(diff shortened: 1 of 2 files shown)


reyang (changes requested):
Doesn't seem to have a good reason to expose new public API from `Batch`.

TimothyMothra:
> Doesn't seem to have a good reason to expose new public API from `Batch`.

changed to internal.
Verified that the tests that need this already have access to internals.

Issue #1276, opened 2020-09-16
Record gRPC message events


JamesNK:
The OpenTelemetry spec for gRPC has been updated with details about the events captured as messages are messages are sent. See https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/rpc.md#events

First step is to figure out how to do this. Currently no diagnostic source method is called or event added to the current activity in [grpc-dotnet](https://github.com/grpc/grpc-dotnet) when a message is sent or received. I suspect support for events will need changes in grpc-dotnet.

Does support for events in opentelemetry-dotnet come from `Activity.Events`? Is it just a matter of adding `Activity.AddEvent` in grpc-dotnet with the appropriate detail?

cijothomas:
> Does support for events in opentelemetry-dotnet come from Activity.Events? Is it just a matter of adding Activity.AddEvent in grpc-dotnet with the appropriate detail? 

Yes. (grpc-dotnet must update to DiagnosticSource version 5.* to leverage Activity.AddEvent though)

JamesNK:
Is there a concern about long running streams that send and receive many messages, and `Activity.Events` acting as a memory leak?

See https://github.com/dotnet/runtime/issues/42347

cijothomas:
closing as no action required in this repo.

PR #3231, opened 2022-04-26
Span attribute count and length limits

alanwest:
Implements [attribute limits](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/README.md#attribute-limits) for span data when using the shim API.

Follow on PR will implement additional [span limits](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#span-limits) for events and links.

I've opened https://github.com/dotnet/runtime/issues/68528 to propose a similar API for `ActivitySource`.

### DIFF
--- a/src/OpenTelemetry.Api/Trace/SpanAttributes.cs
+++ b/src/OpenTelemetry.Api/Trace/SpanAttributes.cs
@@ -43,9 +43,13 @@ namespace OpenTelemetry.Trace
         {
             Guard.ThrowIfNull(attributes);
 
+            var attributeCountLimit = TracerProvider.SpanAttributeCountLimit ?? -1;
             foreach (KeyValuePair<string, object> kvp in attributes)
             {
-                this.AddInternal(kvp.Key, kvp.Value);
+                if (attributeCountLimit-- != 0)
+                {
+                    this.AddInternal(kvp.Key, kvp.Value);
+                }
             }
         }
 
@@ -134,6 +138,26 @@ namespace OpenTelemetry.Trace
         private void AddInternal(string key, object value)
         {
             Guard.ThrowIfNull(key);
+
+            if (TracerProvider.SpanAttributeValueLengthLimit.HasValue)
+            {
+                if (value is string[] strArr)
+                {
+                    var newArr = new string[strArr.Length];
+                    strArr.CopyTo(newArr, 0);
+                    for (var i = 0; i < strArr.Length; ++i)
+                    {
+                        newArr[i] = newArr[i].Substring(0, TracerProvider.SpanAttributeValueLengthLimit.Value);
+                    }
+
+                    value = newArr;
+                }
+
+                if (value is string str)
+                {
+                    value = str.Substring(0, TracerProvider.SpanAttributeValueLengthLimit.Value);
+                }
+            }
 
             this.Attributes[key] = value;
         }
--- a/src/OpenTelemetry.Api/Trace/SpanAttributes.cs
+++ b/src/OpenTelemetry.Api/Trace/SpanAttributes.cs
@@ -139,6 +139,11 @@ namespace OpenTelemetry.Trace
         {
             Guard.ThrowIfNull(key);
 
+            if (this.Attributes.Count == TracerProvider.SpanAttributeCountLimit)
+            {
+                return;
+            }
+
             if (TracerProvider.SpanAttributeValueLengthLimit.HasValue)
             {
                 if (value is string[] strArr)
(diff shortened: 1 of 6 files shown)


src/OpenTelemetry.Api/Trace/TracerProvider.cs
```diff
@@ -35,6 +35,20 @@ protected TracerProvider()
...
+
+        internal static int? SpanAttributeValueLengthLimit { get; set; }
+
+        /// <summary>
+        /// Configures the <see cref="SpanLimits" /> for the default <see cref="TracerProvider" />.
+        /// </summary>
+        /// <param name="spanLimits">The configured <see cref="SpanLimits" />.</param>
+        public static void SetSpanLimits(SpanLimits spanLimits)
+        {
+            SpanAttributeCountLimit = spanLimits.AttributeCountLimit;
```
reyang, line 48: How to handle error when the value is out of range?

src/OpenTelemetry.Api/Trace/SpanAttributes.cs
```diff
@@ -43,9 +43,13 @@ public SpanAttributes(IEnumerable<KeyValuePair<string, object>> attributes)
...
             Guard.ThrowIfNull(attributes);
 
+            var attributeCountLimit = TracerProvider.SpanAttributeCountLimit ?? -1;
             foreach (KeyValuePair<string, object> kvp in attributes)
             {
-                this.AddInternal(kvp.Key, kvp.Value);
+                if (attributeCountLimit-- != 0)
+                {
+                    this.AddInternal(kvp.Key, kvp.Value);
+                }
```
reyang, line 52: Need to exit the loop once the limit is reached (or it will underflow to negative)

alanwest: If negative then it will loop over all attributes and then stop. `attributeCountLimit` is initialized to -1 for this purpose when the user set `TracerProvider.SpanAttributeCountLimit = null`.

Though, for efficiency, this loop could break when the limit has been reached. Also will add tests to demonstrate when limits are not set.

src/OpenTelemetry.Api/Trace/TracerProvider.cs
alanwest: Good question. My thought was that if set to negative then it it won't hurt anything - i.e., the setting would have no effect and nothing would be limited. At least that was my intent... though just realized that setting `SpanAttributeValueLengthLimit` to a negative value will cause an exception upon calling `String.Substring`. Will fix.

I'm open to other opinions, but since this is part of the API and not the SDK, I figured an error here was not appropriate.

src/OpenTelemetry.Api/Trace/SpanLimits.cs
```diff
@@ -0,0 +1,44 @@
...
+
+/// <summary>
+/// Defines the attribute count and length limits for spans.
+/// </summary>
+public class SpanLimits
+{
+    /// <summary>
+    /// Gets or sets the maximum allowed number of attributes on spans.
+    /// </summary>
+    public int? AttributeCountLimit { get; set; } = 128;
```
alanwest, line 27: 128 is the default identified by the spec... but this is a breaking change. So should we default to null - i.e., not set?

reyang (approved):
The actual code logic looks good to me.

I have a design question regarding the goal of these limits. By reading the specification I kind of feel that these limits are intended to provide protection against unbounded memory usage - e.g. if a user keeps adding new attributes, at certain moment the API would become a simple no-op that doesn't has memory cost.

Given .NET is not putting such limits on the Activity APIs, I feel the limits are becoming less useful. Perhaps such logic can be applied to the exporter logic?

alanwest:
> By reading the specification I kind of feel that these limits are intended to provide protection against unbounded memory usage - e.g. if a user keeps adding new attributes, at certain moment the API would become a simple no-op that doesn't has memory cost.

Yes, I agree this is one of the primary goals. Another place this concern was raised was in the context of unbounded events being added to a span here https://github.com/open-telemetry/opentelemetry-dotnet/issues/1276#issuecomment-693690086.

Another thing this feature offers is support to backends that have their own limitations on the number of attributes, attribute length, etc. that spans can have.

> Given .NET is not putting such limits on the Activity APIs

My hope was that .NET would consider implementing configurable limits: https://github.com/dotnet/runtime/issues/68528

> Perhaps such logic can be applied to the exporter logic?

I think I'm liking this idea. I'm going to make this PR a draft for now and try this out.

alanwest:
Closing this will follow up with the plan to handle these limits in the exporters.

PR #3232, opened 2022-04-26
Expose filter for SQL Instrumentation

akorade:
Fixes #.

## Changes
1. Added new Filter option to SQL instrumentation for both .NET core and .NET framework.
2. Updated Unit test cases.
3. Updated Readme file

Creating a new pull request as the old one got stale. https://github.com/open-telemetry/opentelemetry-dotnet/pull/3131


cijothomas:
@akorade sorry to let the old PR go stale - if you want to ressurect a stale PR, simply comment on it, or message in slack, and one of the maintainers can "unstale" it
.

akorade:
> @akorade sorry to let the old PR go stale - if you want to ressurect a stale PR, simply comment on it, or message in slack, and one of the maintainers can "unstale" it .

No worries , I did not see an option to revive that PR so ended up creating a new one.

cijothomas:
@akorade Sorry for getting to this late (and also, I merged a minor test fix, which cause conflict :( )

While the overall idea is good, I'd suggest to:

Add Filter capability only for the .NET Core route. The `Enrich` option is also only available for .NET Core only. The reason is that, for .NET Framework, there is no need of providing Enrich/Filter by the SqlInstrumentation. The general purpose `ActivityProcessor` can be used for the same in .NET Framework. In case of .NET Core, Enrich/Filter by this instrumentation library is useful as it can give access to additional context (i.e the SQLCommand instance itself.)
(This is mentioned in the readme.md file as well)

Could you reduce the scope of this to just .NET Core?

src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientDiagnosticListener.cs
```diff
@@ -56,6 +56,13 @@ public override void OnCustom(string name, Activity activity, object payload)
                 case SqlDataBeforeExecuteCommand:
                 case SqlMicrosoftBeforeExecuteCommand:
                     {
+                        _ = this.commandFetcher.TryFetch(payload, out var command);
```
cijothomas, line 59: I'd suggest to do this and Filter *after* the StartActivity() call. In typical case where user does not have filter, we want to first call StartActivity(), which will run samplers, and decide if this activity should even be collected. If not, we don't need to pay the cost of "fetching" command from payload. This PR alters the flow by always having to pay the cost of "fetching" command.

cijothomas (changes requested):
Overall Good, blocking to address 2 comments:
1. Scope down to .NET Core only
2. Fetch command from payload after samplers are run, i.e after startactivity

akorade:
> Overall Good, blocking to address 2 comments:
> 
> 1. Scope down to .NET Core only
> 2. Fetch command from payload after samplers are run, i.e after startactivity

-> We want the filter to work for .NET Framework , is there any particular reason why we want it for .NET Core only?
-> If we do run Filter after starting activity it already will be sampled and exported. 
     To address this what if we call the Fetch method only if the filter is not null and if filter is null invoke Fetch after the activity is started?

cijothomas:
> -> We want the filter to work for .NET Framework , is there any particular reason why we want it for .NET Core only?

^ See this comment : https://github.com/open-telemetry/opentelemetry-dotnet/pull/3232#issuecomment-1119784595
Filter doesn't add any new value other than already provided by general OpenTelemetry extensibility points.

cijothomas:
> -> If we do run Filter after starting activity it already will be sampled and exported.
> To address this what if we call the Fetch method only if the filter is not null and if filter is null invoke Fetch after the activity is started?

We can stick with
1. Start Activity.
2. If activity is not null and isAllDataRequested=true, then only do anything else, including running the filter.

Issue #3229, opened 2022-04-22
`process.command_args` serializes as `System.String[]`


tillig:
# Bug Report

List of [all OpenTelemetry NuGet
packages](https://www.nuget.org/profiles/OpenTelemetry) and version that you are
using (e.g. `OpenTelemetry 1.0.2`):

- OpenTelemetry.Exporter.Console: 1.3.0-beta.1
- OpenTelemetry.Exporter.Jaeger: 1.3.0-beta.1
- OpenTelemetry.Exporter.OpenTelemetryProtocol: 1.3.0-beta.1
- OpenTelemetry.Exporter.Prometheus: 1.3.0-beta.1
- OpenTelemetry.Extensions.Hosting: 1.0.0-rc9.3
- OpenTelemetry.Instrumentation.AspNetCore: 1.0.0-rc9.3
- OpenTelemetry.Instrumentation.Http: 1.0.0-rc9.3

Runtime version (e.g. `net462`, `net48`, `netcoreapp3.1`, `net6.0` etc. You can
find this information from the `*.csproj` file):

* net6.0

## Symptom

I'm adding the `process.command_args` metadata to a trace resource and when it is serialized to a trace span it appears as `System.String[]` instead of the actual array of values. I see this in both the Console exporter as well as using OpenTelemetry Collector.

**What is the expected behavior?**

I expect the array of arguments to be kept as an array of strings [as defined in the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/process.md#process).

**What is the actual behavior?**

The value is rendered like `.ToString()` was called on the array, so it's `System.String[]` instead of the values.

## Reproduce

Create a resource detector like this:

```c#
using System.Runtime.InteropServices;
using OpenTelemetry.Resources;

namespace MyNamespace;

public class RuntimeResourceDetector : IResourceDetector
{
    public Resource Detect()
    {
        var process = Process.GetCurrentProcess();
        var filename = process.MainModule.FileName;

        // Command line args requires the executable per OpenTelemetry spec.
        var arguments = Environment.GetCommandLineArgs().Prepend(filename).ToArray();
        var resourceData = new Dictionary<string, object>()
        {
          { OpenTelemetrySemanticResourceKey.ProcessCommandArgs, arguments }
        };

        return new Resource(resourceData);
    }
}
```

Add that to your resource when you wire up trace.

```c#
var resourceBuilder = ResourceBuilder
[...]

Issue #2010, opened 2021-04-24
Consistent transform of attribute values


alanwest:
There has been some recent discussion about how to handle the differences in
what the .NET APIs allows as attribute values and what is declared in the OpenTelemetry specification.
Opening this issue to continue the discussion. (#1966, #1973)

The specification states that an attribute values is either:
* A primitive type: string, boolean, double precision floating point (IEEE 754-1985) or signed 64 bit integer.
* An array of primitive type values. The array MUST be homogeneous, i.e. it MUST NOT contain values of different types. For protocols that do not natively support array values such values SHOULD be represented as JSON strings.

The API for setting a tag on a .NET Activity is broader in scope than what the OpenTelemetry specification requires.
A tag's value can be any type.

```csharp
public System.Diagnostics.Activity SetTag (string key, object? value);
```

Currently, the exporters in this project convert tag values to a string, boolean, double, or int64 - or an array of these - as appropriate.

I propose we continue to embrace the flexibility of the .NET Activity API and ensure that we are consistently converting all C# [built-in types](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/built-in-types) to the most appropriate type declared in the OpenTelemetry specification as follows:

* All numeric types which would not be truncated should be converted to a double or int64 value.
    * Examples: short, int, long, float, double
* All numeric types which may be truncated if converted to a double or int64 should be converted to a string.
    * Examples: decimal, ulong
* A bool is a bool 
* All other types would be converted to a string.

For the most part, I think this is pretty close to what the exporters already do, but there are a few open questions:

1) There exist numeric built-in types that are currently converted to a string
    * byte, sbyte, ushort, uint
2) Arrays of built-in types should work
    * Currently a `byte[]`, `char[]`, or `decimal[]`, for example, are converted to the string `System.Byte[]` and `System.Char[]`, and `System.Decimal[]`.
3) ToString can be perilous. Should we be concerned?
    * ToString on an arbitrary type could result in an exception (or worse a deadlock... I've actually seen this). Currently, I think our exporters would throw away an entire batch if an exception were thrown when transforming an attribute value. Should we be more defensive here and discard attributes that cannot be transformed?

alanwest:
[...]

PR #3238, opened 2022-04-28
Fix handling of OTLP ArrayValue attributes

alanwest:
The proto has an [`ArrayValue`](https://github.com/open-telemetry/opentelemetry-proto/blob/17c68a9d2b67d702153d91fa985ecf2b87f74b7c/opentelemetry/proto/common/v1/common.proto#L41-L46). We were not using it. This PR fixes things for `Activity`s.

There's a slew of other bugs in and around handling OTLP attributes. Will continue to follow up with some more PRs.

### DIFF
--- a/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ActivityExtensions.cs
+++ b/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ActivityExtensions.cs
@@ -396,6 +396,8 @@ namespace OpenTelemetry.Exporter.OpenTelemetryProtocol.Implementation
                     this.Created = true;
                 }
 
+                OtlpCommon.ArrayValue arrayValue;
+
                 switch (activityTag.Value)
                 {
                     case string s:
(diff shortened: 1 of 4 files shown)


test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/OtlpTraceExporterTests.cs
```diff
@@ -333,13 +333,12 @@ public void ToOtlpSpanActivitiesWithNullArrayTest()
...
+            var stringArray = otlpSpan.Attributes.FirstOrDefault(kvp => kvp.Key == "stringArray");
 
             Assert.NotNull(stringArray);
-            Assert.Equal(3, stringArray.Count);
-            Assert.Equal("test", stringArray[0].Value.StringValue);
-            Assert.Equal(string.Empty, stringArray[1].Value.StringValue);
-            Assert.Null(stringArray[2].Value);
+            Assert.Equal("test", stringArray.Value.ArrayValue.Values[0].StringValue);
+            Assert.Equal(string.Empty, stringArray.Value.ArrayValue.Values[1].StringValue);
+            Assert.Equal(OtlpCommon.AnyValue.ValueOneofCase.None, stringArray.Value.ArrayValue.Values[2].ValueCase);
```
alanwest, line 341: This may not be right. It seems closest to what we were doing before, but technically an array should be homogenous.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ActivityExtensions.cs
```diff
@@ -416,39 +418,49 @@ public bool ForEach(KeyValuePair<string, object> activityTag)
...
                         foreach (var item in longArray)
                         {
-                            PooledList<OtlpCommon.KeyValue>.Add(ref this.Tags, CreateOtlpKeyValue(key, new OtlpCommon.AnyValue { IntValue = item }));
+                            arrayValue.Values.Add(new OtlpCommon.AnyValue { IntValue = item });
                         }
 
+                        PooledList<OtlpCommon.KeyValue>.Add(ref this.Tags, CreateOtlpKeyValue(key, new OtlpCommon.AnyValue { ArrayValue = arrayValue }));
                         break;
                     default:
                         PooledList<OtlpCommon.KeyValue>.Add(ref this.Tags, CreateOtlpKeyValue(key, new OtlpCommon.AnyValue { StringValue = activityTag.Value.ToString() }));
```
alanwest, line 466: This is a little suspect. I think we should consider dropping things that do not fit the types we want to support.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ActivityExtensions.cs
```diff
@@ -416,39 +418,49 @@ public bool ForEach(KeyValuePair<string, object> activityTag)
...
+                        arrayValue = new OtlpCommon.ArrayValue();
                         foreach (var item in intArray)
                         {
-                            PooledList<OtlpCommon.KeyValue>.Add(ref this.Tags, CreateOtlpKeyValue(key, new OtlpCommon.AnyValue { IntValue = item }));
+                            arrayValue.Values.Add(new OtlpCommon.AnyValue { IntValue = item });
                         }
 
+                        PooledList<OtlpCommon.KeyValue>.Add(ref this.Tags, CreateOtlpKeyValue(key, new OtlpCommon.AnyValue { ArrayValue = arrayValue }));
                         break;
                     case double[] doubleArray:
```
alanwest, line 429: There's a number of types we're not consistently handling across everything. For example, we're not handling `float` or `float[]` in this code. Should we? Will continue discussion here https://github.com/open-telemetry/opentelemetry-dotnet/issues/2010

cijothomas:
https://github.com/open-telemetry/opentelemetry-dotnet/issues/3229 is this related as well?

alanwest:
> #3229 is this related as well?

Yes, definitely. I want to address this as well. Problem is we have a number of disparate code paths that transform attributes to OTLP, and they're not equivalent. I imagine we may have similar issues in some of the other exporters as well (Jaeger/Zipkin).

The code path for resource attributes does not yet handle arrays. The same code path is used by the OTLP logs exporter. I suspect I can refactor things to be a little less disjointed, but I want to do things in a few passes.

I think I'm going to take on the work of #2010 - first in the context of OTLP trace data. That way we can come to a consensus of what types of attributes we want to support and then I can work on making sure everything works uniformly.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ActivityExtensions.cs
cijothomas: Agree. Log and drop would be safer here.

cijothomas (approved):
LGTM. User facing bug fix, so changelog.md also please!

PR #3262, opened 2022-05-09
OTLP exporter: Standardize handling of attributes

### LINKED ISSUE #3229: `process.command_args` serializes as `System.String[]`

alanwest:
Related to #2010.

This PR is specifically in the context of the OTLP exporter but my purpose is to drive the more general conversation for what datatypes all SDK components should handle.

I've taken a pretty broad stance to begin with. This PR ensures that the OTLP exporter handles strings and nearly all [.NET built-in value types](https://docs.microsoft.com/dotnet/csharp/language-reference/builtin-types/built-in-types). Arrays of these types are handled as well (partially addresses #3229 for OTLP exporter).

The `ToOtlpAttribute` method handles attributes on logs, metrics, and resources. Attributes on trace data is handled in a different part of the code and has not yet been expanded to handle the same set of datatypes. I'll wait until we settle on the set of datatypes we want to handle to fix traces.

### DIFF
--- a/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/OtlpCommonExtensions.cs
+++ b/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/OtlpCommonExtensions.cs
@@ -94,11 +94,13 @@ namespace OpenTelemetry.Exporter.OpenTelemetryProtocol.Implementation
                     {
                         var value = ToOtlpValue(item);
 
+#if NETFRAMEWORK || NETSTANDARD
                         // nint[] and nuint[] falls through to this case and ToOtlpValue will return null
                         if (value == null)
                         {
                             return null;
                         }
+#endif
 
                         arrayValue.Values.Add(value);
                     }
--- a/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/OtlpCommonExtensions.cs
+++ b/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/OtlpCommonExtensions.cs
@@ -17,7 +17,6 @@
 using System;
 using System.Collections.Generic;
 using System.Runtime.CompilerServices;
-using Google.Protobuf.Collections;
 using OtlpCommon = Opentelemetry.Proto.Common.V1;
 
 namespace OpenTelemetry.Exporter.OpenTelemetryProtocol.Implementation
--- a/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/OtlpCommonExtensions.cs
+++ b/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/OtlpCommonExtensions.cs
@@ -35,7 +35,7 @@ namespace OpenTelemetry.Exporter.OpenTelemetryProtocol.Implementation
 
             if (value == null)
             {
-                OpenTelemetryProtocolExporterEventSource.Log.UnsupportedAttributeType(kvp.Value.GetType(), kvp.Key);
+                OpenTelemetryProtocolExporterEventSource.Log.UnsupportedAttributeType(kvp.Value.GetType().ToString(), kvp.Key);
                 return null;
             }
 
--- a/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/OtlpCommonExtensions.cs
+++ b/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/OtlpCommonExtensions.cs
@@ -33,9 +33,13 @@ namespace OpenTelemetry.Exporter.OpenTelemetryProtocol.Implementation
 
             var value = ToOtlpValue(kvp.Value);
 
-            return value != null
-                ? new OtlpCommon.KeyValue { Key = kvp.Key, Value = value }
-                : null;
+            if (value == null)
+            {
+                OpenTelemetryProtocolExporterEventSource.Log.UnsupportedAttributeType(kvp.Value.GetType(), kvp.Key);
+                return null;
+            }
+
+            return new OtlpCommon.KeyValue { Key = kvp.Key, Value = value };
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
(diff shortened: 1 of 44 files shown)


src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/OtlpCommonExtensions.cs
```diff
@@ -0,0 +1,116 @@
...
+                    }
+
+                    break;
+
+                // case nint:    Pointer type.
+                // case nuint:   Pointer type.
+                // case ulong:   May throw an exception on overflow.
+                // case decimal: Converting to double produces rounding errors.
+                default:
+                    return null;
```
cijothomas, line 110: this would be a breaking change? as we were doing ToString() before, and now we ignore the value? should we retain the current behavior?
(Or we treat current behavior as a unintentional bug, and this is indeed the fix)

Either way - please add to changelog entry as well.

cijothomas (approved):
LGTM. Need changelog before merge

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/OtlpCommonExtensions.cs
alanwest: > Or we treat current behavior as a unintentional bug, and this is indeed the fix

This is where I'm leaning but want to hear other opinions. While unusual, I have seen a `ToString` overload be susceptible to exceptions, mutating state, and even deadlock.

As an end user, I'd prefer to have to be intentional about how I'm ToStringing my attributes that are not just simple numeric types or strings.

cijothomas: I think this is the right change.

One thing to note is - do we log anywhere about this? i.e if I miss some values, and i enable self-diagnostics, would there be something that tells me exporter is dropping things?

alanwest: At one point, I thought the Jaeger and Zipkin exporters did not ToString arbitrary attributes. I guess they do now. So, I'm ok `ToString`ing everything else if that's what folks prefer.

alanwest: > One thing to note is - do we log anywhere about this? i.e if I miss some values, and i enable self-diagnostics, would there be something that tells me exporter is dropping things?

No logging at the moment - will add.

alanwest: > I think this is the right change.

In that case, we should update Zipkin, Jaeger, Console, Prometheus to behave the same. Maybe there's some way to unify this logic, but I haven't put my mind to this yet.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/OtlpCommonExtensions.cs
```diff
@@ -0,0 +1,123 @@
...
+                        case ushort[]:
+                        case int[]:
+                        case uint[]:
+                        case long[]:
+#if NETSTANDARD || NETFRAMEWORK
+                            if (array is nint[])
+                            {
+                                return null;
+                            }
+#endif
```
alanwest, line 90: I do not understand...

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/OtlpCommonExtensions.cs
cijothomas: From SIG meet: revert to use ToString(), and explore IFormattable in a follow up.

alanwest:
@cijothomas I think this is good to go now. As discussed in the SIG meeting, I've restored the behavior of calling `ToString` on all other types not explicitly supported.

reyang (approved):

PR #3266, opened 2022-05-10
fixing InMemoryExporter & Metrics bug. new class: MetricSnapshot. new ctor: InMemoryExporter(Func)

TimothyMothra:
Bug: #2361 
Supersedes: #3198.

## Changes
- new class `MetricSnapshot`. This is used for a strategic selective copy of the minimum members to facilitate unit testing.
  - other names considered: ExportedMetric, ExportableMetricCopy, InMemoryMetricCopy, ReadOnlyMetric. 
- new constructor added to `InMemoryExporter` to accept `Func<Batch<T>, ExportResult`.
  This allows the export behavior to be fully replaced and provides greater flexibility for unit tests.
- new extension methods `AddInMemoryExporter(ICollection<MetricSnapshot>)` added to `InMemoryExporterMetricsExtensions`.

### Please provide a brief description of the changes here.

`Metric` instances are reused by design. 
As a side effect, the `InMemoryExporter` will repeatedly export duplicates of `Metric` instances.
Because of this, `Metric`s can be modified after export. See also: https://github.com/open-telemetry/opentelemetry-dotnet/issues/2361#issuecomment-1049175566

This change introduces a new class `MetricSnapshot` to be used in unit tests. 
This class is a collection of minimum members needed for existing unit tests.
An end user has the option to use either this copy or the original `Metric`.

This change also introduces a new ctor `public InMemoryExporter(Func<Batch<T>, ExportResult> exportFunc)`.
This is needed to export a type that differs from the `Batch` type, in this case converting `Metric` to `MetricSnapshot`. 

### For significant contributions please make sure you have completed the following items:

## Additional Comment
This PR keeps the existing functionality, allowing users to export `Metric`s via the `InMemoryExporter`.
We have some tests that have a dependency on the `Metric`.
- `PrometheusSerializerTests` needs `Metric` to test `PrometheusSerializer.WriteMetric()`
- `OtlpMetricsExporterTests` needs `Batch<Metric>` to test `ExportMetricsServiceRequest.AddMetrics()`

### DIFF
--- a/src/OpenTelemetry.Exporter.InMemory/InMemoryExporter.cs
+++ b/src/OpenTelemetry.Exporter.InMemory/InMemoryExporter.cs
@@ -23,20 +23,18 @@ namespace OpenTelemetry.Exporter
         where T : class
     {
         private readonly ICollection<T> exportedItems;
-        private readonly ExportDelegate onExport;
+        private readonly Func<Batch<T>, ExportResult> onExport;
 
         public InMemoryExporter(ICollection<T> exportedItems)
         {
             this.exportedItems = exportedItems;
-            this.onExport = this.DefaultExport;
+            this.onExport = (Batch<T> batch) => this.DefaultExport(batch);
         }
 
         public InMemoryExporter(Func<Batch<T>, ExportResult> exportFunc)
         {
-            this.onExport = (in Batch<T> batch) => exportFunc(batch);
+            this.onExport = exportFunc;
         }
-
-        private delegate ExportResult ExportDelegate(in Batch<T> batch);
 
         public override ExportResult Export(in Batch<T> batch) => this.onExport(batch);
 
--- a/src/OpenTelemetry.Exporter.InMemory/InMemoryExporter.cs
+++ b/src/OpenTelemetry.Exporter.InMemory/InMemoryExporter.cs
@@ -14,6 +14,7 @@
 // limitations under the License.
 // </copyright>
 
+using System;
 using System.Collections.Generic;
 
 namespace OpenTelemetry.Exporter
@@ -22,13 +23,24 @@ namespace OpenTelemetry.Exporter
         where T : class
     {
         private readonly ICollection<T> exportedItems;
+        private readonly ExportDelegate onExport;
 
         public InMemoryExporter(ICollection<T> exportedItems)
         {
             this.exportedItems = exportedItems;
+            this.onExport = this.DefaultExport;
         }
 
-        public override ExportResult Export(in Batch<T> batch)
+        public InMemoryExporter(Func<Batch<T>, ExportResult> exportFunc)
+        {
+            this.onExport = (in Batch<T> batch) => exportFunc(batch);
+        }
+
+        private delegate ExportResult ExportDelegate(in Batch<T> batch);
+
+        public override ExportResult Export(in Batch<T> batch) => this.onExport(batch);
+
+        private ExportResult DefaultExport(in Batch<T> batch)
         {
             if (this.exportedItems == null)
             {
--- a/src/OpenTelemetry.Exporter.InMemory/InMemoryExporter.cs
+++ b/src/OpenTelemetry.Exporter.InMemory/InMemoryExporter.cs
@@ -31,7 +31,7 @@ namespace OpenTelemetry.Exporter
             this.onExport = (Batch<T> batch) => this.DefaultExport(batch);
         }
 
-        public InMemoryExporter(Func<Batch<T>, ExportResult> exportFunc)
+        internal InMemoryExporter(Func<Batch<T>, ExportResult> exportFunc)
         {
             this.onExport = exportFunc;
         }
--- a/src/OpenTelemetry.Exporter.InMemory/InMemoryExporterMetricsExtensions.cs
+++ b/src/OpenTelemetry.Exporter.InMemory/InMemoryExporterMetricsExtensions.cs
@@ -33,10 +33,28 @@ namespace OpenTelemetry.Metrics
         /// <summary>
         /// Adds InMemory metric exporter to the <see cref="MeterProviderBuilder"/> using default options.
         /// </summary>
+        /// <remarks>
+        /// Be aware that <see cref="Metric"/> may continue to be updated after export.
+        /// </remarks>
         /// <param name="builder"><see cref="MeterProviderBuilder"/> builder to use.</param>
-        /// <param name="exportedItems">Collection which will be populated with the exported MetricItem.</param>
+        /// <param name="exportedItems">Collection which will be populated with the exported <see cref="Metric"/>.</param>
         /// <returns>The instance of <see cref="MeterProviderBuilder"/> to chain the calls.</returns>
         public static MeterProviderBuilder AddInMemoryExporter(this MeterProviderBuilder builder, ICollection<Metric> exportedItems)
+        {
+            return builder.AddInMemoryExporter(exportedItems: exportedItems, configureMetricReader: null);
+        }
+
+        /// <summary>
+        /// Adds InMemory metric exporter to the <see cref="MeterProviderBuilder"/>.
+        /// </summary>
+        /// <remarks>
+        /// Be aware that <see cref="Metric"/> may continue to be updated after export.
+        /// </remarks>
+        /// <param name="builder"><see cref="MeterProviderBuilder"/> builder to use.</param>
+        /// <param name="exportedItems">Collection which will be populated with the exported <see cref="Metric"/>.</param>
+        /// <param name="configureMetricReader"><see cref="MetricReader"/> configuration options.</param>
+        /// <returns>The instance of <see cref="MeterProviderBuilder"/> to chain the calls.</returns>
+        public static MeterProviderBuilder AddInMemoryExporter(this MeterProviderBuilder builder, ICollection<Metric> exportedItems, Action<MetricReaderOptions> configureMetricReader)
         {
             Guard.ThrowIfNull(builder);
             Guard.ThrowIfNull(exportedItems);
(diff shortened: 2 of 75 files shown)


TimothyMothra:
@cijothomas, this PR is meant to address your ask here: https://github.com/open-telemetry/opentelemetry-dotnet/pull/3198#discussion_r869274010

src/OpenTelemetry.Exporter.InMemory/MetricCopy.cs
```diff
@@ -0,0 +1,58 @@
...
+using System.Collections.Generic;
+
+namespace OpenTelemetry.Metrics
+{
+    /// <summary>
+    /// This class represents a selective copy of <see cref="Metric"/>.
+    /// This contains the minimum fields and properties needed for most
+    /// unit testing scenarios.
+    /// </summary>
+    public class MetricCopy
```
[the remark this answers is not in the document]
TimothyMothra: Seems like there is a precedence to use "ReadOnly".

- In this repo: [ReadOnlyTagCollection](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/ReadOnlyTagCollection.cs)
- .NET: ReadOnlySpan, ReadOnlyCollection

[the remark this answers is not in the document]
cijothomas: ReadOnlyMetric won't be needed - the original Metric is also read only for the user.

I think MetricSnapShot would be more accurate - it is the snapshot of the metric at a point, and won't be affected by anything occuring after that point.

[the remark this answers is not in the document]
TimothyMothra: renamed to MetricSnapshot

src/OpenTelemetry.Exporter.InMemory/InMemoryExporter.cs
```diff
@@ -22,13 +23,24 @@ public class InMemoryExporter<T> : BaseExporter<T>
         where T : class
     {
         private readonly ICollection<T> exportedItems;
+        private readonly ExportDelegate onExport;
```
utpilla, line 26: Use `Func<Batch<T>, ExportResult>` instead of a `delegate`? I think that should simplify the assignment operation in the ctor.

src/OpenTelemetry.Exporter.InMemory/InMemoryExporterMetricsExtensions.cs
```diff
@@ -93,5 +135,33 @@ public static MeterProviderBuilder AddInMemoryExporter(this MeterProviderBuilder
...
+            ICollection<MetricSnapshot> exportedItems,
+            MetricReaderOptions metricReaderOptions,
+            Action<MetricReaderOptions> configureMetricReader)
+        {
+            configureMetricReader?.Invoke(metricReaderOptions);
+
+            var metricExporter = new InMemoryExporter<Metric>(
+                exportFunc: metricBatch =>
+                {
+                    foreach (var metric in metricBatch)
```
utpilla, line 150: Add a `null` check for `exportedItems` before the `foreach` loop.

test/OpenTelemetry.Tests/Metrics/InMemoryExporterTests.cs
```diff
@@ -45,9 +47,9 @@ public void InMemoryExporterShouldDeepCopyMetricPoints()
             meterProvider.ForceFlush();
 
             var metric = exportedItems[0]; // Only one Metric object is added to the collection at this point
-            var metricPointsEnumerator = metric.GetMetricPoints().GetEnumerator();
+            var metricPointsEnumerator = metric.MetricPoints.GetEnumerator();
             Assert.True(metricPointsEnumerator.MoveNext()); // One MetricPoint is emitted for the Metric
-            ref readonly var metricPointForFirstExport = ref metricPointsEnumerator.Current;
```
utpilla, line 50: We should let the `ref` and `ref readonly` semantics be here as we want to check that that `MetricPoint` in the `ICollection` is not updated.

src/OpenTelemetry.Exporter.InMemory/InMemoryExporter.cs
TimothyMothra: I can do this change and remove the `Delegate`. 
Note that this will complicate the other ctor as now the `DefaultExport()` method needs to be encapsulated in a func.
`this.onExport = (Batch<T> batch) => this.DefaultExport(batch);`

test/OpenTelemetry.Tests/Metrics/InMemoryExporterTests.cs
TimothyMothra: with this change, `MetricPoints` is a `IReadOnlyList<MetricPoint>`. It's enumerator cannot be returned by reference:

>Error	CS8156	An expression cannot be used in this context because it may not be passed or returned by reference

src/OpenTelemetry.Exporter.InMemory/InMemoryExporterMetricsExtensions.cs
```diff
@@ -93,5 +135,38 @@ public static MeterProviderBuilder AddInMemoryExporter(this MeterProviderBuilder
...
+        private static MeterProviderBuilder AddInMemoryExporter(
+            MeterProviderBuilder builder,
+            ICollection<MetricSnapshot> exportedItems,
+            MetricReaderOptions metricReaderOptions,
+            Action<MetricReaderOptions> configureMetricReader)
+        {
+            configureMetricReader?.Invoke(metricReaderOptions);
+
+            var metricExporter = new InMemoryExporter<Metric>(
+                exportFunc: metricBatch =>
```
utpilla, line 148: I don't think the `Batch<Metric>` is passed here with `in` modifier constraint.

test/OpenTelemetry.Tests/Metrics/InMemoryExporterTests.cs
utpilla: In that case, we should enumerate over the first MetricSnapshot's `MetricPoints` again to confirm that it has not changed otherwise we are testing it on the copy that was created before the second snapshot was added.

TimothyMothra: Is there any reason to keep using Enumerators here?
I think we can accomplish the same thing just interacting with the list.

utpilla: Not anymore now that you have changed the type to a list in `MetricSnapshot`.

src/OpenTelemetry.Exporter.InMemory/InMemoryExporterMetricsExtensions.cs
TimothyMothra: func doesn't support the `in` keyword, but the encapsulated method still uses `in`:
https://github.com/open-telemetry/opentelemetry-dotnet/blob/04b894e1fd1d6580ff4adf575bc1eb7fe5254446/src/OpenTelemetry.Exporter.InMemory/InMemoryExporter.cs#L39

I just debugged and confirmed that any changes made to the batch within the func don't persist. 
I think I can author a test that validates that the access modifier still applies here.

Would that address your concern here?

utpilla: The main benefit of using `in` here is to avoid copying the `Batch` struct. If it's not passed with the `in` modifier, the method would get a copy of the struct.

test/OpenTelemetry.Tests/Metrics/InMemoryExporterTests.cs
```diff
@@ -39,30 +41,31 @@ public void InMemoryExporterShouldDeepCopyMetricPoints()
...
+            var metric2 = exportedItems[1]; // Second Metric object is added to the collection at this point
+            Assert.Single(metric2.MetricPoints);
+            var metricPoint2 = metric2.MetricPoints[0];
+            Assert.Equal(25, metricPoint2.GetSumLong());
 
-            // MetricPoint.LongValue for the first exporter metric should still be 10
-            Assert.Equal(10, metricPointForFirstExport.GetSumLong());
+            // TEST 3: Verify first exported metric is unchanged
+            // MetricPoint.LongValue for the first exported metric should still be 10
+            Assert.Equal(10, metricPoint1.GetSumLong());
```
utpilla, line 68: You should use `metric1.MetricPoints[0].GetSumLong()` here. We want to ensure that it doesn't change with each export (after `ForceFlush`).

`metricPoint1` still has the copy of the `MetricPoint` before the second export happened.

utpilla (approved):

src/OpenTelemetry.Exporter.InMemory/InMemoryExporterMetricsExtensions.cs
```diff
@@ -93,5 +135,40 @@ public static MeterProviderBuilder AddInMemoryExporter(this MeterProviderBuilder
...
+                metricReaderOptions,
+                DefaultExportIntervalMilliseconds,
+                DefaultExportTimeoutMilliseconds);
+
+            return builder.AddReader(metricReader);
+        }
+
+        private static ExportResult ExportMetricSnapshot(in Batch<Metric> batch, ICollection<MetricSnapshot> exportedItems)
+        {
+            if (exportedItems == null)
```
cijothomas, line 161: given the Guards to the addinmemoryexporter methods, this should never be null

src/OpenTelemetry.Exporter.InMemory/InMemoryExporter.cs
```diff
@@ -22,13 +23,22 @@ public class InMemoryExporter<T> : BaseExporter<T>
...
+        private readonly Func<Batch<T>, ExportResult> onExport;
 
         public InMemoryExporter(ICollection<T> exportedItems)
         {
             this.exportedItems = exportedItems;
+            this.onExport = (Batch<T> batch) => this.DefaultExport(batch);
         }
 
-        public override ExportResult Export(in Batch<T> batch)
+        public InMemoryExporter(Func<Batch<T>, ExportResult> exportFunc)
```
cijothomas, line 34: we can leave this internal as we don't expect people to use this directly. (they use it via the extension methods only.)

TimothyMothra:
@cijothomas, i just updated the changelog;
https://github.com/open-telemetry/opentelemetry-dotnet/blob/f2673c7cf42f70cd5812321710358cf5c6cf2b54/src/OpenTelemetry.Exporter.InMemory/CHANGELOG.md?plain=1#L5-L7

src/OpenTelemetry.Exporter.InMemory/InMemoryExporterMetricsExtensions.cs
utpilla: In the `AddInMemoryExporter` method, we provide a reference to the collection. This could be set to `null` later on externally. We don't want `NullReferenceException` in that case.

TimothyMothra: I agree with @utpilla here. restored the if check

PR #3274, opened 2022-05-11
Use ToOtlpAttribute method for traces

alanwest:
`ToOtlpAttribute` method is now used for transforming attributes on all the things: traces, metrics, logs, resources.

There will be a failing test on this PR which will be resolved by merging #3273.

### DIFF
--- a/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ActivityExtensions.cs
+++ b/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ActivityExtensions.cs
@@ -360,14 +360,13 @@ namespace OpenTelemetry.Exporter.OpenTelemetryProtocol.Implementation
                 {
                     PooledList<OtlpCommon.KeyValue>.Add(ref this.Tags, attribute);
 
-                    switch (attribute.Value.ValueCase)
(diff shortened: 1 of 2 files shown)


src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ActivityExtensions.cs
```diff
@@ -361,74 +355,19 @@ public bool ForEach(KeyValuePair<string, object> activityTag)
                     this.Created = true;
                 }
 
-                OtlpCommon.ArrayValue arrayValue;
+                var attribute = activityTag.ToOtlpAttribute();
```
cijothomas, line 358: the peer serviceResolver part is another switch..case. It might be better to switch to simpler If..Else.

alanwest: Not sure this is what you had in mind, but refactored it a bit to only to the peer service stuff if the attribute was successfully transformed.

cijothomas (approved):

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ActivityExtensions.cs
alanwest: changed to if/else

Issue #3171, opened 2022-04-13
Add `double.NaN` to Histogram Boundaries Tests 


mic-max:
Take tests from this superseded, closed PR https://github.com/open-telemetry/opentelemetry-dotnet/pull/2592

Add `double.NaN` to https://github.com/open-telemetry/opentelemetry-dotnet/blob/cd72a39d99485833040245b8957130f2b7cb374f/test/OpenTelemetry.Tests/Metrics/MetricTestData.cs#L45

PR #3279, opened 2022-05-12
Histogram handle of `double.NaN` bounds and values

mic-max:
Fixes https://github.com/open-telemetry/opentelemetry-dotnet/issues/3171

## Changes

- It will correctly place `double.NaN` values in the final histogram bucket
- Exception will be thrown if `double.NaN` is used as a histogram bound

### DIFF
--- a/src/OpenTelemetry/Metrics/ExplicitBucketHistogramConfiguration.cs
+++ b/src/OpenTelemetry/Metrics/ExplicitBucketHistogramConfiguration.cs
@@ -30,7 +30,7 @@ namespace OpenTelemetry.Metrics
         /// Requirements:
         /// <list type="bullet">
         /// <item>The array must be in ascending order with distinct
-        /// values.</item>
+        /// values. double.NaN is not allowed</item>
         /// <item>An empty array would result in no histogram buckets being
         /// calculated.</item>
         /// <item>A null value would result in default bucket boundaries being
@@ -58,7 +58,7 @@ namespace OpenTelemetry.Metrics
                 {
                     if (!IsSortedAndDistinct(value))
                     {
-                        throw new ArgumentException($"Histogram boundaries are invalid. Histogram boundaries must be in ascending order with distinct values.", nameof(value));
+                        throw new ArgumentException($"Histogram boundaries are invalid. Histogram boundaries must be in ascending order with distinct values. double.NaN is not allowed.", nameof(value));
                     }
 
                     double[] copy = new double[value.Length];
@@ -76,9 +76,14 @@ namespace OpenTelemetry.Metrics
 
         private static bool IsSortedAndDistinct(double[] values)
         {
+            if (double.IsNaN(values[0]))
+            {
+                return false;
+            }
+
             for (int i = 1; i < values.Length; i++)
             {
-                if (values[i] <= values[i - 1])
+                if (double.IsNaN(values[i]) || values[i] <= values[i - 1])
                 {
                     return false;
                 }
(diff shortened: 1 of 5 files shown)


src/OpenTelemetry/Metrics/ExplicitBucketHistogramConfiguration.cs
```diff
@@ -30,7 +30,7 @@ public class ExplicitBucketHistogramConfiguration : MetricStreamConfiguration
         /// Requirements:
         /// <list type="bullet">
         /// <item>The array must be in ascending order with distinct
-        /// values.</item>
+        /// values. double.NaN is not allowed</item>
```
reyang, line 33: What about other IEEE754 subnormal (e.g. -Inf)?

reyang (approved):

mic-max:
@CodeBlanch I changed how recording a `double.NaN` affects the histogram sum. I thought it should not affect it, this fails the prometheus serializer test. How do you think we should handle this value? I don't see a specific mention of it in the [OTel spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#explicit-bucket-histogram-aggregation)

src/OpenTelemetry/Metrics/ExplicitBucketHistogramConfiguration.cs
mic-max: Good callout. I could do a follow up change for other subnormals.
@reyang @alanwest How should we treat subnormals being used as bounds or values.
I assume bounds, we should throw exception. But for values, there's decisions to be made (SDK specific) of whether we should have them affect bucket counts (and which bucket), sum, min, and max for histograms. But I guess this also expands to the other instrument types.

reyang: Check these:

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#numerical-limits-handling

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#exponential-histogram-aggregation-handle-all-normal-values

reyang: The 2nd link applies to exponential histogram, but many ideas might be helpful for explicit bucket histogram, too.

alanwest: Nice. Hadn't seen the new wording for exponential histograms. I think it makes sense to follow this for explicit-bound histograms too. That is, effectively drop all sub-normal values from affecting min, max, and sum. I also think this wording would suggest that it is ok to not modify bucket counts either:

> because these values do not map into a valid bucket

reyang:
> this fails the prometheus serializer test. How do you think we should handle this value?

@mic-max PrometheusExporter is still experimental, feel free to make changes to it (including the tests) if you think it makes sense (based on what the latest Stable SDK spec says).

src/OpenTelemetry/Metrics/ExplicitBucketHistogramConfiguration.cs
reyang: Sharing some additional context/thoughts:

image

reyang: I think for Sum, the challenge is that even if we don't take subnormal as input, the output could still be subnormal:

https://github.com/open-telemetry/opentelemetry-dotnet/blob/d983cb1270fa7a2738f9179d46f581268a4bb449/test/OpenTelemetry.Exporter.Prometheus.Tests/PrometheusSerializerTests.cs#L176-L186

PR #3285, opened 2022-05-17
Add null check when a view does not match an instrument

### LINKED ISSUE #3283: AddView - Returning null results in a NullReferenceException
https://github.com/open-telemetry/opentelemetry-dotnet/blob/879d3c91a0032eee75061871851e2d4febd6899f/src/OpenTelemetry/Metrics/MeterProviderSdk.cs#L170

```
metricStreamConfig = viewConfig(instrument);

// The SDK provides some static MetricStreamConfigurations.
// For example, the Drop configuration. The static ViewId
// should not be changed for these configurations.
if (!metricStreamConfig.ViewId.HasValue)
{
    metricStreamConfig.ViewId = i;
}
```

I suggest it should be like this:
```
if(metricStreamConfig != null && !metricStreamConfig.ViewId.HasValue)
{
    metricStreamConfig.ViewId = i;
}
```

alanwest:
Fixes #3283

Bug was causing superfluous warnings in diagnostic logs when an instrument didn't match a view config. For example:

```text
2022-05-17T19:29:28.8189110Z:View Configuration ignored for Instrument '{0}', Meter '{1}'. Reason: '{2}'. Measurements from the instrument will use default configuration for Aggregation. Suggested action: '{3}'{http.client.duration}{OpenTelemetry.Instrumentation.Http}{Object reference not set to an instance of an object.}{Fix the view configuration.}
```

### DIFF
--- a/src/OpenTelemetry/CHANGELOG.md
+++ b/src/OpenTelemetry/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Fix null reference exception when a metric view does not match an instrument.
+  ([#3285](https://github.com/open-telemetry/opentelemetry-dotnet/pull/3285))
+
 ## 1.3.0-beta.2
 
 Released 2022-May-16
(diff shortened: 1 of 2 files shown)


reyang (approved):

cijothomas (approved):

reyang:
Need a changelog entry?

PR #3307, opened 2022-05-25
Add & use ConfigureResource API.

### LINKED ISSUE #2909: SetResourceBuilder is overly restrictive, no way to cooperatively initialize Resources
# Feature Request

## Problem

When initializing a TracerProviderBuilder, some instrumentations might want to add resources (e.g. by interpreting environment variables), e.g. in their `AddFooInstrumentation(this TracerProviderBuilder builder)` method. Currently the only API for this is `SetResourceBuilder`, which *overwrites* all resources that might already exist.

## Proposed solution

One nice way would be an AddResourceBuilder API that merges the resources into existing ones. Another option would be to provide a GetResourceBuilder API on the TracerProviderBuilder.

## Alternatives considered (currently possible workarounds)

Currently, I can see two ways for any package to work around this:

1. Provide two separate functions for initialization. `AddFooInstrumentation(this TracerProviderBuilder builder)` which sets up everything *except* resources, and a separate `AddFooResources(this ResourceBuilder builder)` which adds *just* the resources. That makes initialization more complicated as now the user has to call two functions (four if you add the construction of ResourceBuilder and SetResourceBuilder which now the user also has to do themselves) instead of one. It works with multiple cooperating packages though.
2. Allow an additional Resource or ResourceBuilder to be passed as argument to the AddFooInstrumentation function. This is simple, but does not work across multiple packages that all want to add resources.

## Additional Context

Note that this is *not* about adding resources *after initialization*, but just to have a less awkward API to add resources from multiple sources during initialization to the TracerProviderBuilder.

Oberon00:
Fixes #2909.

## Changes

Added ConfigureResources which can replace SetResourceBuilder more succinctly in most cases
and has greater flexibility (applies to TracerProvicerBuilder, MeterProviderBuilder, OpenTelemetryLoggingOptions).

* ~~[ ] Design discussion issue #~~ Although this is very similar to the design proposed in #2909


cijothomas:
@Oberon00 Thanks for the PR. I am going to hold off for a week, as we'll be doing 1.3.0 stable very soon, and want to include this to 1.4 series, so as to have some time to gather feedbacks before calling it stable.

examples/AspNetCore/Program.cs
```diff
@@ -24,27 +24,28 @@
...
 
+Action<ResourceBuilder> configureResources = r => r.AddService(
+    serviceName, serviceVersion: assemblyVersion, serviceInstanceId: Environment.MachineName);
+
 // Traces
 builder.Services.AddOpenTelemetryTracing(options =>
 {
     options
-        .SetResourceBuilder(resourceBuilder)
+        .ConfigureResources(configureResources)
```
alanwest, line 48: Make the name singular `ConfigureResource`?

I think this better matches things conceptually in that each tracer/meter/logging provider has a single resource associated with it once it's built. I think we create multiple `Resource` objects as the resource is being built, but then they all get merged into one.

alanwest (approved):
I like this PR @Oberon00, big improvement! Just have one comment to discuss naming, but otherwise things seem solid.

docs/logs/customizing-the-sdk/Program.cs
```diff
@@ -30,9 +30,7 @@ public static void Main()
             builder.AddOpenTelemetry(options =>
             {
                 options.IncludeScopes = true;
-                options.SetResourceBuilder(ResourceBuilder.CreateDefault().AddService(
-                    serviceName: "MyService",
-                    serviceVersion: "1.0.0"));
+                options.ConfigureResources(r => r.AddService(serviceName: "MyService", serviceVersion: "1.0.0"));
```
alanwest, line 33: We'll need to make sure we follow up with opentelemetry.io docs as well. Quite a few examples using `SetResourceBuilder` like: https://opentelemetry.io/docs/instrumentation/net/getting-started/#console-application

examples/AspNetCore/Program.cs
Oberon00: No strong opinion, but the plural was not an accident. Aren't the semantic conventions for resources each called a resoruce? E.g. you add the Telemetry SDK resource, you add the service resource and the process.runtime resource. Of course in the end it's merged in a single resource.

Also, the primary feature of this new method is that it can be called multiple times additively. I felt that this is a bit better conveyed by the plural.

Singular also makes sense though, and I'm happy to change it if you think that's better.

joaopgrassi: I think if the idea is to be able to call it multiple times to configure multiple resources, having it singular seems more natural to me. Also if it's plural, I would assume to pass a list of things so another point for singular.

alanwest: > Aren't the semantic conventions for resources each called a resoruce? E.g. you add the Telemetry SDK resource, you add the service resource and the process.runtime resource.

Hmm, interesting. It hasn't occurred to me to think of it this way. Maybe there's not a "correct" interpretation here, but it's language like this from the [Resource SDK spec](https://github.com/open-telemetry/opentelemetry-specification/blob/9abbdd39d0b35f635f833f072013431da419894e/specification/resource/sdk.md) that has led me to my interpretation

> A [Resource](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/overview.md#resources) is an immutable representation of the entity producing telemetry as [Attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/README.md#attribute). For example, a process producing telemetry that is running in a container on Kubernetes has a Pod name, it is in a namespace and possibly is part of a Deployment which also has a name. All three of these attributes can be included in the Resource.

That's to say a single resource is defined and is comprised of attributes from potentially multiple bodies of semantic conventions.

cijothomas:
@Oberon00 could you fix the conflicts?
We just finished 1.3.0 release, and I am back to reviewing/merging all pending PRs..:)

Oberon00:
Make sure to change the name (or let me change the name) before releasing if desired.

Oberon00:
Build failures seem unrelated to this PR.

cijothomas:
@Oberon00 you can close/open the PR (or push a dummy commit) to re-trigger CI.

cijothomas:
D:\a\opentelemetry-dotnet\opentelemetry-dotnet\src\OpenTelemetry\.publicApi\net462\PublicAPI.Unshipped.txt(2,1): error RS0017: Symbol 'OpenTelemetry.Logs.OpenTelemetryLoggerOptions.ConfigureResources(System.Action<OpenTelemetry.Resources.ResourceBuilder!>! configure) -> OpenTelemetry.Logs.OpenTelemetryLoggerOptions!' is part of the declared API, but is either not public or could not be found [D:\a\opentelemetry-dotnet\opentelemetry-dotnet\src\OpenTelemetry\OpenTelemetry.csproj]


^^ This looks like this PR itself. Could you check this?

Oberon00:
Hopefully I now covered all target frameworks.

src/OpenTelemetry/Trace/TracerProviderBuilderExtensions.cs
```diff
@@ -69,7 +72,27 @@ public static TracerProviderBuilder SetResourceBuilder(this TracerProviderBuilde
...
+        }
+
+        /// <summary>
+        /// Modify the <see cref="ResourceBuilder"/> from which the Resource associated with
+        /// this provider is built from in-place.
+        /// </summary>
+        /// <param name="tracerProviderBuilder">TracerProviderBuilder instance.</param>
+        /// <param name="configure">An action which modifies the provided <see cref="ResourceBuilder"/> in-place.</param>
+        /// <returns>Returns <see cref="TracerProviderBuilder"/> for chaining.</returns>
+        public static TracerProviderBuilder ConfigureResource(this TracerProviderBuilder tracerProviderBuilder, Action<ResourceBuilder> configure)
```
cijothomas, line 88: Thoughts on the name `ConfigureResource` vs `ConfigureResourceBuilder`? @alanwest @Oberon00 ?

cijothomas (approved):
Looks good overall. I would like to get more thoughts on `ConfigureResource` vs `ConfigureResourceBuilder` as the name, but this can be decided before we ship a stable version, so wont block this PR.

PR #3319, opened 2022-05-31
Update protos 0.18.0

alanwest:
Changed the build of the OTLP exporter so that the protos are downloaded rather than maintaining a copy of them in the repo.


alanwest:
Meh. Seems like it build different via CI vs. locally... I'll figure this out later. In the meantime I'll open another PR updating things the old school way.

PR #3329, opened 2022-06-02
Add isRemote check for context propagation

### LINKED ISSUE #3315: Propagation in case of remote parent
Currently, TraceProviderSDK forces activity creation by returning PropagationOnly sampling result for root activities by checking 
[isRootSpan = traceId == default](https://github.com/open-telemetry/opentelemetry-dotnet/blob/4e6d07367249a06603fafbba55f1fde941b49835/src/OpenTelemetry/Trace/TracerProviderSdk.cs#L402). However, in case of remote parent the `traceId` is not default (e.g. asp.net core request with traceparent) which will result in loss of context propagation if the sampling results in data drop.

Proposing to add `isRemote` check as well so that the incoming `traceId` can be propagated to next hop.

#3310 covers the unit test for the scenario. Will create a follow up PR of TracerProviderSDK changes.

vishweshbankwar:
Fixes #3315 

## Changes


src/OpenTelemetry/Trace/TracerProviderSdk.cs
```diff
@@ -393,13 +393,13 @@ protected override void Dispose(bool disposing)
...
-            return PropagateOrIgnoreData(options.Parent.TraceId);
+            return PropagateOrIgnoreData(options.Parent.TraceId, options.Parent.IsRemote);
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        private static ActivitySamplingResult PropagateOrIgnoreData(ActivityTraceId traceId)
+        private static ActivitySamplingResult PropagateOrIgnoreData(ActivityTraceId traceId, bool isParentRemote)
         {
-            var isRootSpan = traceId == default;
+            var isRootSpan = traceId == default || isParentRemote;
```
reyang, line 402: This is semantically buggy, I guess?

Maybe:
```csharp
var isRootSpan = traceId == default;

return isRootSpan || isParentRemote
    ? ...
```

src/OpenTelemetry/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+* Included `isRemote` check in `PropagateOrIgnoreData` in `TracerProviderSDK` to
```
cijothomas, line 5: TracerProvider modified to create activity irrespective of SamplingResult, for spans with Remote Parent, to maintain context propagation.

^ or similar wording.

cijothomas:
@vishweshbankwar will hold off from merging for a day, as we are doing 1.3.0 stable release, and want to avoid any functional changes.

utpilla (approved):

cijothomas:
@vishweshbankwar Please fix the conflict, and we are good to go

src/OpenTelemetry/Trace/TracerProviderSdk.cs
```diff
@@ -393,17 +393,17 @@ protected override void Dispose(bool disposing)
                 return activitySamplingResult;
             }
 
-            return PropagateOrIgnoreData(options.Parent.TraceId);
+            return PropagateOrIgnoreData(options.Parent.TraceId, options.Parent.IsRemote);
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        private static ActivitySamplingResult PropagateOrIgnoreData(ActivityTraceId traceId)
+        private static ActivitySamplingResult PropagateOrIgnoreData(ActivityTraceId traceId, bool isParentRemote)
```
utpilla, line 400: We could probably explore updating the method signature here to something like this:
`private static ActivitySamplingResult PropagateOrIgnoreData(in ActivityContext parentContext)` and pass in `options.Parent` to this method. `ActivityContext` is a `readonly struct` so using the `in` modifier might give some minor perf gains.

PR #3334, opened 2022-06-03
ConsoleMetricExporter to export all resource attributes

cijothomas:
MetricExporter was only printing the servicename attribute. This fixes it to print all resource attributes, just like log and activity exporter.


CodeBlanch (approved):
LGTM

Issue #3353, opened 2022-06-09
Improve Code Coverage


TimothyMothra:
I want to improve code coverage in this repo.
My plan is to use this issue to link PRs and track overall progress.

## CodeCoverage
Measured by [CodeCov](https://app.codecov.io/gh/open-telemetry/opentelemetry-dotnet/)

<details><summary>Screenshot (2022-06-08)</summary>

![image](https://user-images.githubusercontent.com/28785781/172737969-47464b1e-5566-4d55-a239-76e55fa549d4.png)

</details>

FYI, these measurements will be heavily influenced by #3317, which is moving certain projects to the contrib repo. 
I'm using the Progress below to track overall changes to code coverage.

## Getting Started
As of today, 4 packages are below 80%
- ~OpenTelemetry.Instrumentation.AspNet.TelemetryHttpModule 50.37%~ This will be migrated to Contrib repo
- OpenTelemetry.Extensions.Hosting 75.00%
- OpenTelemetry.Exporter.InMemory 78.48%
- OpenTelemetry.Exporter.Prometheus 79.15%

My plan is to start with these. 
As I find areas to improve, I'll search for similar areas in other packages. 

## Progress

- 2022-06-03: 85.44%
- 2022-06-06: 85.24%
- 2022-06-08: 85.8%
  - InMemoryExporter: 78.8%
      - #3345
- 2022-06-10: 86.15%
  - InMemoryExporter: 89.87%
    - #3344
- 2022-07-22: 86.11%
- 2022-08-03: 86.81%
- 2022-08-04: 87.08%

PR #3362, opened 2022-06-13
improve test coverage: OpenTelemetry.Api `CallerArgumentExpressionAttribute`

TimothyMothra:
Contributes to #3353 
As of today;
 - Overall: 85.80%
 - Project: OpenTelemetry.Api: 86.85%
 - File: Guard.cs: 0%

This value for Guard isn't correct. Seems that CodeCov is only counting the first class in this file.

<details><summary>Screenshot</summary>

image

</details>

## Changes
- Update `CallerArgumentExpressionAttribute` with a link to the original implementation. [link](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Runtime/CompilerServices/CallerArgumentExpressionAttribute.cs).
- Add new tests for `CallerArgumentExpressionAttributeTests`, borrowed from [link](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Runtime/tests/System/Runtime/CompilerServices/CallerArgumentExpressionAttributeTests.cs)
- Cleanup GuardTest.cs
  - move all extra classes inside `public class GuardTest`.
  - remove `#pragma warning disable SA1402 // File may only contain a single type`
  - remove `#pragma warning disable SA1649 // File name should match first type name`


test/OpenTelemetry.Tests/Internal/CallerArgumentExpressionAttributeTests.cs
```diff
@@ -0,0 +1,50 @@
+// <copyright file="CallerArgumentExpressionAttributeTests.cs" company="OpenTelemetry Authors">
```
reyang, line 1: I guess the general practice in this repo is to have the test file name reflecting the source file name? (e.g. in this case it'll be `GuardTests.cs`)

TimothyMothra: General practice is also one class per file. ;)
I have no problem moving this to the GuardTests.cs. I'll get that done today :)

test/OpenTelemetry.Tests/Internal/GuardTest.cs
```diff
@@ -171,5 +163,38 @@ public void ZeroTest()
...
+            public string Bar { get; set; }
+        }
+
+        /// <summary>
+        /// Borrowed from: <see href="https://github.com/dotnet/runtime/blob/main/src/libraries/System.Runtime/tests/System/Runtime/CompilerServices/CallerArgumentExpressionAttributeTests.cs"/>.
+        /// </summary>
+        public class CallerArgumentExpressionAttributeTests
+        {
+#if !NETCOREAPP3_0_OR_GREATER
+        [Theory]
```
TimothyMothra, line 178: The dotnet-format CI is angry with me here.
It's not apparent what needs to be changed.

> D:\a\opentelemetry-dotnet\opentelemetry-dotnet\test\OpenTelemetry.Tests\Internal\GuardTest.cs(176,10): error WHITESPACE: Fix whitespace formatting. Replace 2 characters with '\r\n'. [D:\a\opentelemetry-dotnet\opentelemetry-dotnet]
> D:\a\opentelemetry-dotnet\opentelemetry-dotnet\test\OpenTelemetry.Tests\Internal\GuardTest.cs(17[8](https://github.com/open-telemetry/opentelemetry-dotnet/runs/6988755470?check_suite_focus=true#step:5:9),9): error WHITESPACE: Fix whitespace formatting. Insert '\s\s\s\s'. [D:\a\opentelemetry-dotnet\opentelemetry-dotnet]

alanwest: Agree the errors are confusing. Though just looking at this definitely seems some indentation is off.

From `[Theory]` on down needs another four space indent.

Probably not a problem, but might also move `#if !NETCOREAPP3_0_OR_GREATER` to outside the class definition.

TimothyMothra: Thanks, the indention was exactly the problem.
Because of the preprocessor, the format document wasn't working. 🤦‍♂️

I initially did have the preprocessor outside of the class, but then the formatter was identifying the xml comment `///` as an invalid formatted comment `//`. 

I got busy with other work this week. I should be able to circle back to this by Monday.

TimothyMothra: Fixed the preprocessor.

alanwest (approved):
LGTM.

I am curious about

> This value for Guard isn't correct. Seems that CodeCov is only counting the first class in this file.

If `CallerArgumentExpressionAttribute` was moved to its own file does CodeCov account for things correctly? Might be worth experimenting, but I don't feel strongly it has to block this PR.

TimothyMothra:
> We could experiment with moving `CallerArgumentExpressionAttribute` inside of `Guard`. I can look for other nested classes and see if CodeCov reports correctly.

I found an example that appears to be reporting correctly.
`SqlClientInstrumentationOptions` has an internal class `SqlConnectionDetails`.
https://github.com/open-telemetry/opentelemetry-dotnet/blob/7de3f001bd76a32cd9581a5e3ac5c71a06391f7b/src/OpenTelemetry.Instrumentation.SqlClient/SqlClientInstrumentationOptions.cs#L29
https://github.com/open-telemetry/opentelemetry-dotnet/blob/7de3f001bd76a32cd9581a5e3ac5c71a06391f7b/src/OpenTelemetry.Instrumentation.SqlClient/SqlClientInstrumentationOptions.cs#L261


CodeCov is showing 98.65%
image


As opposed to Guard.cs which is still incorrectly reporting 0%
image

TimothyMothra:
> I can look for other nested classes and see if CodeCov reports correctly.

I reverted the change to Guard.cs.
All the related tests failed. 
We should investigate this in a separate PR.

cijothomas (approved):

Issue #1453, opened 2020-11-04
SDK configuration from ENV Variable


cijothomas:
Spec has documented a list of ENV variable names from which SDK should read config, *if* SDK chose to support ENV var based config.
https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/sdk-environment-variables.md

cijothomas:
Given this is not a *MUST*, we can add this after-ga.

pellared:
I would like to discuss the potential API and behavior changes that are expected as part of this issue.

I see two "types" of configuration that can be done via environmental variables. 

The first one is about configuring a specific component like [Zipkin Exporter configuration](https://github.com/open-telemetry/opentelemetry-specification/blob/a5d34467b954866ffff90a1c0ef29a43a040b961/specification/sdk-environment-variables.md#jaeger-exporter). The logic here can be simply encapsulated into existing packages like https://www.nuget.org/packages/OpenTelemetry.Exporter.Zipkin/. No API changes needed.

The second one is harder as it is about enabling components e.g. [Exporter selection](https://github.com/open-telemetry/opentelemetry-specification/blob/a5d34467b954866ffff90a1c0ef29a43a040b961/specification/sdk-environment-variables.md#exporter-selection). I think this type of setting may even need some at least high-level design to make sure we have something that covers `OTEL_PROPAGATORS`, `OTEL_TRACES_SAMPLER`. Taking the `OTEL_TRACES_EXPORTER` as an example I imagine we could add a new extension method to https://www.nuget.org/packages/OpenTelemetry.Exporter.Zipkin/ with something like (I am bad at naming):

```
public static TracerProviderBuilder AddZipkinExporterEnvVarSelection(TracerProviderBuilder builder) {
    if (Environment.GetEnvironmentVariable == "jaeger") {
       return builder.AddZipkinExporter();
    } 
    return builder;
}
```

For this second type, it may be worth creating a convenience NuGet package with a method like `AddCommonEnvVarSelections` which would "register" all env vars defined by the [specs](https://github.com/open-telemetry/opentelemetry-specification/blob/a5d34467b954866ffff90a1c0ef29a43a040b961/specification/sdk-environment-variables.md),

pellared:
[...]

PR #3376, opened 2022-06-17
OTLP exporter support for limiting activity tags, events, and links

### LINKED ISSUE #3297: Ability to set a maximum attribute length
# Feature Request

NewRelic limits their attribute length to 4096 characters and if an attribute is larger than that, it will drop the data. It would be nice to have an option in the client that truncates any attribute values that are longer than a given size.

alanwest:
Fixes #3297
Related to #1453

This PR adds _some_ support for the attribute limit configuration from the spec:

* [General limits](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#attribute-limits)
* [Span-specific limits](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#span-limits)

One reason these limits exist in the spec is to offer users a way to control the memory consumption of instrumentation. This PR **does not** help with that. In order to control memory consumption, we'd need a way for the API to honor the limits. Currently, we cannot do this. So...

The configuration is only supported by the OTLP exporter for now and therefore the limits only impact the data sent over the wire. This is beneficial for sending data to backends that have their own limitations for number and length of attributes. My plan is to follow up with this support for the other exporters.

---

Tangential to the primary motivation of this PR, I've introduced an `SdkConfiguration` class. For now, I've dropped it in the OTLP exporter project and kept it internal. Ultimately, I'd like to move this class to the SDK project, but I did not wish to hash out its design and public API as part of this PR. Also, this means the attribute limit configuration can currently only be set via environment variables. Once moved to the SDK programatic configuration would also be possible.


src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Configuration/SdkConfiguration.cs
```diff
@@ -0,0 +1,69 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+
+namespace OpenTelemetry.Configuration;
+
+internal class SdkConfiguration
```
alanwest, line 19: Ideally I think it'd be nice to have a configuration class like this in the main SDK project. Don't want to take on the work of figuring out that API in the scope of this PR, so putting it here for now.

src/OpenTelemetry.Api/Internal/ActivityHelperExtensions.cs
```diff
@@ -306,12 +323,31 @@ public static void Enumerate(ActivityEvent activityEvent, ref TState state)
...
                 ActivityTagsCollectionEnumerator(
                     tags,
                     ref state,
                     ForEachTagValueCallbackRef);
             }
 
+            // TODO: When a limit has been configured an allocation-free enumerator is not used.
+            // Need to either:
+            //     1) modify the dynamically generated code to only enumerate up to the max number of items, or
+            //     2) wait until .NET 7 is released and do this more easily with the new enumerator functions
```
alanwest, line 341: If we think we'll do another stable release prior to .NET 7 then need to decide if it's worth modifying the dynamically generated allocation-free enumerator to support configured limits.

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/Configuration/SdkConfigurationTests.cs
```diff
@@ -0,0 +1,133 @@
...
+            var config = SdkConfiguration.Instance;
+
+            Assert.Null(config.AttributeValueLengthLimit);
+            Assert.Equal(128, config.AttributeCountLimit);
+            Assert.Null(config.SpanAttributeValueLengthLimit);
+            Assert.Equal(128, config.SpanAttributeCountLimit);
+            Assert.Equal(128, config.SpanEventCountLimit);
+            Assert.Equal(128, config.SpanLinkCountLimit);
+            Assert.Equal(128, config.EventAttributeCountLimit);
+            Assert.Equal(128, config.LinkAttributeCountLimit);
```
alanwest, line 49: This is a breaking change. I suppose the specification is a bit ambiguous here https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#attribute-limits. An SDK implementing the environment variables is a `SHOULD` per the spec, but it's not clear if the default values are also a `SHOULD`.

We have two options. Make the breaking change and consider this a bug fix. Or maintain the prior behavior and not introduce these new defaults.

src/OpenTelemetry.Api/Internal/ActivityHelperExtensions.cs
alanwest: Since this performance hit only affects users who set a limit, my proposal is to address this TODO in a follow up PR since we may be dropping support for DiagnosticSource v6.

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/Configuration/SdkConfigurationTests.cs
alanwest: I'm going to revert this change to defaults for now so we can come back to this conversation. Setting these new defaults mean everyone takes the perf hit described here https://github.com/open-telemetry/opentelemetry-dotnet/pull/3376#discussion_r924971691

src/OpenTelemetry/Internal/TagTransformer.cs
```diff
@@ -33,7 +34,7 @@ public bool TryTransformTag(KeyValuePair<string, object> tag, out T result)
         {
             case char:
             case string:
-                result = this.TransformStringTag(tag.Key, Convert.ToString(tag.Value));
+                result = this.TransformStringTag(tag.Key, TruncateString(Convert.ToString(tag.Value), maxLength));
```
utpilla, line 37: Do we need to do this for the `default` case as well?

src/OpenTelemetry.Api/Internal/ActivityHelperExtensions.cs
```diff
@@ -109,14 +109,15 @@ public static bool TryCheckFirstTag(this Activity activity, string tagName, out
         /// <typeparam name="T">The struct <see cref="IActivityEnumerator{T}"/> implementation to use for the enumeration.</typeparam>
         /// <param name="activity">Activity instance.</param>
         /// <param name="tagEnumerator">Tag enumerator.</param>
+        /// <param name="maxTags">Maximum number of tags to enumerate.</param>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Design", "CA1062:Validate arguments of public methods", Justification = "ActivityProcessor is hot path")]
-        public static void EnumerateTags<T>(this Activity activity, ref T tagEnumerator)
+        public static void EnumerateTags<T>(this Activity activity, ref T tagEnumerator, int? maxTags = null)
```
utpilla, line 115: Would it better to add an overload which takes in an additional non-nullable `int maxTags` as a parameter?

utpilla: I guess you'll also have to set the default values for these parameters these config variables then.

utpilla (approved):
Overall direction looks good to me.

src/OpenTelemetry/Internal/TagTransformer.cs
alanwest: Yes, good catch thanks. Fixed and added a test.

src/OpenTelemetry.Api/Internal/ActivityHelperExtensions.cs
alanwest: Yes? 😄 

I'm usually not a fan of optional parameters - especially for public APIs, but since this is internal I was a bit lazy.

Curious what your thoughts are. Do you similarly not prefer optional parameters, or is there another reason you think a proper overload would be better in this case? I'm fine changing this either way.

utpilla: I was suggesting that because that wat we could get rid of the extra `if` check for regular scenarios. 😄

src/OpenTelemetry.Api/Internal/ActivityHelperExtensions.cs
```diff
@@ -265,7 +270,7 @@ private static class ActivityTagsEnumeratorFactory<TState>
             private static readonly DictionaryEnumerator<string, object, TState>.ForEachDelegate ForEachTagValueCallbackRef = ForEachTagValueCallback;
 
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
-            public static void Enumerate(Activity activity, ref TState state)
+            public static void Enumerate(Activity activity, ref TState state, int? maxTags)
```
utpilla, line 273: If we don't go with the overload approach, then maybe assign `maxTags` a default value of `null`.

src/OpenTelemetry.Api/Internal/ActivityHelperExtensions.cs
alanwest: Oh yea, I see. Since a lot of this code will change or go away if we drop DiagnosticSource 6, I'm inclined to just leave it as is for the moment. What do you think?

utpilla: Yeah, I think that should be fine.

utpilla (approved):

cijothomas (approved):

PR #3381, opened 2022-06-17
Remove SetHttpFlavor from Http instrumentations

### LINKED ISSUE #3380: Remove SetHttpFlavor option from httpclient instrumentations
Related to https://github.com/open-telemetry/opentelemetry-dotnet/issues/3373

HttpClient instrumentations current have an option [SetHttpFlavor](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Instrumentation.Http/HttpClientInstrumentationOptions.cs#L34) which determines if 'http.flavor' is populated or not. This, I believe, was added in early days of instrumentation, when this was the only tag that was optional. Conventions have grown significantly since then, so maintaining on/off flag for each of them would be impractical.

'http.flavor' is listed as [Recommended](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/attribute-requirement-level.md#recommended) in the [Spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#common-attributes), so it should be populated by default. SDK may provide an option to turn it off.

Given the fact that this tag's value is readily available, and is a Recommended tag, proposing to remove this On/Off option.
If any user does not want this tag, they may remove it by doing `activity.SetTag("http.flavor", null)` with an `ActivityProcessor`.

As part of overall issue tracked by https://github.com/open-telemetry/opentelemetry-dotnet/issues/3373, we can consider exposing on/off flags in the future.

cijothomas:
Fixes #3380 

## Changes


alanwest (approved):
I never found the option to be very flavorful anyways...

reyang (approved):

PR #3375, opened 2022-06-16
[Prometheus] Split up projects based on hosting mechanism.

Yun-Ting:
Fixes https://github.com/open-telemetry/opentelemetry-dotnet/issues/3057.

## Changes

The updated project setup is as follows:

- `OpenTelemetry.Exporter.Prometheus.Shared` contains the PrometheusExporter, PrometheusExporterOptions and serialization methods for all the shared classes to both AspNetCore and HttpListener projects. This project is *not* expected to be consumed directly by the end users. It serves as a clean way to group shared code for the below 2 projects for semantic reasons. As a result, I've deleted the Implementation folder that was in `OpenTelemetry.Exporter.Prometheus` project, and moved out the shared files which was previously in the folder to be immediate children of the project.

- `OpenTelemetry.Exporter.Prometheus.HttpListener` contains the PrometheusHttpListener and its options class. It aims to provide end users a quick setup experience for using OTel metrics with Prometheus Exporter locally. Please refer to [readme](https://github.com/open-telemetry/opentelemetry-dotnet/blob/304f27ba3adbec7eda63abb21682f14f6377b2ff/src/OpenTelemetry.Exporter.Prometheus.HttpListener/README.md) for more details.

- `OpenTelemetry.Exporter.Prometheus.AspNetCore` contains the middleware and its extensions for registering. This is for users trying to run PrometheusExporter in production environment. Refer to [readme](https://github.com/open-telemetry/opentelemetry-dotnet/blob/304f27ba3adbec7eda63abb21682f14f6377b2ff/src/OpenTelemetry.Exporter.Prometheus.AspNetCore/README.md) for details.

TODO:
Currently, OpenTelemetry.Exporter.Prometheus.AspNetCore.Tests, OpenTelemetry.Exporter.Prometheus.HttpListener.Tests, and OpenTelemetry.Exporter.Prometheus.Shared.Tests are green. I'll be working on fixing the CIs (not sure if I broke other tests with bad merge.)

For significant contributions please make sure you have completed the following items:

* [ ] Appropriate `CHANGELOG.md` updated for non-trivial changes
* [ ] Design discussion issue #
* [ ] Changes in public API reviewed

PR #3382, opened 2022-06-18
Update Readme Prometheus exporter

damikun:
Add exact details about 2 options of scrapping endpoint... (middelware and builded-in http-listener)

- Maybe in future it is also good to extend the implementation by Socket scraping endpoint  because HttpListener is `Http.sys` and this can have issues with SSL...

Http listener itself is proprietaly implementation from .net FW and also have some complications in crossplatform env. sinc it is originaly bind to be used on Windows....

Fixes #.

## Changes


cijothomas:
@damikun Thanks for your contribution.!

I'd suggest to wait till https://github.com/open-telemetry/opentelemetry-dotnet/pull/3375 is done, as it is splitting up Prometheus exporter into different projects.

Issue #3019, opened 2022-03-10
[.NET7.0] Update AspNetCore instrumentation to enable activity creation using ActivitySource


vishweshbankwar:
similar to #3018

vishweshbankwar:
Completed via https://github.com/open-telemetry/opentelemetry-dotnet/pull/3391

PR #3391, opened 2022-06-20
[.NET7.0] AspNetCore ActivitySource Migration

vishweshbankwar:
Fixes #.
#3019 

TODO: Add additional tests.

## Changes


src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs
```diff
@@ -35,6 +35,8 @@ namespace OpenTelemetry.Instrumentation.AspNetCore.Implementation
     internal class HttpInListener : ListenerHandler
     {
         internal const string ActivityOperationName = "Microsoft.AspNetCore.Hosting.HttpRequestIn";
+        internal static readonly bool IsNet7OrGreater = typeof(HttpRequest).Assembly.GetName().Version.Major >= 7;
```
cijothomas, line 38: consider putting this inside try..catch.

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs
```diff
@@ -35,6 +35,8 @@ namespace OpenTelemetry.Instrumentation.AspNetCore.Implementation
...
+        internal static readonly string FrameworkActivitySourceName = "Microsoft.AspNetCore";
```
cijothomas, line 39: add link to where this source name is coming from.

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs
vishweshbankwar: Added

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs
```diff
@@ -177,6 +203,11 @@ public override void OnStartActivity(Activity activity, object payload)
 
         public override void OnStopActivity(Activity activity, object payload)
         {
+            if (IsNet7OrGreater && string.IsNullOrEmpty(activity.Source.Name))
```
cijothomas, line 206: do we need this check? in which situation will be path be hit?

src/OpenTelemetry.Instrumentation.AspNetCore/AspNetCoreInstrumentation.cs
```diff
@@ -25,9 +25,33 @@ internal class AspNetCoreInstrumentation : IDisposable
...
+
+            return true;
+        };
+
         public AspNetCoreInstrumentation(HttpInListener httpInListener)
         {
-            this.diagnosticSourceSubscriber = new DiagnosticSourceSubscriber(httpInListener, null);
+            // For .NET7.0 activity will be created using activitySource.
+            // https://github.com/dotnet/aspnetcore/blob/main/src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs#L327
+            // However, in case when activity creation returns null (due to sampling)
```
cijothomas, line 42: Is it possible to return null activity, as this is a root activity?

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs
alanwest: Might wanna make this a permalink. Linked from main the line number could change. You can highlight a line in github and then click the 3 little dots when hovering over the line to select copy permalink.

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/BasicTests.cs
```diff
@@ -348,7 +348,7 @@ public async Task ExtractContextIrrespectiveOfSamplingDecision(SamplingDecision
                 var expectedTraceId = ActivityTraceId.CreateRandom();
                 var expectedParentSpanId = ActivitySpanId.CreateRandom();
                 var expectedTraceState = "rojo=1,congo=2";
-                var activityContext = new ActivityContext(expectedTraceId, expectedParentSpanId, ActivityTraceFlags.Recorded, expectedTraceState);
+                var activityContext = new ActivityContext(expectedTraceId, expectedParentSpanId, ActivityTraceFlags.Recorded, expectedTraceState, true);
```
alanwest, line 351: Was setting `isRemote = true` required for this test?

vishweshbankwar: Yes - the extracted context should have `isRemote = true` in case of ASP.NET Core incoming request. This change is not completely related to .NET7.0. Even without that it should use `isRemote = true`. It relates to the check [here] (https://github.com/open-telemetry/opentelemetry-dotnet/blob/f471a9f197d797015123fe95d3e12b6abf8e1f5f/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs#L94)

src/OpenTelemetry.Instrumentation.AspNetCore/TracerProviderBuilderExtensions.cs
```diff
@@ -53,8 +53,19 @@ public static class TracerProviderBuilderExtensions
...
             AspNetCoreInstrumentation instrumentation)
         {
-            builder.AddSource(HttpInListener.ActivitySourceName);
-            builder.AddLegacySource(HttpInListener.ActivityOperationName); // for the activities created by AspNetCore
+            // For .NET7.0 onwards activity will be created using activitySource.
+            // https://github.com/dotnet/aspnetcore/blob/main/src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs#L327
+            // For .NET6.0 and below, we will continue to use legacy way.
+            if (HttpInListener.IsNet7OrGreater)
+            {
+                builder.AddSource(HttpInListener.FrameworkActivitySourceName);
```
CodeBlanch, line 61: There is a subtle potential bug/issue with how this is being done which may or may not be a concern. If we consider what [AspNetCore is doing](https://github.com/dotnet/aspnetcore/blob/main/src/Hosting/Hosting/src/WebHostBuilder.cs#L291)...

```csharp
services.TryAddSingleton(sp => new ActivitySource("Microsoft.AspNetCore"));
```

...that is not a fixed/set in stone mechanism. It is a _try_. Users may register their own activity source before that one, or they may change it after.

Maybe we shouldn't access by name at all and instead do something like...

```csharp
var activitySource = _applicationServices.GetRequiredService<ActivitySource>();
```

...to access the runtime registered `ActivitySource` which is what AspNetCore itself is doing: https://github.com/dotnet/aspnetcore/blob/0e5cbf447d329a1e7d69932c3decd1c70a00fbba/src/Hosting/Hosting/src/Internal/WebHost.cs#L152

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/BasicTests.cs
alanwest: Can you remind me of the significance of that check? 😆

Pretty sure I understood it at one point, but I'm staring at it and it's purpose is no longer obvious. Is the problem that IsRemote is not properly handled from a parent context provided when ASP.NET Core creates its activity?

Also, it looks like the test app does not return the value of IsRemote https://github.com/open-telemetry/opentelemetry-dotnet/blob/3a90287360604762db9ac077223fe3b45c266650/test/TestApp.AspNetCore.7.0/Controllers/ChildActivityController.cs#L34-L36. If we're setting it shouldn't we assert on it?

src/OpenTelemetry.Instrumentation.AspNetCore/TracerProviderBuilderExtensions.cs
```diff
@@ -53,8 +53,19 @@ public static class TracerProviderBuilderExtensions
             this TracerProviderBuilder builder,
             AspNetCoreInstrumentation instrumentation)
         {
-            builder.AddSource(HttpInListener.ActivitySourceName);
-            builder.AddLegacySource(HttpInListener.ActivityOperationName); // for the activities created by AspNetCore
+            // For .NET7.0 onwards activity will be created using activitySource.
+            // https://github.com/dotnet/aspnetcore/blob/main/src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs#L327
```
alanwest, line 57: Same permalink thingy here.

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/BasicTests.cs
vishweshbankwar: Slight correction: I added `isRemote = true` due to following reason:

The  `isRemote = true` is needed as that impacts whether we create the activity or not in case when the sampler decides to drop the activity (we have that case in this test). The change is explained in #3315. So I think any tests we do where are comparing contexts of the extracted parent we should add this. (`isRemote` should always be true for parent context of activity created by ASP.NET Core).

**Explanation of check**: The check is used in case when a custom propagator is used by user e.g. B3. In this case the parent extracted by ASP.NET Core library will not have correct value as it only understands TraceContextPropagator. If that check determines the parent context extracted by custom propagator is different than the one extracted by library, it will create a new activity with correct parent context set. (In this case the sampler will re-run again with correct parent context passed in as one of the [parameters](https://github.com/open-telemetry/opentelemetry-dotnet/blob/f471a9f197d797015123fe95d3e12b6abf8e1f5f/src/OpenTelemetry/Trace/TracerProviderSdk.cs#L371). This is done so that the correct context is propagated to next hop.

Let me know if that makes sense :)


For the assert part, I don't think we can check that as we are creating a child activity of activity created by framework (which will have `HasRemoteParent` set as true). Not sure why we create new activity there and not do Activity.Current. Will have to check.

CodeBlanch (approved):
LGTM

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/BasicTests.cs
```diff
@@ -575,8 +575,13 @@ private static void WaitForActivityExport(List<Activity> exportedItems, int coun
         private static void ValidateAspNetCoreActivity(Activity activityToValidate, string expectedHttpPath)
         {
             Assert.Equal(ActivityKind.Server, activityToValidate.Kind);
+#if NET7_0_OR_GREATER
+            Assert.Equal(HttpInListener.FrameworkActivitySourceName, activityToValidate.Source.Name);
+            Assert.Empty(activityToValidate.Source.Version);
```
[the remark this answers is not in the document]
vishweshbankwar: That is a good point.  As per my understanding that is not in the plan right now (ASP.NET Core adding attributes). Opened #3445 to track this (Please review).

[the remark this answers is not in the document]
alanwest: Looks good, thanks for opening the issue.

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/BasicTests.cs
alanwest: > Let me know if that makes sense :)

Yes, thank you that's helpful for jogging my memory. Regarding the new Activity, I _think_ it might be because if ASP.NET Core also extracted a valid context that is different from the one we extract we don't want to clobber it. Though I don't really recall why that would be bad. In some scenarios it almost seems like it could be useful. Anyways... this is off topic. I might study this code more and consider adding a comment to remind my future self 😄.

alanwest (approved):

src/OpenTelemetry.Instrumentation.AspNetCore/TracerProviderBuilderExtensions.cs
```diff
@@ -51,22 +53,45 @@ public static class TracerProviderBuilderExtensions
...
+            IServiceProvider serviceProvider = null)
         {
-            builder.AddSource(HttpInListener.ActivitySourceName);
-            builder.AddLegacySource(HttpInListener.ActivityOperationName); // for the activities created by AspNetCore
+            // For .NET7.0 onwards activity will be created using activitySource.
+            // https://github.com/dotnet/aspnetcore/blob/bf3352f2422bf16fa3ca49021f0e31961ce525eb/src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs#L327
+            // For .NET6.0 and below, we will continue to use legacy way.
+            if (HttpInListener.IsNet7OrGreater)
+            {
+                var activitySourceService = serviceProvider?.GetService<ActivitySource>();
```
cijothomas, line 64: Not sure if this is safe. Anyone can add `ActivitySource` to DI, and how do we know if this is the one added by asp.net core?

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs
```diff
@@ -135,8 +157,11 @@ public override void OnStartActivity(Activity activity, object payload)
                     return;
                 }
 
-                ActivityInstrumentationHelper.SetActivitySourceProperty(activity, ActivitySource);
-                ActivityInstrumentationHelper.SetKindProperty(activity, ActivityKind.Server);
+                if (!IsNet7OrGreater)
```
cijothomas, line 160: why we have used pre-processor directives (NET7_0_OR_GREATER) in some places and the static variable in other place?

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs
```diff
@@ -35,6 +35,10 @@ namespace OpenTelemetry.Instrumentation.AspNetCore.Implementation
     internal class HttpInListener : ListenerHandler
     {
         internal const string ActivityOperationName = "Microsoft.AspNetCore.Hosting.HttpRequestIn";
+        internal static readonly bool IsNet7OrGreater;
+
+        // https://github.com/dotnet/aspnetcore/blob/bf3352f2422bf16fa3ca49021f0e31961ce525eb/src/Hosting/Hosting/src/WebHostBuilder.cs#L291
+        internal static readonly string FrameworkActivitySourceName = "Microsoft.AspNetCore";
```
cijothomas, line 41:
```suggestion
        internal static readonly string AspNetCoreActivitySourceName = "Microsoft.AspNetCore";
```

src/OpenTelemetry.Instrumentation.AspNetCore/TracerProviderBuilderExtensions.cs
vishweshbankwar: This is how library passes the activitysource as well to diagnostics module. https://github.com/dotnet/aspnetcore/blob/0e5cbf447d329a1e7d69932c3decd1c70a00fbba/src/Hosting/Hosting/src/Internal/WebHost.cs#L152

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs
```diff
@@ -35,6 +35,12 @@ namespace OpenTelemetry.Instrumentation.AspNetCore.Implementation
     internal class HttpInListener : ListenerHandler
     {
         internal const string ActivityOperationName = "Microsoft.AspNetCore.Hosting.HttpRequestIn";
+        internal static readonly bool IsNet7OrGreater;
```
cijothomas, line 38: unused?

src/OpenTelemetry.Instrumentation.AspNetCore/TracerProviderBuilderExtensions.cs
```diff
@@ -51,22 +55,44 @@ public static class TracerProviderBuilderExtensions
...
-            AspNetCoreInstrumentation instrumentation)
+            AspNetCoreInstrumentation instrumentation,
+            IServiceProvider serviceProvider = null)
         {
+            // For .NET7.0 onwards activity will be created using activitySource.
+            // https://github.com/dotnet/aspnetcore/blob/bf3352f2422bf16fa3ca49021f0e31961ce525eb/src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs#L327
+            // For .NET6.0 and below, we will continue to use legacy way.
+#if NET7_0_OR_GREATER
+            // TODO: Check with .NET team to see if this can be prevented
+            // as this allows user to override the ActivitySource.
```
cijothomas, line 66: would be good to write a unittest which simulates this.

cijothomas (approved):

src/OpenTelemetry.Instrumentation.AspNetCore/TracerProviderBuilderExtensions.cs
vishweshbankwar: Added

Issue #3373, opened 2022-06-16
Update instrumentation libraries with latest semantic convention spec


cijothomas:
Most of the instrumentation libraries were written when semantic conventions were in early stages, and a lot changes have occurred since then. Opening an issue to track updating instrumentation libraries in this repo (httpclients,aspnetcore,sqlclient,grpc), to match the latest spec.

An open question, which is to be discussed: How much control these instrumentations offer to user for turning on/off individual tags? Do we want a turn on/off flag for each tag? Like https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Instrumentation.Http/HttpClientInstrumentationOptions.cs#L34.

or a set of tags? What is the criteria to make that decision (perf/security/privacy/spec-recommendedvsrequiresvsoption)

cijothomas:
https://docs.microsoft.com/en-us/aspnet/core/fundamentals/http-logging/?view=aspnetcore-6.0 Might be able to borrow some ideas from this, regd. what tags gets populated / not.

vishweshbankwar:
closing this as the changes are tracked with https://github.com/open-telemetry/opentelemetry-dotnet/issues/4484

PR #3392, opened 2022-06-20
Asp.Net Core trace instrumentation to populate http schema tag

cijothomas:
Part of https://github.com/open-telemetry/opentelemetry-dotnet/issues/3373

Asp.Net Core was missing http.schema attribute, which is listed as a Recommended tag in the sem. conventions.
(the metrics instrumentation was already populating this)


src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
```diff
@@ -9,6 +9,9 @@
 * Tracing instrumentation to populate 'http.flavor' tag.
   ([3372](https://github.com/open-telemetry/opentelemetry-dotnet/pull/3372))
 
+* Tracing instrumentation to populate 'http.schema' tag.
```
reyang, line 12: minor: it seems the convention is to avoid an empty line between bullet points.

reyang (approved):

PR #3407, opened 2022-06-24
HttpClient: Invoke Enrich when SocketException = HostNotFound

### LINKED ISSUE #3406: Enrich not invoked in HttpClient instrumentation when HostNotFound
While implementing an enrich function, I noticed that HTTP Socket Exceptions with the error code `HostNotFound` do not trigger the enrich function on exception because of [these lines in `HttpHandlerDiagnosticListener`](https://github.com/open-telemetry/opentelemetry-dotnet/blob/2a97920ff0a603837a1121204824955bba739c57/src/OpenTelemetry.Instrumentation.Http/Implementation/HttpHandlerDiagnosticListener.cs#L218-L226).

This prevents additional information from being captured in the activity when the exception is a `HostNotFound` Socket Exception. 

Can someone shed some light on this particular behaviour? I could not find anything in the documentation.

Thank you.

_Originally posted by @tl-amanda-oh in https://github.com/open-telemetry/opentelemetry-dotnet/discussions/3208_

alanwest:
Fixes #3406.


src/OpenTelemetry.Instrumentation.Http/Implementation/HttpHandlerDiagnosticListener.cs
```diff
@@ -200,11 +200,10 @@ public override void OnException(Activity activity, object payload)
...
                                 activity.SetStatus(Status.Error.WithDescription(exc.Message));
-                                return;
+                                break;
                         }
                     }
-
-                    if (exc.InnerException != null)
+                    else if (exc.InnerException != null)
                     {
                         activity.SetStatus(Status.Error.WithDescription(exc.Message));
```
alanwest, line 208: Not related to this PR really.. but this is suspect. Did we mean to set the status description with the InnerException message? _And_ do we really want to be grabbing the innermost exception here?

cijothomas: good catch! I think its the outermost we should use for status message. we can address this separately.

cijothomas (approved):
LGTM with a changelog entry as this is user facing change

src/OpenTelemetry.Instrumentation.Http/Implementation/HttpHandlerDiagnosticListener.cs
```diff
@@ -200,11 +200,10 @@ public override void OnException(Activity activity, object payload)
                         {
                             case SocketError.HostNotFound:
                                 activity.SetStatus(Status.Error.WithDescription(exc.Message));
-                                return;
+                                break;
                         }
                     }
-
-                    if (exc.InnerException != null)
+                    else if (exc.InnerException != null)
```
reyang, line 206: The flow seems a bit weird. Consider moving the `exc.InnerException is SocketException exception` check inside `exc.InnerException != null`?

reyang (approved):

src/OpenTelemetry.Instrumentation.Http/Implementation/HttpHandlerDiagnosticListener.cs
cijothomas: yup. 
Also there is no need of a Switch case... as there is a single check..

alanwest: So I moved the check, because I agree the `else if` was wonky. Though, yes as the code currently stands, there's zero need to explicitly check for the SocketException. It seems there was once some thought that we'd set the status description differently based on the type of exception, but clearly that was not actually happening in practice. So I guess I don't know what the original intent was here 😆.

I'll see if the history can shed any light on this...

alanwest: Ah ha ok, all of this dates back to the time that Span.Status was more than just `OK`, `ERROR`, or `UNSET`

https://github.com/open-telemetry/opentelemetry-dotnet/blob/1ed4a289276d2bcd8f19d540faa1fc3e845280d9/src/OpenTelemetry.Instrumentation.Dependencies/Implementation/HttpHandlerDiagnosticListener.cs#L161-L177

I simplified it.

src/OpenTelemetry.Instrumentation.Http/Implementation/HttpHandlerDiagnosticListener.cs
```diff
@@ -194,20 +194,7 @@ public override void OnException(Activity activity, object payload)
...
-                                activity.SetStatus(Status.Error.WithDescription(exc.Message));
-                                return;
-                        }
-                    }
-
-                    if (exc.InnerException != null)
-                    {
-                        activity.SetStatus(Status.Error.WithDescription(exc.Message));
-                    }
+                    activity.SetStatus(Status.Error.WithDescription(exc.Message));
```
alanwest, line 197: I guess this kind of broadens the scope where we record an error status + description to `HttpRequestException` where `InnerException` is null... seems kinda odd we'd limit it in this way.

cijothomas (approved):

PR #3411, opened 2022-06-27
Rename ConfigureResource -> ConfigureResourceBuilder

alanwest:
Staging this up based on open question here https://github.com/open-telemetry/opentelemetry-dotnet/pull/3307#discussion_r907543231.

### DIFF
--- a/examples/Console/TestConsoleExporter.cs
+++ b/examples/Console/TestConsoleExporter.cs
@@ -39,7 +39,7 @@ namespace Examples.Console
             // and use Console exporter.
             using var tracerProvider = Sdk.CreateTracerProviderBuilder()
                     .AddSource("Samples.SampleClient", "Samples.SampleServer")
-                    .SetResourceBuilder(ResourceBuilder.CreateDefault().AddService("console-test"))
+                    .ConfigureResource(res => res.AddService("console-test"))
                     .AddProcessor(new MyProcessor()) // This must be added before ConsoleExporter
                     .AddConsoleExporter()
                     .Build();
--- a/examples/Console/TestConsoleExporter.cs
+++ b/examples/Console/TestConsoleExporter.cs
@@ -39,7 +39,7 @@ namespace Examples.Console
             // and use Console exporter.
             using var tracerProvider = Sdk.CreateTracerProviderBuilder()
                     .AddSource("Samples.SampleClient", "Samples.SampleServer")
-                    .ConfigureResource(res => res.AddService("console-test"))
+                    .ConfigureResourceBuilder(res => res.AddService("console-test"))
                     .AddProcessor(new MyProcessor()) // This must be added before ConsoleExporter
                     .AddConsoleExporter()
                     .Build();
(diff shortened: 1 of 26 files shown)


src/OpenTelemetry/CHANGELOG.md
```diff
@@ -25,6 +25,11 @@
 * Handle possible exception when initializing the default service name.
   ([#3405](https://github.com/open-telemetry/opentelemetry-dotnet/pull/3405))
 
+* Add `ConfigureResourceBuilder` which can replace SetResourceBuilder more
+  succinctly in most cases and has greater flexibility (applies to
+  TracerProvicerBuilder, MeterProviderBuilder, OpenTelemetryLoggingOptions).
```
cijothomas, line 30:
```suggestion
  TracerProviderBuilder, MeterProviderBuilder, OpenTelemetryLoggingOptions).
```

Rs4178:
> Staging this up based on open question here https://github.com/open-telemetry/opentelemetry-dotnet/pull/3307#discussion_r907543231.

cijothomas (approved):
I like this name better. Lets wait couple days to give more people to share feedbacks before merge.

examples/Console/TestConsoleExporter.cs
```diff
@@ -39,7 +39,7 @@ private static object RunWithActivitySource()
             // and use Console exporter.
             using var tracerProvider = Sdk.CreateTracerProviderBuilder()
                     .AddSource("Samples.SampleClient", "Samples.SampleServer")
-                    .ConfigureResource(res => res.AddService("console-test"))
+                    .ConfigureResourceBuilder(res => res.AddService("console-test"))
```
reyang, line 42: It seems the current code are bit random on the local variable naming.
Maybe use `resourceBuilder` or `resBldr` consistently across the repo?

reyang (approved):

Oberon00:
I liked the old/current name a bit more. Just as in ASP.NET one uses ConfigureServices, not ConfigureServiceCollection. I could argue, I don't want to configure the builder, but use the builder to configure the resource(s) that will be created. But this is not a strong opinion.

alanwest:
> I liked the old/current name a bit more. Just as in ASP.NET one uses ConfigureServices, not ConfigureServiceCollection. Or this method here, another example where one uses a Builder to configure something but which doesn't have Builder in the name: https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.testing.webapplicationfactory-1.configurewebhost?view=aspnetcore-6.0 I could argue, I don't want to configure the builder, but use the builder to configure the resource(s) that will be created. But I do not have a very strong opinion.

I agree, these are good examples and a good argument for keeping the name as is: `ConfigureResource`.

cijothomas:
> I liked the old/current name a bit more. Just as in ASP.NET one uses ConfigureServices, not ConfigureServiceCollection. Or this method here, another example where one uses a Builder to configure something but which doesn't have Builder in the name: https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.testing.webapplicationfactory-1.configurewebhost?view=aspnetcore-6.0 I could argue, I don't want to configure the builder, but use the builder to configure the resource(s) that will be created. But I do not have a very strong opinion.

Thanks for explanation. It looks reasonable to me to keep existing name, and abort this PR.

PR #3430, opened 2022-07-06
[Prometheus] Split up projects based on hosting mechanism.

Yun-Ting:
Fixes https://github.com/open-telemetry/opentelemetry-dotnet/issues/3057.

## Changes

Addressing some comments in https://github.com/open-telemetry/opentelemetry-dotnet/pull/3375.
There are some big commits got merged over the course of the development of PR3375 and I figured starting from the more current main line is easier than resolving merge conflicts.

The updated project setup is as follows:
- `OpenTelemetry.Exporter.Prometheus.Shared` contains the PrometheusExporter, PrometheusExporterOptions and serialization methods for all the shared classes to both AspNetCore and HttpListener projects. This project is *not* expected to be consumed directly by the end users. It serves as a clean way to group shared code for the below 2 projects for semantic reasons. As a result, I've deleted the Implementation folder that was in `OpenTelemetry.Exporter.Prometheus` project, and moved out the shared files which was previously in the folder to be the immediate children of the project.

- `OpenTelemetry.Exporter.Prometheus.HttpListener` contains the PrometheusHttpListener and its options class. It aims to provide end users a quick setup experience for using OTel metrics with Prometheus Exporter locally.

- `OpenTelemetry.Exporter.Prometheus.AspNetCore` contains the middleware and its extensions for registering. This is for users trying to run PrometheusExporter in production environment.

* [x] Design discussion issue #


src/OpenTelemetry.Exporter.Prometheus.HttpListener/OpenTelemetry.Exporter.Prometheus.HttpListener.csproj
```diff
@@ -0,0 +1,39 @@
...
+    <MinVerTagPrefix>core-</MinVerTagPrefix>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <NoWarn>$(NoWarn),1591</NoWarn>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <NoWarn>$(NoWarn),1702</NoWarn>
+  </PropertyGroup>
```
CodeBlanch, line 17: I think separator here is `;` (but not totally sure). You can also combine these two blocks into one like...

```xml
  <PropertyGroup>
    <NoWarn>$(NoWarn);1702;1591</NoWarn>
  </PropertyGroup>
```

Also might be nice to add a comment about why these particular warnings are being suppressed.

src/OpenTelemetry.Exporter.Prometheus.HttpListener/OpenTelemetry.Exporter.Prometheus.HttpListener.csproj
```diff
@@ -0,0 +1,39 @@
...
+  <!--Do not run ApiCompat as this package has never released a stable version.
+  Remove this property once we have released a stable version.-->
+  <PropertyGroup>
+    <RunApiCompat>false</RunApiCompat>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="$(RepoRoot)\src\OpenTelemetry\OpenTelemetry.csproj" />
+  </ItemGroup>
+  <ItemGroup>
```
CodeBlanch, line 28: nit: Put an empty line in between these blocks.

src/OpenTelemetry.Exporter.Prometheus.AspNetCore/OpenTelemetry.Exporter.Prometheus.AspNetCore.csproj
```diff
@@ -0,0 +1,41 @@
...
+  <PropertyGroup>
+    <!-- OmniSharp/VS Code requires TargetFrameworks to be in descending order for IntelliSense and analysis. -->
+    <TargetFrameworks>netcoreapp3.1</TargetFrameworks>
+    <Description>Prometheus exporter for AspNetCore for OpenTelemetry .NET</Description>
+    <PackageTags>$(PackageTags);prometheus;metrics</PackageTags>
+    <MinVerTagPrefix>core-</MinVerTagPrefix>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <NoWarn>$(NoWarn),1591</NoWarn>
```
CodeBlanch, line 12: I think separator here needs to be `;`.

src/OpenTelemetry.Exporter.Prometheus.AspNetCore/OpenTelemetry.Exporter.Prometheus.AspNetCore.csproj
```diff
@@ -0,0 +1,41 @@
...
+  </PropertyGroup>
+
+  <!--Do not run ApiCompat as this package has never released a stable version.
+  Remove this property once we have released a stable version.-->
+  <PropertyGroup>
+    <RunApiCompat>false</RunApiCompat>
+  </PropertyGroup>
+  <ItemGroup>
```
CodeBlanch, line 20: nit: Add an empty line in between these.

src/OpenTelemetry.Exporter.Prometheus.HttpListener/PrometheusExporterHttpListenerMeterProviderBuilderExtensions.cs
```diff
@@ -0,0 +1,84 @@
...
+using System;
+using OpenTelemetry.Exporter;
+using OpenTelemetry.Exporter.Prometheus.HttpListener;
+using OpenTelemetry.Internal;
+
+namespace OpenTelemetry.Metrics
+{
+    public static class PrometheusExporterHttpListenerMeterProviderBuilderExtensions
+    {
+        public static MeterProviderBuilder AddPrometheusHttpListener(
```
CodeBlanch, line 26: nit: Add XML comments on the public stuff.

src/OpenTelemetry.Exporter.Prometheus.HttpListener/PrometheusHttpListener.cs
```diff
@@ -18,37 +18,37 @@
...
 using OpenTelemetry.Internal;
 
-namespace OpenTelemetry.Exporter.Prometheus
+namespace OpenTelemetry.Exporter.Prometheus.HttpListener
 {
-    /// <summary>
-    /// An HTTP listener used to expose Prometheus metrics.
-    /// </summary>
-    internal sealed class PrometheusExporterHttpServer : IDisposable
+    public sealed class PrometheusHttpListener : IDisposable
```
CodeBlanch, line 26: Does this need to be public?

examples/Console/TestPrometheusExporter.cs
```diff
@@ -48,15 +48,17 @@ internal static object Run(int port)
...
+            .AddPrometheusHttpListener(
+                exporterOptions =>
             {
-                options.StartHttpListener = true;
-                options.HttpListenerPrefixes = new string[] { $"http://localhost:{port}/" };
-                options.ScrapeResponseCacheDurationMilliseconds = 0;
-            })
+                exporterOptions.ScrapeResponseCacheDurationMilliseconds = 0;
+            },
+                configureListenerOptions: configureListenerOptions)
```
CodeBlanch, line 61: nit: It might read a bit better like...

```csharp
            .AddMeter(MyMeter.Name)
            .AddMeter(MyMeter2.Name)
            .AddPrometheusHttpListener(
                exporterOptions => exporterOptions.ScrapeResponseCacheDurationMilliseconds = 0,
                listenerOptions => listenerOptions.Prefixes = new string[] { $"http://localhost:{port}/" })
```

src/OpenTelemetry.Exporter.Prometheus.HttpListener/PrometheusHttpListener.cs
```diff
@@ -105,13 +105,23 @@ public void Stop()
...
+        {
+            if (!this.disposed)
             {
-                this.Stop();
-                this.httpListener.Close();
+                if (disposing && this.httpListener != null && this.httpListener.IsListening)
+                {
+                    this.Stop();
+                    this.httpListener.Close();
+                    this.Dispose();
```
CodeBlanch, line 121: * Not totally sure if `this.Stop()` should be part of `disposing` check or not. If you make a mistake and don't call Dispose and never Stop the listener does the process hang?
* Won't `this.Dispose()` call here cause an infinite loop?

src/OpenTelemetry.Exporter.Prometheus.Shared/AssemblyInfo.cs
```diff
@@ -0,0 +1,29 @@
...
+[assembly: InternalsVisibleTo("OpenTelemetry.Exporter.Prometheus.AspNetCore, PublicKey=002400000480000094000000060200000024000052534131000400000100010051c1562a090fb0c9f391012a32198b5e5d9a60e9b80fa2d7b434c9e5ccb7259bd606e66f9660676afc6692b8cdc6793d190904551d2103b7b22fa636dcbb8208839785ba402ea08fc00c8f1500ccef28bbf599aa64ffb1e1d5dc1bf3420a3777badfe697856e9d52070a50c3ea5821c80bef17ca3acffa28f89dd413f096f898")]
+[assembly: InternalsVisibleTo("OpenTelemetry.Exporter.Prometheus.HttpListener, PublicKey=002400000480000094000000060200000024000052534131000400000100010051c1562a090fb0c9f391012a32198b5e5d9a60e9b80fa2d7b434c9e5ccb7259bd606e66f9660676afc6692b8cdc6793d190904551d2103b7b22fa636dcbb8208839785ba402ea08fc00c8f1500ccef28bbf599aa64ffb1e1d5dc1bf3420a3777badfe697856e9d52070a50c3ea5821c80bef17ca3acffa28f89dd413f096f898")]
+#else
```
CodeBlanch, line 24: I don't think you need the `InternalsVisibleTo` for `AspNetCore` and `HttpListener` because they don't reference this, they link in the files.

src/OpenTelemetry.Exporter.Prometheus.Shared/OpenTelemetry.Exporter.Prometheus.Shared.csproj
```diff
@@ -3,15 +3,15 @@
   <PropertyGroup>
     <!-- OmniSharp/VS Code requires TargetFrameworks to be in descending order for IntelliSense and analysis. -->
     <TargetFrameworks>netcoreapp3.1;net462</TargetFrameworks>
-    <Description>Prometheus exporter for OpenTelemetry .NET</Description>
+    <TargetFrameworks Condition="$(OS) == 'Windows_NT'">$(TargetFrameworks);net462</TargetFrameworks>
+    <Description>Prometheus exporter shared code for both Prometheus exporter HttpListener and Prometheus Exporter AspNetCore</Description>
     <PackageTags>$(PackageTags);prometheus;metrics</PackageTags>
     <MinVerTagPrefix>core-</MinVerTagPrefix>
```
CodeBlanch, line 9: Maybe we should remove `MinVerTagPrefix` because we don't want to accidentally release this and that might be a good block for that.

src/OpenTelemetry.Exporter.Prometheus.Shared/PrometheusExporter.cs
```diff
@@ -28,32 +28,16 @@ public class PrometheusExporter : BaseExporter<Metric>, IPullMetricExporter
...
         private Func<int, bool> funcCollect;
         private Func<Batch<Metric>, ExportResult> funcExport;
-        private bool disposed;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="PrometheusExporter"/> class.
         /// </summary>
         /// <param name="options">Options for the exporter.</param>
-        public PrometheusExporter(PrometheusExporterOptions options)
+        internal PrometheusExporter(PrometheusExporterOptions options)
```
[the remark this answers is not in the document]
CodeBlanch: One extra thing I just thought of...

Anything in "Shared" project which is public and linked into some other project could become a problem. For example if users reference both projects (HttpListener & AspNetCore) and then they do `new PrometheusExporter(...)` the compiler won't know which one to use. Might not be something to worry about. We _could_ get around that by switching the namespaces based on compiler constants. Something like this...

```csharp
#if PROMETHEUS_ASPNETCORE
namespace OpenTelemetry.Exporter.Prometheus.AspNetCore
#elif PROMETHEUS_ASPNETCORE_HTTPLISTENER
namespace OpenTelemetry.Exporter.Prometheus.HttpListener
#else
namespace OpenTelemetry.Exporter.Prometheus.Shared
#endif
```

Then define `PROMETHEUS_ASPNETCORE` in the aspnetcore project csproj, `PROMETHEUS_ASPNETCORE_HTTPLISTENER` in the httplistener project csproj.

```xml
  <PropertyGroup>
    <DefineConstants>$(DefineConstants);PROMETHEUS_ASPNETCORE</DefineConstants>
  </PropertyGroup>
```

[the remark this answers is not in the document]
Yun-Ting: This is a very good point; I actually encountered something similar at first when I was trying to put tests for both HttpListener and AspNetCore in the same project. But I later figured out putting them in the same project may not be a common use case. :-)

src/OpenTelemetry.Exporter.Prometheus.HttpListener/OpenTelemetry.Exporter.Prometheus.HttpListener.csproj
```diff
@@ -0,0 +1,39 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <!-- OmniSharp/VS Code requires TargetFrameworks to be in descending order for IntelliSense and analysis. -->
+    <TargetFrameworks>net462</TargetFrameworks>
```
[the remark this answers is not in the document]
Yun-Ting: Thanks for the explanation; I've updated the TargetFrameworks here to include both net462 and netstandard2.0.

src/OpenTelemetry.Exporter.Prometheus.HttpListener/OpenTelemetry.Exporter.Prometheus.HttpListener.csproj
Yun-Ting: It turns out that there's really no good reason to suppress these warnings and I've removed the suppression and fixed them properly.

src/OpenTelemetry.Exporter.Prometheus.HttpListener/PrometheusHttpListener.cs
Yun-Ting: Not really. Extension methods should be the way to go. Fixed.

examples/Console/TestPrometheusExporter.cs
Yun-Ting: Fixed here and also in other places with similar patterns. Thanks!

src/OpenTelemetry.Exporter.Prometheus.HttpListener/PrometheusHttpListener.cs
```diff
@@ -105,13 +105,23 @@ public void Stop()
             }
         }
 
-        /// <inheritdoc/>
         public void Dispose()
         {
-            if (this.httpListener != null && this.httpListener.IsListening)
+            this.Dispose(true);
+        }
```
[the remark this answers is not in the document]
Yun-Ting: Removed the usage of disposal pattern as sealed class does not really need it.

src/OpenTelemetry.Exporter.Prometheus.Shared/AssemblyInfo.cs
Yun-Ting: Fixed.

src/OpenTelemetry.Exporter.Prometheus.Shared/PrometheusExporter.cs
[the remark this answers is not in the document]
Yun-Ting: Updated the ctor to be public.

src/OpenTelemetry.Exporter.Prometheus.HttpListener/PrometheusHttpListener.cs
Yun-Ting: I've moved out `this.Stop()` out of the if check and also removed this.Dipose() which is buggy.

Regarding this:
> If you make a mistake and don't call Dispose and never Stop the listener does the process hang?

-> I think the destructor will take care of it?

CodeBlanch: We would have to do something like this...

```csharp
        ~PrometheusHttpListener()
        {
            this.Stop();
        }
```

But that is kind of frowned upon.

I played with it a bit. Today when MeterProvider is disposed, exporter is disposed, http listener is disposed here:

https://github.com/open-telemetry/opentelemetry-dotnet/blob/be588bdcd3a3006a61f9f8f4c957ef4ba66f0a8a/src/OpenTelemetry.Exporter.Prometheus/PrometheusExporter.cs#L85

That is missing in the new flow. Actually, in the new flow there is no way to stop/dispose the listener at all because it never escapes the scope of `AddPrometheusHttpListener`.

I just pushed a change to the branch that ties the disposal of `PrometheusHttpListener` to `PrometheusExporter` which basically restores the old behavior. Let me know what you think about that.

src/OpenTelemetry.Exporter.Prometheus.Shared/PrometheusExporter.cs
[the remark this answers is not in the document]
CodeBlanch: @alanwest @cijothomas @utpilla @reyang 

Anyone have an opinion on what we should do here? To summarize the situation...

* OpenTelemetry.Exporter.Prometheus.Shared.PrometheusExporter is linked into OpenTelemetry.Exporter.Prometheus.AspNetCore & OpenTelemetry.Exporter.Prometheus.HttpListener as `public OpenTelemetry.Exporter.PrometheusExporter`.
* It is unlikely(?) users will ref both packages, but shipping like this could create some problems down the road because 2 assemblies are exposing the same public type in the same namespace.

3 potential solutions I can think of:

* Make _everything_ linked `internal`. Today `PrometheusExporter` is `public` so this could upset some users.
* Actually ship `OpenTelemetry.Exporter.Prometheus.Shared` as a package and just reference it normally. This means users have another assembly to deal with, may or may not be a concern.
* Use the namespace trick from above so that each linked copy ends up distinct in its final assembly.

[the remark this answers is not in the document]
cijothomas: Is there any need of exposing `PrometheusExporterOptions` at all? If only `ScrapeEndpointPath` is required, then we can make it separately in each of the Exporters.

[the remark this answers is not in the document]
cijothomas: I have used the difference ns trick in ApplicationInsights https://github.com/microsoft/ApplicationInsights-dotnet/blob/main/NETCORE/src/Shared/Extensions/ApplicationInsightsServiceOptions.cs#L1-L4.

There was some issue about this.. trying to recollect what that is. will post here, if i find the actual issue..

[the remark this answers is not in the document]
Yun-Ting: > Is there any need of exposing `PrometheusExporterOptions` at all? If only `ScrapeEndpointPath` is required, then we can make it separately in each of the Exporters.

I kinda like the idea of exposing PrometheusExporterOptions mainly because it provides a clear semantic distinction between listener and prometheusExporter. 
And since we anyway put the shared code in one project, referencing it should not be a big problem. From the consumer's perspective, I don't think keeping these 2 separate options would be too complicated, rather it provides semantic clarity.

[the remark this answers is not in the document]
reyang: Sharing my position - I think all shared files should only have `internal` / `private` classes, if the intention is to share the types, it should be a binary/package dependency (or text template files, which have their own set of problems).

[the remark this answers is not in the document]
reyang: > 3 potential solutions I can think of:
> 
> * Make _everything_ linked `internal`. Today `PrometheusExporter` is `public` so this could upset some users.
> * Actually ship `OpenTelemetry.Exporter.Prometheus.Shared` as a package and just reference it normally. This means users have another assembly to deal with, may or may not be a concern.
> * Use the namespace trick from above so that each linked copy ends up distinct in its final assembly.

@CodeBlanch I agree with your ordering. The 1st one sounds the best to me. The 2nd one doesn't sound as good, because `OpenTelemetry.Exporter.Prometheus.Shared` is like an overkill (it's not a super complex project, it makes the package dependency more complicated, it's just hard to justify why we need to have such one more package).

src/OpenTelemetry.Exporter.Prometheus.HttpListener/PrometheusHttpListener.cs
Yun-Ting: I'm not sure whether [the binding](https://github.com/open-telemetry/opentelemetry-dotnet/blob/17fa0b73bd4dd4950236255aad20dd8394ffa2e7/src/OpenTelemetry.Exporter.Prometheus.HttpListener/PrometheusExporterHttpListenerMeterProviderBuilderExtensions.cs#:~:text=exporter.OnDispose%20%3D%20()%20%3D%3E%20listener.Dispose()%3B) should be called out explicitly as I learned that:
> If disposable object x contains or “wraps” or “possesses” disposable object y, x’s Dispose method automatically calls y’s Dispose method—unless instructed otherwise.
image
From: [C# 5.0 in a Nutshell, 5th Edition](https://learning.oreilly.com/library/view/c-5-0-in/9781449334192/)
[Chapter 12. Disposal and Garbage Collection](https://learning.oreilly.com/library/view/c-5-0-in/9781449334192/ch12.html#:-:text=According%20to%20rule%203,otherwise%20in%0A%20%20%20%20%20%20the%20constructor.)

Other than that, 17fa0b73bd4dd4950236255aad20dd8394ffa2e7 looks good to me!
Thank you Blanch! :smiley:

docs/metrics/getting-started-prometheus-grafana/getting-started-prometheus-grafana.csproj
```diff
@@ -1,5 +1,8 @@
 <Project Sdk="Microsoft.NET.Sdk">
-  <ItemGroup>
-    <ProjectReference Include="$(RepoRoot)\src\OpenTelemetry.Exporter.Prometheus\OpenTelemetry.Exporter.Prometheus.csproj" />
-  </ItemGroup>
+<PropertyGroup>
+  <TargetFrameworks>netstandard2.0</TargetFrameworks>
+</PropertyGroup>
+<ItemGroup>
+  <ProjectReference Include="$(RepoRoot)\src\OpenTelemetry.Exporter.Prometheus.HttpListener\OpenTelemetry.Exporter.Prometheus.HttpListener.csproj" />
+</ItemGroup>
```
alanwest, line 7: I think some indentation was lost

```suggestion
  <PropertyGroup>
    <TargetFrameworks>netstandard2.0</TargetFrameworks>
  </PropertyGroup>
  <ItemGroup>
    <ProjectReference Include="$(RepoRoot)\src\OpenTelemetry.Exporter.Prometheus.HttpListener\OpenTelemetry.Exporter.Prometheus.HttpListener.csproj" />
  </ItemGroup>
```

src/OpenTelemetry.Exporter.Prometheus.AspNetCore/CHANGELOG.md
```diff
@@ -2,9 +2,11 @@
 
 ## Unreleased
 
-* Added `IEndpointRouteBuilder` extension methods to help with Prometheus
-  middleware configuration on ASP.NET Core
-  ([#3295](https://github.com/open-telemetry/opentelemetry-dotnet/pull/3295))
```
alanwest, line 7: Was the removal of this changelog entry intentional?

src/OpenTelemetry.Exporter.Prometheus.AspNetCore/OpenTelemetry.Exporter.Prometheus.AspNetCore.csproj
```diff
@@ -0,0 +1,42 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <!-- OmniSharp/VS Code requires TargetFrameworks to be in descending order for IntelliSense and analysis. -->
+    <TargetFrameworks>netcoreapp3.1</TargetFrameworks>
```
alanwest, line 5: Can probably deal with this later when we merge in the `net7.0` branch, but netcoreapp3.1 won't be supported.

src/OpenTelemetry.Exporter.Prometheus.AspNetCore/CHANGELOG.md
Yun-Ting: Good catch! I removed that entry for the listener project changelog and did a copy/paste of the changelog to this project. Fixed!

src/OpenTelemetry.Exporter.Prometheus.Shared/PrometheusExporter.cs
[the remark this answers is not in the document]
Yun-Ting: Based on the feedback - I've updated the `PrometheusExporter` to be internal but will keep this conversation open for future reference.

src/OpenTelemetry.Exporter.Prometheus.HttpListener/PrometheusHttpListener.cs
Yun-Ting: Chatted with Utkarsh offline and figured out that only .NET framework classes will do this auto-binding thing. We're all good now :-).

reyang (approved):
Overall direction looks good to me.
The configuration part (extension methods and option types) might need to be improved in a follow up PR - this seems a bit unnecessarily complicated for an end-user:

```csharp
.AddPrometheusHttpListener(
    exporterOptions => exporterOptions.ScrapeResponseCacheDurationMilliseconds = 0,
    listenerOptions => listenerOptions.Prefixes = new string[] { $"http://localhost:{port}/" })
```

alanwest (approved):
LGTM

test/Benchmarks/Benchmarks.csproj
```diff
@@ -26,9 +26,9 @@
     <ProjectReference Include="$(RepoRoot)\src\OpenTelemetry.Exporter.InMemory\OpenTelemetry.Exporter.InMemory.csproj" />
     <ProjectReference Include="$(RepoRoot)\src\OpenTelemetry.Exporter.Jaeger\OpenTelemetry.Exporter.Jaeger.csproj" Aliases="Jaeger" />
     <ProjectReference Include="$(RepoRoot)\src\OpenTelemetry.Exporter.OpenTelemetryProtocol\OpenTelemetry.Exporter.OpenTelemetryProtocol.csproj" Aliases="OpenTelemetryProtocol" />
-    <ProjectReference Include="$(RepoRoot)\src\OpenTelemetry.Exporter.Prometheus\OpenTelemetry.Exporter.Prometheus.csproj" Aliases="Prometheus" />
     <ProjectReference Include="$(RepoRoot)\src\OpenTelemetry.Exporter.Zipkin\OpenTelemetry.Exporter.Zipkin.csproj" Aliases="Zipkin" />
     <ProjectReference Include="$(RepoRoot)\src\OpenTelemetry.Instrumentation.AspNetCore\OpenTelemetry.Instrumentation.AspNetCore.csproj" />
     <ProjectReference Include="$(RepoRoot)\src\OpenTelemetry.Instrumentation.Http\OpenTelemetry.Instrumentation.Http.csproj" />
+    <ProjectReference Include="..\..\src\OpenTelemetry.Exporter.Prometheus.Shared\OpenTelemetry.Exporter.Prometheus.Shared.csproj" />
```
utpilla, line 32: nit: Use `$(RepoRoot)`.

docs/metrics/getting-started-prometheus-grafana/README.md
```diff
@@ -46,12 +46,13 @@ And replace the below line:
...
-.AddPrometheusExporter(options => { options.StartHttpListener = true; })
+.AddPrometheusHttpListener()
 ```
 
-With `AddPrometheusExporter()`, OpenTelemetry `PrometheusExporter` will export
+`PrometheusHttpListener` is a wrapper that contains `PrometheusExporter`.
+With `AddPrometheusHttpListener()`, OpenTelemetry `PrometheusExporter` will export
 data via the endpoint defined by
-[PrometheusExporterOptions.HttpListenerPrefixes](../../../src/OpenTelemetry.Exporter.Prometheus/README.md#httplistenerprefixes),
+[PrometheusExporterOptions.Prefixes](../../../src/OpenTelemetry.Exporter.Prometheus.HttpListener/README.md#prefixes),
```
utpilla, line 55: `PrometheusHttpListenerOptions.Prefixes`?

src/OpenTelemetry.Exporter.Prometheus.Shared/PrometheusExporter.cs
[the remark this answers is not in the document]
utpilla: I would also agree for shared packages to only contain `internal` things but did we decide to go that route?

We still have `PrometheusExporterOptions` class marked as `public` in the shared project.

There is also the `public` extension method `AddPrometheusExporter(Action<PrometheusExporterOptions>)` in the shared project. If this extension method is only used the AspNetCore project, wouldn't it better to move it there?

src/OpenTelemetry.Exporter.Prometheus.Shared/PrometheusExporterEventSource.cs
```diff
@@ -18,12 +18,14 @@
...
 
-namespace OpenTelemetry.Exporter.Prometheus
+namespace OpenTelemetry.Exporter.Prometheus.Shared
 {
     /// <summary>
     /// EventSource events emitted from the project.
     /// </summary>
     [EventSource(Name = "OpenTelemetry-Exporter-Prometheus")]
+
+    // TODO: make it internal and use Internal Visible to the httpListener
```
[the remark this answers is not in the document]
Yun-Ting: not anymore; removed

src/OpenTelemetry.Exporter.Prometheus.Shared/PrometheusExporter.cs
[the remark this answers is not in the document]
Yun-Ting: > We still have `PrometheusExporterOptions` class marked as `public` in the shared project.

Right, I've updated to use `$<DefinedConstants>` to get away with it.
 
> There is also the `public` extension method `AddPrometheusExporter(Action<PrometheusExporterOptions>)` in the shared project. If this extension method is only used the AspNetCore project, wouldn't it better to move it there?

Yes, `.AddPrometheusExporter()` is move away from the shared project and now resides in AspNetCore project.

src/OpenTelemetry.Exporter.Prometheus.Shared/PrometheusSerializerExt.cs
```diff
@@ -16,7 +16,13 @@
 
 using OpenTelemetry.Metrics;
 
+#if PROMETHEUS_ASPNETCORE
+namespace OpenTelemetry.Exporter.Prometheus.AspNetCore
```
reyang, line 20: Curious - why do we need to have different namespaces (for something internal)? I feel this is breaking the abstraction (e.g. the PrometheusSerializer should have no knowledge about the exproter/listener).

src/OpenTelemetry.Exporter.Prometheus.HttpListener/OpenTelemetry.Exporter.Prometheus.HttpListener.csproj
```diff
@@ -0,0 +1,36 @@
...
+  </PropertyGroup>
+
+  <!--Do not run ApiCompat as this package has never released a stable version.
+  Remove this property once we have released a stable version.-->
+  <PropertyGroup>
+    <RunApiCompat>false</RunApiCompat>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <DefineConstants>$(DefineConstants);PROMETHEUS_ASPNETCORE_HTTPLISTENER</DefineConstants>
```
utpilla, line 18: I think we don't need the term `ASPNETCORE` here.

src/OpenTelemetry.Exporter.Prometheus.Shared/README.md
```diff
@@ -11,8 +11,18 @@
 
 ### Step 1: Install Package
 
+Depending on the hosting mechanism you will be using:
```
cijothomas, line 14: why would shared project has a readme? we dont release it anyway...

test/OpenTelemetry.Exporter.Prometheus.HttpListener.Tests/OpenTelemetry.Exporter.Prometheus.HttpListener.Tests.csproj
```diff
@@ -0,0 +1,31 @@
...
+    <PackageReference Include="xunit.runner.visualstudio" Version="$(XUnitRunnerVisualStudioPkgVer)">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
+    </PackageReference>
+    <DotNetCliToolReference Include="dotnet-xunit" Version="$(DotNetXUnitCliVer)" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="$(RepoRoot)\src\OpenTelemetry.Exporter.InMemory\OpenTelemetry.Exporter.InMemory.csproj" />
+    <ProjectReference Include="$(RepoRoot)\src\OpenTelemetry.Extensions.Hosting\OpenTelemetry.Extensions.Hosting.csproj" />
```
cijothomas, line 23: does this need OTel.Extensions.Hosting?

cijothomas (approved):
left couple of nits.

src/OpenTelemetry.Exporter.Prometheus.AspNetCore/README.md
```diff
@@ -0,0 +1,105 @@
+# Prometheus Exporter AspNetCore for OpenTelemetry .NET
+
```
alanwest, line 2: Most of our projects have a bit of a description here about the package. Suggestion:

```suggestion

[![NuGet](https://img.shields.io/nuget/v/OpenTelemetry.Exporter.Prometheus.AspNetCore.svg)](https://www.nuget.org/packages/OpenTelemetry.Exporter.Prometheus.AspNetCore)
[![NuGet](https://img.shields.io/nuget/dt/OpenTelemetry.Exporter.Prometheus.AspNetCore.svg)](https://www.nuget.org/packages/OpenTelemetry.Exporter.Prometheus.AspNetCore)

An [OpenTelemetry Prometheus exporter](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk_exporters/prometheus.md) for configuring an ASP.NET Core application with an endpoint for Prometheus to scrape.

```

src/OpenTelemetry.Exporter.Prometheus.HttpListener/README.md
```diff
@@ -0,0 +1,65 @@
+# Prometheus Exporter HttpListener for OpenTelemetry .NET
+
```
alanwest, line 2:
```suggestion

[![NuGet](https://img.shields.io/nuget/v/OpenTelemetry.Exporter.Prometheus.HttpListener.svg)](https://www.nuget.org/packages/OpenTelemetry.Exporter.Prometheus.HttpListener)
[![NuGet](https://img.shields.io/nuget/dt/OpenTelemetry.Exporter.Prometheus.HttpListener.svg)](https://www.nuget.org/packages/OpenTelemetry.Exporter.Prometheus.HttpListener)

An [OpenTelemetry Prometheus exporter](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk_exporters/prometheus.md) that configures an [HttpListener](https://docs.microsoft.com/dotnet/api/system.net.httplistener) instance for Prometheus to scrape.

```

alanwest (approved):

test/OpenTelemetry.Exporter.Prometheus.HttpListener.Tests/OpenTelemetry.Exporter.Prometheus.HttpListener.Tests.csproj
Yun-Ting: nope and removed

src/OpenTelemetry.Exporter.Prometheus.Shared/README.md
Yun-Ting: removed

src/OpenTelemetry.Exporter.Prometheus.Shared/PrometheusSerializerExt.cs
Yun-Ting: removed

src/OpenTelemetry.Exporter.Prometheus.Shared/PrometheusExporter.cs
```diff
@@ -69,23 +60,22 @@ public PrometheusExporter(PrometheusExporterOptions options)
...
         public override ExportResult Export(in Batch<Metric> metrics)
         {
             return this.OnExport(metrics);
         }
 
+        /// <inheritdoc/>
         protected override void Dispose(bool disposing)
         {
-            if (!this.disposed)
+            if (disposing)
```
utpilla, line 76: I'm sorry I might have missed this so let me know if this has been discussed before. Why are we getting rid of check on `disposed` here?

src/OpenTelemetry.Exporter.Prometheus.Shared/.publicApi/net462/PublicAPI.Unshipped.txt
```diff
@@ -0,0 +1,6 @@
+OpenTelemetry.Exporter.Prometheus.Shared.PrometheusExporterOptions
```
utpilla, line 1: We don't need to run publicApi analyzer for the shared project. This might require adding some new condition in `Common.prod.props`. Currently, publicApi analyzer is run for every project under `src/`.

src/OpenTelemetry.Exporter.Prometheus.Shared/PrometheusExporterOptions.cs
```diff
@@ -15,61 +15,25 @@
...
 using System;
-using System.Collections.Generic;
 using OpenTelemetry.Internal;
 
-namespace OpenTelemetry.Exporter
+#if PROMETHEUS_ASPNETCORE
+namespace OpenTelemetry.Exporter.Prometheus.AspNetCore
+#elif PROMETHEUS_HTTPLISTENER
+namespace OpenTelemetry.Exporter.Prometheus.HttpListener
+#else
```
utpilla, line 24: This `#else` check is not required.

utpilla (approved):
LGTM. Left some minor suggestions.

src/OpenTelemetry.Exporter.Prometheus.Shared/PrometheusExporterEventSource.cs
```diff
@@ -18,12 +18,13 @@
...
 using OpenTelemetry.Internal;
 
-namespace OpenTelemetry.Exporter.Prometheus
+namespace OpenTelemetry.Exporter.Prometheus.Shared
 {
     /// <summary>
     /// EventSource events emitted from the project.
     /// </summary>
     [EventSource(Name = "OpenTelemetry-Exporter-Prometheus")]
+
```
CodeBlanch, line 27: nit: Remove this blank line

src/OpenTelemetry.Exporter.Prometheus.Shared/PrometheusExporterOptions.cs
CodeBlanch: If we remove it I don't think the shared project will build by itself.

src/OpenTelemetry.Exporter.Prometheus.AspNetCore/OpenTelemetry.Exporter.Prometheus.AspNetCore.csproj
```diff
@@ -0,0 +1,42 @@
...
+    <Description>Prometheus exporter for AspNetCore for OpenTelemetry .NET</Description>
```
CodeBlanch, line 6: nit: I feel like we could be a little more descriptive here. How about...

`AspNetCore middleware for hosting OpenTelemetry .NET Prometheus exporter`

src/OpenTelemetry.Exporter.Prometheus.AspNetCore/OpenTelemetry.Exporter.Prometheus.AspNetCore.csproj
```diff
@@ -0,0 +1,42 @@
...
+    <PackageTags>$(PackageTags);prometheus;metrics</PackageTags>
+    <MinVerTagPrefix>core-</MinVerTagPrefix>
+  </PropertyGroup>
+
+
```
CodeBlanch, line 11: nit: Extra blank line

src/OpenTelemetry.Exporter.Prometheus.Shared/OpenTelemetry.Exporter.Prometheus.Shared.csproj
```diff
@@ -3,13 +3,8 @@
   <PropertyGroup>
     <!-- OmniSharp/VS Code requires TargetFrameworks to be in descending order for IntelliSense and analysis. -->
     <TargetFrameworks>netcoreapp3.1;net462</TargetFrameworks>
-    <Description>Prometheus exporter for OpenTelemetry .NET</Description>
+    <Description>Prometheus exporter shared code for both Prometheus exporter HttpListener and Prometheus Exporter AspNetCore</Description>
```
CodeBlanch, line 6: nit: `Prometheus exporter AspNetCore` (lowercase 'e' to be consistent)

src/OpenTelemetry.Exporter.Prometheus.HttpListener/OpenTelemetry.Exporter.Prometheus.HttpListener.csproj
```diff
@@ -0,0 +1,36 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <!-- OmniSharp/VS Code requires TargetFrameworks to be in descending order for IntelliSense and analysis. -->
+    <TargetFrameworks>netstandard2.0;net462</TargetFrameworks>
+    <Description>Prometheus exporter HttpListener for OpenTelemetry .NET</Description>
```
CodeBlanch, line 6: nit: Maybe...

`Stand-alone HttpListener for hosting OpenTelemetry .NET Prometheus exporter`

🤷‍♀️

src/OpenTelemetry.Exporter.Prometheus.Shared/PrometheusExporter.cs
CodeBlanch: Not sure! @Yun-Ting Was this intentional? I might have done it, can't remember 😄 But probably safe to put back?

CodeBlanch (approved):
A few nits but LGTM

src/OpenTelemetry.Exporter.Prometheus.Shared/.publicApi/net462/PublicAPI.Unshipped.txt
Yun-Ting: https://github.com/open-telemetry/opentelemetry-dotnet/pull/3497

src/OpenTelemetry.Exporter.Prometheus.Shared/PrometheusExporter.cs
Yun-Ting: Adding the private bool back as `Dispose` could be called multiple times.

src/OpenTelemetry.Exporter.Prometheus.Shared/PrometheusExporter.cs
```diff
@@ -69,23 +61,25 @@ public PrometheusExporter(PrometheusExporterOptions options)
...
         {
             if (!this.disposed)
             {
                 if (disposing)
                 {
-                    this.metricsHttpServer?.Dispose();
+                    this.OnDispose?.Invoke();
                 }
-
-                this.disposed = true;
```
utpilla, line 88: Set `this.disposed` to `true`?

Yun-Ting: holy.... don't tell anyone :sweat_smile:




</body>
</html>

Kielek (approved):

src/OpenTelemetry.Exporter.Prometheus.Shared/PrometheusExporter.cs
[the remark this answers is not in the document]
Yun-Ting: Hey @alanwest @cijothomas @CodeBlanch @Kielek @reyang @utpilla:
I've tried out another projects layout given the suggestion:
https://github.com/open-telemetry/opentelemetry-dotnet/pull/3497#discussion_r931572797
with this commit: https://github.com/open-telemetry/opentelemetry-dotnet/pull/3497#discussion_r931572797

With this way, it cleans up the namespaces a little bit and works around the public api documentation issue as was mentioned here: https://github.com/open-telemetry/opentelemetry-dotnet/pull/3430#discussion_r931522363

Please let me know how you feel about this change! Thank you.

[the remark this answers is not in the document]
reyang: > Hey @alanwest @cijothomas @CodeBlanch @Kielek @reyang @utpilla: I've tried out another projects layout given the suggestion: [#3497 (comment)](https://github.com/open-telemetry/opentelemetry-dotnet/pull/3497#discussion_r931572797) with this commit: [#3497 (comment)](https://github.com/open-telemetry/opentelemetry-dotnet/pull/3497#discussion_r931572797)

With **this** commit https://github.com/open-telemetry/opentelemetry-dotnet/pull/3430/commits/3b562b97d3902e59acfc8886e69d977ae83e8c1c?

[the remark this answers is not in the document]
Yun-Ting: > > Hey @alanwest @cijothomas @CodeBlanch @Kielek @reyang @utpilla: I've tried out another projects layout given the suggestion: [#3497 (comment)](https://github.com/open-telemetry/opentelemetry-dotnet/pull/3497#discussion_r931572797) with this commit: [#3497 (comment)](https://github.com/open-telemetry/opentelemetry-dotnet/pull/3497#discussion_r931572797)
> 
> With **this** commit [3b562b9](https://github.com/open-telemetry/opentelemetry-dotnet/commit/3b562b97d3902e59acfc8886e69d977ae83e8c1c)?

yes, my bad.

src/OpenTelemetry.Exporter.Prometheus.Shared/PrometheusExporterEventSource.cs
```diff
@@ -0,0 +1,82 @@
...
+using System;
+using System.Diagnostics.Tracing;
+using OpenTelemetry.Internal;
+
+namespace OpenTelemetry.Exporter.Prometheus.Shared
+{
+    /// <summary>
+    /// EventSource events emitted from the project.
+    /// </summary>
+    [EventSource(Name = "OpenTelemetry-Exporter-Prometheus")]
```
cijothomas, line 26: may need to be changed to more specific

src/OpenTelemetry/AssemblyInfo.cs
```diff
@@ -18,9 +18,12 @@
 
 [assembly: InternalsVisibleTo("OpenTelemetry.Tests" + AssemblyInfo.PublicKey)]
 [assembly: InternalsVisibleTo("OpenTelemetry.Exporter.InMemory" + AssemblyInfo.PublicKey)]
-[assembly: InternalsVisibleTo("OpenTelemetry.Exporter.Prometheus" + AssemblyInfo.PublicKey)]
 [assembly: InternalsVisibleTo("OpenTelemetry.Exporter.Prometheus.Tests" + AssemblyInfo.PublicKey)]
 [assembly: InternalsVisibleTo("OpenTelemetry.Extensions.EventSource" + AssemblyInfo.PublicKey)]
+[assembly: InternalsVisibleTo("OpenTelemetry.Exporter.Prometheus.Shared" + AssemblyInfo.PublicKey)]
+[assembly: InternalsVisibleTo("OpenTelemetry.Exporter.Prometheus.AspNetCore" + AssemblyInfo.PublicKey)]
+[assembly: InternalsVisibleTo("OpenTelemetry.Exporter.Prometheus.AspNetCore.Tests" + AssemblyInfo.PublicKey)]
```
cijothomas, line 25: questionable why aspnetcore.tests need special access.

src/OpenTelemetry.Exporter.Prometheus.HttpListener/PrometheusExporter.cs
```diff
@@ -15,23 +15,24 @@
...
 
 using System;
-using OpenTelemetry.Exporter.Prometheus;
+#if PROMETHEUS_ASPNETCORE
+using OpenTelemetry.Exporter.Prometheus.AspNetCore;
+#endif
 using OpenTelemetry.Metrics;
 
-namespace OpenTelemetry.Exporter
+namespace OpenTelemetry.Exporter.Prometheus
```
cijothomas, line 23: not sure ns need a change here? The class name already have Prometheus in it..

src/OpenTelemetry.Exporter.Prometheus.Shared/PrometheusExporterOptions.cs
```diff
@@ -15,61 +15,25 @@
...
+namespace OpenTelemetry.Exporter.Prometheus.Shared
+#endif
 {
     /// <summary>
-    /// <see cref="PrometheusExporter"/> options.
+    /// Prometheus exporter options.
```
cijothomas, line 29: the ns already has prometheus in it, so class name again having prometheus is bit looong.

cijothomas:
Merging to make progress. Rest of the comments can be addressed as follow ups.

src/OpenTelemetry/AssemblyInfo.cs
Yun-Ting: It's because of this https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/test/OpenTelemetry.Exporter.Prometheus.AspNetCore.Tests/PrometheusCollectionManagerTests.cs#:~:text=if%20(!provider.TryFindExporter(out%20PrometheusExporter%20exporter))

src/OpenTelemetry.Exporter.Prometheus.Shared/PrometheusExporter.cs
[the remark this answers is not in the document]
Yun-Ting: Please disregard my previous comments about the 3b562b9 commit.

I've made a new PR to try out 2 project setup for prometheus here:
https://github.com/open-telemetry/opentelemetry-dotnet/pull/3503

Please take a look.

PR #3446, opened 2022-07-14
Fix exact match in wildcard

### LINKED ISSUE #3443: Generated WildcardRegex will match substrigs of non registered sources.
## Symptom

When adding one activity source with wildcard the [TracerProviderSdk](https://github.com/open-telemetry/opentelemetry-dotnet/blob/f471a9f197d797015123fe95d3e12b6abf8e1f5f/src/OpenTelemetry/Trace/TracerProviderSdk.cs#L219) will build a regex for all the registered activity sources.

Having the activities: `CustomActivity` and `CustomWildcardActivity.*` it will generate the following regex: 
```
^(?:CustomActivity)|(?:CustomWildcardActivity\..*.)$
```

The problem is that this regex will match also activities like `CustomActivity.AnyString` which was not registered and `CustomActivity` dose not contain wildcard.

Sub activities names like `CustomActivity.AnyString` should not be exported.

Sub activities `CustomActivity.Validation` are exported.

## Reproduce
- register the following sources
```csharp
service.AddOpenTelemetryTracing(tracer =>
{
    tracer
        .AddSource("CustomActivity")
        .AddSource("CustomWildcardActivity.*")
        .AddConsoleExporter();
});
```
- generate some activities with `CustomActivity.AnyString`

glucaci:
Fixes #3443.


src/OpenTelemetry/Internal/WildcardHelper.cs
```diff
@@ -41,7 +41,17 @@ public static Regex GetWildcardRegex(IEnumerable<string> patterns = default)
 
         var convertedPattern = string.Join(
             "|",
-            from p in patterns select "(?:" + Regex.Escape(p).Replace("\\*", ".*").Replace("\\?", ".") + ')');
+            from p in patterns select "(?:" + ReplacePatterns(p) + ')');
         return new Regex('^' + convertedPattern + '$', RegexOptions.Compiled | RegexOptions.IgnoreCase);
```
reyang, line 45: I think the fix should be this one line change:

```suggestion
        return new Regex("^(" + convertedPattern + ")$", RegexOptions.Compiled | RegexOptions.IgnoreCase);
```

reyang (changes requested):
https://github.com/open-telemetry/opentelemetry-dotnet/pull/3446/files#r921378089

src/OpenTelemetry/Internal/WildcardHelper.cs
reyang: Or maybe:

```suggestion
        return new Regex("^(?:" + convertedPattern + ")$", RegexOptions.Compiled | RegexOptions.IgnoreCase);
```

@Yun-Ting what's your suggestion?

Yun-Ting: Non-capturing group should be sufficient here.

test/OpenTelemetry.Tests/Internal/WildcardHelperTests.cs
```diff
@@ -0,0 +1,38 @@
...
+// limitations under the License.
+// </copyright>
+
+using Xunit;
+
+namespace OpenTelemetry.Internal.Tests
+{
+    public class WildcardHelperTests
+    {
+        [Theory]
```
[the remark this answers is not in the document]
glucaci: Sure, done

reyang (approved):
LGTM with a changelog entry.

Yun-Ting (approved):

glucaci:
Please let me know if the changelog entry is good enough to understand the issue which was fixed.

PR #3457, opened 2022-07-19
EnvironmentVariableHelper support for boolean

Kielek:
Fixes: N/A

## Changes

Support for boolean values in `EnvironmentVariableHelper`.
Boolean is special type of Enum.

The plan is to copy this file also to contrib repo. The same approach was used for https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/d9cc2811c384943dc7aa5004af7c89dd0ed6bebc/src/OpenTelemetry.Internal/Guard.cs

In AutoIntrumentation we need to configure some env. variables for MySql.Data instrumentation. https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/519

### DIFF
--- a/src/OpenTelemetry/Internal/EnvironmentVariableHelper.cs
+++ b/src/OpenTelemetry/Internal/EnvironmentVariableHelper.cs
@@ -111,5 +111,33 @@ namespace OpenTelemetry.Internal
 
             return true;
         }
+
+        /// <summary>
+        /// Reads an environment variable and parses it as a <see cref="bool" />.
+        /// </summary>
+        /// <param name="envVarKey">The name of the environment variable.</param>
+        /// <param name="result">The parsed value of the environment variable.</param>
+        /// <returns>
+        /// Returns <c>true</c> when a non-empty value was read; otherwise, <c>false</c>.
+        /// </returns>
+        /// <exception cref="FormatException">
+        /// Thrown when failed to parse the non-empty value.
+        /// </exception>
+        public static bool LoadBoolean(string envVarKey, out bool result)
+        {
+            result = default;
+
+            if (!LoadString(envVarKey, out string value))
+            {
+                return false;
+            }
+
+            if (!bool.TryParse(value, out result))
+            {
+                throw new FormatException($"{envVarKey} environment variable has an invalid value: '${value}'");
+            }
+
+            return true;
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


src/OpenTelemetry/Internal/EnvironmentVariableHelper.cs
```diff
@@ -111,5 +111,33 @@ public static bool LoadUri(string envVarKey, out Uri result)
...
+        public static bool LoadBoolean(string envVarKey, out bool result)
+        {
+            result = default;
+
+            if (!LoadString(envVarKey, out string value))
+            {
+                return false;
+            }
+
+            if (!bool.TryParse(value, out result))
```
pellared, line 135: I would rather handle only `true` and `false` - case sensitive - just like other enum values like `OTEL_EXPORTER_JAEGER_PROTOCOL`. I like consistency 😉

pellared:
> In AutoIntrumentation we need to configure some env. variables for MySql.Data instrumentation.

Do you mean that you want to add a possibility to configure [OpenTelemetry.Instrumentation.MySqlData](https://www.nuget.org/packages/OpenTelemetry.Instrumentation.MySqlData) using some env vars? I think it may be good first to create an issue for it and link it here.

It looks like there is no need for a boolean env var in this repo. Therefore the PR in this repo may be a YAGNI.

src/OpenTelemetry/Internal/EnvironmentVariableHelper.cs
pellared: The advantage of starting with a minimal subset of supported values is that we won't make a little breaking change if we later decide that this should be case-sensitive. Personally, I think having "strict" requirements for values may even occur more maintainable for the users, because it makes their configuration consistent.

PS. It is not a strong opinion 😉

Kielek:
> > In AutoIntrumentation we need to configure some env. variables for MySql.Data instrumentation.
> 
> Do you mean that you want to add a possibility to configure [OpenTelemetry.Instrumentation.MySqlData](https://www.nuget.org/packages/OpenTelemetry.Instrumentation.MySqlData) using some env vars? I think it may be good to create an issue for it and link it here.
> 
> Moreover, it looks like there is no need for a boolean env var in this repo. Therefore the PR in this repo may be a YAGNI.

https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/519
I think that it will be great to introduce similar variables also to other instrumentations. SqlClient instrumentation in this repository is potential candidate. As there is a lot of bools to configure, the method will be usefull.

pellared (approved):
One personal opinion https://github.com/open-telemetry/opentelemetry-dotnet/pull/3457#discussion_r924262146

Approving

cijothomas (approved):

Kielek:
@cijothomas, could you please merge?

PR #3475, opened 2022-07-22
Fix ASP.NET Core ExceptionFilter prevents recording exception

### LINKED ISSUE #2853: ExceptionFilter in ASP.NET Core prevents setting error status
# Bug Report

version that you are using (e.g. `OpenTelemetry 1.0.2`):

* OpenTelemetry.Instrumentation.AspNetCore 1.0.0-rc8

* netcoreapp3.1
* net5.0

## Symptom

Exception message is not set to activity status when there is an exception filter added to MVC pipeline, whether this filter handles exception or not.

Exception message appears as a value for activity tag otel.status_description

Tag otel.status_description is absent in activity tag object

## Reproduce

https://github.com/VitalyMargelov/OpenTelemetryBugRepro

## Additional Context

I thought of this behavior being a design feature, but the implementation of it makes me think it was not intended like that. The reason why presence of an ExceptionFilter prevents status description from being recorded is that it yields BeforeOnException event. This is event is caught [by DiagnosticSourceListener](https://github.com/open-telemetry/opentelemetry-dotnet/blob/3de96a1c16493116780777ae401c09533dde2345/src/OpenTelemetry/DiagnosticSourceInstrumentation/DiagnosticSourceListener.cs#L62), then dispatched [to HttpInListener](https://github.com/open-telemetry/opentelemetry-dotnet/blob/3de96a1c16493116780777ae401c09533dde2345/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs#L273) and then is put to stopExceptionFetcher, which can't find [Exception property there](https://docs.microsoft.com/en-gb/dotnet/api/microsoft.aspnetcore.mvc.diagnostics.beforeexceptionfilteronexception?view=aspnetcore-6.0) and basically corrupts the inner fetcher, so when it tries to handle this "*.UnhandledException" event it fails to retrieve exception from there to set its message to status description.

al-mac:
Fixes #2853

ASP.NET Core exception filters fire events before and after filter execution.

Before:
https://github.com/dotnet/aspnetcore/blob/36df8e5dc487e27cb85674ef3f144dca07f05c20/src/Mvc/Mvc.Core/src/MvcCoreDiagnosticListenerExtensions.cs#L396-L402

After:
https://github.com/dotnet/aspnetcore/blob/c85baf8db0c72ae8e68643029d514b2e737c9fae/src/Mvc/Mvc.Core/src/MvcCoreDiagnosticListenerExtensions.cs#L422-L434

The ASP.NET Core instrumentation's `OnException` event is invoked when these events fire. The payload of the event differs from that of the regular exception path and precludes the exception from being recorded (if enabled) and the span status from being set.

## Changes


src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs
```diff
@@ -45,6 +45,7 @@ internal class HttpInListener : ListenerHandler
         private readonly PropertyFetcher<HttpContext> startContextFetcher = new("HttpContext");
         private readonly PropertyFetcher<HttpContext> stopContextFetcher = new("HttpContext");
         private readonly PropertyFetcher<Exception> stopExceptionFetcher = new("Exception");
+        private readonly PropertyFetcher<Exception> stopExceptionFilterFetcher = new("ExceptionContext.Exception");
```
alanwest, line 48: Nice, thanks for discovering this @al-mac!

You got me curious, so I looked and found that we were indeed missing the events that are written before and after exception filters are invoked:

Before:
https://github.com/dotnet/aspnetcore/blob/36df8e5dc487e27cb85674ef3f144dca07f05c20/src/Mvc/Mvc.Core/src/MvcCoreDiagnosticListenerExtensions.cs#L396-L402

After:
https://github.com/dotnet/aspnetcore/blob/c85baf8db0c72ae8e68643029d514b2e737c9fae/src/Mvc/Mvc.Core/src/MvcCoreDiagnosticListenerExtensions.cs#L422-L434

I suspect that both of these events will cause our instrumentation to be invoked potentially duplicating the work we do in `OnException`.

Would you be able to write a test to validate this behavior?

We have some test apps for testing the ASP.NET Core instrumentation. I think the relevant place to configure an exception filter is here:

https://github.com/open-telemetry/opentelemetry-dotnet/blob/720cbdd2f8172569861e23783d8a06b6d5950f30/test/TestApp.AspNetCore.6.0/Startup.cs#L38

And here's an example of a test that exercises the test app:

https://github.com/open-telemetry/opentelemetry-dotnet/blob/720cbdd2f8172569861e23783d8a06b6d5950f30/test/OpenTelemetry.Instrumentation.AspNetCore.Tests/BasicTests.cs#L156-L166

We might need a way to conditionally set up an exception filter in the test app for the purposes of testing your work in this PR. If you need some help tooling this into our test infrastructure, I can help out more next week.

alanwest: > It's possible that something else is going on. Are you sure OnException is not also getting invoked with a payload with an Exception property?

My memory clicked in and I think I know what the problem is. I'm nearly certain that `OnException` is invoked both for Exception filters _and_ the regular exception path. `PropertyFetcher` effectively caches the payload type it receives. All subsequent calls to the `PropertyFetcher` expect that type.  The problem is the first time `OnException` is invoked it is invoked due to an Exception filter.

We're not interested in the exception from exception filters because of the concerns I raised before (i.e., there may be multiple filters and they may handle the exception).

@CodeBlanch ran into this same issue in #2698.

The alternative he considered was maintaining a map of inner `PropertyFetcher`s based on type, but that has downsides for performance. Though in this case, since I don't think we're interested in the Exception filter path, I think we could modify `PropertyFetcher` to ignore it. If the payload from the events raised by Exception filters do not have a property named `Exception` then I think we should prevent the PropertyFetcher from caching anything.

I still think the first step is to write a test that uses Exception filters since this is a use case that affects the behavior of the instrumentation.

al-mac: Hey, i'm working on the stuff you asked and came to the same conclusion! You hinted that the exception should be handled regardless of the `ExceptionFilter` being there or not. I've investigated it and found that the reason for this behaviour was the `innerFetcher` was being populated with the generic one (The one that always returns `null` on any fetch) before the actual object that has the exception property was handled in the pipeline. But when it's their turn to be handled, the `innerFetcher` is not `null` anymore, so the correct `TypedPropertyFetch<,>` is never being constructed.

Another way to fix it is to change make the `CreateFetcherFromProperty` method return `null` when this happens, and to return `default` in the `TryFetch` method if the innerFetcher is still `null` after trying to create it. I still didn't validate if this will make other parts of the code break, but I'm working on it right now.

al-mac: Hi! I've changed the code to reflect what I described above and wrote a few tests to validate this behaviour.

src/OpenTelemetry/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+* Fix issue where when an application has an ExceptionFilter, the exception data
+  wouldn't be collected.
+  ([#3475](https://github.com/open-telemetry/opentelemetry-dotnet/pull/3475))
```
alanwest, line 7: This should be moved to the changelog for the aspnetcore instrumentation https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/ExceptionFilterTests.cs
```diff
@@ -0,0 +1,185 @@
...
+        private readonly WebApplicationFactory<Startup> factory;
+        private TracerProvider tracerProvider = null;
+
+        public ExceptionFilterTests(WebApplicationFactory<Startup> factory)
+        {
+            this.factory = factory;
+        }
+
+        [Fact]
+        public async Task ShouldExportExceptionWithNoExceptionFilters()
```
alanwest, line 60: Not sure we need this test. We do have an existing test for the "normal" exception scenario here https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/test/OpenTelemetry.Instrumentation.AspNetCore.Tests/IncomingRequestsCollectionsIsAccordingToTheSpecTests.cs#L51-L52.

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/ExceptionFilterTests.cs
```diff
@@ -0,0 +1,185 @@
...
+
+                WaitForActivityExport(exportedItems, 1);
+            }
+
+            // Assert
+            AssertException(exportedItems);
+        }
+
+        [Fact]
+        public async Task ShouldExportActivityWithOneExceptionFilter()
```
alanwest, line 81: Could move this test to `BasicTests.cs` to avoid have to maintain a copy of the helper methods `ValidateAspNetCoreActivity` and `WaitForActivityExport`.

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/ExceptionFilterTests.cs
```diff
@@ -0,0 +1,185 @@
...
+                .Build();
+        }
+
+        private void AssertException(List<Activity> exportedItems)
+        {
+            Assert.Single(exportedItems);
+            var activity = exportedItems[0];
+
+            var exMessage = "something's wrong!";
+            Assert.Equal("System.Exception", activity.Events.First().Tags.FirstOrDefault(t => t.Key == SemanticConventions.AttributeExceptionType).Value);
```
alanwest, line 176: Adding an `Assert.Single(activity.Events)` ensures we have get a single Event and not doing additional work for repeat invocations of the `OnException` event.

src/OpenTelemetry/DiagnosticSourceInstrumentation/PropertyFetcher.cs
```diff
@@ -76,6 +76,12 @@ public bool TryFetch(object obj, out T value, bool skipObjNullCheck = false)
                 this.innerFetcher = PropertyFetch.Create(obj.GetType().GetTypeInfo(), this.propertyName);
             }
 
+            if (this.innerFetcher == null)
+            {
+                value = default;
+                return false;
+            }
```
alanwest, line 83: My quick implementation in #3479 differs a bit from yours, but that's ok. I'm going to close #3479.

Though would you mind adding the following test to your PR? https://github.com/open-telemetry/opentelemetry-dotnet/pull/3479/files#diff-35b3db9f257a0d783ab58f02a27c135395048aa3acd17cc54cbb47436370ede4R70-R79

I think it's good to have a test that just validates the underlying changes to `PropertyFetcher`.

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/ExceptionFilterTests.cs
al-mac: I removed the ExceptionFilterTests.cs from the project and moved both of them to BasicTests to avoid having to maintain a copy of `AssertException` on both of those

alanwest:
A few small CI failures to resolve

Static analysis enforces style guidelines. You should be able to produce this locally, but might need to do a release build

```
D:\a\opentelemetry-dotnet\opentelemetry-dotnet\test\OpenTelemetry.Instrumentation.AspNetCore.Tests\BasicTests.cs(665,29): error SA1204: Static members should appear before non-static members [D:\a\opentelemetry-dotnet\opentelemetry-dotnet\test\OpenTelemetry.Instrumentation.AspNetCore.Tests\OpenTelemetry.Instrumentation.AspNetCore.Tests.csproj]
```

We use markdownlint to enforce style in markdown files. You could run this locally too following [instructions here](https://github.com/open-telemetry/opentelemetry-dotnet/blob/8511e92f062f1a8dbef6e248addf3eeb3417c34b/CONTRIBUTING.md#how-to-send-pull-requests), but in your case it looks you just need to add a newline after a header.

```
src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md:3 MD022/blanks-around-headings/blanks-around-headers Headings should be surrounded by blank lines [Expected: 1; Actual: 0; Below] [Context: "## Unreleased"]
```

al-mac:
Fixed! Sorry about that.

alanwest (approved):
LGTM, thanks @al-mac!

If you can make it, we have a .NET SIG meeting tomorrow at 11 am Pacific. I plan to discuss your PR. Part of me wonders if we should do something different in the future to prevent `OnException` and other methods from being invoked in scenarios we did not originally intend for.

My thought is maybe being explicit with the event names we handle versus doing an `EndsWith("Exception")` here
https://github.com/open-telemetry/opentelemetry-dotnet/blob/6f2b1a0b3412f2860d065c81b058441977927584/src/OpenTelemetry/DiagnosticSourceInstrumentation/DiagnosticSourceListener.cs#L60-L63

cijothomas:
Thanks for this PR @al-mac !

@alanwest Please do add this to SIG discussion. I was chatting offline with @vishweshbankwar for some potential improvements in this area as well. (Its related to how leveraging Activity.Current in non Start/End path could result in fetching unintended Activity.) 
Given we have reference to AspNetCore.App (for new runtimes..), we should discuss if we can use the Public types which are the payload, instead of relying on reflection.

al-mac:
@alanwest thanks! Unfortunately I won't be able to attend the meeting. Nice catch on the potential change on the DiagnosticSourceListener.cs. The specific scenario that we worked on this PR will probably never happen if this changes to something like `else if (string.Compare(value.Key, "OnException", StringComparison.OrdinalIgnoreCase) == 0)`. The only concern is if it could break some other use-case, but at this point is valid to consider adding some more `else ifs` to the code in order to cover those.

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/BasicTests.cs
```diff
@@ -550,6 +552,48 @@ void ConfigureTestServices(IServiceCollection services)
             Assert.Equal(shouldEnrichBeCalled, enrichCalled);
         }
 
+        [Fact]
+        public async Task ShouldExportActivityWithOneExceptionFilter()
```
vishweshbankwar, line 556: Minor: Both tests could be combined into one like

```csharp
[Theory]
[InlineData(1)]
[InlineData(2)]
public async Task ShouldExportSingleActivityWithOneOrMoreExceptionFilters(int mode)
.
.
.
```

vishweshbankwar:
Related issue: #2853

vishweshbankwar (approved):

Kielek (approved):

alanwest:
> Related issue: https://github.com/open-telemetry/opentelemetry-dotnet/issues/2853

👍 ah, good find! I've updated the description since this PR should resolve the issue.

alanwest:
@al-mac apologies for the delay on this PR! We just discussed it today and would like to get it merged. Would you mind resolving the conflicts?

al-mac:
@alanwest not at all! i'll work on it later today

test/TestApp.AspNetCore.3.1/Controllers/ErrorController.cs
```diff
@@ -0,0 +1,31 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+using System;
+using Microsoft.AspNetCore.Mvc;
+
+namespace TestApp.AspNetCore._3._1.Controllers
```
alanwest, line 19: Ah yea, all the `TestApp.AspNetCore.3.1` and `TestApp.AspNetCore.6.0` directories and all files can be deleted now. We just have one directory now `TestApp.AspNetCore`. The instrumentation now only supports .NET 6+ so there was no longer a need for separate test app projects.

PR #3477, opened 2022-07-22
Exponential Bucket Histogram - part 4

reyang:
Following #3462.

## Changes

* Added corner case handling for integer overflow.
* Improved the code based on feedback from @CodeBlanch.


CodeBlanch (approved):
LGTM

PR #3479, opened 2022-07-23
PropertyFetcher: Ignore payload type when expected property name does not exist

alanwest:
Might fix the issue #3475


alanwest:
Converted to a draft, as this should get handled by #3475. This PR just highlights what I believe is the root issue in #3475.

PR #2327, opened 2021-09-09
Revive using Batch for metric export

alanwest:
Opening as draft for now...

Reviving @cijothomas's work from #2295 for using Batch<Metric> for metric export.

As is, this PR breaks the OTLP, Prometheus, and in-memory exporters and only applies the solution to the console exporter for now. Will fix the other exporters if we want to continue to pursue this path.

One goofy thing is that `MetricReader` has a handle on a `BaseExporter<Metric>` which meant I needed to move `GetAggregationTemporality` to `BaseExporter` - I'm still considering if there's a better way to do this.

PR #3484, opened 2022-07-25
Simplify code by using struct default initialization

reyang:
Follow up https://github.com/open-telemetry/opentelemetry-dotnet/pull/2327#discussion_r706335131.

### DIFF
--- a/src/OpenTelemetry/Batch.cs
+++ b/src/OpenTelemetry/Batch.cs
@@ -33,9 +33,9 @@ namespace OpenTelemetry
     public readonly struct Batch<T> : IDisposable
         where T : class
     {
-        private readonly T? item;
-        private readonly CircularBuffer<T>? circularBuffer;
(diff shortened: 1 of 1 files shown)


reyang:
Flaky test:

image

Kielek (approved):

alanwest:
> Flaky test:

This is not the test you're looking for 😆

The test you screenshotted is not a flaky, but it catches my eye _all_ the time, too! It's a test of an ASP.NET Core app under a scenario where an exception is thrown... so the log out put shows the exception and in an unfortunate catchy 🔴 color, but the that test does in fact pass. It's probably worth seeing if we can suppress the log output for this test so it stops fooling us.

Here's the flaky test you're looking for:

Screen Shot 2022-07-25 at 9 07 05 AM

alanwest (approved):

PR #3490, opened 2022-07-26
Asp.Net Core Unit test [Activity started in middleware is modified by instrumentation]

vishweshbankwar:
Fixes #.

## Changes
ASP.NET Core instrumentation subscribes to `Microsoft.AspNetCore.Mvc.BeforeAction` event to extract route information of the request. This event is written by framework after executing the middleware request pipeline. So, any activity that would be started inside middleware will be updated by instrumentation instead of the one started by framework. This happens due to our dependency on `Activity.Current` [here](https://github.com/open-telemetry/opentelemetry-dotnet/blob/6f2b1a0b3412f2860d065c81b058441977927584/src/OpenTelemetry/DiagnosticSourceInstrumentation/DiagnosticSourceListener.cs#L66). This change covers the unit test which shows the current behavior. Will follow up with the fix in separate PR.

### DIFF
--- a/test/OpenTelemetry.Instrumentation.AspNetCore.Tests/BasicTests.cs
+++ b/test/OpenTelemetry.Instrumentation.AspNetCore.Tests/BasicTests.cs
@@ -565,7 +565,6 @@ namespace OpenTelemetry.Instrumentation.AspNetCore.Tests
                     {
                         services.AddSingleton<ActivityMiddleware.ActivityMiddlewareImpl>(new TestActivityMiddlewareImpl(activitySourceName, activityName));
                         services.AddOpenTelemetryTracing((builder) => builder.AddAspNetCoreInstrumentation()
-                        .SetSampler(new AlwaysOnSampler())
                         .AddSource(activitySourceName)
                         .AddInMemoryExporter(exportedItems));
                     }))
--- a/test/OpenTelemetry.Instrumentation.AspNetCore.Tests/BasicTests.cs
+++ b/test/OpenTelemetry.Instrumentation.AspNetCore.Tests/BasicTests.cs
@@ -550,7 +550,7 @@ namespace OpenTelemetry.Instrumentation.AspNetCore.Tests
             Assert.Equal(shouldEnrichBeCalled, enrichCalled);
         }
 
-        [Fact]
+        [Fact(Skip = "Changes pending on instrumentation")]
         public async Task ActivitiesStartedInMiddlewareShouldNotBeUpdatedByInstrumentation()
         {
             var exportedItems = new List<Activity>();
--- a/test/OpenTelemetry.Instrumentation.AspNetCore.Tests/BasicTests.cs
+++ b/test/OpenTelemetry.Instrumentation.AspNetCore.Tests/BasicTests.cs
@@ -21,8 +21,6 @@ using System.Linq;
 using System.Net.Http;
 using System.Threading;
 using System.Threading.Tasks;
-using Microsoft.AspNetCore.Builder;
-using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc.Testing;
 using Microsoft.AspNetCore.TestHost;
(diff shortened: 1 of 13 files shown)


test/TestApp.AspNetCore.3.1/ActivityMiddleware.cs
```diff
@@ -0,0 +1,61 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+
+namespace TestApp.AspNetCore._3._1
+{
+    public class ActivityMiddleware
```
utpilla, line 22: It's probably a good idea to create a shared library under `test/` to keep these helper classes such as `ActivityMiddleware` and `CallbackMiddleware` and have these two projects refer to the shared library.

utpilla (approved):
Left a suggestion for refactoring. Could be considered in a follow-up PR.

test/TestApp.AspNetCore.3.1/ActivityMiddleware.cs
vishweshbankwar: Yes - planning to do this after .NET7.0 branch. Could use the same app for multiple targets as well.

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/BasicTests.cs
```diff
@@ -550,6 +550,40 @@ void ConfigureTestServices(IServiceCollection services)
...
+            var activityName = "TestMiddlewareActivity";
+
+            // Arrange
+            using (var client = this.factory
+                .WithWebHostBuilder(builder =>
+                    builder.ConfigureTestServices((IServiceCollection services) =>
+                    {
+                        services.AddSingleton<ActivityMiddleware.ActivityMiddlewareImpl>(new TestActivityMiddlewareImpl(activitySourceName, activityName));
+                        services.AddOpenTelemetryTracing((builder) => builder.AddAspNetCoreInstrumentation()
+                        .SetSampler(new AlwaysOnSampler())
```
alanwest, line 568: Is the AlwaysOnSampler necessary for this test?

alanwest (approved):

alanwest:
I had forgotten about this, but when working on the gRPC instrumentation long ago, I recalled the ASP.NET Core gRPC library looks of the activity created by ASP.NET Core here https://github.com/grpc/grpc-dotnet/blob/bfe312a3baeea3dc3676b3770dedd1b927e42fb3/src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs#L408-L434

I wasn't aware of the Feature you all mentioned in the SIG meeting, but interesting that's exactly what the gRPC library uses. Prior to .NET 6 it looped until it found the correct activity. The following comment in the gRPC library makes me question under what circumstances might the feature not be available?

> // If feature isn't available, or not supported, then fallback to Activity.Current.

cijothomas (approved):

vishweshbankwar:
> I had forgotten about this, but when working on the gRPC instrumentation long ago, I recalled the ASP.NET Core gRPC library looks up the activity created by ASP.NET Core here https://github.com/grpc/grpc-dotnet/blob/bfe312a3baeea3dc3676b3770dedd1b927e42fb3/src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs#L408-L434
> 
> I wasn't aware of the Feature you all mentioned in the SIG meeting, but interesting that's exactly what the gRPC library uses. Prior to .NET 6 it looped until it found the correct activity. The following comment in the gRPC library makes me question under what circumstances might the feature not be available?
> 
> > // If feature isn't available, or not supported, then fallback to Activity.Current.
> 
> Will we also need a fallback mechanism that loops looking for `OperationName=Microsoft.AspNetCore.Hosting.HttpRequestIn`?

I think it is referring to prior to .NET6.0?. This is the exact fix I am planning to do. However, I don't understand why it looks for operation name rather than just getting the the root activity.

vishweshbankwar:
> The following comment in the gRPC library makes me question under what circumstances might the feature not be available?
> 
> > // If feature isn't available, or not supported, then fallback to Activity.Current

@alanwest - looked at it again. one possible scenario is this `context.Features.Set<IHttpActivityFeature>(null);`.

alanwest:
> one possible scenario is this context.Features.Set<IHttpActivityFeature>(null);

💡Good thinking! Something made me think it might be more than just .NET 6.

PR #3497, opened 2022-07-27
Adding conditional check for PublicApiAnalyzers.

Yun-Ting:
## Changes
Not all projects under src/ are meant for releasing.
The `OpenTelemetry.Exporter.Prometheus.Shared` project in this PR https://github.com/open-telemetry/opentelemetry-dotnet/pull/3430/files is one example.

PublicApi documents are not required for projects not for release.
Thus, adding conditional check for PublicApiAnalyzers for the projects not meant for releasing to utilize it.

### DIFF
--- a/build/Common.prod.props
+++ b/build/Common.prod.props
@@ -1,7 +1,7 @@
 <Project>
   <Import Project=".\Common.props" />
 
-  <ItemGroup Condition="'$(forRelease)' == 'true'">
+  <ItemGroup Condition="'$(ForRelease)' == 'true'">
     <PackageReference Include="MinVer" Version="$(MinVerPkgVer)" Condition="'$(IntegrationBuild)' != 'true'">
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
       <PrivateAssets>all</PrivateAssets>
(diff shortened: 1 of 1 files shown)


build/Common.prod.props
```diff
@@ -1,7 +1,7 @@
 <Project>
   <Import Project=".\Common.props" />
 
-  <ItemGroup>
+  <ItemGroup Condition="'$(ForRelease)' == 'true'">
```
cijothomas, line 4: where is this defined?

Yun-Ting: It will be referenced in the Prometheus shared project.
See https://github.com/open-telemetry/opentelemetry-dotnet/pull/3430#discussion_r931522363

reyang (changes requested):
Seems to be a very uncommon case.

build/Common.prod.props
reyang: Maybe we can shove the shared files under the Prometheus HttpListener project, and have the ASP.NET Core binding stealing from it?

It looks like we have a problem that is not generic, and here we try to solve it by abstraction and generalization (which doesn't seem to pay off IMHO) 😸

TimothyMothra: would `Condition="$(MSBuildProjectName.EndsWith('.Shared'))"` work?

Yun-Ting: @TimothyMothra, it works!

PR #3498, opened 2022-07-27
#3490 follow up (Fix asp.net core middleware activity modification issue)

vishweshbankwar:
Follow up to #3490 

## Changes


src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs
```diff
@@ -246,6 +246,22 @@ public override void OnCustom(string name, Activity activity, object payload)
...
+                // after activity started by framework resulting in different Activity.Current.
+                // so, we need to first find the activity started by framework.
+                // For .net6.0 onwards we could use IHttpActivityFeature to get the activity created by framework
+                // var httpActivityFeature = context.Features.Get<IHttpActivityFeature>();
+                // activity = httpActivityFeature.Activity;
+                // However, this will not work as in case of custom propagator
+                // we start a new activity during onStart event which is a sibling to the activity created by framework
+                // So, in that case we need to get the activity created by us here.
+                // we can do so only by looping through activity.Parent chain.
+                while (activity.Parent != null)
```
alanwest, line 260: Are we sure that `activity.Parent != null` will always be the activity we're looking for? I'm trying to think of a scenario where an activity could be created before the ASP.NET Core created activity... though I haven't studied the flow of ASP.NET Core enough to think of one.

Also, seems pretty unlikely, but could this also be at risk of an infinite loop? Like A is the parent of B is the parent of A.

alanwest: Was just reviewing the conversation on the PR for grpc-dotnet that does the same thing you're doing. https://github.com/grpc/grpc-dotnet/pull/341#discussion_r298774343.

Seems if infinite loop was a concern it would have been a concern for them as well. However, I'm not finding a good reason why they're looking for the activity by OperationName versus the simpler thing you're doing here.

vishweshbankwar: > Also, seems pretty unlikely, but could this also be at risk of an infinite loop? Like A is the parent of B is the parent of A.

-- Do you know any specific case when this could be possible? not able to come up with a specific case.

> I'm trying to think of a scenario where an activity could be created before the ASP.NET Core created activity... though I haven't studied the flow of ASP.NET Core enough to think of one.

-- I am not completely sure but here is what I was looking at - ASP.NET Core extracts parent from remote `traceparent` and sets the parent properties on activity it starts. If there was a possibility of the activity being created before asp.net core one it would have some check [here](https://github.com/dotnet/aspnetcore/blob/93d88ac001b54db169aec5c088a8469309aaf30e/src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs#L311-L341)?

vishweshbankwar: > Was just reviewing the conversation on the PR for grpc-dotnet that does the same thing you're doing. [grpc/grpc-dotnet#341 (comment)](https://github.com/grpc/grpc-dotnet/pull/341#discussion_r298774343).
> 
> Seems if infinite loop was a concern it would have been a concern for them as well. However, I'm not finding a good reason why they're looking for the activity by OperationName versus the simpler thing you're doing here.

did not see this ^ before my previous comment

> why they're looking for the activity by OperationName versus the simpler thing you're doing here

I have same question too :)

cijothomas:
```
 private Activity? GetHostActivity()
        {
            var activity = Activity.Current;
            while (activity != null)
            {
                // We only want to add gRPC metadata to the host activity
                // Search parent activities in case a new activity was started in middleware before gRPC endpoint is invoked
                if (string.Equals(activity.OperationName, GrpcServerConstants.HostActivityName, StringComparison.Ordinal))
                {
                    return activity;
                }

                activity = activity.Parent;
            }

            return null;
        }
```

https://github.com/grpc/grpc-dotnet/pull/341/files#diff-2f74a9233c4e6d1aa1bee085bb9c98881e6bb2c2c30f62132c5423d9f4642d02R384 This seems the right one. We need to find the activity created by asp.net hosting, and we are willing to walk the activity parent chain until we find it or we give up when we no longer have a parent.

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/BasicTests.cs
```diff
@@ -550,7 +550,7 @@ void ConfigureTestServices(IServiceCollection services)
             Assert.Equal(shouldEnrichBeCalled, enrichCalled);
         }
 
-        [Fact(Skip = "Changes pending on instrumentation")]
+        [Fact]
```
cijothomas, line 553: also add a test where user manually overrides Activity.Current. In that case, we wont find the asp.net core activity.

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs
vishweshbankwar: @cijothomas - My assumption is the root activity will always be the asp.net core activity. Is there any case this will not be true?

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/BasicTests.cs
vishweshbankwar: not sure if I follow - Is there a way to manually override the `Activity.Current` without making existing `Activity.Current`  as parent.

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs
alanwest: I'm not familiar enough with the use case for why someone would do this, but it looks like you can configure the service with a custom `IHttpContextFactory`:

https://github.com/dotnet/aspnetcore/blob/93d88ac001b54db169aec5c088a8469309aaf30e/src/Hosting/Hosting/src/Internal/WebHost.cs#L154-L155

It's used to create the context here - which happens before the ASP.NET Core activity is created.

https://github.com/dotnet/aspnetcore/blob/93d88ac001b54db169aec5c088a8469309aaf30e/src/Hosting/Hosting/src/Internal/HostingApplication.cs#L76

alanwest: Ok so maybe this is super unusual, but I was able to produce this situation.. if you modify your test setup with

```csharp
                    builder.ConfigureTestServices((IServiceCollection services) =>
                    {
                        services.AddSingleton<IHttpContextFactory, TestHttpContextFactory>();
                        services.AddSingleton<ActivityMiddleware.ActivityMiddlewareImpl>(new TestActivityMiddlewareImpl(activitySourceName, activityName));
                        services.AddOpenTelemetryTracing((builder) => builder.AddAspNetCoreInstrumentation()
                        .AddSource(activitySourceName)
                        .AddSource("HttpContextFactoryActivitySource")
                        .AddInMemoryExporter(exportedItems));
                    }))
```

Here's what `TestHttpContextFactory` looks like

```csharp
        private class TestHttpContextFactory : IHttpContextFactory
        {
            private IHttpContextFactory innerFactory;
            private ActivitySource activitySource;

            public TestHttpContextFactory(IServiceProvider serviceProvider)
            {
                this.innerFactory = new DefaultHttpContextFactory(serviceProvider);
                this.activitySource = new ActivitySource("HttpContextFactoryActivitySource");
            }

            public HttpContext Create(IFeatureCollection featureCollection)
            {
                var activity = activitySource.StartActivity("ActivityFromHttpContextFactory");
                var httpContext = this.innerFactory.Create(featureCollection);
                httpContext.Items["MyActivity"] = activity;
                return httpContext;
            }

            public void Dispose(HttpContext httpContext)
            {
                (httpContext.Items["MyActivity"] as Activity).Stop();
                this.innerFactory.Dispose(httpContext);
            }
        }
```

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/BasicTests.cs
alanwest: One way you achieve this is in the custom middleware like

```csharp
public override void PreProcess(HttpContext context)
{
    Activity.Current = null;
    this.activity = this.activitySource.StartActivity(this.activityName);
}
``

File: src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs (hunk 1 above)
Comment (vishweshbankwar) [reply src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs]: Interesting!. I have updated the logic to now check for operation name. Thanks for helping with clarification.

File: test/OpenTelemetry.Instrumentation.AspNetCore.Tests/BasicTests.cs (hunk 2 above)
Comment (cijothomas) [reply test/OpenTelemetry.Instrumentation.AspNetCore.Tests/BasicTests.cs:553]: I think the key principle we need to follow here is:
Optimize for normal path.
Users can mess with things (like set Activity.Current = null etc.), and we can chose to *not* collect traces in those situations.
We should try not to modify the activity created by user. (Checking operation name is a good option. Users can still mess with it, by create Activity with operationname, and even with same activitysource as the asp.net core framework.)

Comment (vishweshbankwar) [reply test/OpenTelemetry.Instrumentation.AspNetCore.Tests/BasicTests.cs:553]: Makes sense.

**Review by alanwest — APPROVED** [review]

File: src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs [hunk 3]
```diff
@@ -246,6 +246,32 @@ public override void OnCustom(string name, Activity activity, object payload)
         {
             if (name == "Microsoft.AspNetCore.Mvc.BeforeAction")
             {
+                // We cannot rely on Activity.Current here
+                // There could be activities started by middleware
+                // after activity started by framework resulting in different Activity.Current.
+                // so, we need to first find the activity started by framework.
```
cijothomas, line 252:
```suggestion
                // so, we need to first find the activity started by Asp.Net Core.
```

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/BasicTests.cs
```diff
@@ -732,6 +787,30 @@ public override void OnStopActivity(Activity activity, object payload)
...
+
+            public TestNullHostActivityMiddlewareImpl(string activitySourceName, string activityName)
+            {
+                this.activitySource = new ActivitySource(activitySourceName);
+                this.activityName = activityName;
+            }
+
+            public override void PreProcess(HttpContext context)
+            {
+                Activity.Current = null;
```
cijothomas, line 804: why setting Activity.Current to null here?

vishweshbankwar: Was recommended to add this one. 
https://github.com/open-telemetry/opentelemetry-dotnet/pull/3498#discussion_r932650509

cijothomas: please describe it in the test case to make it clear what are we testing here.

cijothomas (approved):

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/BasicTests.cs
vishweshbankwar: done

PR #3505, opened 2022-07-29
Exponential Bucket Histogram - part 8

reyang:
Following #3462.

## Changes

* Minor cleanup (e.g. remove the `ToString` method which is not intended for production).
* Wired up `CircularBufferBuckets` and `ExponentialBucketHistogram` (it should handle 99% of the inputs correctly, the precision fix-up table - which will be covered in a future PR - will make it 100%).


utpilla (approved):

utpilla:
Merging this to unblock progress.

PR #3507, opened 2022-07-30
Dust off Prometheus Exporters

reyang:


Kielek (approved):

Kielek:
Not directly related to the PR, do you have a plan to release prometheus exporters based on 1.3.0? Or only on 1.4.0?
In AutoIntrumentation we have reference to old package, if we will have it based 1.3.0 we can schedule upgrade earlier.

cijothomas (approved):

cijothomas:
> Not directly related to the PR, do you have a plan to release prometheus exporters based on 1.3.0? Or only on 1.4.0? In AutoIntrumentation we have reference to old package, if we will have it based 1.3.0 we can schedule upgrade earlier.

We are doing a release shortly! (was supposed to be last week, but delayed.!). I think it'll be starting off 1.4.x series. Do you need 1.3.0 based one prometheus? (lets slack chat on this)

PR #3509, opened 2022-08-01
[Instrumentation.Http] Change Activity Source Name for HttpWebRequests

Kielek:
Fixes N/A.

## Changes

[Instrumentation.Http] Change Activity Source Name for `HttpWebRequest`s from `OpenTelemetry.HttpWebRequest` to `OpenTelemetry.Instrumentation.Http`.

If we need a different name than for other HttpWebRequests we could use `OpenTelemetry.Instrumentation.HttpWebRequest`


utpilla:
I think we should have the distinction between HttpWebRequest and the newer HttpClient instrumentations. So, we could instead just rename the ActivitySource to `OpenTelemetry.Instrumentation.HttpWebRequest`.

@cijothomas  Thoughts?

Kielek:
@utpilla, @cijothomas, there is a new PR #3515.
I do not have strong opinion which version will be better.
Please keep in mind that this field is serialized as `otel.library.name`, so I would prefer this PR.

Please merge one of them, and close the second one.

alanwest:
> Please keep in mind that this field is serialized as `otel.library.name`, so I would prefer this PR.

Non-OTLP exporters (e.g Zipkin and Jaeger) produce this attribute as a way to map the name and version of an ActivitySource. This used to referred to as the "instrumentation library" name and version. Though there was a somewhat recent change to instead refer to this as "instrumentation scope".

As such, non-OTLP exporters are supposed to also map this to `otel.scope.name` and `otel.scope.version` now. See: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/mapping-to-non-otlp.md#instrumentationscope

> I think we should have the distinction between HttpWebRequest and the newer HttpClient instrumentations. So, we could instead just rename the ActivitySource to OpenTelemetry.Instrumentation.HttpWebRequest.

I agree maintaining a distinction is useful. My vote is #3515.

Kielek:
Closing. Changes should be handled in scope of #3515.

PR #3515, opened 2022-08-02
[Instrumentation.Http] Change Activity Source Name for HttpWebRequest…

Kielek:
Alternative PR to #3509

## Changes

[Instrumentation.Http] Change Activity Source Name for `HttpWebRequest`s from `OpenTelemetry.HttpWebRequest` to `OpenTelemetry.Instrumentation.HttpWebRequest`.


src/OpenTelemetry.Instrumentation.Http/Implementation/HttpWebRequestActivitySource.netfx.cs
```diff
@@ -36,7 +36,7 @@ namespace OpenTelemetry.Instrumentation.Http.Implementation
     /// </remarks>
     internal static class HttpWebRequestActivitySource
     {
-        public const string ActivitySourceName = "OpenTelemetry.HttpWebRequest";
+        public const string ActivitySourceName = "OpenTelemetry.Instrumentation.HttpWebRequest";
```
alanwest, line 39: Just tossing one more idea out. When an instrumentation library defines multiple activity sources like, we could adopt the pattern of `AssemblyName`[.{`SomeAdditionalIdentifier`}]. The `SomeAddionalIdentifier` is optional and would only be used when there are multiple activity sources.

So in this case the two ActivitySources in this library would be:

* `OpenTelemetry.Instrumentation.Http.HttpClient` and
* `OpenTelemetry.Instrumentation.Http.HttpWebRequest`

Kielek: @open-telemetry/dotnet-approvers, what do you think? I will be glad to prepare changes, but I would like to have agreement on that. :)

utpilla: I think this is a good idea.

cijothomas: agree

cijothomas: with .NET 7, we wont control the ActivitySource name, as httpclient already creates activity with its ActivitySource.

cijothomas (approved):

src/OpenTelemetry.Instrumentation.Http/Implementation/HttpWebRequestActivitySource.netfx.cs
Kielek: Fixed, for currently supported set of .NET version.

I think that change will be useful as long OTel support .NET 6 and .NET Framework.

utpilla (approved):

vishweshbankwar (approved):

utpilla:
Please resolve the merge conflicts.

cijothomas (approved):

Issue #3495, opened 2022-07-26
Instrumentation libraries improvements


vishweshbankwar:
# Instrumentation libraries improvements

## Subscribing to DiagnosticSource events

Instrumentation libraries subscribe to the diagnostic source events by looking at event names
that ends with Start, Stop, Exception etc [ref](https://github.com/open-telemetry/opentelemetry-dotnet/blob/6f2b1a0b3412f2860d065c81b058441977927584/src/OpenTelemetry/DiagnosticSourceInstrumentation/DiagnosticSourceListener.cs#L52-L67). This could result in subscribing to events which we did not
intend to. One such example is #3475.

**Discussion**: Should we match with explicit names rather than pattern matching?

## Relying on `Activity.Current`

When the diagnostic source events are written by the respective framework libraries. Usually the activity is 
not part of the payload or activity itself is not started by the framework e.g. SQL instrumentation. In order
to enrich/modify activities instrumentation libraries rely on `Activity.Current` [ref](https://github.com/open-telemetry/opentelemetry-dotnet/blob/6f2b1a0b3412f2860d065c81b058441977927584/src/OpenTelemetry/DiagnosticSourceInstrumentation/DiagnosticSourceListener.cs#L52-L67) which may cause issues e.g. #3490. 

**Action** : Review any more cases like the one shown in example.

## Documentation

We do not have a detailed inventory of events that the instrumentation libraries in this
repo subscribes to. Having this documented for each instrumentation would help provide more clarity to users. Also, this would help us analyze and make decision on whether we need to support more events or update any existing ones. 

**Action**: Update `Readme` documents for each instrumentation library with the event names it subscribes to.

## Avoid reflection (ASP.NET Core)

Reflection is used to extract properties from events payload ([e.g.](https://github.com/open-telemetry/opentelemetry-dotnet/blob/44ac86cb741544b4e5b5ff043bb71454b6f9d8ae/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs#L251-L255)). This was done in order to avoid nuget reference. Now that we have FrameworkReference `Microsoft.AspNetCore.App`, could we avoid reflection?

vishweshbankwar:
@cijothomas @alanwest - let me know if I missed something or need to add more details.

Kielek:
[...]

PR #3519, opened 2022-08-02
[ASP.NET Core] Enable diagnostic source events using explicit names

vishweshbankwar:
Towards #3495 .

## Changes

This change will enable diagnostic source event subscription for asp.net core for the specified list of events.

ASP.NET Core library is using the [best practices](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Diagnostics.DiagnosticSource/src/DiagnosticSourceUsersGuide.md#filtering) defined to write diagnostic source events.

Reference:
https://github.com/dotnet/aspnetcore/blob/main/src/Mvc/Mvc.Core/src/MvcCoreDiagnosticListenerExtensions.cs
https://github.com/dotnet/aspnetcore/blob/main/src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs

This may also help with some perf as it will prevent the payload creation and writing of non-desired events.

Also, this can be extended to other instrumentation libraries based on DiagnosticSource events.


src/OpenTelemetry.Instrumentation.AspNetCore/AspNetCoreInstrumentation.cs
```diff
@@ -23,11 +24,23 @@ namespace OpenTelemetry.Instrumentation.AspNetCore
...
+
+        private readonly Func<string, object, object, bool> isEnabled = (eventName, obj1, obj2)
+            => DiagnosticSourceEvents.Contains(eventName);
+
         private readonly DiagnosticSourceSubscriber diagnosticSourceSubscriber;
 
         public AspNetCoreInstrumentation(HttpInListener httpInListener)
         {
-            this.diagnosticSourceSubscriber = new DiagnosticSourceSubscriber(httpInListener, null);
+            this.diagnosticSourceSubscriber = new DiagnosticSourceSubscriber(httpInListener, this.isEnabled);
```
alanwest, line 43: 👍 I think using an `isEnabled` delegate to define the events we're interested in is a good step in the right direction.

If we were to do this to all our instrumentation then in the end I think we'd still want to do away with pattern matching here https://github.com/open-telemetry/opentelemetry-dotnet/blob/27658f058cc248500805d66ae2ab65c2099d6400/src/OpenTelemetry/DiagnosticSourceInstrumentation/DiagnosticSourceListener.cs#L43.

That is, I no longer see the value this really brings. I think it makes more sense to change [ListenerHandler](https://github.com/open-telemetry/opentelemetry-dotnet/blob/27658f058cc248500805d66ae2ab65c2099d6400/src/OpenTelemetry/DiagnosticSourceInstrumentation/ListenerHandler.cs#L23) to just have one method like the following and have it switch on event name - much like `OnCustom` does today.

```csharp
public virtual void OnEventWritten(Activity activity, string eventName, object payload)
```

All that said, I think we could do this in stages. So leave `ListenerHandler` alone for now. Do what you've done here to define the events we want. Then come back later and decide whether we want to refactor `ListenerHander` and friends.

cijothomas: good that all these DS helpers are kept internal. we can break them apart.

cijothomas: even Activity is not required to be part of this OnEventWritten. It gives wrong information that Activity is somehow part of the event. It is not. It is part of ambient context, and anyone interested can use Activity.Current to get it. Or use other mechanism (like from inside httpcontext somewhere etc) if Activity.Current is not reliable.

vishweshbankwar: I will do a follow up PR for moving all events to custom callback.

utpilla (approved):

test/Benchmarks/Instrumentation/AspNetCoreInstrumentationBenchmarks.cs
```diff
@@ -27,17 +27,17 @@
...
+  [Host] : .NET 6.0.9 (6.0.922.41905), X64 RyuJIT
 
 Job=InProcess  Toolchain=InProcessEmitToolchain
 
 |                                      Method |     Mean |   Error |  StdDev |  Gen 0 | Allocated |
 |-------------------------------------------- |---------:|--------:|--------:|-------:|----------:|
-|                 UninstrumentedAspNetCoreApp | 155.6 us | 2.63 us | 2.33 us | 0.9766 |      5 KB |
-| InstrumentedAspNetCoreAppWithDefaultOptions | 176.8 us | 3.24 us | 2.70 us | 1.2207 |      7 KB |
+|                 UninstrumentedAspNetCoreApp | 164.7 us | 1.66 us | 1.39 us | 0.9766 |      5 KB |
+| InstrumentedAspNetCoreAppWithDefaultOptions | 178.7 us | 2.65 us | 2.35 us | 0.9766 |      5 KB |
```
cijothomas, line 40: :100: ❤️

cijothomas (approved):
LGTM.
Might be good to add a entry in changelog to highlight the improved performance to motivate ppl to update to newer version. We usually do not add changelog for internal only changes, but the perf gain here should be worth mentioning.

src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
```diff
@@ -2,15 +2,18 @@
 
 ## Unreleased
 
-* Fix issue where when an application has an ExceptionFilter, the exception data
-  wouldn't be collected.
-  ([#3475](https://github.com/open-telemetry/opentelemetry-dotnet/pull/3475))
+* Updated DiagnosticSource event subscription to specific set of events.
```
utpilla, line 5:
```suggestion
* AspNetCore instrumentation causes no extra allocation now as we have now updated DiagnosticSource event subscription to a specific set of events.
```

utpilla: > AspNetCore instrumentation causes no extra allocation

^ Ony if this is factually correct

vishweshbankwar: There is some extra allocation - it is not exactly 0. I will try with `params` to see if we can show the difference.

utpilla (approved):

CodeBlanch:
@vishweshbankwar Hey I think this PR might have broken stuff. I'm testing the AspNetCore instrumentation using the example app. Not seeing _any_ spans when running in net6.0. Took me a while but I got here:

image

Looks like a mismatch from what is being used vs what is defined?

PR #3517, opened 2022-08-02
PrometheusExporterOptions namespace

cijothomas:
Following existing naming convention:
eg: https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Exporter.Jaeger/JaegerExporterOptions.cs

https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpExporterOptions.cs

PR #3530, opened 2022-08-03
PrometheusExporter namespace cleanup.

Yun-Ting:
Finishing up https://github.com/open-telemetry/opentelemetry-dotnet/pull/3517
Creating this new PR due to push permission issue.

Following existing naming convention:
eg: https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Exporter.Jaeger/JaegerExporterOptions.cs

https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpExporterOptions.cs


src/OpenTelemetry.Exporter.Prometheus.AspNetCore/OpenTelemetry.Exporter.Prometheus.AspNetCore.csproj
```diff
@@ -17,11 +17,11 @@
...
     <Compile Include="$(RepoRoot)\src\OpenTelemetry.Api\Internal\Guard.cs" Link="Includes\Guard.cs" />
-    <Compile Include="$(RepoRoot)\src\OpenTelemetry.Exporter.Prometheus.HttpListener\Internal\PrometheusCollectionManager.cs" Link="Includes/PrometheusCollectionManager.cs" />
-    <Compile Include="$(RepoRoot)\src\OpenTelemetry.Exporter.Prometheus.HttpListener\Internal\PrometheusExporter.cs" Link="Includes/PrometheusExporter.cs" />
-    <Compile Include="$(RepoRoot)\src\OpenTelemetry.Exporter.Prometheus.HttpListener\Internal\PrometheusExporterEventSource.cs" Link="Includes/PrometheusExporterEventSource.cs" />
-    <Compile Include="$(RepoRoot)\src\OpenTelemetry.Exporter.Prometheus.HttpListener\Internal\PrometheusSerializer.cs" Link="Includes/PrometheusSerializer.cs" />
-    <Compile Include="$(RepoRoot)\src\OpenTelemetry.Exporter.Prometheus.HttpListener\Internal\PrometheusSerializerExt.cs" Link="Includes/PrometheusSerializerExt.cs" />
+    <Compile Include="$(RepoRoot)\src\OpenTelemetry.Exporter.Prometheus.HttpListener\Implementation\PrometheusCollectionManager.cs" Link="Includes/PrometheusCollectionManager.cs" />
```
reyang, line 20: Curious - why is this called "Implementation"?

alanwest: I vote for getting rid of all of our `Implementation` namespaces 😆.

Though, maybe we need to be careful making this decision across the board. For example, the Azure Function integration with OpenTelemetry depends on the Implementation namespace from our hosting extensions package https://github.com/Azure/azure-functions-host/pull/7253/files.

That said, probably safe to remove the `Implementation` namespace for Prometheus stuff at least.

Yun-Ting: > Curious - why is this called "Implementation"?

I was trying to follow the namespace naming conventions from the other exporters (Jaeger and Zipkin.) 
But I agree that in this case `PrometheusExporter.cs` can not be really counted as implementation...

@alanwest, @cijothomas, @CodeBlanch, @reyang
May I ask for your suggestion on whether we should use `OpenTelemetry.Exporter`, `OpenTelemetry.Exporter.Prometheus`, or something like `OpenTelemetry.Exporter.Prometheus.Internal` as the namespace for the internal stuffs?

cijothomas (approved):
changes to public facing classes ns is good.

Its good to separately deal with internal classes namespaces.

src/OpenTelemetry.Exporter.Prometheus.AspNetCore/OpenTelemetry.Exporter.Prometheus.AspNetCore.csproj
cijothomas: OpenTelemetry.Exporter.Prometheus would be good enough for all non-public stuff. The classes are non-public, so no need to again have a namespace .Internal/.Implementation.

reyang: My current thinking:

1. Here we have two different concepts - **namespace** and **folder name**.
2. Regarding folder name, if we chose to have a folder called "Implementation", we should be able to justify it - e.g. everything inside that folder is implementation, everything outside the folder is not the implementation (e.g. declaration, contract, API). Similarly, if there a folder called "Internal", everything inside that folder should be internal and everything else should not be internal.
3. Regarding namespace - anything publicly exposed should have high bar; things that are not publicly exposed can have freedom to change as the project evolves, so I think the guiding principle is to avoid unnecessary confusion and extra overhead.
4. Do we want folder names and namespaces to be aligned? In general it'll be good to have that, but I think it's a low order bit comparing to 2 and 3.

Yun-Ting: Thank you very much Alan, Cijo and Reiley for the feedback!
I've reverted back the internal/shared files for Prometheus - use the term "Internal".
And updated to use `OpenTelemetry.Exporter.Prometheus` as the internal files namespace as suggested.

cijothomas (approved):

PR #3536, opened 2022-08-04
Avoid redundant assignment of MetricPointStatus for Histogram updates

utpilla:
The method `Update(double number)` assigns `MetricPointStatus` to  `MetricPointStatus.CollectPending` so we need not repeat that for Histograms in the `Update(long number)` method.

## Changes
- Avoid redundant assignment of `MetricPointStatus` to `MetricPointStatus.CollectPending` for Histogram updates

### DIFF
--- a/src/OpenTelemetry/Metrics/MetricPoint.cs
+++ b/src/OpenTelemetry/Metrics/MetricPoint.cs
@@ -264,7 +264,9 @@ namespace OpenTelemetry.Metrics
                 case AggregationType.HistogramSumCount:
                     {
                         this.Update((double)number);
-                        break;
+
+                        // At this point MetricPointStatus is already set to CollectPending so we can simply return.
+                        return;
                     }
             }
 


alanwest (approved):
LGTM.

All the `MetricPointStatus` stuff remains dubious to me. I've wondered if we should consider tearing it all out. It's not actually serving a purpose at the moment.

It was meant for the purpose of identifying points that could be freed from memory - particularly when using delta aggregation temporality.

However, the need is actually more complex. Not all metric points are safe to free even when using delta temporality. For example, observable counters require us to persist the previous value in order to compute the delta.

cijothomas:
> LGTM.
> 
> All the `MetricPointStatus` stuff remains dubious to me. I've wondered if we should consider tearing it all out. It's not actually serving a purpose at the moment.
> 
> It was meant for the purpose of identifying points that could be freed from memory - particularly when using delta aggregation temporality.
> 
> However, the need is actually more complex. Not all metric points are safe to free even when using delta temporality. For example, observable counters require us to persist the previous value in order to compute the delta.

We still need this to avoid sending metrics with `NoCollectPending` for Delta model. 
We definitely have rooms for optimization here. (even for cumulative, we can chose to forget, and reset starttime, etc. )

cijothomas (approved):

reyang (approved):

Issue #3118, opened 2022-03-30
Instrumentation Libraries to leverage Activity Status


cijothomas:
https://github.com/open-telemetry/opentelemetry-dotnet/issues/3112 As a follow up to this, we can now modify the instrumentation libraries to leverage the new Status field on Activity.

Yun-Ting:
Migration on AspNetCore, Http, GrpcNetClient, ASP.NET and SqlClient were done.

PR #3537, opened 2022-08-04
[Do not merge] Http Instrumentation to leverage native Activity Status.

Yun-Ting:
Related to #3118

### DIFF
--- a/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs
+++ b/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs
@@ -188,7 +188,15 @@ namespace OpenTelemetry.Instrumentation.AspNetCore.Implementation
 
                 var response = context.Response;
 
-                activity.SetTag(SemanticConventions.AttributeHttpStatusCode, response.StatusCode);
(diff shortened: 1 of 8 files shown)


src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs
```diff
@@ -188,7 +188,15 @@ public override void OnStopActivity(Activity activity, object payload)
 
                 var response = context.Response;
 
-                activity.SetTag(SemanticConventions.AttributeHttpStatusCode, response.StatusCode);
+                //activity.SetTag(SemanticConventions.AttributeHttpStatusCode, response.StatusCode);
+                if (response.StatusCode > 200 && response.StatusCode <= 299)
+                {
+                    activity.SetStatus(ActivityStatusCode.Ok);
```
cijothomas, line 194: dont set OK. Only set status, if Error. Else leave it. 
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#set-status

"Generally, Instrumentation Libraries SHOULD NOT set the status code to Ok, unless explicitly configured to do so. Instrumentation Libraries SHOULD leave the status code as Unset unless there is an error, as described above."

PR #3351, opened 2022-06-08
Use DiagnosticSource 7.0 and drop netcoreapp3.1 target

alanwest:
Merging this to a new `net7.0` branch for so we have a place to start work that is dependent on DiagnosticSource 7.0.

* UpDownCounter support
* Using the new enumerator APIs e.g., https://docs.microsoft.com/dotnet/api/system.diagnostics.activity.enumeratetagobjects
* Test new ASP.NET Core and HttpClient ActivitySource instrumentation

PR #3539, opened 2022-08-04
Merge net7.0 to main

alanwest:
Work included is primarily from the following PRs
* #3391
* #3415
* #3351

TODO:
* Needs a changelog.
* Clean up the `net7.0` branch CI


.github/workflows/apicompatibility.yml
```diff
@@ -2,7 +2,7 @@ name: API Compatibility
 
 on:
   pull_request:
-    branches: [ main ]
+    branches: [ main, net7.0 ]
```
cijothomas, line 5: we can clean this after the main merge..

cijothomas (approved):

reyang (approved):

alanwest:
Fixed a few small things. I think this is good to merge. We definitely need a changelog - maybe there's some dependency on #3448, but assuming we're dropping `netcoreapp3` as a target, we should probably update the changelog for all relevant packages. I'll stage up a follow up PR so we can continue the discussion.

reyang:
These CI can be removed from the branch policy:

image

reyang (approved):

xiang17:
> Fixed a few small things. I think this is good to merge. We definitely need a changelog - maybe there's some dependency on #3448, but assuming we're dropping `netcoreapp3` as a target, we should probably update the changelog for all relevant packages. I'll stage up a follow up PR so we can continue the discussion.

Will we have a version where we support both .NET Core 3.1 and .NET 7? There might be overlapping time where both are officially supported.

.NET Core 3.1 [ends support in December 13, 2022](https://dotnet.microsoft.com/en-us/download/dotnet).

.NET 7 is [scheduled to release in November 2022](https://github.com/dotnet/core/blob/3b691ae43b40566255df8b01adbd0d59301f0044/roadmap.md).

alanwest:
> Will we have a version where we support both .NET Core 3.1 and .NET 7? There might be overlapping time where both are officially supported.

@xiang17 _Technically_ OpenTelemetry .NET would likely continue to work with netcoreapp3.1 even though we've upgraded to DiagnosticSource v7. A netcoreapp3.1 application could still add us as a dependency because we currently still target netstandard2.0. Though, users will receive a [build warning](https://github.com/dotnet/runtime/pull/72518/files#diff-bde310e498052e67ae73f9f1014b3ee8e067170b61c410b6b3e180fb79ed7278R207). However, one option on the table in #3448 is to drop our netstandard targets.

That said, we haven't settled on a final strategy or timeline yet for how we'll coordinate our next stable release with the release of .NET 7.

xiang17:
Sounds cool!

I was trying to validate what you said but got a build *error* after adding `netcoreapp3.1` in one of the test projects, only to find the PR was only 10 days ago. So, no action for now. :D

PR #3543, opened 2022-08-04
Clean up netcoreapp3.1 and add net7.0 to tests

vishweshbankwar:
There are still few tests missing .NET7.0. Planning to do those separately due to some errors.


test/OpenTelemetry.Extensions.Hosting.Tests/OpenTelemetry.Extensions.Hosting.Tests.csproj
```diff
@@ -31,7 +31,8 @@
     <PackageReference Include="Microsoft.AspNetCore.TestHost" Version="6.0.0" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp3.1'">
-    <PackageReference Include="Microsoft.AspNetCore.TestHost" Version="3.1.0" />
+    <ItemGroup Condition="'$(TargetFramework)' == 'net7.0'">
+    <PackageReference Include="Microsoft.AspNetCore.TestHost" Version="7.0.0-preview.6.22330.3" />
```
xiang17, line 35: Can this be put it in https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/build/Common.nonprod.props#L28?

src/OpenTelemetry.Api/Internal/Guard.cs
```diff
@@ -19,7 +19,7 @@
 using System.Runtime.CompilerServices;
 using System.Threading;
 
-#if !NETCOREAPP3_0_OR_GREATER
+#if !NET6_0_OR_GREATER
```
alanwest, line 22: Nothing that needs to be addressed in this PR, but this directive gives me some pause. Here's the targets the API project are compiled for

https://github.com/open-telemetry/opentelemetry-dotnet/blob/51c6d13460b6ccca49d1792d82bf656b8bc9d48d/src/OpenTelemetry.Api/OpenTelemetry.Api.csproj#L4

This directive is always true as compiled by the API project. I understand this file is referenced from other projects where the directive may be false, but I'm kind of surprised this hasn't caused us some issues.

I guess it's not a problem because our implementation is internal and the [real implementation](https://docs.microsoft.com/dotnet/api/system.runtime.compilerservices.callerargumentexpressionattribute) is not?

Even if it's not necessarily causing us problems, I think it may make sense to add a net6.0 target to the API and SDK projects so that net6.0+ applications get a reference to the API/SDK packages that do not contain unnecessary code.

test/OpenTelemetry.Extensions.Hosting.Tests/OpenTelemetry.Extensions.Hosting.Tests.csproj
```diff
@@ -31,7 +31,8 @@
     <PackageReference Include="Microsoft.AspNetCore.TestHost" Version="6.0.0" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp3.1'">
-    <PackageReference Include="Microsoft.AspNetCore.TestHost" Version="3.1.0" />
+    <ItemGroup Condition="'$(TargetFramework)' == 'net7.0'">
```
xiang17, line 34: nit: seems to be mis-indented

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/IntegrationTests.cs
```diff
@@ -58,7 +58,7 @@ public void Dispose()
         [SkipUnlessEnvVarFoundTheory(CollectorHostnameEnvVarName)]
         public void TraceExportResultIsSuccess(OtlpExportProtocol protocol, string endpoint, ExportProcessorType exportProcessorType, bool forceFlush)
         {
-#if NETCOREAPP3_1
+#if NET6_0_OR_GREATER
```
xiang17, line 61: Does the old NETCOREAPP3_1 apply to .NET 6 or greater?

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/IntegrationTests.cs
```diff
@@ -149,7 +149,7 @@ public void TraceExportResultIsSuccess(OtlpExportProtocol protocol, string endpo
         [SkipUnlessEnvVarFoundTheory(CollectorHostnameEnvVarName)]
         public void MetricExportResultIsSuccess(OtlpExportProtocol protocol, string endpoint, bool useManualExport, bool forceFlush)
         {
-#if NETCOREAPP3_1
+#if NET6_0_OR_GREATER
```
xiang17, line 152: Same as above.

test/OpenTelemetry.Instrumentation.Grpc.Tests/GrpcTests.client.cs
```diff
@@ -142,7 +142,7 @@ public void GrpcAndHttpClientInstrumentationIsInvoked(bool shouldEnrich)
                     .AddProcessor(processor.Object)
                     .Build())
             {
-#if NETCOREAPP3_1
+#if NET6_0_OR_GREATER
```
xiang17, line 145: Same as above.

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/IntegrationTests.cs
vishweshbankwar: it does not - this was a mistake.

src/OpenTelemetry.Api/Internal/Guard.cs
vishweshbankwar: Good point - will revisit this as a follow up.

test/OpenTelemetry.Extensions.Hosting.Tests/OpenTelemetry.Extensions.Hosting.Tests.csproj
vishweshbankwar: we use multiple versions based on tfm.

test/OpenTelemetry.Instrumentation.Http.Tests/HttpClientTests.Basic.netcore31.cs
```diff
@@ -13,7 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 // </copyright>
-#if NETCOREAPP3_1_OR_GREATER
+#if NET6_0_OR_GREATER
```
alanwest, line 16: I guess the name of the file no longer makes sense. One idea:
* rename the file to `HttpClientTests.Basic.cs`
* Change the directive to `#if !NETFRAMEWORK`.

test/OpenTelemetry.Extensions.Hosting.Tests/OpenTelemetry.Extensions.Hosting.Tests.csproj
alanwest: We could centralize the versions in Common.nonprod.props if we defined a property for each version. Something like `MicrosoftAspNetCoreTestPkgVer6` and `MicrosoftAspNetCoreTestPkgVer7`.

We have at least [one more project](https://github.com/open-telemetry/opentelemetry-dotnet/blob/aba6eafb8f62614f4035e280bef77d07c21c569a/test/OpenTelemetry.Exporter.Prometheus.AspNetCore.Tests/OpenTelemetry.Exporter.Prometheus.AspNetCore.Tests.csproj#L28) that uses the `TestHost` package.

Also, the AspNetCore.Tests project uses a different test package (Microsoft.AspNetCore.Mvc.Testing) but it looks like it is versioned in sync with the `TestHost` package.
* https://www.nuget.org/packages/Microsoft.AspNetCore.Mvc.Testing#versions-body-tab
* https://www.nuget.org/packages/Microsoft.AspNetCore.TestHost#versions-body-tab

test/OpenTelemetry.Instrumentation.SqlClient.Tests/OpenTelemetry.Instrumentation.SqlClient.Tests.csproj
```diff
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <Description>Unit test project for OpenTelemetry SqlClient instrumentations</Description>
     <!-- OmniSharp/VS Code requires TargetFrameworks to be in descending order for IntelliSense and analysis. -->
-    <TargetFrameworks Condition="$(TARGET_FRAMEWORK) == ''">net6.0</TargetFrameworks>
+    <TargetFrameworks Condition="$(TARGET_FRAMEWORK) == ''">net6.0;net7.0</TargetFrameworks>
```
alanwest, line 5: What do we want our practice to be?

In the case of the SqlClient instrumentation, here's what the instrumentation currently targets:

https://github.com/open-telemetry/opentelemetry-dotnet/blob/aba6eafb8f62614f4035e280bef77d07c21c569a/src/OpenTelemetry.Instrumentation.SqlClient/OpenTelemetry.Instrumentation.SqlClient.csproj#L5

So both net6.0 and net7.0 test runs will be testing the same instrumentation code - i.e., the netstandard2.0 build. Additional targets to the test project will extend the CI time by a little bit. Is there a benefit? Is it that the test will run on different versions of the runtime?

cijothomas: if we run on net6,7, etc, we need to modify the test to use corresponding version of sqlclient. (I dnt know if sqlclient is version aligned with .net cycles..)

vishweshbankwar: This is what we use https://github.com/open-telemetry/opentelemetry-dotnet/blob/0be837f89592576ced5cd7323d15a69999bb179b/test/OpenTelemetry.Instrumentation.SqlClient.Tests/OpenTelemetry.Instrumentation.SqlClient.Tests.csproj#L18. I don't think it varies with runtime. we could just keep `net6.0` target.

alanwest: In that case, I think it's best to not add the `net7.0` targets to most of the other test projects you've added it to. At least not until we have a good reason to.

vishweshbankwar: Done

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/IntegrationTests.cs
alanwest: Good catch. I think the right thing to do here is just delete these whole blocks. There is no `netcoreapp3.1` target for the `OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests` and `OpenTelemetry.Instrumentation.Grpc.Tests` projects anymore. We no longer need this code.

vishweshbankwar: I will do that as follow up. trying to keep code changes separate :). For now, I have just added it back

cijothomas (approved):

PR #3544, opened 2022-08-04
Clean up github workflows

alanwest:
* Removes net7.0 branch trigger
* Removes .NET 6 SDK install - pretty sure it's not necessary?
* Use setup-dotnet@v2 - don't really think this enables anything for us, but it's what we were using before.

### DIFF
--- a/.github/workflows/linux-ci.yml
+++ b/.github/workflows/linux-ci.yml
@@ -25,10 +25,6 @@ jobs:
 
     - uses: actions/setup-dotnet@v1
       with:
-        dotnet-version: '6.0.x'
-
-    - uses: actions/setup-dotnet@v1
-      with:
         dotnet-version: '7.0.x'
         include-prerelease: true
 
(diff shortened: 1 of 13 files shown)


reyang (approved):

.github/workflows/linux-ci.yml
```diff
@@ -23,11 +23,7 @@ jobs:
       with:
         fetch-depth: 0 # fetching all
 
-    - uses: actions/setup-dotnet@v1
```
vishweshbankwar, line 26: I think we need this. IIRC this issue is specific to linux.

alanwest: haha, yea I just recalled that you had explained this to me before. Seeing the problem first hand will serve to cement it in my 🧠.

vishweshbankwar (approved):

.github/workflows/linux-ci.yml
mic-max: Maybe add a comment with a link to the issue if you have it

cijothomas (approved):

PR #3558, opened 2022-08-05
Refactor asp.net core test apps

vishweshbankwar:
1)  Move to minimal hosting model
2) Combine multiple apps into one by targeting multiple tfms in single app.


test/TestApp.AspNetCore/AssemblyInfo.cs
```diff
@@ -14,11 +14,10 @@
...
+using System.Runtime.CompilerServices;
 
-[assembly: SuppressMessage(
-    "StyleCop.CSharp.NamingRules",
-    "SA1300",
-    Justification = "Reviewed.",
-    Scope = "namespaceanddescendants",
-    Target = "TestApp.AspNetCore._7._0")]
+#if SIGNED
+[assembly: InternalsVisibleTo("OpenTelemetry.Instrumentation.AspNetCore.Tests, PublicKey=002400000480000094000000060200000024000052534131000400000100010051c1562a090fb0c9f391012a32198b5e5d9a60e9b80fa2d7b434c9e5ccb7259bd606e66f9660676afc6692b8cdc6793d190904551d2103b7b22fa636dcbb8208839785ba402ea08fc00c8f1500ccef28bbf599aa64ffb1e1d5dc1bf3420a3777badfe697856e9d52070a50c3ea5821c80bef17ca3acffa28f89dd413f096f898")]
```
cijothomas, line 20: why we need this trick? everything needed can be public in the test app

test/TestApp.AspNetCore/Program.cs
```diff
@@ -0,0 +1,63 @@
...
+app.MapControllers();
+
+app.UseMiddleware<CallbackMiddleware>();
+
+app.UseMiddleware<ActivityMiddleware>();
+
+app.Run();
+
+public partial class Program
+{
```
cijothomas, line 62: any reason why we need a partial class here?

test/TestApp.AspNetCore/AssemblyInfo.cs
vishweshbankwar: This is not needed. Removed it.

test/TestApp.AspNetCore/Program.cs
vishweshbankwar: This was taken from suggestion here - https://docs.microsoft.com/en-us/aspnet/core/test/integration-tests?view=aspnetcore-6.0#:~:text=Make%20the%20Program%20class%20public%20using%20a%20partial%20class%20declaration

test/TestApp.AspNetCore/TestApp.AspNetCore.csproj
```diff
@@ -1,18 +1,13 @@
...
   <ItemGroup>
-    <Compile Remove="wwwroot\**" />
-    <Content Remove="wwwroot\**" />
-    <EmbeddedResource Remove="wwwroot\**" />
-    <None Remove="wwwroot\**" />
-  </ItemGroup>
-
-  <ItemGroup>
-    <PackageReference Include="Microsoft.VisualStudio.Web.CodeGeneration.Design" Version="7.0.0-preview.4.22259.1" />
+    <PackageReference Include="Swashbuckle.AspNetCore" Version="$(SwashbuckleAspNetCorePkgVer)" />
```
cijothomas, line 10: we should all Swagger stuff from the test app eventually.

cijothomas (approved):

test/TestApp.AspNetCore/Program.cs
vishweshbankwar: @cijothomas - I removed this partial class and added the explicit public Program class. Let me know if this is better.

PR #3559, opened 2022-08-05
[Not for merging] netcoreapp3.1 for Prometheus middleware project

alanwest:
No need for merging or making a decision at the moment. Opening this PR just to illustrate things, but I plan to close it and just open an issue.

This needs to be an additional discussion point as we consider how to handle #3448.

The Prometheus ASP.NET Core middleware project previously targeted `netcoreapp3.1`. Upon [merging our net7.0 branch](https://github.com/open-telemetry/opentelemetry-dotnet/pull/3539/files#diff-133c0a7a48813a00f9f05e4d3e31c0f369cdddad01f332d3745d0b5bd5d29b0bL5-R5), it now just targets `net6.0`.

Do we only want to support net6.0+ with our Prometheus exporter?

Another option may be to "support" it in the way this PR proposes by suppressing the build warnings for the purposes of our build, though I believe this would still result in end users getting the warnings.

Maybe we'd take the stance that we don't really support the netcoreapp3.1 version even though we offer a version that targets it?

### DIFF
--- a/src/OpenTelemetry.Api/Internal/Guard.cs
+++ b/src/OpenTelemetry.Api/Internal/Guard.cs
@@ -19,7 +19,7 @@ using System.Diagnostics;
 using System.Runtime.CompilerServices;
 using System.Threading;
 
-#if !NET6_0_OR_GREATER
+#if !NET6_0_OR_GREATER && !NETCOREAPP3_0_OR_GREATER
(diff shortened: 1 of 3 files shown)


alanwest:
Closing to continue discussion on #3560

Issue #3561, opened 2022-08-06
Investigating Codecov


TimothyMothra:
I'm investigating why Codecov often reports Impacted Files that are not modified by the PR.

## Take this PR as an example: #3546.
This PR changes 3 files related to Serilog, but Codecov is reporting 19 Impacted Files from unrelated projects! ([Codecov](https://app.codecov.io/gh/open-telemetry/opentelemetry-dotnet/compare/3546/changes))

<details>
<summary>SCREENSHOT: CODECOV FOR PR 3546</summary>

![image](https://user-images.githubusercontent.com/28785781/183222351-50f50314-247f-4ac1-9a4a-61c58178a8db.png)
</details>

### Two things stand out to me...

1. Codecov is reporting all the files that I recently increased coverage to 100%. ([PR 3476](https://github.com/open-telemetry/opentelemetry-dotnet/pull/3476))
    The reason for this is that **3546** was forked off an earlier version of `main`, before **3476** was merged.
    I think it's safe to ignore these as they should be omitted from future PRs, or if the current PR is rebased.

    <details>
    <summary>SCREENSHOT: COMMIT IDs</summary>
    
    Codecov shows the base commit (a789bc0):
    ![image](https://user-images.githubusercontent.com/28785781/183222912-eb1ea442-1c05-4ba1-ab79-719b41ce03cc.png)
    
    Commit history:
    ![image](https://user-images.githubusercontent.com/28785781/183223047-e70c004f-0f52-4946-bb75-360891e54127.png)
    </details>

2. There are a few EventSource classes in this report! 
    This is reporting that some methods were added to coverage, and some were removed.

    <details>
    <summary>SCREENSHOT: src/OpenTelemetry/Internal/OpenTelemetrySdkEventSource.cs</summary>

    ![image](https://user-images.githubusercontent.com/28785781/183224628-805ab88c-0be5-4aff-99b1-b361f1e0e311.png)
    </details>

    <details>
    <summary>SCREENSHOT: src/OpenTelemetry.Exporter.ZPages/Implementation/ZPagesExporterEventSource.cs </summary>

    ![image](https://user-images.githubusercontent.com/28785781/183224691-289f02f5-df04-4182-9712-3cd2494bbf6d.png)
    </details>

    <details>
    <summary>SCREENSHOT: src/OpenTelemetry.Exporter.Prometheus.HttpListener/Internal/PrometheusExporterEventSource.cs</summary>

[...]

PR #598, opened 2020-04-13
Fix EventSource by replacing ReadOnlySpan with string

cijothomas:
EventSource is not able to serialize ReadOnlySpan, 

Fix1: (This PR)
`ReadOnlySpan` is accepted in NonEvent method, which calls ToString() and corresponding event.

Alternate Fix:
Change the source to call .ToString() itself. This results in .ToString() penality even if no one listening. (its not clear if checking if someone is listening is costlier than ToString() itself)

PR #3586, opened 2022-08-18
improve eventsource tests - part 1

TimothyMothra:
#3561.

## Changes
- fixed `OpenTelemetrySdkEventSource.ExporterErrorResult` by removing ToString.
  This method must call WriteEvent with the same types in the method signature.
- removed if statement from `EventSourceTestHelper`, 
   previously `OpenTelemetrySdkEventSource.ExporterErrorResult` was being excluded from tests.
- add missing test for `PrometheusExporterEventSource`.
- use `EventSource.SetCurrentThreadActivityId` to set a unique ActivityId before each EventSource invoke.
  This allows us to search the listener by the ActivityId. 
  This is important because EventSource will write an error if the log fails and we need that error message to fix issues.

## Todo
- after these changes are accepted, need to make the same changes in the [Contrib repo ](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/test/OpenTelemetry.Contrib.Tests.Shared/EventSourceTestHelper.cs)


test/OpenTelemetry.Exporter.Prometheus.HttpListener.Tests/EventSourceTest.cs
```diff
@@ -0,0 +1,29 @@
+// <copyright file="EventSourceTest.cs" company="OpenTelemetry Authors">
```
alanwest, line 1: Makes sense to move this file to the new common test project so it runs both `net462` and `net6.0` targets.

alanwest (approved):
LGTM. Just a small comment now that there's an `OpenTelemetry.Exporter.Prometheus.Tests` project

cijothomas:
Failed OpenTelemetry.Tests.EventSourceTest.EventSourceTest_OpenTelemetrySdkEventSource [59 ms]
  Error Message:
   System.Exception : Method 'OpenTelemetrySdkEventSource.ShutdownEvent' is implemented incorrectly.
---- System.Exception : VerifyEventId Failed: expected: '2' actual: '40'
  Stack Trace:
     at OpenTelemetry.Tests.EventSourceTestHelper.VerifyMethodImplementation(EventSource eventSource, MethodInfo eventMethod) in D:\a\opentelemetry-dotnet\opentelemetry-dotnet\test\OpenTelemetry.Tests\Shared\EventSourceTestHelper.cs:line 68
   at OpenTelemetry.Tests.EventSourceTestHelper.MethodsAreImplementedConsistentlyWithTheirAttributes(EventSource eventSource) in D:\a\opentelemetry-dotnet\opentelemetry-dotnet\test\OpenTelemetry.Tests\Shared\EventSourceTestHelper.cs:line 32
   at OpenTelemetry.Tests.EventSourceTest.EventSourceTest_OpenTelemetrySdkEventSource() in D:\a\opentelemetry-dotnet\opentelemetry-dotnet\test\OpenTelemetry.Tests\EventSourceTest.cs:line 40
----- Inner Stack Trace -----
   at OpenTelemetry.Tests.EventSourceTestHelper.AssertEqual[T](String methodName, T expected, T actual) in D:\a\opentelemetry-dotnet\opentelemetry-dotnet\test\OpenTelemetry.Tests\Shared\EventSourceTestHelper.cs:line 134
   at OpenTelemetry.Tests.EventSourceTestHelper.VerifyEventId(MethodInfo eventMethod, EventWrittenEventArgs actualEvent) in D:\a\opentelemetry-dotnet\opentelemetry-dotnet\test\OpenTelemetry.Tests\Shared\EventSourceTestHelper.cs:line 106
   at OpenTelemetry.Tests.EventSourceTestHelper.VerifyMethodImplementation(EventSource eventSource, MethodInfo eventMethod) in D:\a\opentelemetry-dotnet\opentelemetry-dotnet\test\OpenTelemetry.Tests\Shared\EventSourceTestHelper.cs:line 60

Failed!  - Failed:     1, Passed:   792, Skipped:     3, Total:   796, Duration: 1 m 43 s - OpenTelemetry.Tests.dll (net6.0)
@TimothyMothra

reyang:
What's the overall direction regarding internal troubleshooting logs?
If the answer is "it's clear that we should move to ILogger" (#1258), would this PR be a throwaway work?
If the answer is "it's clear that we will stick with EventSource", could someone reply to #1258 and resolve it?
If the answer is "it's unclear, we don't know what's the direction", does it make sense to clarify the direction first, or just add some tests right now - without knowing whether these are throwaway work or not?

TimothyMothra:
> What's the overall direction regarding internal troubleshooting logs? If the answer is "it's clear that we should move to ILogger" (#1258), would this PR be a throwaway work? If the answer is "it's clear that we will stick with EventSource", could someone reply to #1258 and resolve it? If the answer is "it's unclear, we don't know what's the direction", does it make sense to clarify the direction first, or just add some tests right now - without knowing whether these are throwaway work or not?

I read the thread on #1258 and I didn't see any consensus to switch to ILogger. I believe that there is no change, and the current direction is to use EventSource. I think this idea has merit, but would need more of a design spec and/or POC to continue the conversation.

My PR here assumes there is no change to internal logging. I'm fixing a bug in one EventSource class, modifying the eventsourcetest helper to better catch bugs, and adding a missing eventsource test class. :)

TimothyMothra:
> Failed OpenTelemetry.Tests.EventSourceTest.EventSourceTest_OpenTelemetrySdkEventSource [59 ms] Error Message: System.Exception : Method 'OpenTelemetrySdkEventSource.ShutdownEvent' is implemented incorrectly. ---- System.Exception : VerifyEventId Failed: expected: '2' actual: '40' 

This occurred in the CodeCov run but not the other test runs, which suggests it's in intermittent bug probably due to concurrent tests.
I'm not sure how this could happen because I expected setting an ActivityId on the current test's thread would make this situation impossible.

I can make more changes to the TestHelper to make this less likely.

src/OpenTelemetry/Internal/OpenTelemetrySdkEventSource.cs
```diff
@@ -195,7 +195,7 @@ public void ShutdownEvent(int spansLeftUnprocessed)
         [Event(3, Message = "Exporter returned error '{0}'.", Level = EventLevel.Warning)]
         public void ExporterErrorResult(ExportResult exportResult)
         {
-            this.WriteEvent(3, exportResult.ToString());
+            this.WriteEvent(3, exportResult);
```
alanwest, line 198: Does this fix the behavior we're seeing here https://github.com/open-telemetry/opentelemetry-dotnet/issues/3619#issuecomment-1234724122? If so, we should get this landed.

cijothomas: https://github.com/open-telemetry/opentelemetry-dotnet/pull/598 and old PR doing similar stuff.

TimothyMothra: @alanwest, I don't have a straightforward repo to confirm if this is the issue seen in #3619. 

@cijothomas, I verified that my change works, and can confirm that the `ExportResult` enum is successfully converted to string.
I don't think the `[NonEvent]` pattern in #598 is necessary. But I'm happy to change if that's the preferred pattern. :)

cijothomas: nonevent pattern is to save string alloc if none listening.

Does this PR's change create the string if none is listening?

TimothyMothra: changed to `NonEvent` pattern

PR #3604, opened 2022-08-25
[ASP.NET Core] Remove reflection for diagnostic source event payload

### LINKED ISSUE #3495: Instrumentation libraries improvements

vishweshbankwar:
Partially Fixes #3495 

## Changes


src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs
```diff
@@ -76,7 +72,7 @@ public override void OnStartActivity(Activity activity, object payload)
                 return;
             }
 
-            _ = this.startContextFetcher.TryFetch(payload, out HttpContext context);
+            HttpContext context = payload as DefaultHttpContext;
```
alanwest, line 75: I don't think this is guaranteed to always be of type `DefaultHttpContext`. Might be a bit of an unusual case but I share this example here https://github.com/open-telemetry/opentelemetry-dotnet/pull/3498#discussion_r932771470. `DefaultHttpContext` is derived from `HttpContext`.

vishweshbankwar: Not sure if I understood. With the example (adding HttpContextFactory), I am getting the payload as DefaultHttpContext type.

alanwest: Oh yea you're right, that example uses the DefaultHttpContextFactory. My presumption is that you could use a different factory that might generate a totally different type.

alanwest: Not a fully working example, but something like 

```csharp
private class TestHttpContextFactory : IHttpContextFactory
 {
    public HttpContext Create(IFeatureCollection featureCollection)
    {
        return new MyHttpContext();
    }
}

private class MyHttpContext : HttpContext
{
}
```

vishweshbankwar: changed to HttpContext.

alanwest (approved):
Like a 🧛 whose reflection can't be seen in a 🪞. I no longer see reflection in this PR.

reyang (approved):

src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Removes reflection for extracting required objects from DiagnosticSource event
```
cijothomas, line 5: no user facing change here, so we dont need this changelog right?

cijothomas (approved):
LGTM. wondering if it is time to resurrect the perf benchmarks and see whether these have any impact on them.

src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
reyang: +1

PR #3610, opened 2022-08-27
Add support for samplers to modify Tracestate

### LINKED ISSUE #1708: Samplers should be allowed to modify tracestate
https://github.com/open-telemetry/opentelemetry-specification/pull/988 Spec allows samplers to modify trace state and have it reflected in the created Span.
This requires support from .NET runtime, so need to brought up to .NET repo.  Opening here to visibility and to see if any workarounds can be used until then.

cijothomas:
Fixes #1708
To maintain backward compatibility, `SamplingResult` will default to `null` for `TraceStateString`. Any non-null value set by samplers, will be set as the tracestate of the created activity.
If sampler intents to not modify tracestate, it can return null. (Default)
Else, it can populate tracestate with desired value.
Given tracestate is implemented as a opaque string, the samper is responsible for ensuring the tracestate semantics are followed.


vishweshbankwar (approved):

src/OpenTelemetry/Trace/SamplingResult.cs
```diff
@@ -58,6 +60,38 @@ public SamplingResult(SamplingDecision decision, IEnumerable<KeyValuePair<string
...
+            this.Decision = decision;
+            this.Attributes = Enumerable.Empty<KeyValuePair<string, object>>();
+            this.TraceStateString = traceStateString;
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="SamplingResult"/> struct.
+        /// </summary>
+        /// <param name="decision">indicates whether an activity object is recorded and sampled.</param>
+        /// <param name="attributes">Attributes associated with the sampling decision. Attributes list passed to
```
reyang, line 82:
```suggestion
        /// <param name="attributes">attributes associated with the sampling decision. Attributes list passed to
```

src/OpenTelemetry/Trace/SamplingResult.cs
```diff
@@ -70,6 +104,11 @@ public SamplingResult(SamplingDecision decision, IEnumerable<KeyValuePair<string
         /// </summary>
         public IEnumerable<KeyValuePair<string, object>> Attributes { get; }
 
+        /// <summary>
+        /// Gets the Tracestate.
```
reyang, line 108:
```suggestion
        /// Gets the tracestate.
```

reyang (approved):
Added couple minor suggestions.
LGTM with a changelog entry.

src/OpenTelemetry/Trace/SamplingResult.cs
```diff
@@ -42,6 +43,7 @@ public SamplingResult(bool isSampled)
         {
```
utpilla, line 43: Do we not need an overload for this one?

cijothomas: its optional. if there is a need, we can always add more overloads in future.

PR #3613, opened 2022-08-29
Benchmarks for sampler accessing and modifying tracestate

cijothomas:
Adding benchmarks for samplers, specifically targetting the impact of accessing/modifying tracestate inside sampler.
Related to https://github.com/open-telemetry/opentelemetry-dotnet/pull/3610


test/Benchmarks/Trace/SamplerBenchmarks.cs
```diff
@@ -0,0 +1,65 @@
...
+                .SetSampler(new TraceIdRatioBasedSampler(.5))
+                .AddSource(this.source.Name)
+                .AddProcessor(new DummyActivityProcessor())
+                .Build();
+        }
+
+        [Benchmark]
+        public void TraceIdRatioSampler()
+        {
+            using var activity = this.source.StartActivity("Benchmark");
```
cijothomas, line 58: ratio won't matter as all activity is anyway created due to being root.

test/Benchmarks/Trace/SamplerBenchmarks.cs
```diff
@@ -0,0 +1,65 @@
...
+BenchmarkDotNet=v0.13.1, OS=Windows 10.0.19044.1889 (21H2)
+Intel Core i7-4790 CPU 3.60GHz (Haswell), 1 CPU, 8 logical and 4 physical cores
+.NET SDK=7.0.100-preview.7.22377.5
+  [Host]     : .NET 6.0.8 (6.0.822.36306), X64 RyuJIT
+  DefaultJob : .NET 6.0.8 (6.0.822.36306), X64 RyuJIT
+
+
+|              Method |     Mean |    Error |  StdDev |  Gen 0 | Allocated |
+|-------------------- |---------:|---------:|--------:|-------:|----------:|
+| TraceIdRatioSampler | 642.5 ns | 11.22 ns | 9.95 ns | 0.0992 |     416 B |
```
reyang, line 34: I guess this is primarily contributed by the Activity creation/dispose?

test/Benchmarks/Trace/SamplerBenchmarks.cs
```diff
@@ -0,0 +1,65 @@
...
+
+*/
+
+namespace Benchmarks.Trace
+{
+    public class SamplerBenchmarks
+    {
+        private readonly ActivitySource source = new("SamplerBenchmarks");
```
reyang, line 42: Consider something like https://github.com/open-telemetry/opentelemetry-dotnet/blob/e6d39b8b1459aa0066249c6a771f1a27bdff71a2/test/OpenTelemetry.Tests/Metrics/InMemoryExporterTests.cs#L33?

reyang (approved):

test/Benchmarks/Trace/SamplerBenchmarks.cs
cijothomas: yes. this is just to have something get started. the main goal is to test how modifying tracestate impacts sampler perf/allocs.

test/Benchmarks/Trace/SamplerBenchmarks.cs
```diff
@@ -0,0 +1,65 @@
...
+
+        public SamplerBenchmarks()
+        {
+            // TODO: Parameterize ratio.
+            // TODO: Have sampler which modify tracestate
+            Sdk.CreateTracerProviderBuilder()
+                .SetSampler(new TraceIdRatioBasedSampler(.5))
+                .AddSource(this.source.Name)
+                .AddProcessor(new DummyActivityProcessor())
```
utpilla, line 51: Why add a processor at all?

test/Benchmarks/Trace/SamplerBenchmarks.cs
utpilla, line 34: It would be good to see perf impact caused by this change: https://github.com/open-telemetry/opentelemetry-dotnet/pull/3610/files#diff-1d903debe2b97cb3adcadddd601aa37285eb72f25b8dd54aeb58be3afbe9a8b9R383

cijothomas: no noticeable diff.
I am actually modifying this benchmark to do test for the effect of modifying tracestate vs not touching it.

utpilla: That's interesting. I thought since we are now copying the struct on every sample call we might incur some cost.

utpilla (approved):

Kielek (approved):

PR #3615, opened 2022-08-30
Improve Sampler tests to validate initial tags and links are passed

cijothomas:
Fixes #.

## Changes

### DIFF
--- a/src/OpenTelemetry/Trace/SamplingParameters.cs
+++ b/src/OpenTelemetry/Trace/SamplingParameters.cs
@@ -50,7 +50,7 @@ namespace OpenTelemetry.Trace
             this.TraceId = traceId;
             this.Name = name;
             this.Kind = kind;
-            this.Tags = tags ?? Enumerable.Empty<KeyValuePair<string, object>>();
+            this.Tags = tags;
             this.Links = links;
         }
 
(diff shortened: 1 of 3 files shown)


src/OpenTelemetry/Trace/SamplingParameters.cs
```diff
@@ -42,14 +43,14 @@ namespace OpenTelemetry.Trace
...
-            IEnumerable<KeyValuePair<string, object>> tags = null, // TODO: Empty
+            IEnumerable<KeyValuePair<string, object>> tags = null,
             IEnumerable<ActivityLink> links = null)
         {
             this.ParentContext = parentContext;
             this.TraceId = traceId;
             this.Name = name;
             this.Kind = kind;
-            this.Tags = tags;
+            this.Tags = tags ?? Enumerable.Empty<KeyValuePair<string, object>>();
```
cijothomas, line 53: Is this considered a breaking change? If not provided, Tags used to be null. Now it'll be Empty collection. It makes writing sampler easy.
`foreach(var tag in tags) {}`
vs
`if (tags != null) {foreach(var tag in tags) {}}`

reyang: Would this for-each introduce perf overhead?

cijothomas: yes a minor one. (the null check would likely be the fastest, but empty is easy to write code)

I'll keep the existing `null` as it is possible that some is making sampling decision based on it being `null`, changing it can be breaking behavior.

cijothomas:
image
coverage is shown as improving for this file, as expected. But overall it shows decreasing! FYI @TimothyMothra

utpilla (approved):

PR #3634, opened 2022-09-07
Make publish workflow work based on tag/branch selected

alanwest:
Adds a `ref` parameter to the workflow that must be specified when running it. When running the workflow manually, specifying the `ref` is required. For our releases, we create both a "core" and "non-core" tag at the same SHA, and it doesn't matter which of the two tags you run the workflow with.

Looks like this:

Screen Shot 2022-09-07 at 3 38 30 PM

The nightly build will continue to use `main`.

---

There is an alternative we could explore. The pack and publish workflow could be triggered when we draft a new release. If triggered this way, then the workflow would use the SHA of the release. No need to manually run the workflow and type in the tag name.

### DIFF
--- a/.github/workflows/publish-packages-1.0.yml
+++ b/.github/workflows/publish-packages-1.0.yml
@@ -19,18 +19,12 @@ on:
 
 jobs:
   build-pack:
-    runs-on: ${{ matrix.os }}
-
-    strategy:
-      matrix:
-        os: [windows-latest]
-        branches: [main]
+    runs-on: windows-latest
 
     steps:
     - uses: actions/checkout@v3
       with:
         fetch-depth: 0 # fetching all
-        ref: ${{ matrix.branches }}
 
     - uses: actions/setup-dotnet@v2
       with:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-packages-1.0.yml
```diff
@@ -19,18 +19,12 @@ on:
...
-      matrix:
-        os: [windows-latest]
-        branches: [main]
+    runs-on: windows-latest
 
     steps:
     - uses: actions/checkout@v3
       with:
         fetch-depth: 0 # fetching all
-        ref: ${{ matrix.branches }}
```
alanwest, line 33: This was forcing a build and publish from `main` _always_.

.github/workflows/publish-packages-1.0.yml
```diff
@@ -49,7 +43,7 @@ jobs:
     - name: Publish Artifacts
       uses: actions/upload-artifact@v3
       with:
-        name: ${{ matrix.os }}-packages
+        name: windows-latest-packages
```
alanwest, line 46: Meh? I mean I think this is the name that was being used before, but the prefix `windows-latest` doesn't really mean anything here. Probably can just be 

```suggestion
        name: published-packages
```

or something.

utpilla:
Closing this in favor of #3638

PR #3638, opened 2022-09-07
Trigger pack and publish when release is created

alanwest:
I think I like this approach better than #3634.

The pack and publish job is kicked off automatically upon creating a release in GitHub.

### Step 1 - Create and push the tags just as before. Then click create release here:

Screen Shot 2022-09-07 at 4 10 50 PM

### Step 2 - Fill out the details of the release as usual:

Screen Shot 2022-09-07 at 4 12 34 PM

### Step 3 - Click "Publish release" and the job is kicked off like so:

Screen Shot 2022-09-07 at 4 13 25 PM


TimothyMothra:
This looks like the better option to me.
Triggering the release from the Tag seems straight forward and less error prone.

build/RELEASING.md
```diff
@@ -86,12 +86,16 @@ Only for Maintainers.
...
-    and manually trigger a build after selecting the tag that was pushed in Step
-    6 (if you pushed both core and non-core tags, use any one of them). At the
-    end of the workflow run, the packages would be published to MyGet. The
-    package names will be suffixed with the tag names used in Step 1.
+ 7. Go to the [list of
+    tags](https://github.com/open-telemetry/opentelemetry-dotnet/tags)
+    and find the tag created for the core components. Click the three
+    dots next to the tag and choose `Create release`.
+      * Give the release a name based on the tags created
+      (e.g., `1.4.0-beta.1 / 1.0.0-rc9.7`).
```
alanwest, line 94: As part of this new flow, I'm also suggesting that a single release is created for both core and non-core components. This is similar to what some of the other projects do e.g., [opentelemetry-go](https://github.com/open-telemetry/opentelemetry-go/releases/tag/v1.8.0).

cijothomas: since they have different tags, how do we do it in one-shot? We currently do github release for the core-1.3. tag, following by 1.0.0-rc9... for non-core

alanwest: There's no technical reason we need two release entries in GitHub. A release is associated to a SHA. In practice, the SHA has been the same between our core and non-core tags. Furthermore the pack and publish job has always created all the artifacts in one shot.

Though, if we want two, it'd be purely based on opinion. The opinion I'm putting forth is mostly an aesthetic one. 

I appreciate the way the opentelemetry-go project has made use of the GitHub release functionality. When they do a release, they have one entry that highlights all the changes across both stable and unstable components. If you really dig into the weeds you'll note that they actually have [_many_ more tags than we do](https://github.com/open-telemetry/opentelemetry-go/tags) from the SHA they base their release entry from.

I just think it provides a more user-friendly experience when folks look at the releases page to catch up on the latest new stuff from the OpenTelemetry .NET project.

cijothomas: image
when we do the GH Release, which tag we pick?

alanwest: If you do things in the way I documented:

> Go to the [list of tags](https://github.com/open-telemetry/opentelemetry-dotnet/tags) and find the tag created for the core components.

That drop down will have the `core-xxx` pre-selected. Though, you could do so manually as well.

utpilla (approved):

cijothomas (approved):

PR #3643, opened 2022-09-09
Change StatusCode to 200 when no metrics are collected

dszmigielski:
Fixes #3582.

## Changes

If Prometheus Exporter have no metrics to collect it reports StatusCode 200 now instead of 204.

For significant contributions please make sure you have completed the following items:

* [x] Appropriate `CHANGELOG.md` updated for non-trivial changes
~* [ ] Design discussion issue #~
~* [ ] Changes in public API reviewed~

PR #3654, opened 2022-09-13
Output EOF following the OpenMetrics exposition recommendation

reyang:
## Changes

* Added support for EOF.

Reference: https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#overall-structure

image


src/OpenTelemetry.Exporter.Prometheus.HttpListener/Internal/PrometheusSerializer.cs
```diff
@@ -270,6 +270,15 @@ public static int WriteMetricName(byte[] buffer, int cursor, string metricName,
             return cursor;
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static int WriteEof(byte[] buffer, int cursor)
+        {
+            cursor = WriteAsciiStringNoEscape(buffer, cursor, "# EOF");
```
alanwest, line 276: I don't know enough about prometheus, but I assume this won't cause an older prometheus scrapper to blow up?

Also, is simply adding `# EOF` enough to warrant a change to the content-type to `application/openmetrics-text`?

https://github.com/open-telemetry/opentelemetry-dotnet/blob/a87f8f7373b13dad994e5a9414aed356ec712f67/src/OpenTelemetry.Exporter.Prometheus.AspNetCore/PrometheusExporterMiddleware.cs#L76

reyang: > I don't know enough about prometheus, but I assume this won't cause an older prometheus scrapper to blow up?

I've tested this with older version of scrappers.

image

> Also, is simply adding `# EOF` enough to warrant a change to the content-type to `application/openmetrics-text`?
> 
> https://github.com/open-telemetry/opentelemetry-dotnet/blob/a87f8f7373b13dad994e5a9414aed356ec712f67/src/OpenTelemetry.Exporter.Prometheus.AspNetCore/PrometheusExporterMiddleware.cs#L76

I guess the answer is no, I keep learning new things while reading the OpenMetrics spec, e.g. https://github.com/open-telemetry/opentelemetry-dotnet/pull/3651. I think we need to do a full scan of the spec, maybe it's okay to flip the content-type earlier?

alanwest: > maybe it's okay to flip the content-type earlier?

Probably fine to wait since content-type might be a thing that would affect older scrapers. But may become necessary with exemplar support?

src/OpenTelemetry.Exporter.Prometheus.HttpListener/Internal/PrometheusCollectionManager.cs
```diff
@@ -191,27 +191,37 @@ private ExportResult OnCollect(Batch<Metric> metrics)
...
+                    {
+                        if (!this.IncreaseBufferSize())
+                        {
+                            throw;
                         }
                     }
                 }
 
-                this.previousDataView = new ArraySegment<byte>(this.buffer, 0, Math.Max(cursor - 1, 0));
+                this.previousDataView = new ArraySegment<byte>(this.buffer, 0, cursor);
```
utpilla, line 224: Do we not have to decrement the cursor now?

utpilla (approved):

Kielek:
I do not check the specification, but it is worth to double check how it should behave when there is no metrics to export.
Should it be 200 with empty response or 200 with `# EOF\n`?

Ref: #3643

reyang:
> I do not check the specification, but it is worth to double check how it should behave when there is no metrics to export. Should it be 200 with empty response or 200 with `# EOF\n`?
> 
> Ref: #3643

Good point, the current approach is return HTTP 200 without content/content-type, so technically it is not a Prometheus nor OpenMetrics exposition format (which means OpenMetrics spec won't apply).

alanwest (approved):

PR #3670, opened 2022-09-16
Fix environment variable when contains URI with splitter sign = inside

jkljajic:
Fixes #.
When OTEL_RESOURCE_ATTRIBUTES contains KeyValuePair key=http://localhost/key=value&key2=value2 it does not detect very well because it splits with "=" and  get more then length of 2

## Changes

```
#if NET6_0 || NETSTANDARD2_1_OR_GREATER
                string[] keyValuePair = rawKeyValuePair.Split(AttributeKeyValueSplitter,2);
#else
                string[] keyValuePair = rawKeyValuePair.Split(AttributeKeyValueSplitter);
                if(keyValuePair.Length > 2)
                {
                    keyValuePair = new string[] { keyValuePair[0], string.Join(AttributeKeyValueSplitter.ToString(), keyValuePair.Skip(1)) };
                }
#endif
```

trying to spit only on the first occurrence of AttributeKeyValueSplitter and because that is supported in net6 and NETSTANDARD2_1_OR_GREATER I need to add additional work around wor previus Dotnet release

### DIFF
--- a/src/OpenTelemetry/Resources/OtelEnvResourceDetector.cs
+++ b/src/OpenTelemetry/Resources/OtelEnvResourceDetector.cs
@@ -15,6 +15,8 @@
 // </copyright>
 
 using System.Collections.Generic;
+using System.Linq;
+
 using OpenTelemetry.Internal;
 
 namespace OpenTelemetry.Resources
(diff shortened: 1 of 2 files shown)


reyang:
It's unclear how `OTEL_RESOURCE_ATTRIBUTES` should be used when the attribute value has special characters, is there a clear spec or the spec is simply buggy?

PR #3567, opened 2022-08-09
Remove netstandard targets from asp.netcore instrumentation

vishweshbankwar:
Fixes #.

## Changes

As discussed during SIG meeting, removing `netstandard2.0` and `netstandard2.1` targets from the library. 

For significant contributions please make sure you have completed the following items:

* [x] Appropriate `CHANGELOG.md` updated for non-trivial changes
* [ ] Design discussion issue #
* [ ] Changes in public API reviewed

PR #3697, opened 2022-09-26
Non-core packages: Remove netstandard targets. Replace with net6.0/net462 as applicable.

alanwest:
In the previous SIG meeting I had discussed drafting an announcement regarding our stance for supporting .NET Standard.

In drafting that announcement, I believe this PR may help us clarify our current stance.

.NET Standard targets have already been removed from ASP.NET Core and HttpClient instrumentation in https://github.com/open-telemetry/opentelemetry-dotnet/pull/3567 and
https://github.com/open-telemetry/opentelemetry-dotnet/pull/3664, respectively.

This PR poses the question whether we should apply this stance to all our our non-core "unstable" packages. My opinion is that it would be odd for us to claim support for .NET Standard for some but not all of our instrumentation and extensions.

The question of whether we should do anything with our core packages remains a separate conversation. That said, I think it's possible that users will find it odd that the API/SDK continue support .NET Standard - thereby allowing them to use them in .NET 5/.NET Core 3.1 apps - but they are prevented from using the latest instrumentation.

Thoughts?


reyang (approved):
LGTM, this seems like a significant change that needs to be called out in the changelog? (e.g. someone previously targeting netstandard2.0 will now have to introduce multiple targets?)

alanwest:
> LGTM, this seems like a significant change that needs to be called out in the changelog? (e.g. someone previously targeting netstandard2.0 will now have to introduce multiple targets?)

Yes, definitely. I'm drafting a proper announcement that will show up here https://github.com/open-telemetry/opentelemetry-dotnet/discussions. I plan to discuss this PR in the SIG tomorrow to ensure this is really the direction we want to go. If folks feel good about this then I will update the changelogs with a link to the announcement I've drafted. My goal is to get people's attention one way or another.

cijothomas (approved):

cijothomas:
In the announcements, we need to mention why 3.1 is being dropped. (its EOL in Dec 2022....)

utpilla (approved):

alanwest:
Draft announcement https://github.com/open-telemetry/opentelemetry-dotnet/pull/3701

> In the announcements, we need to mention why 3.1 is being dropped. (its EOL in Dec 2022....)

This is described in the "background" section of the announcement.

I propose we settle on the announcement, get it published, then merge this PR with updated changelogs linking to the announcement.

vishweshbankwar (approved):

alanwest:
Discussed in SIG meeting today to keep .NET Standard targets for now. We can always come back to the decision to remove .NET Standard targets at a later date. #3703 is a draft announcement that addresses the warnings end users may receive when upgrading OpenTelemetry dependencies.

PR #3699, opened 2022-09-27
[HttpClient] Unit test to validate span creation in case of redirect

vishweshbankwar:
Fixes #.

## Changes

Follow up https://github.com/open-telemetry/opentelemetry-dotnet/pull/3072#issuecomment-1258451318

In case of .NET Core new span for each redirect is created as per the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#http-redirects-examples). Adding the unit test from original [PR](https://github.com/open-telemetry/opentelemetry-dotnet/pull/3072) created by @denisivan0v to validate that. 

Will do a separate follow up for retries.

### DIFF
--- a/test/OpenTelemetry.Instrumentation.Http.Tests/HttpClientTests.Basic.cs
+++ b/test/OpenTelemetry.Instrumentation.Http.Tests/HttpClientTests.Basic.cs
@@ -39,7 +39,16 @@ namespace OpenTelemetry.Instrumentation.Http.Tests
             this.serverLifeTime = TestHttpServer.RunServer(
                 (ctx) =>
                 {
-                    ctx.Response.StatusCode = 200;
+                    if (ctx.Request.Url.PathAndQuery.Contains("redirect"))
+                    {
+                        ctx.Response.RedirectLocation = "/";
+                        ctx.Response.StatusCode = 302;
+                    }
+                    else
+                    {
+                        ctx.Response.StatusCode = 200;
+                    }
+
                     ctx.Response.OutputStream.Close();
                 },
                 out var host,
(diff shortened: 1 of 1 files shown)


test/OpenTelemetry.Instrumentation.Http.Tests/HttpClientTests.Basic.cs
```diff
@@ -317,6 +326,28 @@ public async Task HttpClientInstrumentationBacksOffIfAlreadyInstrumented()
...
+            using (Sdk.CreateTracerProviderBuilder()
+                       .AddHttpClientInstrumentation()
+                       .AddProcessor(processor.Object)
+                       .Build())
+            {
+                using var c = new HttpClient();
+                await c.GetAsync($"{this.url}redirect");
+            }
+
+            Assert.Equal(7, processor.Invocations.Count); // SetParentProvider/OnStart/OnEnd/OnStart/OnEnd/OnShutdown/Dispose called.
```
cijothomas, line 342: Would be nice to get this replaced with InMemoryExporter..

cijothomas:
The test is passing, so we were already doing this for redirects and the pending PR is for the retry one only?

cijothomas (approved):

vishweshbankwar:
> The test is passing, so we were already doing this for redirects and the pending PR is for the retry one only?

That's right. The [check](https://github.com/open-telemetry/opentelemetry-dotnet/blob/86ee906dcd02c562e85814ca07f2a1a15a83dc98/src/OpenTelemetry.Instrumentation.Http/Implementation/HttpHandlerDiagnosticListener.cs#L119) has no impact on redirect as `Httpclient` [clears](https://github.com/dotnet/runtime/blob/1f688ed75133664aa20c838fa451e2ecf9f1f424/src/libraries/System.Net.Http/src/System/Net/Http/DiagnosticsHandler.cs#L102) the headers for redirects anyways.

test/OpenTelemetry.Instrumentation.Http.Tests/HttpClientTests.Basic.cs
vishweshbankwar: Can do a follow up as this needs to be changed in other tests as well.

PR #3700, opened 2022-09-27
[HttpClient] Export span even if request was instrumented

vishweshbankwar:
Fixes #.

## Changes

Follow up https://github.com/open-telemetry/opentelemetry-dotnet/pull/3072#issuecomment-1258451318. Adding the change from original [PR](https://github.com/open-telemetry/opentelemetry-dotnet/pull/3072) created by @denisivan0v here.

### DIFF
--- a/test/OpenTelemetry.Instrumentation.Http.Tests/HttpClientTests.Basic.cs
+++ b/test/OpenTelemetry.Instrumentation.Http.Tests/HttpClientTests.Basic.cs
@@ -39,7 +39,16 @@ namespace OpenTelemetry.Instrumentation.Http.Tests
             this.serverLifeTime = TestHttpServer.RunServer(
                 (ctx) =>
                 {
-                    ctx.Response.StatusCode = 200;
+                    if (ctx.Request.Url.PathAndQuery.Contains("redirect"))
+                    {
+                        ctx.Response.RedirectLocation = "/";
+                        ctx.Response.StatusCode = 302;
+                    }
+                    else
+                    {
+                        ctx.Response.StatusCode = 200;
+                    }
+
                     ctx.Response.OutputStream.Close();
                 },
                 out var host,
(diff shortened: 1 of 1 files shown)


src/OpenTelemetry.Instrumentation.Http/CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+* Spans will now be exported even if the request was instrumented before. This
```
cijothomas, line 5: this is an instrumentation library. it can produce spans, enrich them, but exporting is not owned by instrumentation.

src/OpenTelemetry.Instrumentation.Http/CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
...
+is to enable exporting spans in case of retries as per the
+[spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#http-retries-examples)
+where same request can be reused. Note: This change will not create a new span
+for each retry as per spec.
+([#3700](https://github.com/open-telemetry/opentelemetry-dotnet/pull/3700))
```
cijothomas, line 10: do callout if this applicable to HttpClient (.NET Core/.NET) only

src/OpenTelemetry.Instrumentation.Http/CHANGELOG.md
vishweshbankwar: It also decides if the activity will be exported or not by setting `activity.IsAllDataRequested`. This is what we were doing before `activity.IsAllDataRequested = false;`. That is the reason to use export in here, let me know if this can be clarified further.

src/OpenTelemetry.Instrumentation.Http/CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+* Spans will now be exported even if the request was instrumented before. This
+is to enable exporting spans in case of retries as per the
+[spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#http-retries-examples)
+where same request can be reused. Note: This change will not create a new span
```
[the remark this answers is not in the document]
vishweshbankwar: The spec states each retry should have a new activity. looking at https://github.com/open-telemetry/opentelemetry-dotnet/pull/2756, I thought the plan was to create these activities inside instrumentation library. 

Just thinking about it again, the change in this PR does not necessarily relates to retry. It would also cover the case where the request was already instrumented before by some other library than OTel. In this case we were not exporting the span, with this change we will.

PR #2179, opened 2021-07-23
Remove upper constraint for Ms.Extensions. dependencies

cijothomas:
Once we get an ack from .NET runtime team that, Msft.Extensions.* packages will always remain backward compatible, we can remove the upper bound restriction on Msft.Extensions package. As of today, we have <6 allowed, blocking .NET 6 preview users.

~~This also ups the minimum version to be 3.1 (instead of 2.1), as this update will be released along 1.2.0 (Nov 2021), and 2.1* packages are out of support then.~~
The minimum version is kept as 2.1, to support Asp.Net Core 2.1 applications which target the .NET Framework.

## Changes

Please provide a brief description of the changes here.

For significant contributions please make sure you have completed the following items:

* [ ] `CHANGELOG.md` updated for non-trivial changes
* [ ] Design discussion issue #
* [ ] Changes in public API reviewed

PR #1769, opened 2021-02-01
Relax System.* package upper bounds.

cijothomas:
Remove upper bound restriction for System.* packages as these are expected to not introduce breaking changes.

Also require 5.0.1 of DS package, as per this discussion: https://github.com/open-telemetry/opentelemetry-dotnet/pull/1745#discussion_r567023082 

## Changes

Please provide a brief description of the changes here.

For significant contributions please make sure you have completed the following items:

* [ ] `CHANGELOG.md` updated for non-trivial changes
* [ ] Design discussion issue #
* [ ] Changes in public API reviewed

PR #3701, opened 2022-09-28
Proposal for dropping .NET Standard

alanwest:
Opening this proposal up as a PR for us to discuss. This PR is not meant to be merged. The plan is to post the contents of this PR as [an announcement](https://github.com/open-telemetry/opentelemetry-dotnet/discussions/categories/announcements).

### DIFF
--- a/netstandard.md
+++ b/netstandard.md
@@ -2,7 +2,7 @@
 
 ## Background
 
-The OpenTelemetry .NET project depends heavily on APIs provided by .NET.
+The OpenTelemetry .NET project depends heavily on APIs provided by [.NET Runtime](https://github.com/dotnet/runtime).
 Specifically, the APIs provided under `System.Diagnostics.DiagnosticSource`.
 
 With the release of .NET 6, [it was announced](https://github.com/dotnet/announcements/issues/190)
--- a/netstandard.md
+++ b/netstandard.md
@@ -18,7 +18,8 @@ Therefore, OpenTelemetry .NET can technically be consumed by projects targeting
 out-of-support frameworks - even as far back as .NET Core 2.0.
 
 However, referencing the latest version of OpenTelemetry .NET by an application
-targeting .NET Core 3.1 or .NET 5 will generate the following build warnings:
+targeting .NET Core 3.1 or .NET 5, for example, will generate the following
+build warnings:
 
 > System.Diagnostics.DiagnosticSource doesn't support netcoreapp3.1. Consider updating your TargetFramework to net6.0 or later.
 
@@ -53,10 +54,10 @@ contains the following targets:
 <TargetFrameworks>net6.0;netstandard2.1;netstandard2.0;net462</TargetFrameworks>
 ```
 
-When .NET 6 reaches end of life November 2024, the net6.0 and .NET Standard targets
+When .NET 6 reaches end of life November 2024, the `net6.0` and .NET Standard targets
 will be removed and we will perform a major version release of the OpenTelemetry
 .NET SDK. Presumably, at that time .NET 8 will have been released, and the
-the project file may look like:
+the project file may look like this:
 
 ```xml
 <TargetFrameworks>net8.0;net462</TargetFrameworks>
@@ -67,7 +68,7 @@ example describes our process for all of our stable packages.
 
 That said, the OpenTelemetry .NET project offers a number of packages that have
 not yet had a stable release. We will be removing .NET Standard builds from the
-following packages ing the next minor release:
+following packages in the next minor release:
 
 * `OpenTelemetry.Exporter.Prometheus.AspNetCore`
 * `OpenTelemetry.Exporter.Prometheus.HttpListener`
--- a/netstandard.md
+++ b/netstandard.md
@@ -44,13 +44,30 @@ frameworks reach end of life, we will remove those targets during a major
 version release roughly aligned with the corresponding major version release
 of .NET.
 
-Given that dropping a framework from a package is a source breaking change, we
-will coordinate dropping .NET Standard targets with a major version release for
-all of our stable packages.
+For example:
+
+Currently the [OpenTelemetry SDK project](/blob/ee11de90a37915c68d9d44cdd283ba6047b394a3/src/OpenTelemetry/OpenTelemetry.csproj#L4)
+contains the following targets:
+
+```xml
+<TargetFrameworks>net6.0;netstandard2.1;netstandard2.0;net462</TargetFrameworks>
+```
+
+When .NET 6 reaches end of life November 2024, the net6.0 and .NET Standard targets
+will be removed and we will perform a major version release of the OpenTelemetry
+.NET SDK. Presumably, at that time .NET 8 will have been released, and the
+the project file may look like:
+
+```xml
+<TargetFrameworks>net8.0;net462</TargetFrameworks>
+```
+
+Since dropping a framework from a package is a source breaking change, the above
+example describes our process for all of our stable packages.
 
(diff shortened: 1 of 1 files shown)


netstandard.md
```diff
@@ -0,0 +1,127 @@
...
+The plan moving forward is to align the support cycle of OpenTelemetry .NET
+with the support cycle of .NET. To achieve this, OpenTelemetry .NET plans to
+drop .NET Standard builds of its artifacts. OpenTelemetry packages will only
+offer implementation specific target frameworks (e.g., `net462` and `net6.0`).
+This will enable us to easily test and validate that OpenTelemetry .NET works
+as expected for all the frameworks we target.
+
+At this time, the frameworks targeted will include at least `net462` and
+`net6.0`, but may include others like `net7.0`, when appropriate. As target
+frameworks reach end of life, we will remove those targets during a major
```
reyang, line 44: major version is something like 2.0 / 3.0?

netstandard.md
```diff
@@ -0,0 +1,127 @@
...
+version release roughly aligned with the corresponding major version release
+of .NET.
+
+For example:
+
+Currently the [OpenTelemetry SDK project](/blob/ee11de90a37915c68d9d44cdd283ba6047b394a3/src/OpenTelemetry/OpenTelemetry.csproj#L4)
```
vishweshbankwar, line 50: invalid link.

netstandard.md
cijothomas: https://github.com/open-telemetry/opentelemetry-dotnet/issues/2138 historically we only did minor version bump when we dropped framework support (when the fw itself went out-of-support)

netstandard.md
```diff
@@ -0,0 +1,127 @@
+# Dropping support for .NET Standard
+
+## Background
+
+The OpenTelemetry .NET project depends heavily on APIs provided by .NET.
```
cijothomas, line 5:
```suggestion
The OpenTelemetry .NET project depends heavily on APIs provided by [.NET Runtime](https://github.com/dotnet/runtime).
```

netstandard.md
alanwest: Yes, this differs from our prior pattern of practice, and I'm suggesting we reconsider our practice for the future. As mentioned in https://github.com/dotnet/announcements/issues/190

> Dropping a framework from a package is a source breaking change.

I think a major version number is the correct tool to communicate this to users clearly. It aligns with the .NET Runtime team's versioning of DiagnosticSource in regards to dropping supported frameworks.

That said, there certainly are other examples of library authors dropping frameworks with minor version releases of their libraries. Though, I do not think a minor version bump communicates very clearly whether the drop in support was intentional or not.

I'd be very interested in some official guidance to library authors in this respect. Maybe here https://learn.microsoft.com/en-us/dotnet/standard/library-guidance/versioning. Something like:

> ✔️ CONSIDER (or DO?) bumping the major version when dropping framework support.

Maybe .NET would consider this overstepping their bound with their guidance, but I'd like to hear their thoughts.

cijothomas: My main worry with major version bump was the above _( worried about upgrading from 2.0 to 3.0 due to the fear of breaking changes)_

I am more inclined to do minor version bump, and document that some runtimes are no longer supported, as opposed to doing major version bump and documenting no-breaking-api-but-drop-certain-runtimes)

alanwest: sorry, didn't see @reyang's comment prior to posting mine.

> For example, a user might be worried about upgrading from 2.0 to 3.0 due to the fear of breaking changes, although what actually happened is just a drop of deprecated runtime version.

Would we anticipate that users upgrading to OTel 1.4 would be worried that it comes with a major version bump of DiagnosticSource 6 -> 7. That is, might they be concerned that 7 introduces breaking changes?

cijothomas: https://github.com/open-telemetry/opentelemetry-dotnet/pull/1769
https://github.com/open-telemetry/opentelemetry-dotnet/pull/2179

We make a bet than System.* and Ms.Extensions.Logging.* will not have breaking API changes *even with major version bump*!
This is one of the reasons library authors should feel more comfortable taking a dependency on DS (as opposed to OTel.API which can have breaking changes with Major ver bump).

reyang: To summarize, I think we have the following options:

1. Bump minor version when we drop old runtime versions. And communicate it clearly (via docs or something else).
2. Bump major version when we drop old runtime versions, and communicate it clearly that OpenTelemetry .NET will never make intentional breaks in it's public API surface and data semantics. Which essentially means https://github.com/open-telemetry/opentelemetry-dotnet/labels/need-majorversion-bump will have to be solved in a backward-compatible way rather than a breaking way.
3. Bump major version when we drop old runtime versions, and have clear communication whether a major version bump was because of pure runtime version drop, or API surface break, or a combination of both.
4. Bump major version whenever we drop old runtime or break API surface, without communicating whether the verion bump was a result of runtime version deprecation or API breaks.

I'm fine with either 1, 2 or 3. I would rank them 1 > 2 > 3.
I think we should avoid 4.

alanwest: > it will need to change the current position: a major version bump is NOT going to introduce breaks to the API surface or any telemetry data semantics

Yes, fair. Personally, I think this would be a worthy goal in the long run. So basically adopting option number 2.

In the short run though, I was more considering option number 3. Version 1.x of OTel .NET has served as a learning grounds as OpenTelemetry has matured. We already have a number of things marked `Obsolete` and we're anticipating some more. I think cleaning these things up in a breaking major release will end up benefitting users in the long run.

---

Though let me ask a different question. Do we all agree we want to drop .NET Standard targets from our core packages in the future? If so, my primary intent with this announcement is to:
1) Give end users a very clear notice of our intended plan to drop .NET Standard, and
2) Some sense of a rough timeline for when we will do this.

Whether this be in a minor or major release is more of a technical detail.

alanwest: In other words, I'm proposing this announcement as our resolution to #3448.

alanwest: #3703 as an alternative resolution to #3448.

noahfalk: > To summarize, I think we have the following options: ...

IMO, option (3) is the best one. I think many people assume minor version number bump = "not a breaking change, may add features, no need to even look at release notes if I don't feel like it." Doing a minor version bump that broke people because it dropped support for the runtime they were using would be unexpected.

For option (2) I'd be careful about promises that you will never make a breaking change, it could become a real straight jacket. I'd suggest something more like you strongly value compatibility and breaking changes should be rare.

alanwest: Thank you for your guidance @noahfalk!

> Doing a minor version bump that broke people because it dropped support for the runtime they were using would be unexpected.

This is the primary reason I'm suggesting changing our practice to a major version bump under this proposal is because with the absence of .NET Standard targets, dropping support for a runtime version means users on old frameworks can not upgrade.

Previously, when we dropped `net461`, `netcoreapp3.1`, and `net5.0` targets we have done it with a minor version bump. Though, in all of these instances, users targeting these frameworks can currently still upgrade as they will fall back to our `netstandard2.x` targets. It is for this reason I was not as inclined to recommend a major version bump.

@cijothomas @CodeBlanch @utpilla Maybe we chat once more in next week's SIG meeting before choosing between this proposal or #3703? If we're set on dropping .NET Standard, we can also further discuss the topic of major vs. minor version bump.

alanwest:
Closing this in favor of #3703. Discussed in SIG meeting today to keep .NET Standard targets for now. We can always come back to the decision to remove .NET Standard targets at a later date.

PR #3744, opened 2022-10-07
Add a separate example project for Logs redaction

utpilla:
It's a common ask from customers to provide examples for log redaction.

1. The current example for showing logs redaction is incorrect.
2. It's also convoluted as we trying to showcase a lot of things at once in `extending-the-sdk` project.

## Changes
- Create a dedicated example project for logs redaction


src/OpenTelemetry.Exporter.Console/ConsoleLogRecordExporter.cs
```diff
@@ -84,7 +84,28 @@ public override ExportResult Export(in Batch<LogRecord> batch)
 
                 if (logRecord.State != null)
                 {
-                    this.WriteLine($"{"LogRecord.State:",-RightPaddingLength}{logRecord.State}");
+                    if (logRecord.State is IReadOnlyList<KeyValuePair<string, object>> listKvp)
```
cijothomas, line 87: we may want to do what OTLP exporter does.
https://github.com/open-telemetry/opentelemetry-dotnet/pull/3186#issuecomment-1104642714

CodeBlanch (approved):
LGTM

docs/logs/redaction/README.md
```diff
@@ -0,0 +1,6 @@
+# Redaction
+
+This program shows an example of how to redact sensitive information from Logs.
```
reyang, line 3:
```suggestion
This example shows how to redact sensitive information from Logs.
```

reyang (approved):

docs/logs/redaction/MyClassWithRedactionEnumerator.cs
```diff
@@ -0,0 +1,61 @@
...
+
+        public int Count => this.state.Count;
+
+        public KeyValuePair<string, object> this[int index]
+        {
+            get
+            {
+                var item = this.state[index];
+                var entryVal = item.Value;
+                if (entryVal != null && entryVal.ToString() != null && entryVal.ToString().Contains("<secret>"))
```
reyang, line 39: Consider a real world example: input contains email address `david@otel.org`, output will change it to `***@otel.org`.

docs/logs/redaction/Program.cs
```diff
@@ -0,0 +1,38 @@
...
+namespace Redaction;
+
+public class Program
+{
+    public static void Main()
+    {
+        using var loggerFactory = LoggerFactory.Create(builder =>
+            builder.AddOpenTelemetry(options =>
+            {
+                options.AddProcessor(new MyRedactionProcessor());
```
cijothomas, line 29: processors are run sync.. so it might affect latency if executing complex regex etc. But this might be okay for now. Unless we can show how to do this in Exporter (which wraps real exporter).

src/OpenTelemetry.Exporter.Console/ConsoleLogRecordExporter.cs
utpilla: Since `ConsoleExporter` is only meant to be used for local testing, I was thinking it's probably a good idea to not impose any option from our end and let the user figure out what works best for them.

docs/logs/redaction/MyClassWithRedactionEnumerator.cs
utpilla: This is probably not as straightforward to do without causing allocation.

PR #3755, opened 2022-10-12
[ASP.NET Core] Add back netstandard2.0 and 2.1 targets

vishweshbankwar:
Fixes #.

## Changes

Adding back `netstandard2.0` and `netstandard2.1` targets as discussed in SIG meeting. It was removed previously in https://github.com/open-telemetry/opentelemetry-dotnet/pull/3567


src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInMetricsListener.cs
```diff
@@ -95,6 +97,17 @@ public override void OnEventWritten(string name, object payload)
...
                     };
                 }
+#else
+                tags = new TagList
+            {
+                { SemanticConventions.AttributeHttpFlavor, context.Request.Protocol },
+                { SemanticConventions.AttributeHttpScheme, context.Request.Scheme },
+                { SemanticConventions.AttributeHttpMethod, context.Request.Method },
+                { SemanticConventions.AttributeHttpHost, host },
+                { SemanticConventions.AttributeHttpStatusCode, context.Response.StatusCode.ToString() },
```
alanwest, line 107: Could be a follow up, but the spec has been clarified a bit recently. Two attributes exist: `http.target` and `http.route`. I think we should be able to set `http.target` in this block. In the `NET6_0_OR_GREATER` block we're currently setting `http.target` but should in fact be setting `http.route`. I think we can actually set both `http.target` and `http.route`.

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#http-server-semantic-conventions

alanwest: Oh wait... this is metrics... we should _only_ be setting `http.route` _not_ `http.target`. What I said above is true for traces, though.

alanwest: This all raises another question... why are we resolving `http.route` differently between metrics and traces?

* metrics: `var target = (context.GetEndpoint() as RouteEndpoint)?.RoutePattern.RawText;`
* traces: `var template = beforeActionEventData.ActionDescriptor?.AttributeRouteInfo?.Template;`

alanwest (approved):
LGTM. There's some clean up we can do to conform to the spec with regards to `http.route`/`http.target` but that's beyond the scope of this PR.

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInMetricsListener.cs
vishweshbankwar: @alanwest - I am planning to revisit the semantic conventions to keep the instrumentations compliant as part of #3373. Will follow up on this with that.

For your last question: why are we resolving http.route differently between metrics and traces? It is because we don't have the same information available in the payload during `OnStop` event as we do in `OnCustom` event in traces. That is why the workaround is used.

alanwest: > It is because we don't have the same information available in the payload during OnStop event as we do in OnCustom event in traces. That is why the workaround is used.

For a given request, I would expect `http.route` to be the same value between both traces and metrics. Do these two different means resolve to the same value?

vishweshbankwar: That is a good question - I am not sure. I will follow up on that as well. Opened #3771 to track this.

cijothomas (approved):

Issue #2215, opened 2021-07-30
Expected behavior of HttpClientInstrumentation RecordException=true


notcool11:
# When should HttpClientInstrumentation create an Activity Event when RecordException=true??

**Describe your environment.**
n/a

Describe any aspect of your environment relevant to the question.

**What are you trying to achieve?**
I'd like to get the exception details from outbound HttpClient requests added as Activity Events when RecordException=true

Here is my startup.cs configuration:
```
            service.AddOpenTelemetryTracing((builder) => builder
                .SetResourceBuilder(
                    ResourceBuilder.
                    CreateDefault().
                    AddService("myapp").
                    AddAttributes(infrastructureTags))                    
                .AddAspNetCoreInstrumentation((options) => 
                    {
                        options.RecordException = true;
                        options.Enrich = EnrichParentSpan;
                    })                    
                .AddHttpClientInstrumentation((options) =>
                    {
                        options.RecordException = true;
                        options.Enrich = EnrichChildSpan;
                    })
                .SetSampler(new AlwaysOnSampler())
                .AddOtlpExporter(options =>
                {
                    options.Endpoint = OtlpUri;
                })
            );
```

If the parent service fails I get an Activity Event with the Exception details. I can also see that the `activityEventName` in my EnrichParentSpan method is `OnException` and it gives me access to the Exception object.

If an outbound request fails (404) the client span does not have an Activity Event with the Exception details, but it will set the `error` tag to true and the `status.code` tag to 2. When debugging my EnrichChildSpan method the `activityEventName` is `OnStopActivity` and I'll get access to the HttpResponseMessage object.

**What did you expect to see?**
I expected the behavior for inbound and outbound Exception processing to be the same, namely that if an error is returned from an outbound call it would log the same as an inbound call.

Is the expectation that I would code the response processing myself in my EnrichChildSpan method and to something like this??

```
                try
                {
                    httpResponseMessage.EnsureSuccessStatusCode();
                }
                catch (Exception exception)
                {
                    activity.RecordException(exception);                    
                    throw;
                }
```

The `HttpHandlerDiagnosticListener` has an exception method that will record the exception, but it doesn't get hit. Not sure if this is intended, thanks for the clarification.

## Additional Context

Add any other context about the problem here. If you followed an existing
documentation, please share the link to it.

lhidalgo89:
@cijothomas  @vishweshbankwar Any updates on this by chance?

PR #3761, opened 2022-10-12
[HttpClient] Add unit tests for `RecordException` case

vishweshbankwar:
Fixes #.

## Changes

Adding few tests to validate `RecordException = true` behavior in different HttpClient and HttpWebRequest methods for both .NET and .NET Framework.


test/OpenTelemetry.Instrumentation.Http.Tests/HttpClientTests.Basic.cs
```diff
@@ -462,6 +466,89 @@ public async Task HttpClientInstrumentationContextPropagation()
...
+
+            using var traceprovider = Sdk.CreateTracerProviderBuilder()
+                   .AddHttpClientInstrumentation(o => o.RecordException = true)
+                   .AddInMemoryExporter(exportedItems)
+                   .Build();
+
+            using var c = new HttpClient();
+            try
+            {
+                await c.GetAsync("https://www.invalidurl.com");
```
TimothyMothra, line 483: This is risky. 
This test may fail if someone registers this domain.

If you're just trying to have the httpclient throw an exception, I think you can use an invalid url. (eg `GetAsync("https://")` ).

If you're trying to force a network error specifically, I think you could use a non-routable ip address (eg `GetAsync("https://10.0.0.0/invalidurl")`).

vishweshbankwar: This one times out https://10.0.0.0/invalidurl. I have updated it to use something random which we use in other tests.

TimothyMothra (approved):
LGTM

cijothomas (approved):
tests LGTM!
We need to update doc to inform users about when can they except exception be recorded by the library.

cijothomas:
Ping from https://github.com/open-telemetry/opentelemetry-dotnet/issues/2215

PR #3778, opened 2022-10-18
Guard.Range now uses invariant culture for error message

Driedas:
## Changes
while working on another PR the
```
public void RangeDoubleTest()
```
kept failing. The Guard.Range method constructs an error message with the current (ambient) culture, thus yielding different results for different culture settings.
In order to reproduce, add e. g. 
```
CultureInfo.CurrentCulture = new CultureInfo("de-DE");
```
to the beginning of the test. Any culture that doesn't use the dot as the decimal separator will break the test.

I've changed the logic that calculates the error message to instead use the invariant culture explicitly (if you prefer another solution, feel free to throw this away)
```
var exMessage = message ?? string.Format(
    CultureInfo.InvariantCulture,
    "Must be in the range: [{0}{1}, {2}{3}]",
    min,
    minMessage,
    max,
    maxMessage);
```

A very minor thing, but I think tests should work regardless of the regional settings...


mic-max:
Is this the test and assert that will fail https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/test/OpenTelemetry.Tests/Internal/GuardTest.cs#L136?

And to confirm it is because in `de-DE` culture the Exception will be created with the message as: `Must be in the range: [0,1: empty, 99,9: full]` using commas instead of periods to separate the whole number and the mantissa?

Thanks for contributing! I think this is a good change that will give us consistent error messages and fix that flaky test :)

PS: Speaking of flaky tests, I think the CI failure is caused by one, please push an empty commit to rerun the CI :)

Driedas:
> Is this the test and assert that will fail https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/test/OpenTelemetry.Tests/Internal/GuardTest.cs#L136?

eeeexactly, that's the one...

>And to confirm it is because in de-DE culture the Exception will be created with the message as: Must be in the range: [0,1: empty, 99,9: full] using commas instead of periods to separate the whole number and the mantissa?

and right again, sorry, should've explained it better in the description :)

Right, will push another foobar commit...

reyang:
Might be a good idea to have CI running using different culture/locale.

reyang (approved):

Kielek (approved):

utpilla (approved):

Issue #3740, opened 2022-10-07
Missing traces after OTEL backend has been unavailable?


zornwal:
# Question/bug?
I am using OTEL .NET to send traces to a Promscale backend directly from my instrumented application (I know it's recommended to use a collector), and am seeing some odd behaviour that I don't know if is intended or not. 
Whenever the OTEL backend (Promscale) has had downtime (e.g. container being restarted), and the instrumented application can not send its traces, the application stops sending traces forever until the application is restarted.

I would expect the client to drop/lose whatever traces are being sent while the OTEL backend is down, and that the client would resume sending traces once the backend is up again. However, it seems that the client never resumes sending traces.
Is this intended? And if so, is there a way to configure it to start sending traces again?

### Nugets & Runtime
```
<PackageReference Include="OpenTelemetry.Exporter.Console" Version="1.4.0-alpha.1" />
<PackageReference Include="OpenTelemetry.Exporter.OpenTelemetryProtocol" Version="1.4.0-alpha.1" />
<PackageReference Include="OpenTelemetry.Extensions.Hosting" Version="1.0.0-rc9.5" />
<PackageReference Include="OpenTelemetry.Instrumentation.AspNetCore" Version="1.0.0-rc9.5" />
<PackageReference Include="OpenTelemetry.Instrumentation.EntityFrameworkCore" Version="1.0.0-beta.3" />
<PackageReference Include="OpenTelemetry.Instrumentation.Http" Version="1.0.0-rc9.6" />
<PackageReference Include="OpenTelemetry.Instrumentation.Runtime" Version="1.0.0" />
<PackageReference Include="OpenTelemetry.Instrumentation.SqlClient" Version="1.0.0-rc9.6" />
```

Targeting .NET 6.0

cijothomas:
I'd expect exporter to resume when endpoint is available again...
Could you collect https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/README.md#self-diagnostics ?

zornwal:
Adding self diagnostics produces the following logs, with the same error repeating at every send attempt. Restarting the instrumented application resumes exporting. Restarting the backend service has no effect.
```
Successfully opened file.
[...]

Issue #1779, opened 2021-02-02
OTLP Exporters to implement Retry mechanism


utpilla:
OTLP Exporter should follow the spec guideline to implement a retry strategy: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#retry

Grpc.Net.Client has an open issue to provide support for Retry: https://github.com/dotnet/core/issues/5493

We could wait for gRPC client to implement support for this which would make it easy for us to comply with the spec requirements for retrying.

FireRose0929:
Money now

vishweshbankwar:
This is now available with https://www.nuget.org/packages/OpenTelemetry.Exporter.OpenTelemetryProtocol/1.8.0-beta.1 as an experimental feature. Users can opt-in by setting `OTEL_DOTNET_EXPERIMENTAL_OTLP_ENABLE_INMEMORY_RETRY` environment variable to `true`.

PR #3788, opened 2022-10-19
Validate OTLP exporter continues to export after RpcException

alanwest:
Attempt to reproduce #3740.

This PR introduces a mock OpenTelemetry collector that can mimic failed requests. It's run via docker-compose, but could have been done using the in-process ASP.NET Core test framework. However, my hope is that we can expand on this to also test .NET Framework.

The test configures the mock collector to first respond with an `Unimplemented` status code followed by an `OK` on the subsequent request.


vishweshbankwar (approved):

alanwest:
> do we plant to merge this, and build on more tests based on this foundation?

Yes, I think this will provide a good foundation for testing retry #1779. We need a "bad acting" collector - one that can return error status codes. I'd like to implement retry soon.

alanwest:
> https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/test/OpenTelemetry.Tests/Shared/TestHttpServer.cs - Can we leverage this?

This is something I looked into in the past when I implemented the original integration test with the real collector. That is, my first choice was to try to avoid using docker. However, as I recall, HttpListener has limitations with regards to HTTP2 and gRPC. I'm trying to dig up my findings from the past, but no luck yet... I'd be happy to dig in again and retrace my steps.

cijothomas:
> , HttpListener has limitations with regards to HTTP2 and gRPC.

Got it, if this is specifically to cover .NET FW apps. Else we can use Asp.Net Core host, which can cover gRPC/Http2.

alanwest:
> Got it, if this is specifically to cover .NET FW apps. Else we can use Asp.Net Core host, which can cover gRPC/Http2.

Correct. the ideal non-docker solution would use ASP.NET Core host for .NET and something like HttpListner for .NET Framework.

I refreshed my memory a bit on HttpListener. On Windows, HttpListener relies on Http.Sys. Http.Sys does support Http/2 on recent versions of Windows. It requires https which creates a bit of a headache for generating test certs for the CI. But the real blocker is that HttpListener provides no way to set trailing metadata and a gRPC status code.

Perhaps we just punt on concerning ourselves with testing .NET Framework.

cijothomas:
> Perhaps we just punt on concerning ourselves with testing .NET Framework

This would be an option as well...

alanwest:
> This would be an option as well...

That's fair. I'll rework things to just test .NET using the in-process host.

alanwest:
@cijothomas decided to just merge it? Good by me, though I was going to un-dockerize it.

cijothomas:
> @cijothomas decided to just merge it? Good by me, though I was going to un-dockerize it.

Sorry I was thinking and my mouse clicked merge.. (have a new feature enable to auto-click...sometimes causes things like this)

alanwest:
> Sorry I was thinking and my mouse clicked merge.. (have a new feature enable to auto-click...sometimes causes things like this)

No worries! I'll open a separate PR with the changes.

FireRose0929 (approved):

PR #3799, opened 2022-10-20
Mark private and internal classes as sealed when appropriate

utpilla:
## Changes
- Mark private and internal classes as sealed when appropriate
- Perf benefits mentioned in this issue: https://github.com/dotnet/runtime/issues/49944


src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInMetricsListener.cs
```diff
@@ -24,7 +24,7 @@
 
 namespace OpenTelemetry.Instrumentation.AspNetCore.Implementation
 {
-    internal class HttpInMetricsListener : ListenerHandler
+    internal sealed class HttpInMetricsListener : ListenerHandler
```
utpilla, line 27: @vishweshbankwar We would have to measure it first but we could get some perf improvement by marking the [HttpListener](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs#L38) class for Traces as sealed.

Currently, a class in one of the unit tests inherits from it so I didn't mark it sealed. If there is a considerable perf improvement by marking the HttpListener class as sealed, we could try to figure out a way to run the test without inheriting from the class.

vishweshbankwar (approved):

Kielek (approved):

cijothomas (approved):

PR #3709, opened 2022-09-29
[ASP.NET Core] Fix broken instrumentation

vishweshbankwar:
Fixes #https://github.com/open-telemetry/opentelemetry-dotnet/pull/3519#issuecomment-1262783396

This surfaced a gap in our tests. From my brief investigation, the issue is coming from [here](https://github.com/dotnet/aspnetcore/blob/01177230911caab93dff106e4f0d973a06f51bf3/src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs#L61). This will enable activity creation and once the activity is created we were subscribing to `.Start` and `.Stop` events.

We probably need to test with and without `builder.Logging.ClearProviders()`. Will look in to this separately and update our tests accordingly.

## Changes

Please provide a brief description of the changes here.

For significant contributions please make sure you have completed the following items:

* [ ] Appropriate `CHANGELOG.md` updated for non-trivial changes
* [ ] Design discussion issue #
* [ ] Changes in public API reviewed

PR #3810, opened 2022-10-24
[ASP.NET Core] Clear logging for testing scenarios

vishweshbankwar:
Follow up from https://github.com/open-telemetry/opentelemetry-dotnet/pull/3709

Enabled logging causes a side effect of span [creation](https://github.com/dotnet/aspnetcore/blob/8fa956fec929cccabda11a051cfcbac94bc94944/src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs#L61) in case of ASP.NET Core even without ActivitySource listener and DiagnosticSourceListener. This change validates that the tracing works even if logging is disabled.

## Changes


test/OpenTelemetry.Instrumentation.AspNetCore.Tests/BasicTests.cs
```diff
@@ -72,7 +74,10 @@ void ConfigureTestServices(IServiceCollection services)
             // Arrange
             using (var client = this.factory
                 .WithWebHostBuilder(builder =>
-                    builder.ConfigureTestServices(ConfigureTestServices))
+                {
+                    builder.ConfigureTestServices(ConfigureTestServices);
+                    builder.ConfigureLogging(loggingBuilder => loggingBuilder.ClearProviders());
```
utpilla, line 79: The instrumentation should work even if logging is enabled, right? If that's the case, maybe have some tests where logging is enabled to test that it works then as well.

vishweshbankwar: Yes, added logging case of one of the test. It will confirm that the activity is created even with logging enabled. We probably don't need to add that for each test case. Let me know what you think.

utpilla (approved):

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/BasicTests.cs
```diff
@@ -57,8 +59,10 @@ public void AddAspNetCoreInstrumentation_BadArgs()
             Assert.Throws<ArgumentNullException>(() => builder.AddAspNetCoreInstrumentation());
         }
 
-        [Fact]
-        public async Task StatusIsUnsetOn200Response()
+        [Theory]
+        [InlineData(true)]
```
cijothomas, line 63: why only for this test we test both? why not for all of them !

cijothomas (approved):
Left a nit. maybe a good follow up if needed.

PR #3820, opened 2022-10-25
Rename AggregationType to better reflect intent

cijothomas:
Sum,Count is the only thing which is always present for a Histogram. Everything else is optional. The internal enum is updated to better reflect this.

Histogram - sum and count
HistogramWithBuckets - sum and count + buckets
HistogramWithMinMax - sum and count + Min/Max
HistogramWithMinMaxBuckets - sum and count + Min/Max + buckets

### DIFF
--- a/src/OpenTelemetry/Metrics/AggregationType.cs
+++ b/src/OpenTelemetry/Metrics/AggregationType.cs
@@ -56,21 +56,21 @@ namespace OpenTelemetry.Metrics
         /// <summary>
         /// Histogram with sum, count, buckets.
         /// </summary>
-        Histogram = 6,
+        HistogramWithBuckets = 6,
 
         /// <summary>
         /// Histogram with sum, count, min, max, buckets.
         /// </summary>
-        HistogramMinMax = 7,
+        HistogramWithMinMaxBuckets = 7,
 
         /// <summary>
         /// Histogram with sum, count.
         /// </summary>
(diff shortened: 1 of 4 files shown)


reyang (approved):

src/OpenTelemetry/Metrics/AggregationType.cs
```diff
@@ -56,21 +56,21 @@ internal enum AggregationType
         /// <summary>
         /// Histogram with sum, count, buckets.
         /// </summary>
-        Histogram = 6,
+        HistogramWithBuckets = 6,
```
utpilla, line 59: Renaming suggestion:

**NonBucketedHistogram** - sum and count
**BucketedHistogram** - sum and count + buckets
**NonBucketedHistogramWithMinMax** - sum and count + Min/Max
**BucketedHistogramWithMinMax**- sum and count + Min/Max + buckets

cijothomas: Starting everything with Histogram is good for auto-completion as well... "non-bucketed " sounds weird, as buckets are anyway optional!
image

utpilla: Auto-complete works with substrings as well.

> "non-bucketed " sounds weird, as buckets are anyway optional
I would say "non-bucketed" explicitly tells you that the buckets option was not enabled.

PR #3847, opened 2022-10-31
Fix `null` handling in counter tags

### LINKED ISSUE #3846: Counters with null tag values stay at value 1
# Bug Report

using:

* `OpenTelemetry 1.4.0-beta.2`
* `OpenTelemetry.Exporter.Prometheus.AspNetCore 1.4.0-beta.2`

Runtime version:

* `net6.0`

## Symptom

Counters of `int` or `long` stay at the value of `1` forever if one of the values for the tags passed is `null`.

The counter should increase to `2` in the Prometheus scraping endpoint after two `Add` calls to the counter, even if one of the tag values are `null`. If tag values of `null` are not supported, I would expect that to be documented somewhere and a clear `ArgumentNullException` on calling `Add`.

The counter keeps the value `1` in the Prometheus scraping endpoint.

## Reproduce

```
        // This test passes correctly.
        [Fact]
        public void CounterWithTags()
        {
            var aggregatorStore = new AggregatorStore("test", AggregationType.LongSumIncomingDelta, AggregationTemporality.Cumulative, 1024, Metric.DefaultHistogramBounds);
            aggregatorStore.Update(1, new ReadOnlySpan<KeyValuePair<string, object>>(new[] {
                new KeyValuePair<string, object>("foo", "junk"),
                new KeyValuePair<string, object>("bar", "other"),
            }));
            aggregatorStore.Update(1, new ReadOnlySpan<KeyValuePair<string, object>>(new[] {
                new KeyValuePair<string, object>("foo", "junk"),
                new KeyValuePair<string, object>("bar", "other"),
            }));

            aggregatorStore.Snapshot();

            var metricPoints = new List<MetricPoint>();
            foreach (var metricPoint in aggregatorStore.GetMetricPoints())
            {
                metricPoints.Add(metricPoint);
            }

            Assert.Single(metricPoints);
            Assert.Equal(2, metricPoints[0].GetSumLong());
        }

        // This test fails. It should pass as well.
        [Fact]
        public void CounterWithNullTags()
        {
            var aggregatorStore = new AggregatorStore("test", AggregationType.LongSumIncomingDelta, AggregationTemporality.Cumulative, 1024, Metric.DefaultHistogramBounds);
            aggregatorStore.Update(1, new ReadOnlySpan<KeyValuePair<string, object>>(new[] {
                new KeyValuePair<string, object>("foo", "junk"),
                new KeyValuePair<string, object>("bar", null),
            }));
            aggregatorStore.Update(1, new ReadOnlySpan<KeyValuePair<string, object>>(new[] {
                new KeyValuePair<string, object>("foo", "junk"),
                new KeyValuePair<string, object>("bar", null),
            }));

            aggregatorStore.Snapshot();

            var metricPoints = new List<MetricPoint>();
            foreach (var metricPoint in aggregatorStore.GetMetricPoints())
            {
                metricPoints.Add(metricPoint);
            }

            Assert.Single(metricPoints);
            Assert.Equal(2, metricPoints[0].GetSumLong());
        }
```

aristotelos:
Fix `null` handling in counter tags. This would cause the Prometheus exporter to keep the value 1 for counters that had `null` values in their tags.

Signed-off-by: Aristotelos <arisvd@gmail.com>

Fixes #3846.

## Changes

Updated the equality comparison in the internal `Tags` struct that could not handle `null` values.


reyang (approved):
LGTM assuming that the changelog would be updated by adding a link to this PR.

Kielek (approved):

src/OpenTelemetry/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+* Fix handling of `null` values for tags in counters. Previously, when the
```
cijothomas, line 5: Not just counter. this is applicable to every type of instrument.

test/OpenTelemetry.Tests/Metrics/AggregatorTest.cs
```diff
@@ -23,6 +24,50 @@ public class AggregatorTest
     {
         private readonly AggregatorStore aggregatorStore = new("test", AggregationType.HistogramWithBuckets, AggregationTemporality.Cumulative, 1024, Metric.DefaultHistogramBounds);
 
+        [Fact]
+        public void CounterWithTags()
```
cijothomas, line 28: lets use existing tests model, and add new use case with null tags. This is not specific to Counters, but applicable to every instrument.

src/OpenTelemetry/CHANGELOG.md
aristotelos: Agreed, pushed an update

test/OpenTelemetry.Tests/Metrics/AggregatorTest.cs
aristotelos: Do you mean adding `[InlineData()]` cases to `MetricApiTest.CounterAggregationTest`, `UpDownCounterAggregationTest`, `ObservableCounterWithTagsAggregationTest`, `ObservableUpDownCounterWithTagsAggregationTest`?

alanwest: Instead of adding new tests, we have an existing test that could cover things. Basically, just add an additional `counter.Add(...)` here:

https://github.com/open-telemetry/opentelemetry-dotnet/blob/a5e5ed8705e46d63102fb3026df6c891a3ea61fd/test/OpenTelemetry.Tests/Metrics/MetricAPITest.cs#L46-L57

That said, the tests you have added do have the benefit of being closer to the unit of code _actually_ being tested - i.e., the `AggregatorStore`. I'd be content with simply removing the word `Counter` from your tests. Maybe just name it something like `LookupAggregatorStoreTest`. `LookupAggregatorStore` is the name of the private method that's actually doing the work being tested.

In the set up of your test you do:
```csharp
var aggregatorStore = new AggregatorStore("test", AggregationType.LongSumIncomingDelta, AggregationTemporality.Cumulative, 1024, Metric.DefaultHistogramBounds);
```

We _could_ use `InlineData` to introduce test cases with different permutations of `AggregationType`, `AggregationTemporality`, etc, but I do not feel strongly about this since varying these things do not affect the aggregator store lookup.

cijothomas:
@utpilla @alanwest Tagging to consider this for resurrection for 1.5.

utpilla:
This issue does not exist starting 1.4.0 version of the SDK.

PR #3848, opened 2022-10-31
Add benchmark numbers and remove invalid case

cijothomas:
Removed the benchmark using CustomProperty. OTel does not use it ever.
Added numbers for quick reference.


CodeBlanch (approved):
LGTM

reyang (approved):

Issue #3393, opened 2022-06-20
Ability to use OTLP servers with self-signed certificates


ejsmith:
# Feature Request

My OTLP server has a self-signed SSL certificate, but there is no way for me to configure the `GrpcChannelOptions` in order to customize how the certificates are validated. I've read through various issues and there seems to be a desire to not expose `GrpcChannelOptions` which is fine, but there needs to be some sort of way to customize how certificates are validated. Ideally, I'd like to check the certificates thumbprint to validate it and I can also see wanting to just ignore cert errors just for the OTLP server requests.

It seems like it would be nice if the `HttpClientFactory` was used to set the `HttpClient` property on the `GrpcChannelOptions`. Then I could configure the client just for that export service.

cijothomas:
@alanwest might be able to recommend the options, if any. I believe this is same as https://github.com/open-telemetry/opentelemetry-dotnet/issues/2009

Maybe related is https://github.com/open-telemetry/opentelemetry-dotnet/issues/2120

alanwest:
> options, if any

Yea, there are no options today. 

> It seems like it would be nice if the HttpClientFactory was used to set the HttpClient property on the GrpcChannelOptions.

Agreed. This occurred to me in the past as well when the HttpClientFactory options was introduced for http/protobuf. I just haven't had a chance to experiment with this idea yet.

The option in #2009 to implement the additional configuration options required by the specification should probably be done anyways just to be consistent with other language SDKs, but as you note, it's a limited solution that doesn't quite get you what you're after.

The other option that might get you what you're after is to expose the GrpcChannelOptions from the exporter's config. We originally had it this way, but scoped it down to just the options defined in the spec prior to the 1.0 release. That said, I'd be keen on exploring the HttpClientFactory idea first.

martincostello:
Support for mTLS (#2009) was added in v1.14. Does that solve this issue?

martincostello:
Likely resolved by #2009. We can create a new issue to fill the specific gap if it didn't.

PR #3849, opened 2022-11-01
Refactor OTLP mock collector test

alanwest:
Follow up from https://github.com/open-telemetry/opentelemetry-dotnet/pull/3788#issuecomment-1289805871


test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/MockCollectorIntegrationTests.cs
```diff
@@ -0,0 +1,147 @@
...
+                }))
+           .StartAsync();
+
+        var httpClient = host.GetTestClient();
+
+        var codes = new[] { Grpc.Core.StatusCode.Unimplemented, Grpc.Core.StatusCode.OK };
+        await httpClient.GetAsync($"/MockCollector/SetResponseCodes/{string.Join(",", codes.Select(x => (int)x))}");
+
+        var exportResults = new List<ExportResult>();
+        var otlpExporter = new OtlpTraceExporter(new OtlpExporterOptions() { HttpClientFactory = () => httpClient });
```
alanwest, line 74: This won't work today. But if #3393 is addressed first, then this test should go green.

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/MockCollectorIntegrationTests.cs
```diff
@@ -0,0 +1,147 @@
...
+                               var codes = responseCodesCsv.Split(",").Select(x => int.Parse(x)).ToArray();
+                               collectorState.SetStatusCodes(codes);
+                           });
+
+                       endpoints.MapGrpcService<MockTraceService>();
+                   });
+                }))
+           .StartAsync();
+
+        var httpClient = host.GetTestClient();
```
cijothomas, line 68: can we use a normal httpclient, not the Test "httpclient"?

alanwest: Yep, we could. That's what we do in the Grpc.Net.Client tests https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/test/OpenTelemetry.Instrumentation.Grpc.Tests/GrpcServer.cs#L71.

Do we have a preference? I kinda like the TestServer/TestClient as I think it makes the test code more concise. That is, I think finding an open port, performing the ConfigureKestrel thing, etc. Not a huge deal, but it's extra noise.

alanwest: Ok, you twisted my arm 😄. Still planning on addressing #3393, but this PR is good to go now.

I hard-coded ports for now making sure to use something that doesn't conflict with our other test with a hard-coded port

https://github.com/open-telemetry/opentelemetry-dotnet/blob/be78372696cfdb4358854fdb32b859b778150661/test/OpenTelemetry.Instrumentation.AspNetCore.Tests/InProcServerTests.cs#L55

cijothomas (approved):

PR #3826, opened 2022-10-26
Metric Perf improvement  when out of MetricPoints

cijothomas:
When we are out of MetricPoints, we can exit early. Currently we do alloc + copying, *before* we confirm we have room to store the new MP. This PR does the check before doing any alloc+copy.

Simple benchmark results:

```
[Benchmark]
        public void CounterWith3LabelsHotPath()
        {
            var tag1 = new KeyValuePair<string, object>("DimName1", Guid.NewGuid().ToString());
            var tag2 = new KeyValuePair<string, object>("DimName2", Guid.NewGuid().ToString());
            var tag3 = new KeyValuePair<string, object>("DimName3", Guid.NewGuid().ToString());
            this.counter.Add(100, tag1, tag2, tag3);
        }
```

Before
```
|                    Method | AggregationTemporality |     Mean |     Error |    StdDev |   Gen0 | Allocated |
|-------------------------- |----------------------- |---------:|----------:|----------:|-------:|----------:|
| CounterWith3LabelsHotPath |             Cumulative | 1.518 us | 0.0163 us | 0.0127 us | 0.1144 |     480 B |
| CounterWith3LabelsHotPath |                  Delta | 1.516 us | 0.0134 us | 0.0112 us | 0.1144 |     480 B |
```
After
```
|                    Method | AggregationTemporality |     Mean |     Error |    StdDev |   Gen0 | Allocated |
|-------------------------- |----------------------- |---------:|----------:|----------:|-------:|----------:|
| CounterWith3LabelsHotPath |             Cumulative | 1.196 us | 0.0075 us | 0.0067 us | 0.0687 |     288 B |
| CounterWith3LabelsHotPath |                  Delta | 1.200 us | 0.0158 us | 0.0148 us | 0.0687 |     288 B |
```

PR #3867, opened 2022-11-04
Minor perf tweaks in Metrics

cijothomas:
Same fix as https://github.com/open-telemetry/opentelemetry-dotnet/pull/3826 applied to single tag case. (To check if we have MetricPoints available, before alloc+copy)

Minor adjustments to the benchmarks (not related to this PRs perf fix)


CodeBlanch (approved):
LGTM

PR #3883, opened 2022-11-09
Add Retries to gRPC version of OTLP Exporter

mic-max:
Towards #1779

## Changes

Enables retry for `OTLP Exporter gRPC` by default.

The default values used to initialize `Grpc.Net.Client.Configuration.RetryPolicy` were taken from the Java SIG's implementation.
  - RetryMaxAttempts = 5
  - RetryInitialBackoff = 1 second
  - RetryMaxBackoff = 5 seconds
  - RetryBackoffMultiplier = 1.5x

- `RetryableStatusCodes` taken from specification of which are retryable
- Added several unit tests with new mock OTLP collector thanks to @alanwest for creating it, letting me know and helping me onboard to it.


src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpExporterOptions.cs
```diff
@@ -179,6 +179,35 @@ public Uri Endpoint
...
 
+        /// <summary>
+        /// Gets or sets the maximum number of call attempts. This value
+        /// includes the original attempt. Used to set
+        /// <c>Grpc.Net.Client.Configuration.RetryPolicy.MaxAttempts</c>.
+        /// </summary>
+        /// <remarks>
+        /// Setting this value to less than 2 will disable retry attempts.
+        /// </remarks>
+        public int RetryMaxAttempts { get; set; } = 5;
```
alanwest, line 190: Do we want to expose these options publicly at this time? I summarized some thoughts and findings awhile back here with the opentelemetry-js project https://github.com/open-telemetry/opentelemetry-js/issues/1233.

Most notably, Java has implemented retry, but has left the configuration internal since the spec has not decided on defaults.

vishweshbankwar: Since the spec does not specify these details it would be better to keep these non-public for now. @alanwest should retries be enabled by default? or there should be some option to enable it?

alanwest: My opinion is to enable it by default. The spec for OTLP/gRPC seems reasonably clear that clients should have a retry mechanism by default. https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#failures

Namely these two statements:

> The server MUST indicate retryable errors using code [Unavailable](https://godoc.org/google.golang.org/grpc/codes) and MAY supply additional [details via status](https://godoc.org/google.golang.org/grpc/status#Status.WithDetails) using [RetryInfo](https://github.com/googleapis/googleapis/blob/6a8c7914d1b79bd832b5157a09a9332e8cbd16d4/google/rpc/error_details.proto#L40) containing 0 value of RetryDelay.

and

> The server MAY use other gRPC codes to indicate retryable and not-retryable errors if those other gRPC codes are more appropriate for a particular erroneous situation. The client SHOULD interpret gRPC status codes as retryable or not-retryable according to the following table:

As for finer-grained configuration of the retry policy, I think we leave that TBD until and if the spec adopts configuration settings.

mic-max: Spec says "Transient errors MUST be handled with a retry strategy" https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#retry so I feel like it should be enabled by default. Maybe not enabled by default yet though..

I think it's fair to leave out the options for now and enable as users ask for more control over the retry logic.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpExporterOptionsExtensions.cs
```diff
@@ -42,7 +48,38 @@ public static Channel CreateChannel(this OtlpExporterOptions options)
...
+                MaxBackoff = RetryMaxBackoff,
+                BackoffMultiplier = RetryBackoffMultiplier,
+
+                // See table showing gRPC codes and if they are retryable:
+                // https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#failures
+                RetryableStatusCodes =
+                {
+                    StatusCode.Cancelled,
+                    StatusCode.DeadlineExceeded,
+                    StatusCode.ResourceExhausted, // TODO: Investigate this one.
```
vishweshbankwar, line 64: looks like we need to follow https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#otlpgrpc-throttling for this one.

alanwest: My initial assumption is that the gRPC client handles this. Can probably validate the behavior conforms with the specs requirements with the mock server.

vishweshbankwar: looks like it does https://github.com/grpc/grpc-dotnet/blob/23047e0d498501dc95bb58b010ed3c11f8cb3cf8/src/Grpc.Net.Client/Internal/Retry/RetryCall.cs#L190
if I am looking at right place 😄

Issue #3421, opened 2022-06-30
Use `Grpc.Net.Client` with `netstandard2.0` for `OpenTelemetry.Exporter.OpenTelemetryProtocol`


ismaelhamed:
# Feature Request

**Is your feature request related to a problem?**

The old `Grpc` package copies `libgrpc_csharp_ext.x64.dylib` and `libgrpc_csharp_ext.x86.so` all over the place. That's ~17MB of unnecessary libs being copied to all of our the services that references `OpenTelemetry.Exporter.OpenTelemetryProtocol` in our solution. 

**Describe the solution you'd like:**

It should be possible to use `Grpc.Net.Client` with `netstandard2.0` instead. In fact, in theory, any `net462` or greater app should be able to reference `netstandard2.0`.

Yun-Ting:
I've chatted with @alanwest, and we are currently thinking to split the existing OTLP package into 2:
1. OpenTelemetry.Exporter.OpenTelemetryProtocol.Grpc, and
2. OpenTelemetry.Exporter.OpenTelemetryProtocol.Http

Please refer to the comment for more details:
https://github.com/open-telemetry/opentelemetry-dotnet/pull/3885#discussion_r1018557835

PR #3885, opened 2022-11-09
Phased out Grpc package for OTLP Exporter.

Yun-Ting:
Fixes https://github.com/open-telemetry/opentelemetry-dotnet/issues/3421.

## Changes
- Removed Grpc package.
- Added `HttpClient` as a hard requirement for configuring `GrpcChannelOptions`.


src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OpenTelemetry.Exporter.OpenTelemetryProtocol.csproj
```diff
@@ -7,26 +7,12 @@
...
-
-  <ItemGroup Condition="'$(TargetFramework)' == 'net462'">
-    <PackageReference Include="Grpc" Version="$(GrpcPkgVer)" />
-  </ItemGroup>
-
   <ItemGroup>
     <ProjectReference Include="$(RepoRoot)\src\OpenTelemetry\OpenTelemetry.csproj" />
     <PackageReference Include="Google.Protobuf" Version="$(GoogleProtobufPkgVer)" />
     <PackageReference Include="Grpc.Tools" Version="$(GrpcToolsPkgVer)" PrivateAssets="all" />
+    <PackageReference Include="Grpc.Net.Client" Version="$(GrpcNetClientPkgVer)" />
```
alanwest, line 14: I think the primary concern here is that this is probably a breaking change for some users. Consider .NET Framework users for whom things are working just fine today, but with this change they now have to configure the HttpClient for things to work. Worst part is that they'll only find this out at runtime.

alanwest: One idea we'd discussed in the past is to introduce separate OTLP/gRPC and OTLP/HTTP exporter packages. Something like

* `OpenTelemetry.Exporter.OpenTelemetryProtocol.Grpc`: This package would _only_ depend on the new `Grpc.Net.Client` as you've done here. The benefit of it being a new package is that it won't break existing users.
* `OpenTelemetry.Exporter.OpenTelemetryProtocol.Http`: This could be done as a separate effort. Mostly just provides symmetry, but also reduces the number of transitive dependencies for users who don't want to use gRPC.

PR #3907, opened 2022-11-15
[HttpClient] Add `net.peer.name` and `net.peer.port` on metric dimensions

vishweshbankwar:
Towards #3373 

## Changes

Added dimensions as per [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/http-metrics.md#attributes)

### DIFF
--- a/src/OpenTelemetry.Instrumentation.Http/Implementation/HttpHandlerMetricsDiagnosticListener.cs
+++ b/src/OpenTelemetry.Instrumentation.Http/Implementation/HttpHandlerMetricsDiagnosticListener.cs
@@ -50,9 +50,6 @@ namespace OpenTelemetry.Instrumentation.Http.Implementation
                     var request = response.RequestMessage;
 
                     TagList tags;
-
-                    // TODO: This is just a minimal set of attributes. See the spec for additional attributes:
-                    // https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/http-metrics.md#http-client
                     if (!request.RequestUri.IsDefaultPort)
                     {
                         tags = new TagList
--- a/src/OpenTelemetry.Instrumentation.Http/Implementation/HttpHandlerMetricsDiagnosticListener.cs
+++ b/src/OpenTelemetry.Instrumentation.Http/Implementation/HttpHandlerMetricsDiagnosticListener.cs
@@ -49,15 +49,33 @@ namespace OpenTelemetry.Instrumentation.Http.Implementation
                 {
                     var request = response.RequestMessage;
 
+                    TagList tags;
+
                     // TODO: This is just a minimal set of attributes. See the spec for additional attributes:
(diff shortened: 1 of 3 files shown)


cijothomas (approved):

src/OpenTelemetry.Instrumentation.Http/Implementation/HttpHandlerMetricsDiagnosticListener.cs
```diff
@@ -49,15 +49,33 @@ public override void OnEventWritten(string name, object payload)
...
+                    if (!request.RequestUri.IsDefaultPort)
+                    {
+                        tags = new TagList
+                        {
+                            new KeyValuePair<string, object>(SemanticConventions.AttributeHttpMethod, HttpTagHelper.GetNameForHttpMethod(request.Method)),
+                            new KeyValuePair<string, object>(SemanticConventions.AttributeHttpScheme, request.RequestUri.Scheme),
+                            new KeyValuePair<string, object>(SemanticConventions.AttributeHttpStatusCode, (int)response.StatusCode),
+                            new KeyValuePair<string, object>(SemanticConventions.AttributeHttpFlavor, HttpTagHelper.GetFlavorTagValueFromProtocolVersion(request.Version)),
+                            new KeyValuePair<string, object>(SemanticConventions.AttributeNetPeerName, request.RequestUri.Host),
+                            new KeyValuePair<string, object>(SemanticConventions.AttributeNetPeerPort, request.RequestUri.Port),
```
Kielek, line 65: from the spec: `Conditionally Required: If not default (80 for http, 443 for https).`

IMO worth to check if it is really needed to add this attribute.

cijothomas: this is already doing that right? port is only added if non-default.

vishweshbankwar: @Kielek - I am checking if it is not a default port first. do you mean the port should not be added at all?

Kielek: Sure... I have overlooked condition. Sorry!

src/OpenTelemetry.Instrumentation.Http/Implementation/HttpHandlerMetricsDiagnosticListener.cs
```diff
@@ -49,15 +49,33 @@ public override void OnEventWritten(string name, object payload)
                 {
                     var request = response.RequestMessage;
 
+                    TagList tags;
+
                     // TODO: This is just a minimal set of attributes. See the spec for additional attributes:
```
cijothomas, line 54: this todo can be removed now, right?

vishweshbankwar: done

PR #4034, opened 2022-12-20
[main-1.3.1] Update publish-package workflow

utpilla:
## Changes
- Update publish-package workflow to match the main branch's workflow


alanwest (approved):
I'm 92% sure this is going to do what you want.

PR #4039, opened 2022-12-22
Refactor unit tests

utpilla:
## Changes
- Update Moq version
- Add using statement for IDisposable objects
- Use method name as ActivitySource


TimothyMothra (approved):

TimothyMothra:
I think the activitySourceName is a good change 👍

PR #4040, opened 2022-12-22
Allow tests from different classes within an assembly to run in parallel

utpilla:
## Changes
- Updated `xunit.runner.json` to allow for tests from different classes within an assembly to run in parallel
- Kept the `xunit.runner.json` file to conveniently make changes in future if required
- Added `AssemblyInfo.cs` to AspNetCore, Http, OTLP, and OpenTelemetry test projects to exclude them from having their tests from different classes run in parallel. The tests in these projects are not setup to be allowed to run in parallel.

This has reduced the run time for the build-test workflow for Windows by one minute.

Note: The changes in this PR do not enable tests from different assemblies to run in parallel. The tests for each assembly would still be run sequentially. It only allows for tests from different tests collections within the same assembly to be run in parallel. Here's the link to the xUnit docs for more information: https://xunit.net/docs/running-tests-in-parallel

### DIFF
--- a/build/xunit.runner.json
+++ b/build/xunit.runner.json
@@ -1,4 +1,2 @@
-﻿{
-  "maxParallelThreads": 1,
-  "parallelizeTestCollections": false
+{
 }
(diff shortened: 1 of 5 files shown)


TimothyMothra (approved):
LGTM. Thanks for the clear explanation in the description!

CodeBlanch (approved):

utpilla:
It looks like the tests are still flaky 😀

Kielek (approved):

utpilla:
It looks like more test projects have to be configured to not allow parallelization. The amount of time saved is probably not worth the new config changes.

PR #4060, opened 2023-01-06
Aggregate count, sum, min, and max for exponential histograms

alanwest:
This is the first small step in breaking up #4053.

This doesn't ship anything public yet. It only adds aggregating count, sum, min and max for exponential histograms. Implementation mirrors what we have for explicit bounds histograms.

NOT in this PR:
* Updating and snapshotting the actual buckets.
* Exemplar support for exponential histograms.
* Further optimizing MetricPointOptionalComponents to perhaps make use of `StructLayout` for exponential vs. explicit-bounds histogram. (this was an offline discussion with @CodeBlanch and @cijothomas)
* Attempting to reduce the gargantuan switch statements in MetricPoint. @CodeBlanch had the thought that much work could be done in the constructor to set function pointers for a MP's update and snapshot methods. This might reduce the number of lengthy switch statements we currently have.


src/OpenTelemetry/Metrics/MetricPoint.cs
```diff
@@ -466,6 +496,58 @@ internal void Update(double number)
...
+                        break;
+                    }
+
+                case AggregationType.ExponentialHistogram:
+                    {
+                        var sw = default(SpinWait);
+                        while (true)
+                        {
+                            if (Interlocked.Exchange(ref this.exponentialBucketHistogram.IsCriticalSectionOccupied, 1) == 0)
+                            {
```
utpilla, line 508: Call `this.exponentialHistogram.Record(value)`?

src/OpenTelemetry/Metrics/MetricPoint.cs
```diff
@@ -726,6 +808,41 @@ internal void TakeSnapshot(bool outputDelta)
...
+                                // Lock acquired
+                                this.snapshotValue.AsLong = this.runningValue.AsLong;
+                                this.exponentialBucketHistogram.SnapshotSum = this.exponentialBucketHistogram.RunningSum;
+                                this.exponentialBucketHistogram.SnapshotMin = this.exponentialBucketHistogram.RunningMin;
```
utpilla, line 825: We could avoid updating `SnapshotMin`, `SnapShotMax`, `RunningMin`, `RunningMax` for `case AggregationType.ExponentialHistogram:`.

src/OpenTelemetry/Metrics/Metric.cs
```diff
@@ -25,6 +25,8 @@ public sealed class Metric
     {
         internal static readonly double[] DefaultHistogramBounds = new double[] { 0, 5, 10, 25, 50, 75, 100, 250, 500, 750, 1000, 2500, 5000, 7500, 10000 };
 
+        internal static readonly int DefaultExponentialHistogramMaxBuckets = 160;
```
utpilla, line 28: Use `const`?

src/OpenTelemetry/Metrics/MetricPoint.cs
alanwest: My plan was to do this in the next PR.

utpilla (approved):

src/OpenTelemetry/Metrics/Metric.cs
```diff
@@ -126,7 +128,7 @@ public sealed class Metric
                 throw new NotSupportedException($"Unsupported Instrument Type: {instrumentIdentity.InstrumentType.FullName}");
             }
 
-            this.aggStore = new AggregatorStore(instrumentIdentity.InstrumentName, aggType, temporality, maxMetricPointsPerMetricStream, histogramBounds ?? DefaultHistogramBounds, tagKeysInteresting);
+            this.aggStore = new AggregatorStore(instrumentIdentity.InstrumentName, aggType, temporality, maxMetricPointsPerMetricStream, histogramBounds ?? DefaultHistogramBounds, DefaultExponentialHistogramMaxBuckets, tagKeysInteresting);
```
utpilla, line 131: Just leaving a note: We would have to set the AggregationType to ExponentialHistogram in the ctor.

You might have planned this for another PR. 🙂

alanwest: Yea good call. Hadn't actually thought about this, but I'll grab it in another PR. I'll be adding tests as I go. This will certainly be caught.

src/OpenTelemetry/Metrics/MetricPoint.cs
```diff
@@ -30,6 +30,8 @@ public struct MetricPoint
 
         private HistogramBuckets histogramBuckets;
 
+        private Base2ExponentialBucketHistogram exponentialBucketHistogram;
```
cijothomas, line 33: won't this increase the size of the MetricPoint struct, and hence affect overall memory, as we pre-allocate 2k of MP?

I am facing similar challenge for Exemplars too (as storing pointer to Exemplar increases MP size!), and as a solution, I am trying to bundle histogramBuckets and exemplars into a new class, and keep its instance in MP, so MP's size remains same.

alanwest: I could try combining HistogramBuckets and Base2ExponentialBucketHistogram into a single class to keep size of MP the same.

Though, do you think it makes sense to further combine exemplars into the same class as histogram buckets? Coupling the memory consumption of enabling exemplars with the recording of a histogram seems a little weird.

cijothomas: > I could try combining HistogramBuckets and Base2ExponentialBucketHistogram into a single class to keep size of MP the same.

^ would be good.

cijothomas: > , do you think it makes sense to further combine exemplars into the same class as histogram buckets?

Most likely yes, but I'll give better answer once I finish my prototype (expect PRs on it in next day or so).

src/OpenTelemetry/Metrics/AggregationType.cs
```diff
@@ -72,5 +72,15 @@ internal enum AggregationType
         /// Histogram with sum, count, min, max.
         /// </summary>
         HistogramWithMinMax = 9,
+
+        /// <summary>
+        /// Exponential Histogram with sum, count.
```
cijothomas, line 77: its implicit that there are buckets, right?

src/OpenTelemetry/Metrics/MetricPoint.cs
```diff
@@ -234,12 +243,16 @@ internal MetricPointStatus MetricPointStatus
...
-                this.aggType != AggregationType.HistogramWithMinMax)
+                this.aggType != AggregationType.HistogramWithMinMax &&
+                this.aggType != AggregationType.Base2ExponentialHistogram &&
+                this.aggType != AggregationType.Base2ExponentialHistogramWithMinMax)
             {
                 this.ThrowNotSupportedMetricTypeException(nameof(this.GetHistogramSum));
             }
 
-            return this.mpComponents.HistogramBuckets.SnapshotSum;
+            return this.mpComponents.HistogramBuckets != null
```
cijothomas, line 253: nit: This method can return based on this.aggtype.

if aggtype = his, return sum from it
if aggtype == exp.his, return sum from it.
esle throw.

cijothomas (approved):

src/OpenTelemetry/Metrics/MetricPoint.cs
alanwest: Yes agreed this would be the more natural way to have written this, and it is what I originally attempted 😆. However, the throw happens inside of method call `this.ThrowNotSupportedMetricTypeException(nameof(this.GetHistogramSum));`, so this wouldn't compile as is. That is you'd get a "not all paths return" error.

Anyways, I don't know how much value we're getting from pushing the exception handling into a separate method. Would probably read better without it.

PR #4063, opened 2023-01-06
Fix duplicate instrument registration with Views

utpilla:
## Changes
- Add the missing `continue` statement when a duplicate `MetricStreamName` is found when using Views. There were multiple registrations being made for the same `MetricStreamName`
- Switch the order of checks to perform the lesser computationally expensive operation to find an existing`MetricStreamName` before finding an existing `MetricStreamIdentity`

### DIFF
--- a/test/OpenTelemetry.Tests/Metrics/MetricAPITest.cs
+++ b/test/OpenTelemetry.Tests/Metrics/MetricAPITest.cs
@@ -405,7 +405,7 @@ namespace OpenTelemetry.Metrics.Tests
             // Ensure that the two instruments have different MetricStreamIdentity but the same MetricStreamName
             var meterProviderBuilder = Sdk.CreateMeterProviderBuilder()
                 .AddMeter(meter.Name)
-                .AddView("duplicateInstrumentName", new MetricStreamConfiguration { Name = "instrumentName", Description = "duplicateInstrument" }) 
+                .AddView("duplicateInstrumentName", new MetricStreamConfiguration { Name = "instrumentName", Description = "duplicateInstrument" })
                 .AddInMemoryExporter(exportedItems);
 
             using var meterProvider = meterProviderBuilder.Build();
--- a/test/OpenTelemetry.Tests/Metrics/MetricAPITest.cs
+++ b/test/OpenTelemetry.Tests/Metrics/MetricAPITest.cs
@@ -396,6 +396,43 @@ namespace OpenTelemetry.Metrics.Tests
         }
 
         [Fact]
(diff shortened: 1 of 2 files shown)


test/OpenTelemetry.Tests/Metrics/MetricAPITest.cs
```diff
@@ -395,6 +395,43 @@ public void DuplicateInstrumentRegistration_NoViews_DuplicateInstruments_Differe
...
+            using var meterProvider = meterProviderBuilder.Build();
+
+            var instrument = meter.CreateCounter<long>("instrumentName");
+            var duplicateInstrument = meter.CreateCounter<long>("duplicateInstrumentName");
+
+            instrument.Add(10);
+            duplicateInstrument.Add(20);
+
+            meterProvider.ForceFlush(MaxTimeToAllowForFlush);
+            Assert.Single(exportedItems);
```
alanwest, line 420: I would expect two items. They should have conflicting names.

See https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#view

> For each View, if the Instrument could match the instrument selection criteria:
Try to apply the View configuration. If applying the View results in [conflicting metric identities](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/data-model.md#opentelemetry-protocol-data-model-producer-recommendations) the implementation SHOULD apply the View and emit a warning.

utpilla:
The existing behavior is expected. Refer to https://github.com/open-telemetry/opentelemetry-dotnet/pull/4063#discussion_r1063873477.

test/OpenTelemetry.Tests/Metrics/MetricAPITest.cs
alanwest: Though _technically_ there's this annoying statement [here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/data-model.md#opentelemetry-protocol-data-model-producer-recommendations)

> If the potential conflict involves a non-identifying property (i.e., description), the producer SHOULD choose the longer string.

We do not do this today.

alanwest: The areas of the spec concerning views and instrument identity still have some rough edges.  The statement above declares description a "non-identifying property". This seems to contradict [this part of the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument) that states:

> Instruments are identified by all of these fields.

That is name, kind, unit, _and_ description are identifying.

My head hurts now. I hope you have a good weekend.

PR #4066, opened 2023-01-09
Support customized OTLP endpoint for log example

ThomsonTan:
## Changes

The current OTLP example doesn't accept endpoint from command line. This PR add the support following the convention from OTLP trace/metrics exporter.

### DIFF
--- a/examples/Console/TestLogs.cs
+++ b/examples/Console/TestLogs.cs
@@ -47,7 +47,7 @@ namespace Examples.Console
                          * Open another terminal window at the examples/Console/ directory and
                          * launch the OTLP example by running:
                          *
-                         *     dotnet run logs --useExporter otlp
+                         *     dotnet run logs --useExporter otlp -e http://localhost:4317
                          *
                          * The OpenTelemetry Collector will output all received logs to the stdout of its terminal.
                          *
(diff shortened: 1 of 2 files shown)


examples/Console/TestLogs.cs
```diff
@@ -58,28 +58,29 @@ internal static object Run(LogsOptions options)
...
-                            });
                         }
+
+                        opt.AddOtlpExporter(otlpOptions =>
+                        {
+                            otlpOptions.Protocol = protocol;
+                            if (!string.IsNullOrWhiteSpace(options.Endpoint))
+                            {
+                                otlpOptions.Endpoint = new Uri(options.Endpoint);
+                            }
```
alanwest, line 82: Hmm, I guess TestMetrics.cs is also missing this functionality. At a minimum, probably makes sense to add it there too. Though, seems like there may be an opportunity to share some code between these examples, but maybe makes sense to consider that separately.

ThomsonTan: Added the same `Endpoint` command line option to metrics example. I did some minor cleanup to remove dup code, but agree that further cleanup could be done separately.

alanwest (approved):
LGTM

reyang (approved):

PR #4095, opened 2023-01-21
fix: change Environment Var parsing to not throw FormatException (part1)

TimothyMothra:
Towards #3690

> According the [OTel spec](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.13.0/specification/sdk-environment-variables.md?plain=1#L24):
> 
> > For variables accepting an enum value, if the user provides a value the SDK does not recognize, the SDK MUST generate a warning and gracefully ignore the setting.

In this PR, I changed the TryParse to NOT throw `FormatException` and instead log to EventSource.
Unit tests and documentation will be fixed in a follow up PR to keep the PR a reasonable size.

## Changes
- `ConfigurationExtensions`
  remove throw FormatException and instead log a Warning in EventSource.
- temporarily disable unit tests.

  

### DIFF
--- a/src/OpenTelemetry/Internal/ConfigurationExtensions.cs
+++ b/src/OpenTelemetry/Internal/ConfigurationExtensions.cs
@@ -67,7 +67,6 @@ internal static class ConfigurationExtensions
         if (!Uri.TryCreate(stringValue, UriKind.Absolute, out value))
         {
             OpenTelemetrySdkEventSource.Log.InvalidEnvironmentVariable(key, stringValue);
-            value = default;
             return false;
         }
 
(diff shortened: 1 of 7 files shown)


src/OpenTelemetry/Internal/ConfigurationExtensions.cs
```diff
@@ -66,7 +66,9 @@ internal static class ConfigurationExtensions
 
         if (!Uri.TryCreate(stringValue, UriKind.Absolute, out value))
         {
-            throw new FormatException($"{key} environment variable has an invalid value: '{stringValue}'");
+            OpenTelemetrySdkEventSource.Log.InvalidEnvironmentVariable(key, stringValue);
+            value = default;
```
reyang, line 70: Should `value` be touched here? (I guess no)

TimothyMothra: good catch! it's an out param so it needs to be set, but it's already been set by the Try method in the if statement.

reyang (approved):
LGTM.

cijothomas (approved):

PR #4102, opened 2023-01-25
feat: generate semconv v1.17.0

### LINKED ISSUE #4101: SemanticConventions attribute name constants differ from spec
# Bug Report

OpenTelemetry.SemanticConventions 1.0.0-rc9.9

## Symptom

I notice that some of the constants in the package don't match the values in the spec; for example [`public const string AttributeMessagingMessageId = "messaging.message_id";`](https://github.com/open-telemetry/opentelemetry-dotnet/blob/bb0c381478a5d6b6a3dd52cc9990d3582e908767/src/OpenTelemetry.SemanticConventions/Trace/TraceSemanticConventions.cs#L695) but the spec refers to [`messaging.message.id`](https://github.com/open-telemetry/opentelemetry-specification/blob/3d594588815fa4564fe35984e8b1a714e54d48f2/specification/trace/semantic_conventions/messaging.md?plain=1#L203).

The [PR that introduced this](https://github.com/open-telemetry/opentelemetry-dotnet/pull/2069) says the values were autogenerated so maybe it's just out of date? I had a quick look at the [build-tools](https://github.com/open-telemetry/build-tools/tree/main/semantic-conventions) and it links to the java package which has a constant that [does match the spec](https://github.com/open-telemetry/opentelemetry-java/blob/6edba790d87782dd2d681ac90beaa51b74e95c78/semconv/src/main/java/io/opentelemetry/semconv/trace/attributes/SemanticAttributes.java#L800).

ttd2089:
Generates SemanticConventions from the v1.17.0 schemas.

Fixes #4101 

There is an open issue in the generator that caused mounting subfolders during the `docker run` steps to fail due to exception.yaml being absent. Mounting that file explicitly made the error go away but it's not clear whether that produces the desired code or just runs without an error. There are at least a few changes in the generated source that don't appear when in the spec changelog when comparing [v1.13.0 to v.1.17.0](https://github.com/open-telemetry/opentelemetry-specification/compare/v1.13.0...v1.17.0):
- code.column
- messaging.source.kind
- messaging.source.template

I'm going to take a closer look at this tomorrow but I figured I'd open the PR for visibility anyway in case someone else can make sense of the differences more quickly than I can.


src/OpenTelemetry.SemanticConventions/scripts/semantic-conventions/generate.ps1
```diff
@@ -3,9 +3,9 @@ $SCRIPT_DIR=$PSScriptRoot
 $ROOT_DIR="${SCRIPT_DIR}/../../"
```
Kielek, line 3: There is also .sh script in the same folder. It should be modified in the same way.

src/OpenTelemetry.SemanticConventions/scripts/semantic-conventions/generate.ps1
```diff
@@ -3,9 +3,9 @@ $SCRIPT_DIR=$PSScriptRoot
...
 
 # freeze the spec & generator tools versions to make SemanticAttributes generation reproducible
-$SPEC_VERSION="v1.13.0"
+$SPEC_VERSION="v1.17.0"
 $SCHEMA_URL="https://opentelemetry.io/schemas/$SPEC_VERSION"
-$GENERATOR_VERSION="0.14.0"
+$GENERATOR_VERSION="0.15.0"
```
Kielek, line 8: I see also 0.15.1 version. Does it fix mentioned issue?

ttd2089: I believe I tried both but I'll double check.

src/OpenTelemetry.SemanticConventions/scripts/semantic-conventions/generate.ps1
ttd2089: Will do.

src/OpenTelemetry.SemanticConventions/scripts/semantic-conventions/generate.ps1
ttd2089: Yea I get the same error with 0.14, 0.15, and 0.15.1; but I updated to 0.15.1 anyway.

pellared (approved):

cijothomas:
@ttd2089 Could you update the public api files, given there are new conventions now. Also add an entry to changelog to indicate the new version.
Thanks!

PR #4113, opened 2023-01-27
ExemplarPart1 - ExemplarFilter classes

cijothomas:
Merging only to main-metrics branch until main is open for contributions after 1.4 is out.

Part1 for Exemplar Support - Add ExemplarFilter classes. Not connected to Metric pipeline.

....
Part2 will add Exemplar data structure
Part3 will add ExemplarReservoirs
.....


src/OpenTelemetry/Metrics/Exemplar/AlwaysOffNoneExemplarFilter.cs
```diff
@@ -0,0 +1,34 @@
...
+// limitations under the License.
+// </copyright>
+
+namespace OpenTelemetry.Metrics;
+
+/// <summary>
+/// An ExemplarFilter which makes no measurements eligible for being an Exemplar.
+/// Using this ExemplarFilter is as good as disabling Exemplar feature.
+/// </summary>
+internal sealed class AlwaysOffNoneExemplarFilter : ExemplarFilter
```
reyang, line 23:
```suggestion
internal sealed class AlwaysOffExemplarFilter : ExemplarFilter
```

src/OpenTelemetry/Metrics/Exemplar/ExemplarFilters.cs
```diff
@@ -0,0 +1,39 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+
+namespace OpenTelemetry.Metrics;
+
+/// <summary>
+/// Provides access to built-in ExemplarFilters.
+/// </summary>
+internal static class ExemplarFilters
```
reyang, line 22:
```suggestion
internal static class BuiltinExemplarFilters
```

cijothomas: removed the extra class and made it part of ExemplarFilter itself.

src/OpenTelemetry/Metrics/Exemplar/ExemplarFilters.cs
```diff
@@ -0,0 +1,39 @@
+// <copyright file="ExemplarFilters.cs" company="OpenTelemetry Authors">
```
reyang: Do we need this at all? (maybe it would break the existing ownership model)

e.g. https://github.com/open-telemetry/opentelemetry-dotnet/blob/5680f837e277a73d7c9060ada9ed170414181d7a/docs/trace/reporting-exceptions/Program.cs#L32

https://github.com/open-telemetry/opentelemetry-dotnet/blob/5680f837e277a73d7c9060ada9ed170414181d7a/src/OpenTelemetry/Trace/TracerProviderSdk.cs#L390

src/OpenTelemetry/Metrics/Exemplar/ExemplarFilter.cs
```diff
@@ -0,0 +1,80 @@
...
+    /// and will be given to an ExemplarReservoir.
+    /// Reservoir may further sample, so a true here does not mean that this
+    /// measurement will become an exemplar, it just means it'll be
+    /// eligible for being Exemplar.
+    /// <c>false</c> to indicate this measurement is not eligible to become Exemplar
+    /// and will not be given to the ExemplarReservoir.
+    /// </returns>
+    public virtual bool ShouldSample(long value, ReadOnlySpan<KeyValuePair<string, object>> tags)
+    {
+        return false;
```
reyang, line 57: Do we want to force the derived class to provide implementation? (I suspect a default implementation which returns false might create confusion)

reyang (approved):
LGTM with minor suggestions.

src/OpenTelemetry/Metrics/Exemplar/ExemplarFilter.cs
cijothomas: making it abstract, and we need to force derived class to provide impln.

alanwest (approved):

PR #4120, opened 2023-01-31
OTLP exporter add User-Agent header

### LINKED ISSUE #3724: OTLP exporters should emit standard user agent string
As per the following change in the specification (1.14 releasing Oct 4):

> OpenTelemetry protocol exporters SHOULD emit a User-Agent header to at a minimum identify the exporter, the language of its implementation, and the version of the exporter. For example, the Python OTLP exporter version 1.2.3 would report the following:
> ```OTel OTLP Exporter Python/1.2.3```

See https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#user-agent

alanwest:
Fixes #3724


src/OpenTelemetry.Exporter.OpenTelemetryProtocol/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Include User-Agent header.
```
vishweshbankwar, line 5: Would it be helpful to include the spec [link](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#user-agent) here?

vishweshbankwar (approved):

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpExporterOptions.cs
```diff
@@ -39,11 +40,20 @@ public class OtlpExporterOptions
...
+        };
+
         internal readonly Func<HttpClient> DefaultHttpClientFactory;
 
         private const string DefaultGrpcEndpoint = "http://localhost:4317";
         private const string DefaultHttpEndpoint = "http://localhost:4318";
         private const OtlpExportProtocol DefaultOtlpExportProtocol = OtlpExportProtocol.Grpc;
+        private const string UserAgentProduct = "OTel-OTLP-Exporter-Dotnet";
+
+        private static readonly AssemblyName AssemblyName = typeof(OtlpExporterOptions).Assembly.GetName();
```
cijothomas, line 55: https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/Resources/ResourceBuilderExtensions.cs#L135 Protect inside try catch?

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpExporterOptions.cs
```diff
@@ -39,11 +40,20 @@ public class OtlpExporterOptions
...
+        {
+            new KeyValuePair<string, string>("User-Agent", UserAgentProductVersion != null ? $"{UserAgentProduct}/{UserAgentProductVersion}" : UserAgentProduct),
+        };
+
         internal readonly Func<HttpClient> DefaultHttpClientFactory;
 
         private const string DefaultGrpcEndpoint = "http://localhost:4317";
         private const string DefaultHttpEndpoint = "http://localhost:4318";
         private const OtlpExportProtocol DefaultOtlpExportProtocol = OtlpExportProtocol.Grpc;
+        private const string UserAgentProduct = "OTel-OTLP-Exporter-Dotnet";
```
cijothomas, line 53: Should we use something to indicate the new dedicated grpc exporter vs current? 
https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/src/OpenTelemetry.Exporter.OpenTelemetryProtocol.Grpc

cijothomas (approved):
left couple of  small comments.

Kielek (approved):

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpExporterOptions.cs
alanwest: Sure, good idea. Though, do you think this name is ok for this one?

The dedicated gRPC one could be `OTel-OTLP-Exporter-gRPC-Dotnet`. Ideally there will also be a dedicated HTTP one in the future as well which could be `OTel-OTLP-Exporter-HTTP-Dotnet`.

cijothomas: agree

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpExporterOptions.cs
alanwest: Ha, got me! I tried slipping this one by you. 😆 

This was on my todo list awhile back to centralize this and apply it across the repo, I'll get on this for post 1.4. In the meantime, I applied the try/catch to just this spot.

Interesting though, I hadn't realized ResourceBuilderExtensions is actually doing things a little differently than most other places.

```csharp
typeof(Resource).Assembly.GetCustomAttribute<AssemblyFileVersionAttribute>()?.Version
```

vs.

```csharp
typeof(Resource).Assembly.GetName().Version
```

Any idea why? I can look into it, but was curious if you knew off hand.

cijothomas: not able to recollect the history....

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpExporterOptions.cs
```diff
@@ -195,5 +203,18 @@ internal static void RegisterOtlpExporterOptionsFactory(IServiceCollection servi
                     configuration,
                     sp.GetRequiredService<IOptionsMonitor<BatchExportActivityProcessorOptions>>().Get(name)));
         }
+
+        private static Version GetAssemblyVersion()
+        {
+            try
+            {
+                var assemblyName = typeof(OtlpExporterOptions).Assembly.GetName();
+                return assemblyName.Version;
```
martincostello, line 212: IMHO it would be better to use either the assembly file version (blue) or even better the assembly informational version (green) instead of the assembly version (red) as it's much more granular:

image

I'd be happy to submit a PR to do that if people agree.

reyang: +1 👍

martincostello: Opened #4143.

PR #4164, opened 2023-02-07
Move namespace of IServiceCollection.AddOpenTelemetry

alanwest:
We discussed this briefly on a [previous PR](https://github.com/open-telemetry/opentelemetry-dotnet/pull/3923#discussion_r1043722341), but I'm raising it once more because there seems to be conflicting guidance for whether to place extension methods in the namespace of the thing they're extending versus some other namespace.

##### DO NOT put our extensions under Microsoft.Extensions.* namespace:
https://learn.microsoft.com/dotnet/core/extensions/options-library-authors#namespace-guidance

##### DO put our extensions under Microsoft.Extensions.* namespace:
https://learn.microsoft.com/dotnet/core/extensions/dependency-injection-usage#register-services-for-di

In general, I don't feel strongly either way. However, we do have an existing precedent for placing our extensions in the namespace of the thing they're extending. So I lean towards us being self-consistent.

##### ILoggingBuilder extensions
https://github.com/open-telemetry/opentelemetry-dotnet/blob/1273e83d557cccf0514a8941988c5fdc894afe04/src/OpenTelemetry/Logs/OpenTelemetryLoggingExtensions.cs#L26-L43

##### Prometheus exporter ApplicationBuilder extensions
https://github.com/open-telemetry/opentelemetry-dotnet/blob/1273e83d557cccf0514a8941988c5fdc894afe04/src/OpenTelemetry.Exporter.Prometheus.AspNetCore/PrometheusExporterApplicationBuilderExtensions.cs#L24-L43


CodeBlanch (approved):
Don't feel passionately that it should be `namespace OpenTelemetry` so if you feel this is better LGTM

cijothomas (approved):
Can you add changelog entry as this is breaking change when bumping from last rc

Kielek (approved):

alanwest:
Holding on this until we've settled on some of the questions about where the AddOpenTelemetry API will live...

alanwest:
Handled by #4174

Issue #3429, opened 2022-07-06
OpenTelemetry .NET SDK is not AOT safe


timmydo:
# Bug Report

When publishing my app with `dotnet publish --self-contained -p:PublishAot=true -r win-x64 -o out` and then running it, the app crashes on startup with an open telemetry stack:

```
Unhandled Exception: System.TypeInitializationException: A type initializer threw an exception. To determine which type, inspect the InnerException's StackTrace property.
 ---> System.TypeInitializationException: A type initializer threw an exception. To determine which type, inspect the InnerException's StackTrace property.
 ---> System.Reflection.MissingMetadataException: 'OpenTelemetry.Context.AsyncLocalRuntimeContextSlot<System.Int32>' is missing metadata. For more information, please visit http://go.microsoft.com/fwlink/?LinkID=392859
   at System.Reflection.Runtime.General.TypeUnifier.WithVerifiedTypeHandle(RuntimeConstructedGenericTypeInfo, RuntimeTypeInfo[]) + 0x15a
   at System.Reflection.Runtime.General.TypeUnifier.GetConstructedGenericTypeWithTypeHandle(RuntimeTypeInfo, RuntimeTypeInfo[]) + 0x3b
   at System.Reflection.Runtime.TypeInfos.RuntimeTypeInfo.MakeGenericType(Type[]) + 0x40d
   at OpenTelemetry.Context.RuntimeContext.RegisterSlot[T](String) + 0x161
   at OpenTelemetry.SuppressInstrumentationScope..cctor() + 0x1d
   at System.Runtime.CompilerServices.ClassConstructorRunner.EnsureClassConstructorRun(StaticClassConstructionContext*) + 0x16a
   --- End of inner exception stack trace ---
   at System.Runtime.CompilerServices.ClassConstructorRunner.EnsureClassConstructorRun(StaticClassConstructionContext*) + 0x26b
   at System.Runtime.CompilerServices.ClassConstructorRunner.CheckStaticClassConstructionReturnGCStaticBase(StaticClassConstructionContext*, Object) + 0x1c
   at OpenTelemetry.SuppressInstrumentationScope.get_IsSuppressed() + 0x1b
   at OpenTelemetry.Sdk.get_SuppressInstrumentation() + 0x15
   at OpenTelemetry.Logs.OpenTelemetryLoggerProvider..cctor() + 0x10
[...]

PR #4191, opened 2023-02-14
Enable AOT analyzer

reyang:
https://github.com/open-telemetry/opentelemetry-dotnet/issues/3429#issuecomment-1430537746

No rush, probably wait after 1.4 release.


build/Common.prod.props
```diff
@@ -1,6 +1,11 @@
 <Project>
   <Import Project=".\Common.props" />
 
+  <PropertyGroup>
+    <IsTrimmable>true</IsTrimmable>
+    <EnableAotAnalyzer>true</EnableAotAnalyzer>
```
eerhardt, line 6: See my comment on https://github.com/open-telemetry/opentelemetry-dotnet/issues/3429#issuecomment-1430540405. 

This isn't actually doing anything because Open Telemetry isn't targeting `net6.0`+. The analyzers only work when the `System.*` assemblies are annotated with the correct attributes. And those attributes are only added to `net6.0` and `net7.0` TFMs.

reyang: I think it is doing lots of things as the builds are failing.

And here is the target https://github.com/open-telemetry/opentelemetry-dotnet/blob/d0829fff2f229b67c884526cb4943b99f6c75f29/src/OpenTelemetry/OpenTelemetry.csproj#L4.

There are other components which might not target net6.0+, and that's something we should look into (maybe in a separate PR).

eerhardt: Sorry - I should have said it isn't going to catch everything. Out of 21 projects under `src`, only 6 of them target net6:

image

So, for example, the [issue that is blocking Open Telemetry from working in NativeAOT](https://github.com/open-telemetry/opentelemetry-dotnet/issues/3429) won't be caught by this because it occurs in `OpenTelemetry.Api` which only targets netstandard2.0 and net fx.

https://github.com/open-telemetry/opentelemetry-dotnet/blob/a878ca3eead7d1b84c3e4c379c81b665b553c41b/src/OpenTelemetry.Api/OpenTelemetry.Api.csproj#L4

TimothyMothra: We could have a build flag (ie -CompileForAnalyzers) that conditionally adds the latest framework to projects.
This would let us run other analyzers. The Microsoft.CodeAnalysis.NetAnalyzers is included in NET5+.

PR #4196, opened 2023-02-15
Update CONTRIBUTING.md with RunApiCompat

mic-max:
New guidance on enabling public api analysis.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -72,24 +72,17 @@ will be easily spotted in pull request. It will also ensure that OpenTelemetry
 doesn't expose APIs outside of the library primary concerns like a generic
 helper methods.
 
-#### How to enable and configure
+#### How to configure
 
 * Create a folder in your project called `.publicApi` with the frameworks that
   as folders you target.
 * Create two files called `PublicAPI.Shipped.txt` and `PublicAPI.Unshipped.txt`
    in each framework that you target.
-* Add the following lines to your csproj:
-
-```xml
-<ItemGroup>
-  <AdditionalFiles Include=".publicApi\$(TargetFramework)\PublicAPI.Shipped.txt" />
-  <AdditionalFiles Include=".publicApi\$(TargetFramework)\PublicAPI.Unshipped.txt" />
-</ItemGroup>
-```
-
 * Use
    [IntelliSense](https://docs.microsoft.com/visualstudio/ide/using-intellisense)
    to update the publicApi files.
+
+Enabled by default, add `<RunApiCompat>false</RunApiCompat>` to your csproj to disable.
 
 ## Pull Requests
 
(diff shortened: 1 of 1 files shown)


TimothyMothra (approved):

CONTRIBUTING.md
```diff
@@ -72,25 +72,18 @@ will be easily spotted in pull request. It will also ensure that OpenTelemetry
...
-  <AdditionalFiles Include=".publicApi\$(TargetFramework)\PublicAPI.Shipped.txt" />
-  <AdditionalFiles Include=".publicApi\$(TargetFramework)\PublicAPI.Unshipped.txt" />
-</ItemGroup>
-```
-
 * Use
    [IntelliSense](https://docs.microsoft.com/visualstudio/ide/using-intellisense)
    to update the publicApi files.
 
+Enabled by default, add `<RunApiCompat>false</RunApiCompat>` to your csproj to disable.
```
cijothomas, line 85: any reason why we want to allow this be disabled? Its probably a good thing in contrib repo, but this repo should not need such bypass options. 
If this is for temporarily bypassing api check while developing locally, be explicit about it.

mic-max: Just calling it out `how` to, I can add a more explicit comment about `why` and `when`

CONTRIBUTING.md
```diff
@@ -72,25 +72,21 @@ will be easily spotted in pull request. It will also ensure that OpenTelemetry
...
-</ItemGroup>
-```
-
 * Use
    [IntelliSense](https://docs.microsoft.com/visualstudio/ide/using-intellisense)
    to update the publicApi files.
 
+It is enabled by default. Add `<RunApiCompat>false</RunApiCompat>` to your
+`.csproj` file to temporarily disable it. This should only be done when
+bypassing api check while developing locally or until a stable version of the
```
cijothomas, line 87: all components in this repo should have api check irrespective of if they are stable.. I'd say - document this as a quick-hack-for-local-testing, and initial PR about the API shape itself.

CONTRIBUTING.md
```diff
@@ -72,25 +72,21 @@ will be easily spotted in pull request. It will also ensure that OpenTelemetry
...
-  <AdditionalFiles Include=".publicApi\$(TargetFramework)\PublicAPI.Shipped.txt" />
-  <AdditionalFiles Include=".publicApi\$(TargetFramework)\PublicAPI.Unshipped.txt" />
-</ItemGroup>
-```
-
 * Use
    [IntelliSense](https://docs.microsoft.com/visualstudio/ide/using-intellisense)
    to update the publicApi files.
 
+It is enabled by default. Add `<RunApiCompat>false</RunApiCompat>` to your
```
Kielek, line 85: As  know PublicApiAnalyzer != RunApiCompat.

You always can execute PublicApiAnalyzer. RunApiCompact, only if previous (stable) package was released for givem TFM.

I think that it could be misleading oyttung ut under PublicAPI header.

Issue #4193, opened 2023-02-15
Integrate Testcontainers into OpenTelemetry .NET tests


HofmeisterAn:
# Feature Request

👋 OpenTelemetry maintainers and community! I wanted to introduce myself as one of the Testcontainers maintainers, working primarily on the .NET implementation. I know that Testcontainers has been adopted in some of the other OpenTelemetry implementations, such as [Go](https://github.com/search?q=org%3Aopen-telemetry+Testcontainers+lang%3AGo+&type=code) and [Java](https://github.com/search?q=org%3Aopen-telemetry+Testcontainers+lang%3AJava&type=code), and I believe that the .NET version could benefit from it as well.

For those who may not be familiar, Testcontainers is an open-source library that makes it easy to spin up and manage disposable Docker containers for testing. By using Testcontainers, you can run your tests in an environment that closely mimics production, without having to manually manage infrastructure or worry about the dependencies of your application. This can lead to more reliable and reproducible test results.

An example configuration for a PostgreSQL or Redis container look like:

```csharp
_postgreSqlContainer = new ContainerBuilder()
    .WithImage("postgres:15.1")
    .WithPortBinding(5432, true)
    .WithEnvironment("POSTGRES_DB", "postgres")
    .WithEnvironment("POSTGRES_USER", "postgres")
    .WithEnvironment("POSTGRES_PASSWORD", Guid.NewGuid().ToString())
    .WithWaitStrategy(Wait.ForUnixContainer().UntilCommandIsCompleted("pg_isready"))
    .Build();

_redisContainer = new ContainerBuilder()
    .WithImage("redis:7.0")
    .WithPortBinding(6379, true)
    .WithWaitStrategy(Wait.ForUnixContainer().UntilCommandIsCompleted("redis-cli", "ping"))
    .Build();
```

Please note that we will be supporting pre-configured modules (e.g. [PostgreSQL](https://github.com/testcontainers/testcontainers-dotnet/blob/fe9f55a84341b57f5e054e631043f590dd2cc74c/tests/Testcontainers.PostgreSql.Tests/PostgreSqlContainerTest.cs#L5)), including best practices in the future. This will reduce the configuration to a bare minimum, typically requiring only a single line:

```csharp
_redisContainer = new RedisBuilder().Build();
_postgreSqlContainer = new PostgreSqlBuilder().Build();
```

[...]

PR #4198, opened 2023-02-16
feat: Enable SuccessfulCommandTest by leveraging Testcontainers

HofmeisterAn:
Relates #4193.

## Changes

I am submitting this pull request to enable `SuccessfulCommandTest` by leveraging Testcontainers. The changes in this PR spin up instances of Microsoft SQL Server that the test can run against. By doing so, we can achieve a broader coverage, reliable and reproducible test results.

With the use of Testcontainers, we can be sure that the test environment is always in the same state, regardless of the underlying host system. This is achieved by running the test against a container, which is created on-demand and destroyed after the test has completed. In addition, it allows the test to run in isolation, reducing the risk of interference from other processes or applications.

To run the tests with Testcontainers, a Moby compatible engine such as Docker is required.


cijothomas:
@HofmeisterAn Thanks. Could you sign the CLA agreement, which is a pre-req before we can consider any contributions. The link from the CLA bot comment has instructions. Your employer might need to be consulted for any company specific instructions.

HofmeisterAn:
I was able to integrate Testcontainers, enable and run the tests within just a few minutes. However, I was not aware of the current pipeline configuration, which adds unnecessary overhead and requires additional configuration when using it with Testcontainers. I'm not sure how we should proceed with the tests. I think the best approach is to move away from the Docker Compose setup and simply run the tests.

Regarding the issue with strong-named assemblies, I am aware of it. The issue has been [addressed](https://github.com/dotnet/Docker.DotNet/pull/567), and I am waiting for a release of Docker.DotNet to resolve it. Although it may not be critical for tests, it's good to know that it has been taken care of.

HofmeisterAn:
@cijothomas any feedback how to continue?

cijothomas:
> @cijothomas any feedback how to continue?

Lets wait for the signing fix, so we can make the CI green before proceeding further.
I like the overall approach, and yes, we can get rid of existing manual docker spin ups once we fully move to this.

HofmeisterAn:
> System.ArgumentException : Cannot detect the Docker endpoint. Use either the environment variables or the ~/.testcontainers.properties file to customize your configuration:

This issue relates to the Docker Compose environment. The container created by Compose does not include Docker itself. We can either adjust the [Compose configuration](https://dotnet.testcontainers.org/examples/compose/) (Docker Desktop requires the env variable too) by mounting the daemon socket, or we can remove Compose since it is not needed. In that case, you can simply run the tests.

.github/workflows/integration.yml
```diff
@@ -20,8 +20,8 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - name: Run sql docker-compose.integration
-        run: docker-compose --file=test/OpenTelemetry.Instrumentation.SqlClient.Tests/docker-compose.yml --file=build/docker-compose.${{ matrix.version }}.yml --project-directory=. up --exit-code-from=tests --build
+      - name: Run sql integration
+        run: dotnet test test/OpenTelemetry.Instrumentation.SqlClient.Tests/OpenTelemetry.Instrumentation.SqlClient.Tests.csproj --framework ${{ matrix.version }} --filter CategoryName=SqlIntegrationTests
```
alanwest, line 24: We can probably get rid of the separate workflow step now as these tests should just run with the rest of the unit tests.

test/OpenTelemetry.Instrumentation.SqlClient.Tests/SqlClientTests.cs
```diff
@@ -35,18 +36,29 @@ public class SqlClientTests : IDisposable
...
-        private const string SqlConnectionStringEnvVarName = "OTEL_SQLCONNECTIONSTRING";
         private const string TestConnectionString = "Data Source=(localdb)\\MSSQLLocalDB;Database=master";
 
-        private static readonly string SqlConnectionString = SkipUnlessEnvVarFoundTheoryAttribute.GetEnvironmentVariable(SqlConnectionStringEnvVarName);
-
         private readonly FakeSqlClientDiagnosticSource fakeSqlClientDiagnosticSource;
 
+        // Please note this image is not compatible with macOS running on Apple silicon.
+        // Maybe we should replace it with mcr.microsoft.com/azure-sql-edge (the SqlEdge module).
+        private readonly MsSqlContainer mssqlContainer = new MsSqlBuilder().Build();
```
alanwest, line 45: If `mcr.microsoft.com/azure-sql-edge` works on Apple silicon, we could use an expression like the following to decide which image to use

```csharp
RuntimeInformation.IsOSPlatform(OSPlatform.OSX) 
&& RuntimeInformation.ProcessArchitecture == Architecture.Arm64
```

alanwest (commented):
@HofmeisterAn I like this! I've been curious about Testcontainers but hadn't had a chance to try them out yet.

.github/workflows/integration.yml
HofmeisterAn: But you still need the filter. I have tried running all tests without the filter, but the tests never complete. It seems like the remaining tests are waiting for the Docker Compose environment to be available. Please note that the filter already exists and is being used [here](https://github.com/open-telemetry/opentelemetry-dotnet/blob/d924663dc3d6bfdf737bc49ceaa1addcec90a2d6/test/OpenTelemetry.Instrumentation.SqlClient.Tests/docker-compose.yml#L20), and was not introduced by me. I believe that there is another stage where the remaining tests run (see the following [Dockerfile](https://github.com/open-telemetry/opentelemetry-dotnet/blob/d924663dc3d6bfdf737bc49ceaa1addcec90a2d6/test/OpenTelemetry.Instrumentation.SqlClient.Tests/Dockerfile)). Thanks to Testcontainers, all these configurations are not necessary anymore.

Please note that the initial goal was to enable the disabled SQL tests and not to refactor all tests, etc. in the first place.

test/OpenTelemetry.Instrumentation.SqlClient.Tests/SqlClientTests.cs
HofmeisterAn: Right now, it requires a bit more abstraction. The modules do not share a common database base class or interface, but this is not much of a problem. Why not use Azure SQL Edge for all platforms?

.github/workflows/integration.yml
HofmeisterAn: Oh, my bad. I believe you are correct in that the tests are already run with `build-test`. However, they fail on Windows because the Docker image is not available (compatible) for the Windows engine.

alanwest: > However, they fail on Windows because the Docker image is not available (compatible) for the Windows engine.

I still haven't had a chance to look too closely at TestContainers for .NET. Is it that they just aren't compatible with Windows at all? Or is it specifically the SQL image that is not compatible?

alanwest: The original integration tests only ran on linux anyways. So maybe what we could do is introduce a new xUnit attribute that will skip the test if not linux. Something similar to the `SkipUnlessEnvVarFoundTheory` but instead `SkipIfNotLinux`.

HofmeisterAn: > I still haven't had a chance to look too closely at TestContainers for .NET. Is it that they just aren't compatible with Windows at all? Or is it specifically the SQL image that is not compatible?

Testcontainers for .NET supports both. I ran it primarily with WCOW to test legacy Windows applications and MSIs. This particular case is all about the Docker image. Multi arch images that follow the [Windows container version compatibility](https://learn.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/version-compatibility) run fine.

> The original integration tests only ran on linux anyways. So maybe what we could do is introduce a new xUnit attribute that will skip the test if not linux.

👍

test/OpenTelemetry.Tests/Shared/SkipOnWindowsRunnerTheoryAttribute.cs
```diff
@@ -0,0 +1,36 @@
...
+/// The GitHub-hosted Windows runners do not support Linux containers. This <see cref="TheoryAttribute" /> skips tests running on Windows runners.
+/// </summary>
+internal class SkipOnWindowsRunnerTheoryAttribute : TheoryAttribute
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="SkipOnWindowsRunnerTheoryAttribute" /> class.
+    /// </summary>
+    public SkipOnWindowsRunnerTheoryAttribute()
+    {
+        if ("Windows".Equals(Environment.GetEnvironmentVariable("RUNNER_OS"), StringComparison.OrdinalIgnoreCase))
```
alanwest, line 31: For tests with this attribute, what will happen when I run them locally?

HofmeisterAn: Hm, it looks like `RUNNER_OS` does not return what the [docs](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables) says, I will check.

alanwest: Ideally, the attribute causes the test to be skipped in all scenarios where either the environment is not compatible (windows, mac) or docker is not running.

HofmeisterAn: The only two cases are Docker is not running or Windows engine is enabled (very rare, macOS is supported too). If you like I can add something like we are doing in Testcontainers. We check if the Docker CLI response with the correct [platform](https://github.com/testcontainers/testcontainers-dotnet/blob/710987a2562f7bb5a58d64bb74cba17cc8681db4/tests/Testcontainers.Commons/DockerCli.cs#L44-L48). This will only enable the tests for environments that are capable of. WDYT?

HofmeisterAn: Lets see if the recent changes work better. BTW, I was trying a few times to test the changes in advance (on CI too), but I have serious issues building the forked version. I run into [this](https://github.com/open-telemetry/opentelemetry-dotnet/pull/3412) issue all the time.

cijothomas:
@HofmeisterAn Consider making a small PR into this repo and get it merged, so that you'll no longer considered "New contriutor". (New contributors require one of the maintainers to trigger the CI builds adding delays to you ability to test quikly)

test/OpenTelemetry.Tests/Shared/SkipOnWindowsRunnerTheoryAttribute.cs
alanwest: You'll probably need fetch tags to your fork. This should allow minver to properly version assemblies.

HofmeisterAn: Ok, I think we need to move the tests to another class, because the remaining tests trigger xUnit.net's `IAsyncLifetime` again. That's why we see the issues on Windows.

alanwest: Gotcha, yea moving the this one test to something like `SqlClientIntegrationTests.cs` would be fine.

Alternatively, instead of using IAsyncLifetime, could you just do the following in this one test?

```csharp
public async Task SuccessfulCommandTest(...)
{
    await this.databaseContainer.StartAsync();
    ...
    await this.databaseContainer.DisposeAsync().AsTask();
}
```

alanwest:
Woot! @HofmeisterAn, I pulled down your branch and ran the tests on my M1 Mac within VSCode.

**Before starting up docker - test is skipped**
```text
OpenTelemetry.Instrumentation.SqlClient.Tests.SqlClientIntegrationTests.SuccessfulCommandTest:
    Outcome: Skipped
    Standard Output Messages:
    The Docker Linux engine is not available.
    
Total tests: 1. Passed: 0. Failed: 0. Skipped: 1
```

**After starting up docker - test runs and passes** (takes about 2 minutes first run. 1 minute subsequent runs.)
```text
[xUnit.net 00:00:01.03]   Starting:    OpenTelemetry.Instrumentation.SqlClient.Tests
[xUnit.net 00:01:57.99]   Finished:    OpenTelemetry.Instrumentation.SqlClient.Tests
----- Test Execution Summary -----

OpenTelemetry.Instrumentation.SqlClient.Tests.SqlClientIntegrationTests.SuccessfulCommandTest:
    Outcome: Passed
    
OpenTelemetry.Instrumentation.SqlClient.Tests.SqlClientIntegrationTests.SuccessfulCommandTest:
    Outcome: Passed
    
OpenTelemetry.Instrumentation.SqlClient.Tests.SqlClientIntegrationTests.SuccessfulCommandTest:
    Outcome: Passed
    
OpenTelemetry.Instrumentation.SqlClient.Tests.SqlClientIntegrationTests.SuccessfulCommandTest:
    Outcome: Passed
    
OpenTelemetry.Instrumentation.SqlClient.Tests.SqlClientIntegrationTests.SuccessfulCommandTest:
    Outcome: Passed
    
OpenTelemetry.Instrumentation.SqlClient.Tests.SqlClientIntegrationTests.SuccessfulCommandTest:
    Outcome: Passed
    
OpenTelemetry.Instrumentation.SqlClient.Tests.SqlClientIntegrationTests.SuccessfulCommandTest:
    Outcome: Passed
    
Total tests: 7. Passed: 7. Failed: 0. Skipped: 0
```

test/OpenTelemetry.Instrumentation.SqlClient.Tests/SqlClientIntegrationTests.cs
```diff
@@ -0,0 +1,131 @@
+﻿// <copyright file="SqlClientIntegrationTests.cs" company="OpenTelemetry Authors">
```
alanwest, line 1: Our dotnet-format CI is failing on this file. It's got a BOM that needs to be removed. File format should be plain UTF-8

reyang: bad 👮‍♂️

alanwest: Call the BOM squad!

alanwest:
Ran on my Windows machine too using Visual Studio. I got the test to pass... kinda. But Visual Studio's test discoverability is confused. Maybe something about the custom theory? It does not seem to find all the test cases - should be 7, only finds 1.

Screenshot 2023-03-24 at 4 45 36 PM

Maybe just my machine, but curious if you also ran into anything funky here.. I assume you're also running on Windows.. cause you know, BOMs 😆

test/OpenTelemetry.Instrumentation.SqlClient.Tests/SqlClientIntegrationTests.cs
```diff
@@ -0,0 +1,131 @@
...
+        {
+            return this.databaseContainer.StartAsync();
+        }
+
+        public Task DisposeAsync()
+        {
+            return this.databaseContainer.DisposeAsync().AsTask();
+        }
+
+        [Trait("CategoryName", "SqlIntegrationTests")]
```
alanwest, line 45: Doesn't need to be in this PR, but this test extends our build/test CI by about 1 minute. This will only get longer as we port our other integration tests to use TestContainers.

We could add filter like `dotnet test --filter Traits\!~IntegrationTests` here to exclude them:

https://github.com/open-telemetry/opentelemetry-dotnet/blob/094124e9ee61466a5bbf182d923658cde2c2dcf5/.github/workflows/ci.yml#L41-L42

Then we could make a separate integration test job that just ran all tests with trait equal to `*IntegrationTests`

alanwest (approved):
Other than the minor dotnet-format issue, LGTM! One more minor thing, I think we can remove the docker-compose and Dockerfile from this test project now, right?

I'm also curious why test discovery is not working.. but this could just be a "me" problem..

test/OpenTelemetry.Instrumentation.SqlClient.Tests/SqlClientIntegrationTests.cs
HofmeisterAn: I saved the file without BOM.

test/OpenTelemetry.Instrumentation.SqlClient.Tests/SqlClientIntegrationTests.cs
HofmeisterAn: Usually, putting tests in different assemblies and running them parallel can improve performance. In some cases, running tests against a singleton instance of a container or parallel against multiple instances can also increase performance, although this depends on the specific test. However, at the moment, I believe that the additional test time will be mostly due to pulling the image.

HofmeisterAn:
> I'm also curious why test discovery is not working.. but this could just be a "me" problem..

Even though I use Windows most of the time, I do not use Visual Studio. While there is an issue with xUnit.net's [Theory](https://xunit.net/faq/theory-data-stability-in-vs) when used with Visual Studio, I do not believe this relates to this issue. Test discovery works fine in Rider (macOS and Windows).

PR #4199, opened 2023-02-16
Added in changes for retrieving a helpful service name when none have been provided

### LINKED ISSUE #2781: Refine the default resource service.name
Ideally, the SDK should try the name of the entry assembly first and only if that fails fallback to the process name. This way it gets better default names for applications started via `dotnet MySvc.dll` which is currently resolves `dotnet` as the process name.

Do you see any problem with such a change?

_Originally posted by @pjanotti in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/312#discussion_r782478960_

Kronos11:
Fixes 
#2781 
https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/2189#issuecomment-1432041198
I have also opened https://github.com/open-telemetry/opentelemetry-specification/issues/3210 if this is required for the change.
## Changes

This includes checking for IIS Hosting and returning site name + virtual directory. This is consistent with how DataDog and Splunk handle this in their custom instrumentation.

Fixed tests to remove looking for unknown_service:

### DIFF
--- a/src/OpenTelemetry/OpenTelemetry.csproj
+++ b/src/OpenTelemetry/OpenTelemetry.csproj
@@ -17,7 +17,11 @@
   <PropertyGroup Condition="'$(TargetFramework)' == 'netstandard2.1' OR '$(TargetFramework)' == 'net6.0'">
     <RunApiCompat>false</RunApiCompat>
   </PropertyGroup>
-
+  
+  <ItemGroup Condition=" $(TargetFramework.StartsWith('net4')) ">
+    <Reference Include="System.Web" />
+  </ItemGroup>
+  
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.Logging.Configuration" Version="$(MicrosoftExtensionsLoggingConfigurationPkgVer)" />
   </ItemGroup>
--- a/src/OpenTelemetry/Resources/ResourceBuilder.cs
+++ b/src/OpenTelemetry/Resources/ResourceBuilder.cs
@@ -33,7 +33,7 @@ namespace OpenTelemetry.Resources
 
         static ResourceBuilder()
         {
-            string serviceName = GetServiceName();
+            string? serviceName = GetServiceName();
 
             DefaultResource = new Resource(new Dictionary<string, object>
             {
@@ -69,7 +69,7 @@ namespace OpenTelemetry.Resources
             return defaultServiceName;
         }
 
-        private static string GetGeneratedServiceName()
+        private static string? GetGeneratedServiceName()
         {
 #if NETFRAMEWORK
             // System.Web.dll is only available on .NET Framework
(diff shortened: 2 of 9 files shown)


src/OpenTelemetry/OpenTelemetry.csproj
```diff
@@ -17,7 +17,11 @@
   <PropertyGroup Condition="'$(TargetFramework)' == 'netstandard2.1' OR '$(TargetFramework)' == 'net6.0'">
     <RunApiCompat>false</RunApiCompat>
   </PropertyGroup>
-
+  
+  <ItemGroup Condition=" $(TargetFramework.StartsWith('net4')) ">
+    <Reference Include="System.Web" />
```
cijothomas, line 22: we are very hesitant to adding dependencies to the core sdk packages, so this direction is a no-no.

src/OpenTelemetry/Resources/ResourceBuilder.cs
```diff
@@ -30,30 +32,72 @@ public class ResourceBuilder
...
 
             try
             {
-                var processName = Process.GetCurrentProcess().ProcessName;
-                if (!string.IsNullOrWhiteSpace(processName))
+                var serviceName = GetGeneratedServiceName();
+                if (string.IsNullOrWhiteSpace(serviceName))
                 {
-                    defaultServiceName = $"{defaultServiceName}:{processName}";
+                    defaultServiceName = $"{defaultServiceName}:{Process.GetCurrentProcess().ProcessName}";
```
cijothomas, line 57: I think this is best suited to be authored as a separate ResourceDetector, which can be either bought by the application or by auto-instrumentation.

src/OpenTelemetry/OpenTelemetry.csproj
Kronos11: How would you like this handled then?

Kielek:
@RassK, please check

src/OpenTelemetry/OpenTelemetry.csproj
Kronos11: Doesn't look like there is anyway to do this without using System.Web

https://github.com/dotnet/aspnetcore/issues/7400
https://github.com/dotnet/aspnetcore/issues/17069

src/OpenTelemetry/Resources/ResourceBuilder.cs
RassK: I think this is best suited for auto-instrumentation, it gives very little to manual instrumentation since you have easy access to the `OpenTelemetry.Resources.ResourceBuilder` anyway.

src/OpenTelemetry/OpenTelemetry.csproj
RassK: auto-instrumentation can bring this dependency (it's actually already there), it's easier for us since we are mixed to the application level. About libs brining extra dependencies, I agree with  @cijothomas .

src/OpenTelemetry/Resources/ResourceBuilder.cs
Kronos11: I think this is a good plan as well. I'll close this for now and move changes into the instrumentation repo.

Kronos11:
closing to move changes into dotnet-instrumentation

PR #4202, opened 2023-02-16
[To metric branch] Exemplar - add filtered tags

cijothomas:
Add support for FilteredTags concept to Exemplars. This is the set of Tags that were dropped (by View config).

### DIFF
--- a/src/OpenTelemetry/Metrics/Exemplar/AlignedHistogramBucketExemplarReservoir.cs
+++ b/src/OpenTelemetry/Metrics/Exemplar/AlignedHistogramBucketExemplarReservoir.cs
@@ -32,14 +32,43 @@ internal sealed class AlignedHistogramBucketExemplarReservoir
         this.snapshotExemplars = new Exemplar[length + 1];
     }
 
-    public void OfferAtBoundary(int index, double value)
+    public void OfferAtBoundary(int index, double value, ReadOnlySpan<KeyValuePair<string, object>> tags)
     {
-        var exemplar = default(Exemplar);
+        ref var exemplar = ref this.runningExemplars[index];
         exemplar.Timestamp = DateTime.UtcNow;
         exemplar.DoubleValue = value;
         exemplar.TraceId = Activity.Current?.TraceId;
         exemplar.SpanId = Activity.Current?.SpanId;
-        this.runningExemplars[index] = exemplar;
+
+        if (tags == default)
+        {
+            // default tag is used to indicate
+            // the special case where all tags provided at measurement
+            // recording time are stored.
+            // In this case, Exemplars does not have to store any tags.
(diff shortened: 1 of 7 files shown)


src/OpenTelemetry/Metrics/Exemplar/Exemplar.cs
```diff
@@ -23,6 +23,8 @@ namespace OpenTelemetry.Metrics
     /// </summary>
     public struct Exemplar
     {
+        internal List<KeyValuePair<string, object>> AllTags;
```
cijothomas, line 26: i just realized we dont need 2 separate lists. one is fine (FilteredTags). It can store all tags internally, and during snaptshot, it can be modified to store the actuall FilteredTags by FIlteredTags - FilteredTags - StoredTags

src/OpenTelemetry/Metrics/Exemplar/AlignedHistogramBucketExemplarReservoir.cs
```diff
@@ -32,26 +32,62 @@ public AlignedHistogramBucketExemplarReservoir(int length)
...
         exemplar.TraceId = Activity.Current?.TraceId;
         exemplar.SpanId = Activity.Current?.SpanId;
-        this.runningExemplars[index] = exemplar;
+
+        if (tags == default)
+        {
+            // default tag is used to indicate
+            // the special case where all tags provided at measurement
+            // recording time are stored.
+            // In this case, Exemplars does not have to store any tags.
```
reyang, line 48:
```suggestion
            // In this case, Exemplars do not have to store any tags.
```

src/OpenTelemetry/Metrics/Exemplar/AlignedHistogramBucketExemplarReservoir.cs
```diff
@@ -32,26 +32,62 @@ public AlignedHistogramBucketExemplarReservoir(int length)
...
+    public void SnapShot(ReadOnlyTagCollection actualTags, bool reset)
     {
         for (int i = 0; i < this.runningExemplars.Length; i++)
         {
             this.snapshotExemplars[i] = this.runningExemplars[i];
+            if (this.runningExemplars[i].FilteredTags != null)
+            {
+                // TODO: Better data structure to avoid this Linq.
+                // This is doing filtered = alltags - storedtags.
+                this.snapshotExemplars[i].FilteredTags = this.runningExemplars[i].FilteredTags.Except(actualTags.KeyAndValues.ToList()).ToList();
```
reyang, line 88: This might be a good fit for lazy evaluation (e.g. the first exporter which triggered the serialization will actually handle evaluation, though the minor downside is that CPU consumption attribution will be harder).

reyang (approved):

PR #4221, opened 2023-02-23
Remove unused eventsource methods

cijothomas:
For furture: May need more systematic way of telling unused stuff....


src/OpenTelemetry/Trace/ExceptionProcessor.cs
```diff
@@ -75,7 +75,7 @@ public override void OnEnd(Activity activity)
 
             if (snapshot != pointers)
             {
-                activity.SetStatus(Status.Error);
+                activity.SetStatus(ActivityStatusCode.Error);
```
cijothomas, line 78:
```suggestion
                activity.SetStatus(Status.Error);
```

utpilla (approved):

PR #4231, opened 2023-02-24
{To metrics} Add ExemplarReservoir base class

cijothomas:
Trying to see how to get closer to spec, and also to add one more ExemplarReservoir for non-histogram-with-bucket metric types.
The code user needs to write for own ExemplarReservoir is quite ugly still, but I'll make them cleaner next.

### DIFF
--- a/src/OpenTelemetry/Metrics/Exemplar/ExemplarReservoir.cs
+++ b/src/OpenTelemetry/Metrics/Exemplar/ExemplarReservoir.cs
@@ -1,0 +1,42 @@
+// <copyright file="ExemplarReservoir.cs" company="OpenTelemetry Authors">
+// Copyright The OpenTelemetry Authors
+//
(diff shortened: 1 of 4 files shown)


reyang (approved):

src/OpenTelemetry/Metrics/Exemplar/ExemplarReservoir.cs
```diff
@@ -0,0 +1,42 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+namespace OpenTelemetry.Metrics;
+
+/// <summary>
+/// The base class for defining Exemplar Reservoir.
+/// </summary>
+public abstract class ExemplarReservoir
```
utpilla, line 21: Why not use an interface instead of an abstract class?

PR #4238, opened 2023-02-25
MetricPoint refactoring to keep its size fixed

cijothomas:
Discussion on this: https://github.com/open-telemetry/opentelemetry-dotnet/pull/4060/files#r1084483713

Benchmarks show hardly ~1-2 nsec diff, so this seems good way to keep MetricPoint Size fixed even with Exponential Histograms, Exemplars coming.

### DIFF
--- a/src/OpenTelemetry/Metrics/MetricPoint.cs
+++ b/src/OpenTelemetry/Metrics/MetricPoint.cs
@@ -28,7 +28,7 @@ namespace OpenTelemetry.Metrics
 
         private readonly AggregationType aggType;
 
-        private HistogramBuckets histogramBuckets;
+        private MetricPointOptionalComponents mpComponents;
 
         // Represents temporality adjusted "value" for double/long metric types or "count" when histogram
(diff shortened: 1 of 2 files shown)


alanwest (approved):

src/OpenTelemetry/Metrics/MetricPoint.cs
```diff
@@ -28,7 +28,7 @@ public struct MetricPoint
 
         private readonly AggregationType aggType;
 
-        private HistogramBuckets histogramBuckets;
+        private MetricPointOptionalComponents mpComponents;
```
utpilla, line 31: nit: Rename this to `optionalComponents`?

utpilla (approved):

src/OpenTelemetry/Metrics/MetricPoint.cs
cijothomas: will do as soon as I get get this back to main branch. (working on it now)

PR #4240, opened 2023-02-27
Metrics branch - enable Public API

cijothomas:
Removes any temp changes done in metrics branch to ensure smooth merge back to main. (and as a side effect this adds public api files now)
Rest of Exempalr work will be done to main branch. following usual procedure.

### DIFF
--- a/src/OpenTelemetry/Metrics/Exemplar/ExemplarReservoir.cs
+++ b/src/OpenTelemetry/Metrics/Exemplar/ExemplarReservoir.cs
@@ -18,7 +18,7 @@ namespace OpenTelemetry.Metrics;
 /// <summary>
 /// The base class for defining Exemplar Reservoir.
 /// </summary>
-public abstract class ExemplarReservoir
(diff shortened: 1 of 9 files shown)


src/OpenTelemetry/Metrics/Exemplar/ExemplarReservoir.cs
```diff
@@ -18,7 +18,7 @@ namespace OpenTelemetry.Metrics;
 /// <summary>
 /// The base class for defining Exemplar Reservoir.
 /// </summary>
-public abstract class ExemplarReservoir
+internal abstract class ExemplarReservoir
```
reyang, line 21: Is this just for now (and later allow custom reservoirs) or the goal is to only allow built-in reservoirs?

cijothomas: just for now! 
Reservoirs would be opened up like Filters.

PR #4256, opened 2023-03-02
Enabling Exemplars for all types of metrics

cijothomas:
Enabling Exemplars for all types of metrics:
Histogram with buckets - same as beofre, it uses AlignedHistogramResrevoir.
All others - SimpleReservoir

BenchmarkDotNet=v0.13.3, OS=Windows 10 (10.0.19045.2604)
Intel Core i7-4790 CPU 3.60GHz (Haswell), 1 CPU, 8 logical and 4 physical cores
.NET SDK=7.0.103
  [Host]     : .NET 7.0.3 (7.0.323.6910), X64 RyuJIT AVX2
  DefaultJob : .NET 7.0.3 (7.0.323.6910), X64 RyuJIT AVX2

With ExemplarFilter = ALwaysOff

|                    Method | AggregationTemporality |      Mean |    Error |   StdDev |
|-------------------------- |----------------------- |----------:|---------:|---------:|
|            CounterHotPath |             Cumulative |  21.51 ns | 0.101 ns | 0.095 ns |
| CounterWith1LabelsHotPath |             Cumulative |  82.38 ns | 0.397 ns | 0.352 ns |
| CounterWith3LabelsHotPath |             Cumulative | 180.80 ns | 1.151 ns | 1.020 ns |
| CounterWith5LabelsHotPath |             Cumulative | 283.25 ns | 3.091 ns | 2.740 ns |
| CounterWith6LabelsHotPath |             Cumulative | 319.57 ns | 5.484 ns | 4.579 ns |
| CounterWith7LabelsHotPath |             Cumulative | 360.49 ns | 4.076 ns | 3.813 ns |
|            CounterHotPath |                  Delta |  23.03 ns | 0.111 ns | 0.099 ns |
| CounterWith1LabelsHotPath |                  Delta |  91.15 ns | 0.792 ns | 0.661 ns |
| CounterWith3LabelsHotPath |                  Delta | 175.83 ns | 1.483 ns | 1.387 ns |
| CounterWith5LabelsHotPath |                  Delta | 288.38 ns | 4.776 ns | 4.468 ns |
| CounterWith6LabelsHotPath |                  Delta | 329.11 ns | 3.964 ns | 3.708 ns |
| CounterWith7LabelsHotPath |                  Delta | 358.58 ns | 4.220 ns | 3.948 ns |

================================
With ExemplarFilter = ALwaysON
=================================

|                    Method | AggregationTemporality |      Mean |    Error |   StdDev |
[...]

PR #4262, opened 2023-03-03
Add dedicated method to enable Exemplar feature

cijothomas:
`EnableExemplar()` - Enables the feature with all defaults.
`TraceBasedExemplarFilter` - default filter.
`AlignedHistogramBucketExemplarReservoir` and `SimpleExemplarReservoir` - default reservoir for histogram with buckets and others respectively.

`EnableExemplar()` + `AlwaysOffExemplatFilter` gives same effect as the feature being disabled completely.

`AlwaysOnExemplatFilter`, etc *alone* is not sufficient to turn on the feature, as it must be done via `EnableExemplar()` call first.

Also addresses https://github.com/open-telemetry/opentelemetry-dotnet/pull/4256#discussion_r1124113958

### DIFF
--- a/docs/metrics/customizing-the-sdk/README.md
+++ b/docs/metrics/customizing-the-sdk/README.md
@@ -411,6 +411,24 @@ context, if any. It also provides "Filtered Tags", which are attributes (Tags)
 that are [dropped by a view](#select-specific-tags). Exemplars are an opt-in
 feature, and allow customization via ExemplarFilter and ExemplarReservoir.
 
+#### Enabling Exemplars
+
+Exemplars feature is not enabled by default. The following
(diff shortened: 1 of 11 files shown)


alanwest:
Maybe an API like the following to ultimately play well with options.

```csharp
public class ExemplarOptions
{
    public ExemplarFilter ExemplarFilter { get; set; } = new TraceBasedExemplarFilter();
}

    .EnableExemplar(options => {
        options.ExemplarFilter = new AlwaysOnExemplarFilter()
    })
```

cijothomas:
I liked the suggestion. Holding off from doing this public api change now, to fix other underlying issues first, and come back to this after 1st alpha release.

PR #4267, opened 2023-03-04
Implement Update and Snapshot for exponential histogram buckets

alanwest:
Follow on from #4060 which excluded recording bucket counts. This PR still does not yet expose any public API, but take a look at the test to get a sense of what I'm proposing will be the public API for exporter authors.

Test still needs to be flushed out, but opening now for folk to review.


src/OpenTelemetry/Metrics/CircularBufferBuckets.cs
```diff
@@ -267,6 +267,37 @@ static void Move(long[] array, uint src, uint dst)
...
+            {
+                this.trait[i] = 0;
+            }
+        }
+    }
+
+    internal CircularBufferBuckets Copy()
+    {
+        // TODO: Consider avoiding allocating a new one every time.
+        // Maybe we can have two instances that we swap between snapshots.
```
cijothomas, line 284: 👍

alanwest: Was planning to grab this in a follow up.

src/OpenTelemetry/Metrics/ExponentialBuckets.cs
```diff
@@ -0,0 +1,72 @@
...
+            this.buckets = buckets;
+            this.index = buckets.Offset;
+            this.Current = default;
+            this.lastIndex = buckets.Size - 1;
+        }
+
+        /// <summary>
+        /// Gets the bucket count at the current position of the enumerator.
+        /// </summary>
+        public long Current { get; private set; }
```
utpilla, line 51: How do we retrieve the actual bucket boundary that this count belongs to?

utpilla: It's probably a good idea to add a property to provide `Base` in `ExponentialBucketSnapshot`.

utpilla (approved):

src/OpenTelemetry/Metrics/ExponentialBuckets.cs
alanwest: Not sure it'll be necessary. `ExponentialBucketSnapshot` is basically what exporters get in order to translate stuff to their wire format. As seen in the proto, there's no need to actually compute bucket boundaries on behalf of the exporter. The back end can compute bucket boundaries simply with scale, offset, and the array of bucket counts.

In fact, one of the benefits of exponential histograms is that for a histogram of the roughly the same size, an exponential histogram requires sending less data over the wire than an explicit bounds histogram. Because you don't send the boundaries.

PR #4279, opened 2023-03-07
[Exporter.OTLP] Scaffolding for separate projects

Kielek:
Towards https://github.com/open-telemetry/opentelemetry-dotnet/issues/3421#issuecomment-1314630702

## Changes

Scaffolding for OTLP over HTTP and gRPC separate projects.
gRPC project was in place.

PR should move forward with support for OTLP environmental variables specific for each signal. I do not see any possibility to do it without introducing breaking changes to current, unified exporter. There were a proposal to split it into GRPC and HTTP. It allows us to introduce breaking changes, see comment from @CodeBlanch: https://github.com/open-telemetry/opentelemetry-dotnet/issues/4154#issuecomment-1453863637.

Second benefit - possibility to use only HTTP version without bringing dependencies to gRPC client. Based on https://github.com/open-telemetry/opentelemetry-dotnet/issues/3421#issue-1290022276 - it can reduce instrumented project by 17MB. It can be significant difference if you are using a lot of different micro-services.

As a results we will have two projects:
 - OpenTelemetry.Exporter.OpenTelemetryProtocol.Grpc - handling grpc protocol
 - OpenTelemetry.Exporter.OpenTelemetryProtocol.Http - nandling http/protobuf

* ~~[ ] Appropriate `CHANGELOG.md` updated for non-trivial changes~~

### DIFF
--- a/OpenTelemetry.sln
+++ b/OpenTelemetry.sln
@@ -251,7 +251,11 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "getting-started-jaeger", "d
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "stratified-sampling-example", "docs\trace\advanced\stratified-sampling-example\stratified-sampling-example.csproj", "{9C99621C-343E-479C-A943-332DB6129B71}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "OpenTelemetry.Exporter.OpenTelemetryProtocol.Http", "src\OpenTelemetry.Exporter.OpenTelemetryProtocol.Http\OpenTelemetry.Exporter.OpenTelemetryProtocol.Http.csproj", "{877F8535-7851-4DC8-984A-461B097DE154}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "OpenTelemetry.Exporter.OpenTelemetryProtocol.Http", "src\OpenTelemetry.Exporter.OpenTelemetryProtocol.Http\OpenTelemetry.Exporter.OpenTelemetryProtocol.Http.csproj", "{877F8535-7851-4DC8-984A-461B097DE154}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "OpenTelemetry.Exporter.OpenTelemetryProtocol.Grpc.Tests", "test\OpenTelemetry.Exporter.OpenTelemetryProtocol.Grpc.Tests\OpenTelemetry.Exporter.OpenTelemetryProtocol.Grpc.Tests.csproj", "{FA9B9576-1158-40A8-9CF3-8878CBDA873A}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "OpenTelemetry.Exporter.OpenTelemetryProtocol.Http.Tests", "test\OpenTelemetry.Exporter.OpenTelemetryProtocol.Http.Tests\OpenTelemetry.Exporter.OpenTelemetryProtocol.Http.Tests.csproj", "{BBCF30A4-D470-4C25-AA03-D151C551225D}"
 EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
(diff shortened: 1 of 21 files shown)


reyang:
@Kielek would you put a brief summary regarding why would we want to have two separate projects/components for OTLP Exporter?

pellared:
@reyang https://github.com/open-telemetry/opentelemetry-dotnet/pull/3885#discussion_r1018557835

reyang:
> @reyang [#3885 (comment)](https://github.com/open-telemetry/opentelemetry-dotnet/pull/3885#discussion_r1018557835)

I looked at the comments and didn't get the clarity. May I get a summary of what does it mean? - e.g. what exact problem do we want to solve

Kielek:
@reyang, I am trying to move forward with support for OTLP environmental variables specific for each signal. I do not see any possibility to do it without introducing breaking changes to current, unified exporter. There were a proposal to split it into GRPC and HTTP. It allows us to introduce breaking changes, see comment from @CodeBlanch: https://github.com/open-telemetry/opentelemetry-dotnet/issues/4154#issuecomment-1453863637.

There is also a second benefit - possibility to use only HTTP version without bringing dependencies to gRPC client. Based on https://github.com/open-telemetry/opentelemetry-dotnet/issues/3421#issue-1290022276 - it can reduce instrumented project by 17MB. It can be significant difference if you are using a lot of different micro-services.

reyang:
My main concern about this PR - it seems we make random moves without clear direction/picture. E.g. What 's our vision? What do we plan to do with https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/file-exporter.md and other potential upcoming "flavors of OTLP exporters"?

> @reyang, I am trying to move forward with support for OTLP environmental variables specific for each signal. I do not see any possibility to do it without introducing breaking changes to current, unified exporter. There were a proposal to split it into GRPC and HTTP. It allows us to introduce breaking changes, see comment from @CodeBlanch: [#4154 (comment)](https://github.com/open-telemetry/opentelemetry-dotnet/issues/4154#issuecomment-1453863637).

I don't think "introducing environment variable" could justify "changing how components are layered/organized".

> There is also a second benefit - possibility to use only HTTP version without bringing dependencies to gRPC client. Based on [#3421 (comment)](https://github.com/open-telemetry/opentelemetry-dotnet/issues/3421#issue-1290022276) - it can reduce instrumented project by 17MB. It can be significant difference if you are using a lot of different micro-services.

I would say the right direction is to work with the gRPC/protobuf owner and see how to implement OTLP/gRPC exporter without having to take 17MB extra packages.

reyang:
@Kielek would you update the PR description to make sure the overall thinking is captured? Here goes a good example from @Yun-Ting's PR https://github.com/open-telemetry/opentelemetry-dotnet/pull/3430.

Kielek:
> @Kielek would you update the PR description to make sure the overall thinking is captured? Here goes a good example from @Yun-Ting's PR #3430.

Description updated.

alanwest:
> My main concern about this PR - it seems we make random moves without clear direction/picture.

I agree we need to consider our direction a bit more deeply prior to continuing work to create separate projects. We have discussed various things we could do, but have not yet devoted the focus necessary to settle on what to actually do.

There are a number of different issues, some of which may be easier to address with separate packages, but I'm not yet convinced it's our only option. 

1. Support for signal-specific environment variables.
    * Even if we did ultimately create separate packages, I'd ideally still want to try supporting signal-specific environment variables with the original package. I think it would be odd to tell users they have to use one of the new packages to get this functionality. Though, to be honest I have thought much about how we'd make this work.

2. Eliminating the use of the [very soon to be deprecated gRPC library](https://grpc.io/blog/grpc-csharp-future/).
    * We have heard this from users as well (#3421). I have not yet studied it closely but my understanding is that opentelemetry-java eliminated the dependency on Java's gRPC and protobuf libraries by hand-rolling their own implementation. We might be able to do something similar.

3. Better support for DI/configuration integration (the single OTLP exporter options class causes us some grief here).

reyang:
> 2. Eliminating the use of the [very soon to be deprecated gRPC library](https://grpc.io/blog/grpc-csharp-future/).
>    
>    * We have heard this from users as well ([Use `Grpc.Net.Client` with `netstandard2.0` for `OpenTelemetry.Exporter.OpenTelemetryProtocol` #3421](https://github.com/open-telemetry/opentelemetry-dotnet/issues/3421)). I have not yet studied it closely but my understanding is that opentelemetry-java eliminated the dependency on Java's gRPC and protobuf libraries by hand-rolling their own implementation. We might be able to do something similar.

This sounds like music 🎵 to me 👍
We should seek some input from @JamesNK.

PR #4293, opened 2023-03-10
Consolidate NullableAttributes across OpenTelemetry and OpenTelemetry.Api

TimothyMothra:
Towards #4338

1. `NullableAttributes` are copied from dotnet runtime ([link](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Diagnostics/CodeAnalysis/NullableAttributes.cs)) to support older frameworks. 
This repo currently has nullable attributes spread across both `OpenTelemetry` project ([link](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/Internal/Shims/NullableAttributes.cs)) and `OpenTelemetry.Api` ([link](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Api/Internal/Guard.cs)).
This PR merges these into a single file.

2. Another issue that needed to be addressed is target versions between `OpenTelemetry` and `OpenTelemetry.Api` projects.
`OpenTelemetry` project targets `net6.0;netstandard2.1;netstandard2.0;net462` and has a dependency on `OpenTelemetry.Api` which targets `netstandard2.0;net462`.
Both projects need access to `NullableAttributes` and `OpenTelemetry.Api` makes its internals visible to `OpenTelemetry`.
Consider that the highest version of OpenTelemetry.Api is netstandard2.0, which means the NullableAttributes are always present. 
The OpenTelemetry project will have access to these NullableAttributes in the Api project and will conflict with the dotnet runtime. 
To address this, OpenTelemetry.Api needs to target netstandard2.1 so it produces a dll without those internals. See also this comment: https://github.com/open-telemetry/opentelemetry-dotnet/pull/3374#discussion_r899649955

## Changes

- add `netstandard2.1` to `OpenTelemetry.Api` project.
- move `NullableAttributes` from OpenTelemetry to OpenTelemetry.Api.
- remove nullable attributes from `OpenTelemetry.Api`'s Guard.cs and add to NullableAttributes.


src/OpenTelemetry.Api/Internal/NullableAttributes.cs
```diff
@@ -0,0 +1,37 @@
+// <copyright file="NullableAttributes.cs" company="OpenTelemetry Authors">
```
Kielek, line 1: Please check https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/Internal/Shims/NullableAttributes.cs

I think that content from this file should be merged with existing one.

TimothyMothra: These should absolutely be merged!

I didn't know this file was here! This is probably what was causing conflicts in my nullable branch. Thanks for pointing this out! :)

TimothyMothra: I've run into an issue merging these two files and I'm still exploring options before I make a recommendation.

Both `OpenTelemetry` and `OpenTelemetry.Api` projects need access to nullables.
OTel has a build dependency on OTel.Api.
OTel.Api makes internals visible to OTel.

The blocker is that OTel targets `net6.0;netstandard2.1;netstandard2.0;net462` where as OTel.Api targets `netstandard2.0;net462`.
OTel (net6 and netstardard2.1) will take a dependency on OTel.Api (netstandard2.0).
And then OTel (net6 and netstardard2.1) has a conflict with runtime. ☹️☹️☹️

> Error	CS0433	The type 'AllowNullAttribute' exists in both 'OpenTelemetry.Api, Version=1.0.0.0, Culture=neutral, PublicKeyToken=7bd6737fe5b67e3c' and 'System.Runtime, Version=6.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'	OpenTelemetry (net6.0)

TimothyMothra: The fix is to add netstandard2.1 to the OTel.Api project. I'll bring this up in the SIG meeting tomorrow.

TimothyMothra: Discussed this in today's SIG and there are no blockers with this approach. :)


It would be preferred if we can move shared files to a separate project, but this only works if we can also remove the internals visible to between OTel and OTel.Api. I'm going to take a day to explore if this would work.

PR #4295, opened 2023-03-11
Add MetricType.ExponentialHistogram

alanwest:
Adds `MetricType.ExponentialHistogram`. Even though it becomes part of the public API in this PR, this PR still does not yet expose the ability to configure the exponential histogram aggregation.

I've also begun refactoring things a bit. I'm not thrilled by how the constructors for Metric, AggregatorStore, and MetricPoint keep growing. With this PR I could have added yet another parameter to the Metric constructor for ExponentialHistogramMaxSize. Later I would then also need to add ExponentialHistogramMaxScale. Instead this information is already on the MetricStreamIdentity. The refactor in this PR is a small step, but there's definitely more we can do to further simplify this code further.

### DIFF
--- a/src/OpenTelemetry/Metrics/AggregatorStore.cs
+++ b/src/OpenTelemetry/Metrics/AggregatorStore.cs
@@ -38,7 +38,7 @@ namespace OpenTelemetry.Metrics
         private readonly int[] currentMetricPointBatch;
         private readonly AggregationType aggType;
         private readonly double[] histogramBounds;
-        private readonly int exponentialHistogramMaxBuckets;
+        private readonly int exponentialHistogramMaxSize;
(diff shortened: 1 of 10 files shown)


utpilla (approved):

src/OpenTelemetry/Metrics/AggregatorStore.cs
```diff
@@ -49,27 +49,24 @@ internal sealed class AggregatorStore
         private bool zeroTagMetricPointInitialized;
 
         internal AggregatorStore(
-            string name,
+            MetricStreamIdentity metricStreamIdentity,
```
cijothomas, line 52: something i have been trying to do but never actually!

cijothomas (approved):

PR #4303, opened 2023-03-14
Add exponential histogram tests and in-memory exporter support

alanwest:
Adds some tests for exponential histogram configuration. Also adds InMemory exporter support.

### DIFF
--- a/src/OpenTelemetry/Metrics/ExponentialHistogramBuckets.cs
+++ b/src/OpenTelemetry/Metrics/ExponentialHistogramBuckets.cs
@@ -43,6 +43,15 @@ public sealed class ExponentialHistogramBuckets
         buckets.Copy(this.buckets);
     }
 
+    internal ExponentialHistogramBuckets Copy()
+    {
+        var copy = new ExponentialHistogramBuckets();
+        copy.Offset = this.Offset;
+        copy.buckets = new long[this.buckets.Length];
+        Array.Copy(this.buckets, copy.buckets, this.buckets.Length);
+        return copy;
+    }
+
     /// <summary>
     /// Enumerates the bucket counts of an exponential histogram.
     /// </summary>
--- a/test/OpenTelemetry.Tests/Metrics/AggregatorTest.cs
+++ b/test/OpenTelemetry.Tests/Metrics/AggregatorTest.cs
@@ -27,6 +27,26 @@ namespace OpenTelemetry.Metrics.Tests
         private static readonly MetricStreamIdentity MetricStreamIdentity = new(Instrument, HistogramConfiguration);
         private readonly AggregatorStore aggregatorStore = new(MetricStreamIdentity, AggregationType.HistogramWithBuckets, AggregationTemporality.Cumulative, 1024);
 
+        internal static void AssertExponentialBucketsAreCorrect(Base2ExponentialBucketHistogram expectedHistogram, ExponentialHistogramData data)
+        {
+            Assert.Equal(expectedHistogram.Scale, data.Scale);
+            Assert.Equal(expectedHistogram.ZeroCount, data.ZeroCount);
+            Assert.Equal(expectedHistogram.PositiveBuckets.Offset, data.PositiveBuckets.Offset);
+            Assert.Equal(expectedHistogram.NegativeBuckets.Offset, data.NegativeBuckets.Offset);
+
+            var index = expectedHistogram.PositiveBuckets.Offset;
+            foreach (var bucketCount in data.PositiveBuckets)
+            {
+                Assert.Equal(expectedHistogram.PositiveBuckets[index++], bucketCount);
+            }
+
+            index = expectedHistogram.NegativeBuckets.Offset;
+            foreach (var bucketCount in data.NegativeBuckets)
+            {
+                Assert.Equal(expectedHistogram.PositiveBuckets[index++], bucketCount);
+            }
+        }
+
         [Fact]
         public void HistogramDistributeToAllBucketsDefault()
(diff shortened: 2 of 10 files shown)


src/OpenTelemetry/Metrics/ExponentialHistogramBuckets.cs
```diff
@@ -43,6 +43,15 @@ internal void SnapshotBuckets(CircularBufferBuckets buckets)
         buckets.Copy(this.buckets);
     }
 
+    internal ExponentialHistogramBuckets Copy()
+    {
+        var copy = new ExponentialHistogramBuckets();
+        copy.Offset = this.Offset;
+        copy.buckets = new long[this.buckets.Length];
+        Array.Copy(this.buckets, copy.buckets, this.buckets.Length);
```
utpilla, line 51: Copy `size`?

test/OpenTelemetry.Tests/Metrics/AggregatorTest.cs
```diff
@@ -188,6 +188,26 @@ public void HistogramWithOnlySumCount()
...
 
+        internal static void AssertExponentialBucketsAreCorrect(Base2ExponentialBucketHistogram expectedHistogram, ExponentialHistogramData data)
+        {
+            Assert.Equal(expectedHistogram.Scale, data.Scale);
+            Assert.Equal(expectedHistogram.ZeroCount, data.ZeroCount);
+            Assert.Equal(expectedHistogram.PositiveBuckets.Offset, data.PositiveBuckets.Offset);
+            Assert.Equal(expectedHistogram.NegativeBuckets.Offset, data.NegativeBuckets.Offset);
+
+            var index = expectedHistogram.PositiveBuckets.Offset;
+            foreach (var bucketCount in data.PositiveBuckets)
```
utpilla, line 199: I think this should be tested by enumerating over the expected data instead of the actual data. That would ensure that each of the expected bucket counts were present in the actual data as well.

If `foreach (var bucketCount in data.PositiveBuckets)` does not return any item or returns lesser items than expected, we would not be able to verify that all the expected buckets were actually present.

test/OpenTelemetry.Tests/Metrics/MetricSnapshotTests.cs
```diff
@@ -163,5 +163,86 @@ public void VerifySnapshot_Histogram()
...
+            meterProvider.ForceFlush();
+
+            // Verify Metric 1
+            Assert.Single(exportedMetrics);
+            var metric1 = exportedMetrics[0];
+            var metricPoints1Enumerator = metric1.GetMetricPoints().GetEnumerator();
+            Assert.True(metricPoints1Enumerator.MoveNext());
+            ref readonly var metricPoint1 = ref metricPoints1Enumerator.Current;
+            Assert.Equal(1, metricPoint1.GetHistogramCount());
+            Assert.Equal(10, metricPoint1.GetHistogramSum());
```
utpilla, line 195: Test for min and max as well.

utpilla: It's probably better to do it in `AssertExponentialBucketsAreCorrect` method.

src/OpenTelemetry/Metrics/ExponentialHistogramBuckets.cs
alanwest: Good catch, will fix.

test/OpenTelemetry.Tests/Metrics/AggregatorTest.cs
alanwest: Yup, this would have caught the missed copy of size.

test/OpenTelemetry.Tests/Metrics/MetricSnapshotTests.cs
alanwest: Yes, should probably will take care of this on this PR. I noted a bug on the explicit bounds histograms where min/max is actually not captured currently in the snapshot, was gonna fix in another PR, but might just do it here since probably small.

src/OpenTelemetry/Metrics/ExponentialHistogramBuckets.cs
alanwest: Turns out this was actually more insidious... I removed `size` altogether. Size was initialized wrong... this was never testing anything correctly.

This was a result of a refactor in my prior PR where I changed `ExponentialHistogramBuckets` from having a handle an instance of `CircularBufferBuckets` to simply a `long[]`.

test/OpenTelemetry.Tests/Metrics/MetricSnapshotTests.cs
alanwest: Min/max is now tested... gotta take off for the evening, so will fix explicit in a follow up

utpilla (approved):

src/OpenTelemetry/Metrics/ExponentialHistogramBuckets.cs
```diff
@@ -85,7 +81,7 @@ internal Enumerator(int size, long[] buckets)
         /// collection.</returns>
         public bool MoveNext()
         {
-            if (this.index < this.size)
+            if (this.index < this.buckets.Length)
```
utpilla, line 84: Shouldn't this be size? We are now always going enumerate over the full length of the `CircularBufferBuckets` instead of just enumerating from `CircularBufferBuckets.begin` to `CircularBufferBuckets.end`

PR #4312, opened 2023-03-17
Sdk internal metrics

vishweshbankwar:
Towards #3880 

## Changes

This PR introduces a new class `SdkHealthReporter`. The idea is to have one `SdkHealthReporter` instance per provider with shared Meter and instruments. The emitted metrics will contain the identity of providers in the form of dimensions/tags.


src/OpenTelemetry/Internal/SdkHealthReporter.cs
```diff
@@ -0,0 +1,32 @@
...
+
+using System.Diagnostics.Metrics;
+
+namespace OpenTelemetry.Internal
+{
+    internal class SdkHealthReporter
+    {
+        public SdkHealthReporter(string providerId)
+        {
+            this.BatchExportProcessorDroppedCount = InternalMeter.CreateCounter<long>($"{providerId}.BatchExportProcessorDroppedCount");
```
alanwest, line 25: A standard prefix for our health metrics would be useful. Something like `dotnet.sdk.*`. So this metric could be `dotnet.sdk.processor.dropped_count`.

One scenario where this becomes useful is if you're using the collector to process these metrics specially. I can configure my collector to match any metric name that starts with `dotnet.sdk`.

Also I think `provider.id=123` and `processor.type=BatchExportProcessor` would be good as attributes.

src/OpenTelemetry/BatchExportProcessor.cs
```diff
@@ -114,6 +118,8 @@ internal bool TryExport(T data)
                 return true; // enqueue succeeded
             }
 
+            this.ParentProvider?.GetSdkHealthReporter()?.BatchExportProcessorDroppedCount.Add(1, this.tags);
```
CodeBlanch, line 121: Probably better perf to pass `TagList` by ref (eg `Add(1, in this.tags)`).

src/OpenTelemetry/BatchExportProcessor.cs
```diff
@@ -46,6 +46,7 @@ public abstract class BatchExportProcessor<T> : BaseExportProcessor<T>
         private long shutdownDrainTarget = long.MaxValue;
         private long droppedCount;
         private bool disposed;
+        private TagList tags = default;
```
CodeBlanch, line 49: Can we add `readonly` here?

src/OpenTelemetry/Internal/SdkHealthReporter.cs
CodeBlanch: * If we don't use providerId in the name than this whole thing could be made static and then that just becomes part of the `TagList` on `BatchExportProcessor`?

* Naming it `dotnet.sdk.processor.dropped_count` are we sure about that? The name seems to indicate other processors should also use it? How do we enforce they write the correct tags? And would that prevent us from changing the tags later? Feel like it should be specific a la `dotnet.sdk.batch_processor.dropped_count` 🤷

alanwest: > Feel like it should be specific a la dotnet.sdk.batch_processor.dropped_count

That name sounds fine. My main point is we should be striving towards a convention with a standard prefix. So something along the lines `dotnet.sdk.[sdkcomponent].[somemeasurement]`.

> The name seems to indicate other processors should also use it?

Sure, and maybe this metric is very specific to batch processors, in which case the name you've proposed is fine. Though, I don't want us to lock ourselves into thinking this level of specificity is required for all health metrics we might come up with. Just an example, let's say we have `dotnet.sdk.sampler.total_spans`.

> How do we enforce they write the correct tags?

I'm not convinced the processor should be using the underlying counter directly. I alluded to this on our call, but I was thinking the processor would use a method like.

```csharp
void RecordBatchExportProcessorDroppedCount(long count, string processorName, string signalName)

// Then a processor would call it like
sdkHeathReporter.RecordBatchExporterProcessorDroppedCount(4, "BatchExportProcessor", "Activity")
```

This makes it impossible to not add the expected tags. Also makes it less possible to generate an explosion of tag values.

vishweshbankwar: > If we don't use providerId in the name than this whole thing could be made static and then that just becomes part of the TagList on BatchExportProcessor?

would like to discuss this part in SIG today if we get time.

vishweshbankwar: @CodeBlanch @alanwest  made some updates. Please have a look when you get chance.

src/OpenTelemetry/Internal/SdkHealthReporter.cs
```diff
@@ -0,0 +1,49 @@
...
+            this.providerName = providerName;
+        }
+
+        private static Meter InternalMeter { get; } = new Meter("OpenTelemetry.Sdk");
+
+        private static Counter<long> BatchExportProcessorDroppedCount { get; } = InternalMeter.CreateCounter<long>("dotnet.sdk.batchprocessor.dropped_count");
+
+        internal void ReportBatchProcessorDroppedCount(long droppedCount, ref TagList droppedCountTags)
+        {
+            droppedCountTags.Add(ProviderIdKey, this.providerId);
```
cijothomas, line 44: this is around 4 tags now. It might be okay this is only for "drop" scenario only and not in hot path.
But I think we should explore a way to model these as resource or meter.attributes or making it all into a long string and store in instrument.Description - these seems static and not going to change. From benchmarks we know every additional Tag eats perf.

CodeBlanch: We add the tags here for each invocation to a field passed by ref. Won't that cause 2 new tags to be added each time this is called?

src/OpenTelemetry/BatchExportProcessor.cs
vishweshbankwar: image

vishweshbankwar: updated to array

src/OpenTelemetry/BatchExportProcessor.cs
```diff
@@ -114,6 +125,8 @@ internal bool TryExport(T data)
                 return true; // enqueue succeeded
             }
 
+            this.HealthReporter?.ReportBatchProcessorDroppedCount(1, ref this.tags);
```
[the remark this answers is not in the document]
vishweshbankwar: Replied on different comment accidentally :)
https://github.com/open-telemetry/opentelemetry-dotnet/pull/4312#discussion_r1146980619

src/OpenTelemetry/Internal/SdkHealthReporter.cs
```diff
@@ -0,0 +1,44 @@
...
+            this.ProviderName = providerName;
+        }
+
+        internal string ProviderId { get; }
+
+        internal string ProviderName { get; }
+
+        private static Meter InternalMeter { get; } = new Meter("OpenTelemetry.Sdk");
+
+        private static Counter<long> BatchExportProcessorDroppedCount { get; } = InternalMeter.CreateCounter<long>("dotnet.sdk.batchprocessor.dropped_count");
```
utpilla, line 37: I think we should mention "otel" in the instrument name as their might be backends that don't use or show Meter name on their UI.

src/OpenTelemetry/BatchExportProcessor.cs
```diff
@@ -95,6 +102,17 @@ public abstract class BatchExportProcessor<T> : BaseExportProcessor<T>
         /// </summary>
         internal long ProcessedCount => this.circularBuffer.RemovedCount;
 
+        /// <summary>
+        /// Gets the SdkHealthReporter object to report health metrics.
+        /// </summary>
+        // TODO: Add Noop reporter for null case.
+        private SdkHealthReporter? HealthReporter => this.sdkHealthReporter ??= this.ParentProvider?.GetSdkHealthReporter();
```
utpilla, line 109: You could probably get rid of this additional field `this.sdkHealthReporter`.

src/OpenTelemetry/Internal/SdkHealthReporter.cs
vishweshbankwar: Added `otel.`

src/OpenTelemetry/Internal/SdkHealthReporter.cs
vishweshbankwar: You are correct. This is now resolved (changed the taglist type to params)

src/OpenTelemetry/Internal/SdkHealthMetricsConstants.cs
```diff
@@ -0,0 +1,26 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+
+namespace OpenTelemetry.Internal
+{
+    internal class SdkHealthMetricsConstants
```
CodeBlanch, line 19: nit: `sealed`

src/OpenTelemetry/Internal/SdkHealthReporter.cs
```diff
@@ -0,0 +1,44 @@
...
+// limitations under the License.
+// </copyright>
+
+#nullable enable
+
+using System.Diagnostics.Metrics;
+
+namespace OpenTelemetry.Internal
+{
+    internal class SdkHealthReporter
```
CodeBlanch, line 23: nit: `sealed`

src/OpenTelemetry/BatchExportProcessor.cs
```diff
@@ -78,6 +82,9 @@ public abstract class BatchExportProcessor<T> : BaseExportProcessor<T>
                 Name = $"OpenTelemetry-{nameof(BatchExportProcessor<T>)}-{exporter.GetType().Name}",
             };
             this.exporterThread.Start();
+
+            this.exporterName = exporter.GetType().Name;
+            this.processorType = typeof(T).Name;
```
reyang, line 87: Is this resulting in the type of the processor or the type of the data (being exported)?

vishweshbankwar: The type of data being exported i.e. `Activity` or `LogRecord`

utpilla (approved):

alanwest (approved):
Approving so you can continue to make progress, but sharing thoughts for us to continue to discuss.

@cijothomas mentions a few ideas here https://github.com/open-telemetry/opentelemetry-dotnet/pull/4312#discussion_r1145235652 that I think are worth tracking for further consideration.

1. Adding these tags as resource attributes
    * I've considered this option a bit. We might find a way to make this work well, but also I'd want to be a bit cautious of this approach because I think it may be a common case that folks will expect their resource attributes to match across all their providers (trace, metrics, logs) for a process.
2. Adding these tags as meter attributes
    * I think this would be a good option, but the Meter API currently does not support scope attributes.
3. Using the instrument description
    * I'm not a fan of this idea. I think it defeats the purpose of being able to easily facet by provider in the event of multiple providers.

Some other thoughts for continued brainstorming:

1. Maybe the inclusion of the provider.id attribute could be configurable (default off). It may often be the case that folks will not have multiple providers, so this tag is just needless noise.
2. We might consider eliminating one tag by moving provider.name back into the name of the metric. Something like `otel.dotnet.activity.batchprocessor.dropped_count` and `otel.dotnet.logrecord.batchprocessor.dropped_count`.
3. We try to constrain ourselves to using a maximum of three tags for our health metrics so that we can use the more [perf friendly Counter.Add methods](https://learn.microsoft.com/dotnet/api/system.diagnostics.metrics.counter-1.add#system-diagnostics-metrics-counter-1-add(-0-system-collections-generic-keyvaluepair((system-string-system-object))-system-collections-generic-keyvaluepair((system-string-system-object))-system-collections-generic-keyvaluepair((system-string-system-object))))

alanwest:
One other question to folks...

Are we ok with the metric names and attributes for health metrics continue to evolve between stable releases? That is, if we shipped these metrics with 1.5, would we be ok changing the metric name or its attributes in 1.6? Or will we consider this a breaking change much like conventions for instrumentation?

src/OpenTelemetry/BatchExportProcessor.cs
reyang: Consider changing the variable name to reflect the intention, e.g. "this.processorType" -> "this.theTypeOfDataBeingExported" (something simpler would be even better).

reyang:
I took a quick glance and this seems good to me (without the issue mentioned https://github.com/open-telemetry/opentelemetry-dotnet/pull/4289#discussion_r1134390550). Do we have stress test to make sure we don't leak memory?

reyang:
> One other question to folks...
> 
> Are we ok with the metric names and attributes for health metrics continue to evolve between stable releases? That is, if we shipped these metrics with 1.5, would we be ok changing the metric name or its attributes in 1.6? Or will we consider this a breaking change much like conventions for instrumentation?

My take - changing the name or adding attributes should be considered as breaking changes unless we made it obvious so the users would opt-in the risk. For example, to me it is _okay_ if we ship this under a feature flag called `EnableExperimentalFeature("SDK.InternalMetrics")`.

vishweshbankwar:
> > One other question to folks...
> > Are we ok with the metric names and attributes for health metrics continue to evolve between stable releases? That is, if we shipped these metrics with 1.5, would we be ok changing the metric name or its attributes in 1.6? Or will we consider this a breaking change much like conventions for instrumentation?
> 
> My take - changing the name or adding attributes should be considered as breaking changes unless we made it obvious so the users would opt-in the risk. For example, to me it is _okay_ if we ship this under a feature flag called `EnableExperimentalFeature("SDK.InternalMetrics")`.

1. Opened https://github.com/open-telemetry/opentelemetry-dotnet/issues/4352 to track this.

vishweshbankwar:
> I took a quick glance and this seems good to me (without the issue mentioned [#4289 (comment)](https://github.com/open-telemetry/opentelemetry-dotnet/pull/4289#discussion_r1134390550)). Do we have stress test to make sure we don't leak memory?

@[#4289 (comment)](https://github.com/open-telemetry/opentelemetry-dotnet/pull/4289#discussion_r1134390550) was considered in this new approach. We do not have a stress test yet. Opened https://github.com/open-telemetry/opentelemetry-dotnet/issues/4353 to track this

vishweshbankwar:
> Approving so you can continue to make progress, but sharing thoughts for us to continue to discuss.
> 
> @cijothomas mentions a few ideas here [#4312 (comment)](https://github.com/open-telemetry/opentelemetry-dotnet/pull/4312#discussion_r1145235652) that I think are worth tracking for further consideration.
> 
> 1. Adding these tags as resource attributes
>    
>    * I've considered this option a bit. We might find a way to make this work well, but also I'd want to be a bit cautious of this approach because I think it may be a common case that folks will expect their resource attributes to match across all their providers (trace, metrics, logs) for a process.
> 2. Adding these tags as meter attributes
>    
>    * I think this would be a good option, but the Meter API currently does not support scope attributes.
> 3. Using the instrument description
>    
>    * I'm not a fan of this idea. I think it defeats the purpose of being able to easily facet by provider in the event of multiple providers.
> 
> Some other thoughts for continued brainstorming:
> 
> 1. Maybe the inclusion of the provider.id attribute could be configurable (default off). It may often be the case that folks will not have multiple providers, so this tag is just needless noise.
> 2. We might consider eliminating one tag by moving provider.name back into the name of the metric. Something like `otel.dotnet.activity.batchprocessor.dropped_count` and `otel.dotnet.logrecord.batchprocessor.dropped_count`.
> 3. We try to constrain ourselves to using a maximum of three tags for our health metrics so that we can use the more [perf friendly Counter.Add methods](https://learn.microsoft.com/dotnet/api/system.diagnostics.metrics.counter-1.add#system-diagnostics-metrics-counter-1-add(-0-system-collections-generic-keyvaluepair((system-string-system-object))-system-collections-generic-keyvaluepair((system-string-system-object))-system-collections-generic-keyvaluepair((system-string-system-object))))

Opened https://github.com/open-telemetry/opentelemetry-dotnet/issues/4354 to further discuss this.

src/OpenTelemetry/BatchExportProcessor.cs
vishweshbankwar: Renamed to `exportedDataType`

src/OpenTelemetry/BatchExportProcessor.cs
```diff
@@ -114,6 +132,8 @@ internal bool TryExport(T data)
                 return true; // enqueue succeeded
             }
 
+            this.HealthReporter?.ReportBatchProcessorDroppedCount(1, this.DroppedCountTags);
```
cijothomas, line 135: given we already keep track of droppedCount here, this feels like better suited for ObservableCounter. Was that considered before?

vishweshbankwar: Yes - We discussed this during one of the SIG meetings

Decided to use Counter as

1) This should not be called many times so it should not have an high impact on perf. 
2) Having ObservableCounter will add additional complexity with respect to https://github.com/open-telemetry/opentelemetry-dotnet/pull/4289#discussion_r1134390550.

Separately, we also discussed the potential to revisit the existing DroppedCount as it is only logged during shutdown https://github.com/open-telemetry/opentelemetry-dotnet/blob/bf9f07dc0ff17a6e9c113c7f46fba50a46e11c00/src/OpenTelemetry/BatchExportProcessor.cs#L222.

reyang (changes requested):
Need to avoid virtual calls to an aggressively inlined function https://github.com/open-telemetry/opentelemetry-dotnet/pull/4312/files#r1161850247.

vishweshbankwar:
> Need to avoid virtual calls to an aggressively inlined function https://github.com/open-telemetry/opentelemetry-dotnet/pull/4312/files#r1161850247.

@reyang Could you please clarify which function you are referring to?

reyang:
> > Need to avoid virtual calls to an aggressively inlined function https://github.com/open-telemetry/opentelemetry-dotnet/pull/4312/files#r1161850247.
> 
> @reyang Could you please clarify which function you are referring to?

image

src/OpenTelemetry/BatchExportProcessor.cs
vishweshbankwar: @reyang @cijothomas Have opened https://github.com/open-telemetry/opentelemetry-dotnet/pull/4409 with alternate approach to use ObservableCounter.

vishweshbankwar:
closing in favor of https://github.com/open-telemetry/opentelemetry-dotnet/pull/4409

PR #4313, opened 2023-03-17
Fix exponential histogram bucket snapshot

alanwest:
As @utpilla points out (https://github.com/open-telemetry/opentelemetry-dotnet/pull/4303#discussion_r1137857027), I have lost all ability to reason over `CircularBufferBuckets`. With some enhancements to my 🧠, I believe this PR makes the world right again.

### DIFF
--- a/src/OpenTelemetry/Metrics/ExponentialHistogramBuckets.cs
+++ b/src/OpenTelemetry/Metrics/ExponentialHistogramBuckets.cs
@@ -20,32 +20,30 @@ namespace OpenTelemetry.Metrics;
 
 public sealed class ExponentialHistogramBuckets
 {
-    private long[] buckets = Array.Empty<long>();
+    private readonly long[] buckets;
+    private int size = 0;
 
-    internal ExponentialHistogramBuckets()
+    internal ExponentialHistogramBuckets(int maxSize)
     {
+        this.buckets = new long[maxSize];
     }
 
     public int Offset { get; private set; }
 
-    public Enumerator GetEnumerator() => new(this.buckets);
+    public Enumerator GetEnumerator() => new(this.buckets, this.size);
 
     internal void SnapshotBuckets(CircularBufferBuckets buckets)
     {
-        if (this.buckets.Length != buckets.Capacity)
-        {
-            this.buckets = new long[buckets.Capacity];
-        }
-
+        this.size = buckets.Size;
         this.Offset = buckets.Offset;
         buckets.Copy(this.buckets);
     }
 
     internal ExponentialHistogramBuckets Copy()
     {
-        var copy = new ExponentialHistogramBuckets();
+        var copy = new ExponentialHistogramBuckets(this.buckets.Length);
+        copy.size = this.size;
         copy.Offset = this.Offset;
-        copy.buckets = new long[this.buckets.Length];
         Array.Copy(this.buckets, copy.buckets, this.buckets.Length);
         return copy;
     }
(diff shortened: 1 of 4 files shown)


src/OpenTelemetry/Metrics/ExponentialHistogramBuckets.cs
```diff
@@ -58,10 +61,12 @@ public struct Enumerator
     {
         private readonly long[] buckets;
         private int index;
+        private int size;
```
utpilla, line 64:
```suggestion
        private readonly int size;
```

src/OpenTelemetry/Metrics/ExponentialHistogramBuckets.cs
```diff
@@ -21,14 +21,15 @@ namespace OpenTelemetry.Metrics;
 public sealed class ExponentialHistogramBuckets
 {
     private long[] buckets = Array.Empty<long>();
+    private int size = 0;
```
utpilla, line 24: nit: This initialization is probably not needed,
```suggestion
    private int size;
```

src/OpenTelemetry/Metrics/ExponentialHistogramBuckets.cs
```diff
@@ -58,10 +61,12 @@ public struct Enumerator
     {
         private readonly long[] buckets;
         private int index;
+        private int size;
 
-        internal Enumerator(long[] buckets)
+        internal Enumerator(long[] buckets, int size)
         {
             this.index = 0;
```
utpilla, line 68: Shouldn't we assign `index` to `Offset`?

src/OpenTelemetry/Metrics/ExponentialHistogramBuckets.cs
```diff
@@ -81,7 +86,7 @@ internal Enumerator(long[] buckets)
         /// collection.</returns>
         public bool MoveNext()
```
utpilla, line 87: I think the unit tests should also cover a scale-down scenario when the CircularBufferBuckets offset is updated.

src/OpenTelemetry/Metrics/ExponentialHistogramBuckets.cs
alanwest: No, not in this case, the values from the circular buffer are always copied starting at index 0 up to the size of the circular buffer.

src/OpenTelemetry/Metrics/ExponentialHistogramBuckets.cs
alanwest: It's not very obvious, but the existing tests actually do cover this scenario.

Even this simple test that only records two values requires a scale down. When the first value is recorded (10) the initial scale of 20 is fine, but when the second value is recorded (5) a scale down is necessary to fit this range of values within the default 160 buckets.

https://github.com/open-telemetry/opentelemetry-dotnet/blob/d924663dc3d6bfdf737bc49ceaa1addcec90a2d6/test/OpenTelemetry.Tests/Metrics/MetricSnapshotTests.cs#L238-L241

utpilla (approved):

src/OpenTelemetry/Metrics/ExponentialHistogramBuckets.cs
alanwest: I'm going to merge this PR, but I will be coming back to the tests. Clearly there's improvements to make as the bug this PR fixes is not really covered by any tests yet.

PR #4332, opened 2023-03-24
Exponential histogram benchmarks

alanwest:
Adds some basic benchmarks measuring exponential histogram performance.

Two benchmarks `HistogramHotPath` and a lower-level benchmark just of `MapToIndex`.

### `HistogramHotPath`

This benchmark provides a point of comparison with our explicit bound histograms.

Here's the perf numbers from my machine for explicit bound histograms for comparison with the numbers in this PR for exp histograms:

#### Explicit bound histogram
```text
|           Method | BoundCount |     Mean |    Error |   StdDev | Allocated |
|----------------- |----------- |---------:|---------:|---------:|----------:|
| HistogramHotPath |         10 | 43.40 ns | 0.126 ns | 0.118 ns |         - |
| HistogramHotPath |         49 | 50.41 ns | 0.119 ns | 0.112 ns |         - |
| HistogramHotPath |         50 | 55.90 ns | 0.078 ns | 0.069 ns |         - |
| HistogramHotPath |       1000 | 73.98 ns | 0.228 ns | 0.213 ns |         - |
```

#### Exponential histogram
```text
|           Method | Scale |     Mean |    Error |   StdDev | Allocated |
|----------------- |------ |---------:|---------:|---------:|----------:|
| HistogramHotPath |   -11 | 29.79 ns | 0.054 ns | 0.042 ns |         - |
| HistogramHotPath |     3 | 32.10 ns | 0.086 ns | 0.080 ns |         - |
| HistogramHotPath |    20 | 32.08 ns | 0.076 ns | 0.063 ns |         - |
```

### `MapToIndex`

I expect `MapToIndex` is the most costly operation. Scales <= 0 are expected to perform better than positive scales. The numbers show this to be true, but not by as much of a margin as I expected.

```text
|     Method | Scale |     Mean |    Error |   StdDev | Allocated |
|----------- |------ |---------:|---------:|---------:|----------:|
| MapToIndex |   -11 | 11.60 ns | 0.057 ns | 0.053 ns |         - |
| MapToIndex |     3 | 14.63 ns | 0.135 ns | 0.126 ns |         - |
| MapToIndex |    20 | 14.40 ns | 0.026 ns | 0.024 ns |         - |
```


test/Benchmarks/Metrics/ExponentialHistogramBenchmarks.cs
```diff
@@ -0,0 +1,99 @@
...
+    {
+        private const int MaxValue = 10000;
+        private readonly Random random = new();
+        private readonly string[] dimensionValues = new string[] { "DimVal1", "DimVal2", "DimVal3", "DimVal4", "DimVal5", "DimVal6", "DimVal7", "DimVal8", "DimVal9", "DimVal10" };
+        private Histogram<long> histogram;
+        private MeterProvider provider;
+        private Meter meter;
+        private Base2ExponentialBucketHistogram exponentialHistogram;
+
+        // Note: Values related to `HistogramBuckets.DefaultHistogramCountForBinarySearch`
```
utpilla, line 56: Remove this comment.

utpilla:
> I expect MapToIndex is the most costly operation. Scales <= 0 are expected to perform better than positive scales. The numbers show this to be true, but not by as much of a margin as I expected.

The scale of exponential histogram could keep changing throughout the updates, right? Does it make sense to add Scale as a benchmark parameter? If I understand it correctly, it's just to set the initial scale and then it's up to the values that we record, right?

test/Benchmarks/Metrics/ExponentialHistogramBenchmarks.cs
```diff
@@ -0,0 +1,99 @@
...
+        [Params(-11, 0, 5, 20)]
+        public int Scale { get; set; }
```
utpilla, line 58: Do we want to add `Params` for buckets size?

alanwest:
> The scale of exponential histogram could keep changing throughout the updates, right? Does it make sense to add Scale as a benchmark parameter? If I understand it correctly, it's just to set the initial scale and then it's up to the values that we record, right?

Correct, the scale will stabilize during benchmark warmup. Given the range of values reported [0, 10000) and a max 160 buckets, it should scale down to 3 during warmup. I benchmarked a couple positive max scales simply to validate that they did not differ in any significant way. If they had, that would have highlighted a problem.

test/Benchmarks/Metrics/ExponentialHistogramBenchmarks.cs
```diff
@@ -0,0 +1,98 @@
...
+    {
+        private const int MaxValue = 10000;
+        private readonly Random random = new();
+        private readonly string[] dimensionValues = new string[] { "DimVal1", "DimVal2", "DimVal3", "DimVal4", "DimVal5", "DimVal6", "DimVal7", "DimVal8", "DimVal9", "DimVal10" };
+        private Histogram<long> histogram;
+        private MeterProvider provider;
+        private Meter meter;
+        private Base2ExponentialBucketHistogram exponentialHistogram;
+
+        [Params(-11, 0, 5, 20)]
```
utpilla, line 56: I think it's a good idea to add this [information](https://github.com/open-telemetry/opentelemetry-dotnet/pull/4332#issuecomment-1485515624)as a comment here.

test/Benchmarks/Metrics/ExponentialHistogramBenchmarks.cs
utpilla: > It's not terribly interesting since compute time should remain constant.

I think this goes back to your [comment](https://github.com/open-telemetry/opentelemetry-dotnet/pull/4332#issuecomment-1485515624) about why use scale as a parameter.

> I benchmarked a couple positive max scales simply to validate that they did not differ in any significant way. If they had, that would have highlighted a problem.

We could do the same thing for buckets then. Should we?

utpilla (approved):
Approved with suggestions.

test/Benchmarks/Metrics/ExponentialHistogramBenchmarks.cs
```diff
@@ -0,0 +1,91 @@
...
+
+
+|           Method | Scale |     Mean |    Error |   StdDev | Allocated |
+|----------------- |------ |---------:|---------:|---------:|----------:|
+| HistogramHotPath |   -11 | 29.79 ns | 0.054 ns | 0.042 ns |         - |
+| HistogramHotPath |     3 | 32.10 ns | 0.086 ns | 0.080 ns |         - |
+| HistogramHotPath |    20 | 32.08 ns | 0.076 ns | 0.063 ns |         - |
+*/
+
+namespace Benchmarks.Metrics
```
reyang, line 38: minor: maybe reduce the indent and save some spaces:

```suggestion
namespace Benchmarks.Metrics;
```

reyang (approved):

PR #4336, opened 2023-03-27
Set Activity.Status when adding SetErrorStatusOnException

### LINKED ISSUE #4333: `SetErrorStatusOnException` doesn't set `Activity.Status`
# Bug Report

* OpenTelemetry 1.4.0

Runtime version (e.g. `net462`, `net48`, `netcoreapp3.1`, `net6.0` etc. You can find this information from the `*.csproj` file):

* net7.0

## Symptom

`SetErrorStatusOnException` doesn't set `Activity.Status`. If I'm reading correctly, it looks like the exception processor calls into an extension method that sets tags on the activity instead:

https://github.com/open-telemetry/opentelemetry-dotnet/blob/094124e9ee61466a5bbf182d923658cde2c2dcf5/src/OpenTelemetry.Api/Trace/ActivityExtensions.cs#L42-L48

Spans that contained an exception should be marked as failed.

Some exporters (e.g. Jaeger via OTLP) recognize these tags and properly display the spans as failed. However, some exporters (e.g. [Azure.Monitor.OpenTelemetry.Exporter](https://www.nuget.org/packages/Azure.Monitor.OpenTelemetry.Exporter)) only look at `Activity.Status` and don't display the spans as failed.

## Reproduce

The following program exhibits this issue where failed spans do not show up as such in Azure:

```csharp
public class Program
{
    private static readonly ActivitySource MyActivitySource = new("MyCompany.MyProduct.MyLibrary");

    public static void Main()
    {
            .AddSource("MyCompany.MyProduct.MyLibrary")
            .SetErrorStatusOnException()
            .AddAzureMonitorTraceExporter()
            .Build();

        try
        {
            using (MyActivitySource.StartActivity("Foo"))
            {
                throw new Exception("Oops!");
            }
        }
        catch (Exception)
        {
            // swallow the exception
        }
    }
}
```

## Additional Context

Is there any reason to avoid setting `Activity.Status` for uncaught exceptions or is this implementation just a historical artifact from earlier versions of .NET?

vishweshbankwar:
Fixes #4333 
Design discussion issue #

## Changes


src/OpenTelemetry.Api/Trace/ActivityExtensions.cs
```diff
@@ -38,13 +38,17 @@ public static class ActivityExtensions
         /// </summary>
         /// <param name="activity">Activity instance.</param>
         /// <param name="status">Activity execution status.</param>
+        [Obsolete("Use https://learn.microsoft.com//dotnet/api/system.diagnostics.activity.setstatus?view=net-7.0")]
```
vishweshbankwar, line 41: Is it ok to do this?

src/OpenTelemetry.Api/Trace/ActivityExtensions.cs
```diff
@@ -38,13 +38,17 @@ public static class ActivityExtensions
...
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static void SetStatus(this Activity activity, Status status)
         {
             Debug.Assert(activity != null, "Activity should not be null");
 
             activity.SetTag(SpanAttributeConstants.StatusCodeKey, StatusHelper.GetTagValueForStatusCode(status.StatusCode));
             activity.SetTag(SpanAttributeConstants.StatusDescriptionKey, status.Description);
+
+            // For processors\exporters that do not look up tags
```
cijothomas, line 50: why are we adding new code to the method being marked obsolete?

cijothomas (changes requested):
Not clear how the linked issue is addressed with this PR. The linked issue can be fixed by setting status on Activity (w/o using tags)

src/OpenTelemetry.Api/Trace/ActivityExtensions.cs
vishweshbankwar: Should we
1) mark method obsolete and not set the status
OR
2) set the status and mark it obsolete some time later.

cijothomas: neither is needed to fix the issue described in the PR

src/OpenTelemetry/Trace/ExceptionProcessor.cs
```diff
@@ -75,7 +75,11 @@ public override void OnEnd(Activity activity)
 
             if (snapshot != pointers)
             {
+                // TODO: Remove this when SetStatus is deprecated
```
utpilla, line 78: The old API call is not needed.

vishweshbankwar: That would be a breaking change for exporters still checking tags for status.

utpilla: The exporters have already been updated to also look for `ActivityStatusCode`.

utpilla: Yes. I think we are stuck with this till 2.0 release then. This does incur some perf cost ☹️

utpilla (approved):

cijothomas (approved):
Please add changelog

src/OpenTelemetry/Trace/ExceptionProcessor.cs
```diff
@@ -75,7 +75,11 @@ public override void OnEnd(Activity activity)
...
                 activity.SetStatus(Status.Error);
```
alanwest, line 79: I assume there's never a time we'd call this version of `SetStatus` and not also want to call the "real" version, right?

Here's another spot

https://github.com/open-telemetry/opentelemetry-dotnet/blob/a274f06b2c4c3e76533695b0bd31e177c86b1a30/src/OpenTelemetry.Shims.OpenTracing/SpanBuilderShim.cs#L176

We could kill two stones with one bird if we moved the call to `activity.SetStatus(ActivityStatusCode)` from our extension method. Though, from what I can tell I think that OpenTracing shim is the only other place... so maybe easier to just duplicate the call there as well.

vishweshbankwar: @alanwest I was not aware of this one, will get it updated. what do you think about marking our extension method obsolete?

alanwest: I would support marking it obsolete.

cijothomas: +1

PR #4340, opened 2023-03-28
added a test for Histogram `Update` and `TakeSnapshot` methods

Yun-Ting:
## Changes
Related to https://github.com/open-telemetry/opentelemetry-dotnet/pull/3534


test/OpenTelemetry.Tests/Metrics/AggregatorTest.cs
```diff
@@ -188,6 +188,40 @@ public void HistogramWithOnlySumCount()
             Assert.False(enumerator.MoveNext());
         }
 
+        [Fact]
+        public void MultiThreadedHistogramUpdateAndSnapShotTest()
```
cijothomas, line 192: https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/test/OpenTelemetry.Tests/Metrics/MetricAPITest.cs#L1629 What about the existing tests?

test/OpenTelemetry.Tests/Metrics/AggregatorTest.cs
```diff
@@ -188,6 +188,42 @@ public void HistogramWithOnlySumCount()
...
+                MreToEnsureAllThreadsStart = new ManualResetEvent(false),
+            };
+
+            var numberOfThreads = 10;
+            var snapshotThread = new Thread(HistogramSnapshotThread);
+            Thread[] updateThreads = new Thread[numberOfThreads];
+            for (int i = 0; i < numberOfThreads; ++i)
+            {
+                updateThreads[i] = new Thread(HistogramUpdateThread);
+                updateThreads[i].Start(argsToThread);
```
reyang, line 208: Create the threads (suspend) first, then start the threads. This will increase the chance of race conditions (since thread creation is very time consuming).

reyang: Never mind, I saw a more complex approach later (which uses `mreToEnsureAllThreadsStart` to coordinate the start).

reyang: I guess a more stepping back question is: how likely this test will catch a race condition if the code is indeed buggy?

utpilla: > This test is not trying to catch a race condition.

I think this test would be more valuable if it could actually catch the race condition induced errors when the code has that bug.

Yun-Ting: > I think this test would be more valuable if it could actually catch the race condition induced errors when the code has that bug.

The same test will become flaky in 1.3.0:
Check out: https://github.com/open-telemetry/opentelemetry-dotnet/compare/main-1.3.0...Yun-Ting:yunl/1.3.0-UpdateAndSnapshotWithCoyote?expand=1

I'll collect  data like below to share: waiting for an API (SpinWait) to be supported in Coyote to get the exact numbers.
image

reyang (approved):
LGTM as an exploration of using Coyote.

test/OpenTelemetry.Tests/Metrics/AggregatorTest.cs
Yun-Ting: https://github.com/microsoft/coyote/pull/469

utpilla (approved):

PR #4375, opened 2023-04-06
Update value of the resource attribute `telemetry.sdk.version`

utpilla:
## Changes
- Update value of the resource attribute `telemetry.sdk.version` to show the tag name (which resembles the package version) instead of showing version number with revision

### Current behavior:
- If you use, `1.4.0` version of the SDK, the value logged for telemetry.sdk.version is `1.4.0.802`
- If you use, `1.4.0-rc.4` version of the SDK, the value logged for telemetry.sdk.version is `1.4.0.788`

### With this PR:
- If you use, `1.4.0` version of the SDK, the value logged for telemetry.sdk.version is `1.4.0`
- If you use, `1.4.0-rc.4` version of the SDK, the value logged for telemetry.sdk.version is `1.4.0-rc.4`

Note: You would see this behavior going forward once we start to make newer releases.

### DIFF
--- a/src/OpenTelemetry/CHANGELOG.md
+++ b/src/OpenTelemetry/CHANGELOG.md
@@ -2,13 +2,16 @@
 
 ## Unreleased
 
-* The default resource provided by `ResourceBuilder.CreateDefault()` now
-  adds the `telemetry.sdk.*` attributes defined in the
+* The default resource provided by `ResourceBuilder.CreateDefault()` now adds
+  the `telemetry.sdk.*` attributes defined in the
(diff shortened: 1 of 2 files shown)


vishweshbankwar (approved):

Kielek (approved):

reyang (approved):

reyang:
A non-blocking question - is there a way to show the difference when this is running from a local build (potentially with code changes that are not yet committed)?

alanwest (approved):

utpilla:
> A non-blocking question - is there a way to show the difference when this is running from a local build (potentially with code changes that are not yet committed)?

When you run this from a local build (assuming you haven't added a new tag at the current commit), it would include Git Height either after the `patchVersion` or after the `preReleaseVersion`. This is what it logs when I run it from this feature branch:

telemetry.sdk.version: 1.5.0-alpha.1.**41**

That ".41" is the git height which would otherwise be missing if you used a NuGet package (stable/prerelease).

Issue #4348, opened 2023-03-30
Investigating use of a shared project for common internal classes


TimothyMothra:
## Background

This repo doesn't have a designated location for shared internal classes.
For example, nullable attributes are spread across both the OpenTelemetry project ([link](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/Internal/Shims/NullableAttributes.cs)) and the OpenTelemetry.Api project ([link](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Api/Internal/Guard.cs)).
This creates confusion of where to store new classes and can cause compile-time conflicts.

## Goal

The goal is to have a common shared library, similar to the Contrib Repo's [OpenTelemetry.Contrib.Shared project](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.Contrib.Shared).
This would not be a shipped library, but a location for files included at compile-time.

For example;
```xml
<ItemGroup>
  <Compile Include="$(RepoRoot)\src\SharedInternals\SharedInternals\**" Link="SharedInternals\%(Filename)%(Extension)" />
</ItemGroup>
```

## Prerequisites

- Removing `InternalsVisibleTo` between projects.
  The OpenTelemetry.Api project makes its internals visible to the OpenTelemetry project. 

  - This is mostly used to share common internals.
  - OpenTelemetry SDK.cs ([link](https://github.com/open-telemetry/opentelemetry-dotnet/blob/533513f65b5abb007f51eb6a6c9335053522a200/src/OpenTelemetry/Sdk.cs#L34)) accesses OpenTelemetry.Api Propagators.DefaultTextMapPropagator internal set ([link](https://github.com/open-telemetry/opentelemetry-dotnet/blob/533513f65b5abb007f51eb6a6c9335053522a200/src/OpenTelemetry.Api/Context/Propagation/Propagators.cs#L32)).

- Both OpenTelemetry and OpenTelemetry.Api make their internals visible to the OpenTelemetry.Tests project. 
  This test project would need to be split into two test projects.

## TODO
- Review `ActivityHelperExtensions.GetTagValue()`. See comment: https://github.com/open-telemetry/opentelemetry-dotnet/pull/4368#discussion_r1159154210

TimothyMothra:
There is now a Shared directory.
[...]

PR #4378, opened 2023-04-10
new shared project (OpenTelemetry.Shared) and move Guard.cs

TimothyMothra:
Towards #4348.

## Background
I'm working to create a single shared project for common internal classes.
This will hold internal classes meant to be referenced by libraries in this repo.
This should eliminate the need for InternalsVisableTo and prevent conflicts when expanding internal classes.

This work was inspired by the Contrib repo, however the Contrib repo has two shared projects ([OpenTelemetry.Contrib.Shared](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.Contrib.Shared) and [OpenTelemetry.Internal](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.Internal)).

## Changes

- new project `OpenTelemetry.Shared`. 
  Shared files from OpenTelemetry.csproj or OpenTelemetry.Api.csproj will be moved to this project in a future PR.
- create subdirectory `Internals`.
  This will hold files intended to be included in all other projects.

- move `Guard.cs` from OpenTelemetry.Api to OpenTelemetry.Shared\Internals.
- split `Guard.cs` into three files, one for each class: `Guard`, `NullableAttributes`, and `CallerArgumentExpressionAttribute`

- update other projects with references to `Guard.cs`.
  


src/OpenTelemetry.Api/OpenTelemetry.Api.csproj
```diff
@@ -15,4 +15,9 @@
   <ItemGroup>
     <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="$(SystemDiagnosticSourcePkgVer)" />
   </ItemGroup>
+
+  <ItemGroup>
```
utpilla, line 19: Include these files for the SDK and ProviderBuilderExtensions project as well?

utpilla:
> I'm working to create a single shared project for common internal classes.
This will hold internal classes meant to be referenced by libraries in this repo.
**This should eliminate the need for InternalsVisibleTo and prevent conflicts when expanding internal classes.**

Is the `InternalsVisibleTo` only required for the `Guard` class? The SDK also uses a few other API internal stuff as well such as the setter for `Propagators.DefaultTextMapPropagator`: https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/Sdk.cs#L58 

I don't think we would want to extract out `Propagators` class just to get rid of the `InternalsVisibleTo`.

reyang (changes requested):
Need to discuss about the direction.

reyang:
> * create subdirectory `Internals`.
>   This will hold files intended to be included in all other projects.

Why is such folder needed?

> * split `Guard.cs` into three files, one for each class: `Guard`, `NullableAttributes`, and `CallerArgumentExpressionAttribute`

Why is this split needed?

reyang:
> I don't think we would want to extract out `Propagators` class just to get rid of the `InternalsVisibleTo`.

+1

PR #3789, opened 2022-10-19
Bump System.Text.Json version due to CVE-2021-26701

reyang:
Fixes #3735.

Read more: https://github.com/dotnet/runtime/issues/49377

PR #4390, opened 2023-04-14
Add direct reference to System.Text.Encodings.Web version 4.7.2 due to CVE-2021-26701

### LINKED ISSUE #3735: Snyk.io reports Critical Security Vulnerability in all versions of OpenTelementry.Exporter.Console
Remote Code Execution (RCE) vulnerability
Introduced through
[OpenTelemetry.Exporter.Console@1.3.0](mailto:OpenTelemetry.Exporter.Console@1.3.0)
Fixed in
[System.Text.Encodings.Web@4.5.1](mailto:System.Text.Encodings.Web@4.5.1), @4.7.2, @5.0.1
Exploit maturity
NO KNOWN EXPLOIT
Show less detail 
Detailed paths
Introduced through: [Centene.Observability@*](mailto:Centene.Observability@*) › [OpenTelemetry.Exporter.Console@1.3.0](mailto:OpenTelemetry.Exporter.Console@1.3.0) › [System.Text.Json@4.7.0](mailto:System.Text.Json@4.7.0) › [System.Text.Encodings.Web@4.7.0](mailto:System.Text.Encodings.Web@4.7.0)
Fix: No remediation path available.

Same happens in the latest version - 1.4.0-beta.1. We need to switch to System.Text.Json version 4.7.2.

jrebagliatti:
Fixes #3735 

## Changes

Added direct reference to package `System.Text.Encodings.Web` version `4.7.2` to deal with [CVE-2021-26701](https://github.com/advisories/GHSA-ghhp-997w-qr28). This overrides dependency path to a vulnerable version: `System.Text.Json/4.7.2` -> `SystemText.Encodings.Web/4.7.1`. 

See [my comment](https://github.com/open-telemetry/opentelemetry-dotnet/pull/3789#issuecomment-1506921921) in #3789.

### DIFF
--- a/build/Common.props
+++ b/build/Common.props
@@ -49,6 +49,7 @@
     <StyleCopAnalyzersPkgVer>[1.2.0-beta.435,2.0)</StyleCopAnalyzersPkgVer>
     <SystemCollectionsImmutablePkgVer>1.4.0</SystemCollectionsImmutablePkgVer>
     <SystemDiagnosticSourcePkgVer>7.0.0</SystemDiagnosticSourcePkgVer>
+    <SystemTextEncodingsWebPkgVer>4.7.2</SystemTextEncodingsWebPkgVer>
     <SystemTextJsonPkgVer>4.7.2</SystemTextJsonPkgVer>
     <SystemThreadingTasksExtensionsPkgVer>4.5.4</SystemThreadingTasksExtensionsPkgVer>
   </PropertyGroup>
--- a/build/Common.props
+++ b/build/Common.props
@@ -49,7 +49,7 @@
     <StyleCopAnalyzersPkgVer>[1.2.0-beta.435,2.0)</StyleCopAnalyzersPkgVer>
     <SystemCollectionsImmutablePkgVer>1.4.0</SystemCollectionsImmutablePkgVer>
     <SystemDiagnosticSourcePkgVer>7.0.0</SystemDiagnosticSourcePkgVer>
-    <SystemTextJsonPkgVer>4.7.2</SystemTextJsonPkgVer>
+    <SystemTextJsonPkgVer>5.0.2</SystemTextJsonPkgVer>
     <SystemThreadingTasksExtensionsPkgVer>4.5.4</SystemThreadingTasksExtensionsPkgVer>
   </PropertyGroup>
 
--- a/build/Common.props
+++ b/build/Common.props
@@ -49,7 +49,7 @@
     <StyleCopAnalyzersPkgVer>[1.2.0-beta.435,2.0)</StyleCopAnalyzersPkgVer>
     <SystemCollectionsImmutablePkgVer>1.4.0</SystemCollectionsImmutablePkgVer>
     <SystemDiagnosticSourcePkgVer>7.0.0</SystemDiagnosticSourcePkgVer>
-    <SystemTextJsonPkgVer>5.0.2</SystemTextJsonPkgVer>
+    <SystemTextJsonPkgVer>4.7.2</SystemTextJsonPkgVer>
     <SystemThreadingTasksExtensionsPkgVer>4.5.4</SystemThreadingTasksExtensionsPkgVer>
   </PropertyGroup>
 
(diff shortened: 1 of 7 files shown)


build/Common.props
```diff
@@ -49,7 +49,7 @@
     <StyleCopAnalyzersPkgVer>[1.2.0-beta.435,2.0)</StyleCopAnalyzersPkgVer>
     <SystemCollectionsImmutablePkgVer>1.4.0</SystemCollectionsImmutablePkgVer>
     <SystemDiagnosticSourcePkgVer>7.0.0</SystemDiagnosticSourcePkgVer>
-    <SystemTextJsonPkgVer>4.7.2</SystemTextJsonPkgVer>
+    <SystemTextJsonPkgVer>5.0.2</SystemTextJsonPkgVer>
```
reyang, line 52: I haven't looked into the versioning story. Do we know if `5.*.*` is backward compatible with `4.7.*` and if folks can just bump the version?

If the answer is no, we probably need different versions depending on the target runtime version:

image

reyang: I got some advice from the owner of `System.Text.Json` and `System.Text.Encoding.Web`:

1. The proposed version bump of `System.Text.Json` might give trouble to users who rely on certain features that were affected by these [breaking changes](https://devblogs.microsoft.com/dotnet/whats-next-for-system-text-json/#breaking-changes).
2. There is much higher confidence that `System.Text.Encoding.Web` version bump is very backward compatible (so we don't expect any surprise due to breaking changes).

Given the above information, I suggest that we don't make the `System.Text.Json` version bump, instead, we add explicit dependency to `System.Text.Encoding.Web` and enforce the versions which have fixed the security vulnerabilities.

jrebagliatti: @reyang based on your comments, I've reverted the version bump of `System.Text.Json` and added a direct reference to `System.Text.Encodings.Web` version `4.7.2` to the following projects:
* `Exporter.Console`
* `Exporter.Jaeger`
* `Exporter.Zipkin`

`Instrumentation.Http.Tests` is also referencing `System.Text.Json`, but apparently it's using version `6.0.5` which references version `>6.0.0` of `System.Text.Encodings.Web` which is not affected by the vulnerability.

I see, however, that **_there's at least one other library referencing a vulnerable version of `System.Text.Encodings.Web`_**:
* `OpenTelemetry.Instrumentation.AspNetCore` -> `Microsoft.AspNetCore.Http.Abstractions/2.1.1` -> `System.Text.Encodings.Web/4.5.0`

In this case I think the solution should be to add a direct reference to `System.Text.Encodings.Web/4.5.1` (patched version) as there's no other `Microsoft.AspNetCore.Http.Abstractions/2.1.x` minor release that fix the issue (actually, latest version, `2.2.0`, doesn't fix it either).

If you want me to do the change, please advise how should I proceed as I'm not familiarized with the project's dependency management.

reyang (approved):

build/Common.props
reyang: > I see, however, that **_there's at least one other library referencing a vulnerable version of `System.Text.Encodings.Web`_**:
> 
> * `OpenTelemetry.Instrumentation.AspNetCore` -> `Microsoft.AspNetCore.Http.Abstractions/2.1.1` -> `System.Text.Encodings.Web/4.5.0`
> 
> In this case I think the solution should be to add a direct reference to `System.Text.Encodings.Web/4.5.1` (patched version) as there's no other `Microsoft.AspNetCore.Http.Abstractions/2.1.x` minor release that fix the issue (actually, latest version, `2.2.0`, doesn't fix it either).
> 
> If you want me to do the change, please advise how should I proceed as I'm not familiarized with the project's dependency management.

@jrebagliatti I suggest that we address this in a separate PR.
@vishweshbankwar Would you look into the `OpenTelemetry.Instrumentation.AspNetCore` part?

image

Kielek:
@jrebagliatti, could you please update PR tile to reflect current content?

Kielek (approved):
AutoInstrumentation - as we should avoid bumping libraries, the CVE is good explanation to do it.

Issue #4395, opened 2023-04-17
OTLP Exporter: Consider internalizing Google.Protobuf


alanwest:
#4201 upgraded the OTLP exporter's dependency on Google.Protobuf from 3.19.4 to 3.22.0.

This causes issues for the opentelemetry-dotnet-instrumentation project (see: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/2220). Currently, their project is tightly coupled to the SDK project and any changes in transitive dependencies can cause them issues.

They have requested that we either:
1. Downgrade back to 3.19.4 
2. Internalize Google.Protobuf

Downgrading is not desirable as consumers of the OTLP exporter would see degraded performance unless they were to explicitly reference Google.Protobuf 3.22 or greater. Therefore, this issue is for discussing the possibility of internalizing Google.Protobuf.

Ultimately, this should be an issue for the opentelemetry-dotnet-instrumentation project to solve. It is too prohibitive for the SDK project to not be able to upgrade its transitive dependencies or add new ones as necessary. However, it may be reasonable to address the problem only for Google.Protobuf because:

1. Google.Protobuf is only used by the internal code generated from the OTLP protos.
2. In the longer term, we want to try eliminating our dependency on Google.Protobuf altogether, and implement our own protobuf serialization. The opentelemetry-java project has done this.

As such, an intermediate step towards eliminating Google.Protobuf could be to first internalize it. This would be solely for offering an immediate solution unblocking the opentelemetry-dotnet-instrumentation project.

There are two potential methods for internalizing Google.Protobuf.

1. Forklift Google.Protobuf code into the OTLP exporter project.
2. Use a tool like [ILRepack](https://github.com/gluck/il-repack).

Forklifting Google.Protobuf may be the more desirable option. It would probably enable us to more easily analyze it for compatibility for things like AOT, trimming, and single-file deploys. However, it is a lot of code (~50k lines). It is possible to identify and remove unused code, but this could be very time consuming and more difficult to maintain.

rajkumar-rangaraj:
Fixed with 1.11.0-rc.1 version of OpenTelemetry.Exporter.OpenTelemetryProtocol package.

PR #4396, opened 2023-04-17
OTLP Exporter: Forklift and internalize Google.Protobuf

alanwest:
Solution for #4395 by forklifting Google.Protobuf into the OTLP exporter project.

This PR copies the code from [Google.Protobuf @ v3.22](https://github.com/protocolbuffers/protobuf/tree/v22.0/csharp/src/Google.Protobuf). The code is modified to suppress Stylecop and mark all public items internal.

Options to consider:

1) Accept this PR as is.
    * Pros: If we ever need to update to a later version of Google.Protobuf, the maintenance burden may be lower. Basically replace the code with newer version, suppress stylecop, and make everything internal.
    * Cons: It's a lot of code. Binary size increases by about 425KB. This might be offset by the fact that consumers may no longer need to also pull the Google.Protobuf dependency.
2) Attempt to trim out unused code.
    * Pros: Makes the binary smaller. I'm not sure by how much without doing the work. I suspect not by much. I've reviewed things a bit and it seems the code generated from the OTLP protos depends on quite a bit from Google.Protobuf.
    * Cons: Tedious. I have not found good tooling that helps automate this much. The maintenance burden would definitely be higher in the event we ever wanted to update to a later version.
3) Reject this PR and pursue the option using ILRepack.
    * Pros: Significantly less code to maintain. ILRepack is a tool that merges the MSIL from two binaries. It allows you to completely internalize a dependency. I have a [branch](https://github.com/open-telemetry/opentelemetry-dotnet/compare/main...alanwest:opentelemetry-dotnet:alanwest/protobuf-ilrepack) that demonstrates its use. This option also makes updating to a later version trivial.
    * Cons: The binary size increases by about 425KB just as though we had copied the code and compiled it ourselves.
4) Reject internalizing Google.Protobuf altogether. This may limit or hinder the opentelemetry-dotnet-instrumentation project from upgrading to the 1.5.0 SDK.


reyang:
5. use https://github.com/protobuf-net/protobuf-net
6. use https://github.com/protobuf-net/protobuf-net and work with the owner to find a proper solution (maybe codegen?)

reyang (changes requested):
Two reasons to block the PR for now:
1. Need to figure out whether copying more than one hundred files is the right path forward (considering all the future obligations of security hotfixes and patches).
2. Need to understand the licensing / legal implications.

alanwest:
> 5. use https://github.com/protobuf-net/protobuf-net
> 6. use https://github.com/protobuf-net/protobuf-net and work with the owner to find a proper solution (maybe codegen?)

I suspect these options would present the same concern for the opentelemetry-dotnet-instrumentation project. Though, @Kielek can weigh in.

> 1. Need to figure out whether copying more than one hundred files is the right path forward (considering all the future obligations of security hotfixes and patches).

For the reason of hotfixes/patches, if we seek to internalize anything - whether it be Google.Protobuf of protobuf-net - I'm beginning to lean more towards looking at ILRepack. Though, I'm not interested in spending much more time on this if I know up front that we're not keen on internalizing anything. Maybe @Kielek can share how this would affect the auto-instrumentation project.

> 2. Need to understand the licensing / legal implications.

This is likely a concern whether we copy code or use a tool like ILRepack. BSD-3 is a permissive license as I understand it, but I'm not a lawyer. Does the CNCF have a process for this?

reyang:
7. handcraft a tailored version of serializer, similar to what https://github.com/open-telemetry/opentelemetry-dotnet/blob/9a7bd9f55ad3f643484fe7d62d92c2553c623da1/src/OpenTelemetry.Exporter.Prometheus.HttpListener/Internal/PrometheusSerializer.cs is doing.

reyang:
> > 7. handcraft a tailored version of serializer
> 
> This is actually option 4 😆 - that is, don't internalize anything. My thought behind #4395 was to have the discussion about the viability of internalizing Google.Protobuf out in the open with the intent that it would be an intermediate step towards "option 7". I agree option 7 is where we want to go in the long run.

Here is my suggestion (from most preferred to least preferred):
1. Don't reinvent wheel if there is already a good solution. Seek guidance from the maintainers of https://github.com/protobuf-net/protobuf-net - as the issue we're facing here is not something specific to us, most likely many folks have resolved this issue already.
2. If there is no good solution from 1), we can explore a handcrafted version of serializer, make sure perf is awesome (specifically, make sure we do great job on memory utilization). https://github.com/protobuf-net/protobuf-net/issues/754
3. If we need to steal code or have package dependency "smoothed out" by ILRepack. The first thing we should be very clear about, is **which** one to bet on - https://stackoverflow.com/questions/64359801/what-nuget-package-should-i-use-between-protobuf-net-and-google-protobuf-for-a-n.

reyang:
Some updates from my conversation with the .NET team:
* If we have to choose between `Google.Protobuf` and `protobuf-net`, we should use `protobuf-net` since it is the one that works well with AOT.
* Most likely we'll end up with a handcrafted version, however we might want to seek some guidance from Mark Gravell (I have yet to search his GitHub alias) who is the author of protobuf.net.
* If we need to look into gRPC client lib, we might want to consult @JamesNK who did the ASP.NET Core implementation of gRPC.

JamesNK:
> If we have to choose between Google.Protobuf and protobuf-net, we should use protobuf-net since it is the one that works well with AOT.

??? Google.Protobuf is the library that works with AOT. protobuf-net uses a lot of dynamic code generation

> If we need to look into gRPC client lib, we might want to consult @JamesNK who did the ASP.NET Core implementation of gRPC.

gRPC code generation uses Google.Protobuf. I don't know whether it would be possible to use an internal copy.

reyang:
> > If we have to choose between Google.Protobuf and protobuf-net, we should use protobuf-net since it is the one that works well with AOT.
> 
> ??? Google.Protobuf is the library that works with AOT. protobuf-net uses a lot of dynamic code generation
> 
> > If we need to look into gRPC client lib, we might want to consult @JamesNK who did the ASP.NET Core implementation of gRPC.
> 
> gRPC code generation uses Google.Protobuf. I don't know whether it would be possible to use an internal copy.

Appreciate your input @JamesNK 👍 I should clarify that I was relaying messages based on what I heard from several folks, I have no expertise in protobuf so you need to verify what I put above - FYI @alanwest

mgravell:
> If we have to choose between Google.Protobuf and protobuf-net, we should use protobuf-net since it is the one that works well with AOT.

To echo James' point: yes, *currently* protobuf-net uses ref-emit at runtime, because it was originally designed for code-first scenarios; AOT focus is in progress, via Roslyn analyzers for code-first, and the regular code-gen for contract-first, but: the AOT story is not complete today.

> Most likely we'll end up with a handcrafted version, however we might want to seek some guidance from Mark Gravell (I have yet to search his GitHub alias) who is the author of protobuf.net.

That would be @mgravell

alanwest:
Yes, thank you for both your input, @JamesNK and @mgravell!

> AOT focus is in progress, via Roslyn analyzers for code-first, and the regular code-gen for contract-first, but: the AOT story is not complete today.

@mgravell in #4430, I'm using the contract-first approach and using [protobuf-net.BuildTools to perform code-gen](https://github.com/open-telemetry/opentelemetry-dotnet/pull/4430/files#diff-94154eaf1371ec02c1f01166462762840492afb26a7eca7f60f70a79efb41dabR46-R51). Will AOT support be a concern here?

PR #4409, opened 2023-04-19
Sdk internal metrics - BatchExportProcessorDroppedCount via ObservableCounter

vishweshbankwar:
Towards #3880 

## Changes
Address https://github.com/open-telemetry/opentelemetry-dotnet/pull/4312#pullrequestreview-1377793906

Adding `otel.dotnet.sdk.batchprocessor.dropped_count` metric with following dimensions

`batchprocessor.data.type` - Type of data that is being exported e.g. `Activity` or `LogRecord`.
`batchprocessor.exporter.name` - Name of the exporter e.g. `OtlpExporter`.

This PR shows an alternate approach to the one in https://github.com/open-telemetry/opentelemetry-dotnet/pull/4312. It uses `ObservableCounter` so that we can use existing `DroppedCount` on `BatchExportProcessor`. The overall idea is to add a callback to the list of `otel.dotnet.sdk.batchprocessor.dropped_count` callbacks everytime a new `BatchExportProcessor` is created. 

**Memory Management**
The callback will be removed during the `Dispose` of processor which will also address https://github.com/open-telemetry/opentelemetry-dotnet/pull/4289#discussion_r1134390550. Confirmed this behavior by taking memory snapshots during and after execution of following function

MicrosoftTeams-image

Here is the diff
MicrosoftTeams-image (1)
This confirms that the instance is properly disposed and not held by `SdkHealthReporter`.

In case if we don't clean up the Callbacks during `dispose` the diff will look like this
MicrosoftTeams-image (2)
Meaning the object could not be collected due to an active reference

===============
**Example Output**
image

**NOTE:**
Current changes does not consider a case of multiple instances of Providers and BatchExportProcessors of same data type. This can be handled separately once we have a basic working version.


src/OpenTelemetry/BatchExportProcessor.cs
```diff
@@ -72,6 +77,11 @@ public abstract class BatchExportProcessor<T> : BaseExportProcessor<T>
             this.scheduledDelayMilliseconds = scheduledDelayMilliseconds;
             this.exporterTimeoutMilliseconds = exporterTimeoutMilliseconds;
             this.MaxExportBatchSize = maxExportBatchSize;
+
+            this.exporterName = exporter.GetType().Name;
+            this.exportedDataType = typeof(T).Name;
+            this.processorId = Interlocked.Increment(ref processorCount);
```
reyang, line 83: What does this id mean? (e.g. how would the user use it?)

vishweshbankwar: This is to be added as a dimension on metric (In a scenario where they have same exporterName and exportedDataType).

reyang (approved):
The code looks good to me, regarding the ID, if the intention is to distinguish different instances, maybe a GUID would serve the purpose better than an ordinal number (both would have unlimited cardinality and in practice should not be a big issue since we don't expect many instances from a single process).

vishweshbankwar:
> The code looks good to me, regarding the ID, if the intention is to distinguish different instances, maybe a GUID would serve the purpose better than an ordinal number (both would have unlimited cardinality and in practice should not be a big issue since we don't expect many instances from a single process).

@reyang Could you please elaborate on how GUID will be a better dimension than ordinal number?

reyang:
> @reyang Could you please elaborate on how GUID will be a better dimension than ordinal number?

For example if there are two processes sending the metrics and they both have ID = 0, would the user be confused.

vishweshbankwar:
> > @reyang Could you please elaborate on how GUID will be a better dimension than ordinal number?
> 
> For example if there are two processes sending the metrics and they both have ID = 0, would the user be confused.

Thanks!, Updated it to guid.

vishweshbankwar:
> > > @reyang Could you please elaborate on how GUID will be a better dimension than ordinal number?
> > 
> > 
> > For example if there are two processes sending the metrics and they both have ID = 0, would the user be confused.
> 
> Thanks!, Updated it to guid.

@reyang, I reverted it for now to integer. Added description for each dimension that can be used as reference for further discussion.

reyang:
> > > > @reyang Could you please elaborate on how GUID will be a better dimension than ordinal number?
> > > 
> > > 
> > > For example if there are two processes sending the metrics and they both have ID = 0, would the user be confused.
> > 
> > 
> > Thanks!, Updated it to guid.
> 
> @reyang, I reverted it for now to integer. Added description for each dimension that can be used as reference for further discussion.

👍

vishweshbankwar:
@open-telemetry/dotnet-maintainers  - Please have a look.

PR #4429, opened 2023-04-25
Update diagnostic logging for MeterProviderSdk

utpilla:
## Changes
- Update diagnostic logging for MeterProviderSdk to clearly log if SDK enabled measurements for the instrument


src/OpenTelemetry/Metrics/MeterProviderSdk.cs
```diff
@@ -309,9 +313,13 @@ internal sealed class MeterProviderSdk : MeterProvider
                                     listener.EnableMeasurementEvents(instrument, metrics);
                                 }
                             }
-                        }
 
-                        OpenTelemetrySdkEventSource.Log.MeterProviderSdkEvent($"Completed publishing Instrument = \"{instrument.Name}\" of Meter = \"{instrument.Meter.Name}\".");
+                            OpenTelemetrySdkEventSource.Log.MeterProviderSdkEvent($"Enabled measurements for Instrument = \"{instrument.Name}\" of Meter = \"{instrument.Meter.Name}\".");
```
vishweshbankwar, line 317: Should not log if [this](https://github.com/open-telemetry/opentelemetry-dotnet/blob/ef0c9217087bfca0d6fe9206ccde343eb4e5738c/src/OpenTelemetry/Metrics/MeterProviderSdk.cs#L298) return `null`

utpilla: Good catch. Fixed this.

vishweshbankwar (approved):

cijothomas (approved):
Good improvement! Suggested to make the wording even more user friendly.

src/OpenTelemetry/Metrics/MeterProviderSdk.cs
```diff
@@ -307,11 +321,19 @@ internal sealed class MeterProviderSdk : MeterProvider
...
                         }
 
-                        OpenTelemetrySdkEventSource.Log.MeterProviderSdkEvent($"Completed publishing Instrument = \"{instrument.Name}\" of Meter = \"{instrument.Meter.Name}\".");
+                        if (enabledMeasurements)
+                        {
+                            OpenTelemetrySdkEventSource.Log.MeterProviderSdkEvent($"Enabled measurements for Instrument = \"{instrument.Name}\" of Meter = \"{instrument.Meter.Name}\".");
+                        }
+                        else
+                        {
+                            OpenTelemetrySdkEventSource.Log.MeterProviderSdkEvent($"Measurements not enabled for Instrument = \"{instrument.Name}\" of Meter = \"{instrument.Meter.Name}\".");
```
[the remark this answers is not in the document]
utpilla: I have updated the log message.

src/OpenTelemetry/Metrics/MeterProviderSdk.cs
```diff
@@ -242,11 +245,19 @@ internal sealed class MeterProviderSdk : MeterProvider
...
                                     listener.EnableMeasurementEvents(instrument, metricsSuperList);
+                                    enabledMeasurements = true;
                                 }
                             }
                         }
 
-                        OpenTelemetrySdkEventSource.Log.MeterProviderSdkEvent($"Completed publishing Instrument = \"{instrument.Name}\" of Meter = \"{instrument.Meter.Name}\".");
+                        if (enabledMeasurements)
+                        {
+                            OpenTelemetrySdkEventSource.Log.MeterProviderSdkEvent($"Measurements for Instrument = \"{instrument.Name}\" of Meter = \"{instrument.Meter.Name}\" are enabled and would be processed and aggregated by the SDK.");
```
reyang, line 255: I found this a bit hard to understand - what does "measurements are enabled" mean?

utpilla: It means we have called this [MeterListener.EnableMeasurementEvents(Instrument, Object)](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.metrics.meterlistener.enablemeasurementevents?view=net-8.0) for the given instrument.

reyang: ^ this is an internal thing, how would a user interpret this? (or the log is not intended to be used by the user?)

reyang: > > ^ this is an internal thing, how would a user interpret this?
> 
> The full message tells what this means for the user: "SDK will not process or it would drop the measurements."
> 
> > or the log is not intended to be used by the user?
> 
> I believe that just because we log some internal things doesn't mean that it cannot be used by users. We could still ask them to collect logs and look for a particular message. Btw we have been logging internal stuff in our logs. See this for [example](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/Metrics/PeriodicExportingMetricReader.cs#L141).

Makes sense, in a nutshell we treat all error/warning logs as user-facing (so they have to be understood by the users, with clear action items), while info/verbose/debug logs are for OpenTelemetry developers (it is okay that the users don't understand the message) - is this aligned with your understanding?

utpilla: Yes, I agree.

reyang (approved):

utpilla:
There is scope to improve the diagnostic logging further by clearly telling which `MeterProvider` are we logging for in case of multiple MeterProvider instances. The current logging would still be helpful if multiple MeterProvider instances are created sequentially and not in parallel. This could be covered in a future PR.

PR #4431, opened 2023-04-25
TraceIdRatioBasedSampler enhancement

### LINKED ISSUE #4393: TraceIdRatioBasedSampler question
https://github.com/open-telemetry/opentelemetry-dotnet/blob/30b7b79e9c170e3dd58f3b80a458dcb47cc0bde0/src/OpenTelemetry/Trace/TraceIdRatioBasedSampler.cs#L67-L80

Is it possible to return when the `idUpperBound` is `long.MaxValue`/`long.MinValue`
to reduce the unnecessary `stackalloc` and comparison

``` c#
public override SamplingResult ShouldSample(in SamplingParameters samplingParameters)
{
    if (idUpperBound == long.MinValue) return new SamplingResult(false);
    if (idUpperBound == long.MaxValue) return new SamplingResult(true);
    // ...
}
```

WeihanLi:
Fixes #4393 
Design discussion issue #

## Changes

Return result without calculating when the ratio is the upper bound or lower bound

### DIFF
--- a/src/OpenTelemetry/Trace/TraceIdRatioBasedSampler.cs
+++ b/src/OpenTelemetry/Trace/TraceIdRatioBasedSampler.cs
@@ -62,12 +62,16 @@ namespace OpenTelemetry.Trace
                 this.idUpperBound = (long)(probability * long.MaxValue);
             }
         }
+        
+        /// <summary>
+        /// The desired probability of sampling. This must be between 0.0 and 1.0.
+        /// Higher the value, higher is the probability of a given Activity to be sampled in.
+        /// </summary>
+        public double Probability => probability;
 
         /// <inheritdoc />
         public override SamplingResult ShouldSample(in SamplingParameters samplingParameters)
         {
-            if (idUpperBound == long.MinValue) return new SamplingResult(false);
-            if (idUpperBound == long.MaxValue) return new SamplingResult(true);
             // Always sample if we are within probability range. This is true even for child activities (that
             // may have had a different sampling decision made) to allow for different sampling policies,
             // and dynamic increases to sampling probabilities for debugging purposes.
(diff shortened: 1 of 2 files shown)


src/OpenTelemetry/Trace/TraceIdRatioBasedSampler.cs
```diff
@@ -66,6 +66,8 @@ public TraceIdRatioBasedSampler(double probability)
         /// <inheritdoc />
         public override SamplingResult ShouldSample(in SamplingParameters samplingParameters)
         {
+            if (idUpperBound == long.MinValue) return new SamplingResult(false);
```
utpilla, line 69: This would not help in most cases as the expected usage of `TraceIdRatioBasedSampler` when the probability is set to something in (0.0,1.0) (excluding 0.0 and 1.0: we have dedicated `AlwaysOffSampler` and `AlwaysOnSampler` for them respectively). In fact, this is only going to lower the performance by adding two more if checks in the hot-path for most of the cases.

reyang: @WeihanLi how important is this optimization?
If it is indeed a common path with good perf yield, this could be the right place to put the logic https://github.com/open-telemetry/opentelemetry-dotnet/blob/1143c52e739a8af22da0ccd213ba6ba16b8f9241/src/OpenTelemetry/Trace/TracerProviderSdk.cs#L234.

WeihanLi: We're using the `TraceIdRatioBasedSampler` in our code, and for the test environment, we're using `1.0` for the sampling rate, while for production maybe `0.1` or `0.01` would be used, for this case I think it's a point can be improved

PR #4442, opened 2023-04-26
[Do not merge] Add measurement dropped metric

vishweshbankwar:
Fixes #
Sample output
image

## Changes


src/OpenTelemetry/Metrics/MetricReaderExt.cs
```diff
@@ -72,6 +72,10 @@ internal Metric AddMetricWithNoViews(Instrument instrument)
                     try
                     {
                         metric = new Metric(metricStreamIdentity, this.GetAggregationTemporality(metricStreamIdentity.InstrumentType), this.maxMetricPointsPerMetricStream, exemplarFilter: this.exemplarFilter);
+                        if (this.parentProvider?.GetSdkHealthReporter() != null)
+                        {
+                            metric.AddAggStoreCallback();
```
utpilla, line 77: You could probably mark `Metric.aggStore` internal and avoid adding these new methods on the Metric class.

src/OpenTelemetry/Metrics/MetricReader.cs
```diff
@@ -226,6 +226,19 @@ public bool Shutdown(int timeoutMilliseconds = Timeout.Infinite)
...
+            if (this.metrics != null)
+            {
+                foreach (var metric in this.metrics)
+                {
+                    if (metric == null)
+                    {
+                        break;
+                    }
+
+                    metric.AggStore.RemoveMeasurementDroppedCallbacks();
```
utpilla, line 238: This should be done in the `MeterListener.MeasurementsCompleted` callback.

utpilla, line 238: This would not work for multiple `MetricReader` scenarios.

utpilla: This should work on a combination of `Instrument` and `MetricReader`. So we should also remove the entry from the dictionary when the instrument's `Meter` is disposed.

src/OpenTelemetry/Internal/SdkHealthReporter.cs
```diff
@@ -0,0 +1,65 @@
...
+
+namespace OpenTelemetry.Internal
+{
+    internal sealed class SdkHealthReporter
+    {
+        private static readonly Dictionary<int, Func<Measurement<long>>> MeasurementDroppedCountCallbacks = new();
+
+        private static readonly Meter InternalMeter = new Meter("OpenTelemetry.Sdk");
+
+        private static readonly ObservableCounter<long> MeasurementDroppedCountObs = InternalMeter.CreateObservableCounter("otel.dotnet.sdk.measurement.dropped_count", GetMeasurementDroppedCounts);
```
utpilla, line 29: We should avoid any SDK specific features such as `Views`, `SetMaxMetrics`, `SetMaxMetricStreamPoints` to be applied to any of our internal metrics. This would have to be done in the `MeterProviderSdk`.

src/OpenTelemetry/Metrics/MetricReader.cs
vishweshbankwar: Added the removal on MeasureComplete callbacks.

src/OpenTelemetry/Metrics/MetricReaderExt.cs
```diff
@@ -157,6 +161,13 @@ internal List<Metric> AddMetricsListWithViews(Instrument instrument, List<Metric
                     else
                     {
                         Metric metric = new(metricStreamIdentity, this.GetAggregationTemporality(metricStreamIdentity.InstrumentType), this.maxMetricPointsPerMetricStream, this.exemplarFilter);
+                        if (this.parentProvider?.GetSdkHealthReporter() != null)
```
utpilla, line 164: Is this check needed?

vishweshbankwar: Yes - as we are planning to add this feature as experimental, we would only add it if the `SdkHealthReporter` is initialized for the provider.

utpilla: Add this here as well then? https://github.com/open-telemetry/opentelemetry-dotnet/pull/4442/files#diff-8650a8083e99ad454eac9d5bef1a6061cc0c1ab31085a1d25e5c4f57ed16f74fR77

vishweshbankwar: Oh, I missed it there. Will add it

PR #4465, opened 2023-05-04
Exponential histogram LowerBoundary implementation

alanwest:
Provides an implementation of the LowerBoundary method [described in the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/data-model.md#exponential-buckets) (i.e., the inverse of `MapToIndex`).

This method is implemented in its own class separate from the `Base2ExponentialHistogram` class because the `LowerBoundary` implementation will only be consumed by something like the console exporter which needs to display bucket boundaries in a human readable way.

A follow up PR will integrate `LowerBoundary` into the console exporter.


src/OpenTelemetry/Internal/Base2ExponentialHistogramHelper.cs
```diff
@@ -0,0 +1,120 @@
...
+            // It should return values >= double.Epsilon (2 ^ -1074).
+            // n < -1074 occurs at the minimum index of a scale.
+            // e.g., At scale -1, minimum index is -538. -538 << 1 = -1075
+            if (n < -1074)
+            {
+                return double.Epsilon;
+            }
+
+#if NET6_0_OR_GREATER
+            return Math.ScaleB(1, n);
```
utpilla, line 55: Left-shift is probably faster?
```suggestion
            return 2 << (n - 1);
```

src/OpenTelemetry/Internal/Base2ExponentialHistogramHelper.cs
```diff
@@ -0,0 +1,120 @@
...
+#else
+            return ScaleB(1, n);
+#endif
+        }
+    }
+
+#if !NET6_0_OR_GREATER
+    // Math.ScaleB was introduced in .NET Core 3.0.
+    // This is the implementation of from:
```
utpilla, line 64:
```suggestion
    // This implementation is from:
```

src/OpenTelemetry/Internal/Base2ExponentialHistogramHelper.cs
alanwest: Math.ScaleB(1, n) != 2 << (n - 1) for n < 1. Some math involving a right shift might work? I don't think this is a super perf critical path.

utpilla: Got it.

utpilla (approved):
Approving to unblock progress.

PR #4494, opened 2023-05-16
Make exponential histogram negative buckets internal

### LINKED ISSUE #4489: Should NegativeBuckets of ExponentialHistogram be made public?
The [spec ](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#record:~:text=The%20value%20is%20expected%20to%20be%20non%2Dnegative.%20This%20API%20SHOULD%20be%20documented%20in%20a%20way%20to%20communicate%20to%20users%20that%20this%20value%20is%20expected%20to%20be%20non%2Dnegative.%20This%20API%20SHOULD%20NOT%20validate%20this%20value%2C%20that%20is%20left%20to%20implementations%20of%20the%20API.)currently says that `Histogram.Record` should only be used for non-negative values. However, the [spec ](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/data-model.md#exponential-buckets:~:text=The%20positive%20and%20negative%20ranges%20of%20the%20histogram%20are%20expressed%20separately.%20Negative%20values%20are%20mapped%20by%20their%20absolute%20value%20into%20the%20negative%20range%20using%20the%20same%20scale%20as%20the%20positive%20range.%20Note%20that%20in%20the%20negative%20range%2C%20therefore%2C%20histogram%20buckets%20use%20lower%2Dinclusive%20boundaries.) also mentions how to deal with negative values in case of ExponentialHistograms. The current implementation of ExponentialHistogram does process negative values and puts them into `NegativeBuckets`.

Do we want to make `NegativeBuckets` available to exporters considering we should also not expect people to use `Histogram.Record` for negative values?

alanwest:
Fixes #4489

This behavior differs subtlety with our explicit bucket histogram implementation.

Negative values recorded using the explicit bucket aggregation will contribute to the sum, count, min, max, and (-infinity, 0] bucket count.

Negative values recorded using the exponential bucket aggregation are ignored.

### DIFF
--- a/test/OpenTelemetry.Tests/Metrics/AggregatorTest.cs
+++ b/test/OpenTelemetry.Tests/Metrics/AggregatorTest.cs
@@ -291,7 +291,11 @@ namespace OpenTelemetry.Metrics.Tests
             foreach (var value in valuesToRecord)
             {
                 aggregatorStore.Update(value, Array.Empty<KeyValuePair<string, object>>());
-                expectedHistogram.Record(value);
+
+                if (value >= 0)
+                {
+                    expectedHistogram.Record(value);
+                }
             }
 
             aggregatorStore.Snapshot();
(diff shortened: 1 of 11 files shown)


reyang (approved):

test/OpenTelemetry.Tests/Metrics/AggregatorTest.cs
```diff
@@ -291,7 +291,11 @@ internal void ExponentialHistogramTests(AggregationType aggregationType, Aggrega
             foreach (var value in valuesToRecord)
             {
                 aggregatorStore.Update(value, Array.Empty<KeyValuePair<string, object>>());
-                expectedHistogram.Record(value);
+
+                if (value >= 0)
```
utpilla, line 295: I think we want to remove this `if` check in the test to make sure that negative values are not aggregated by the SDK.

utpilla (approved):

PR #4509, opened 2023-05-18
Analysis of the imprecision of  MapToIndex

alanwest:
I do not necessarily intend for this PR to be merged. It's purpose is simply to drive conversation regarding the imprecision of the exponential histogram's `MapToIndex` function. Though, if folks would find it useful to merge, I'm ok with that.

At positive scales, `MapToIndex` can be imprecise for values near a bucket boundary.

https://github.com/open-telemetry/opentelemetry-dotnet/blob/1510b76634f4fdfbf79cd73e4b71aaf474e40a46/src/OpenTelemetry/Metrics/Base2ExponentialBucketHistogram.cs#L153-L166

There are two TODOs in the above code related to the imprecision of `MapToIndex`.

1. The code block before the first TODO ensures that values that are exact powers of two map to the correct index. This may only be marginally useful, however the [spec does suggest](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/data-model.md#all-scales-use-the-logarithm-function):

    > However, when inputs are an exact power of two, it is possible to calculate the exactly correct bucket index. Since it is relatively simple to check for exact powers of two, implementations SHOULD apply such a special case

2. The second TODO addresses the general issue that values near a bucket boundary that are not exact powers of two may get mapped to the incorrect index. This imprecision is due to intermediate rounding that occurs in the computation of `Math.Log(value) * this.scalingFactor`.

---

This PR contains a test for analyzing the imprecision of `MapToIndex` at scale one for values near indexes within the range [-50, 50]. It only tests buckets at indexes that have a lower bound which is an exact power of 2 (every two buckets at scale one).

At scale one, this range of indexes can record values in the range [2**-25, 2**26] or approximately [0.000000029802322387695312, 67108864].

The test generates the following output:
```text
Successes: 18
Failures: 33
Average number of values near a bucket boundary that are off by one: 2.878787878787879
Average range of values near a bucket boundary that are off by one: 3.0880480139955346E-09
```

That is, for 18 of the tested indexes, values around that index are mapped precisely. For 33 of the tested indexes there exist values that are mapped to an index off by one.

For the 33 indexes tested with values mapped incorrectly, there are on average ~2.9 values near the index that are mapped off by one. The average range of these 2.9 values is 3.0880480139955346E-09.

---

There may exist use cases where this margin of error would be considered unacceptable. Though, one use case where I would speculate it would be acceptable is for monitoring the latency of a web server. For example, assuming OpenTelemetry's default unit of seconds for `http.server.duration` the margin of error in this test's scenario is ~3 nanoseconds.

We could always attempt to pursue an exact mapping function, but for now, I'd assert we can move forward with the current implementation. I'm all for documenting the limitations/imprecision of the implementation so that users can decide for themselves if this is the correct tool for their purposes. The analysis in this PR may be useful for articulating the limitations.

Lastly, there is a precedence for other imprecise implementations of MapToIndex. For example, I have customers using  the following implementations today: [OpenTelemetry Java](https://github.com/open-telemetry/opentelemetry-java/blob/93bcdf28706c1b6830394dc3a6d4dae1844e1c90/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/aggregator/Base2ExponentialHistogramIndexer.java#L82-L84) and [NrSketch](https://github.com/newrelic-experimental/newrelic-sketch-java/blob/d87cb49233ab65dd5c0c3cd9253c77c981ad25a9/src/main/java/com/newrelic/nrsketch/indexer/LogIndexer.java#L29-L33).

@utpilla @reyang Please share your thoughts/critiques of this analysis. I'd be happy to expand on this analysis if you'd like (e.g., different scales, different ranges of index, etc). I have observed ranges where the margin of error is greater especially for very large and very small double values.


reyang:
> There may exist use cases where this margin of error would be considered unacceptable. Though, one use case where I would speculate it would be acceptable is for monitoring the latency of a web server. For example, assuming OpenTelemetry's default unit of seconds for `http.server.duration` the margin of error in this test's scenario is ~3 nanoseconds.

+100, I think as long as we set clear expectations and users clearly understand the consequences, the current version is good enough for common scenarios.

reyang (changes requested):
I think this test is great and we should merge it. In the future if someone wants to make it mathematically precise (e.g. for a nuclear powerplant), they would find the test/analysis helpful.

PR #4512, opened 2023-05-21
Support custom header in JaegerPropagator

### LINKED ISSUE #3836: JaegerPropagator support custom values for JaegerHeader
# Feature Request

Currently, JaegerPropagator checks the header "uber-trace-id" for existing trace values. The issue is that this header is not configurable and stored as a const, so I would like to suggest making this configurable, but keep the current value as default.

Yes, the consumer of my application uses the header "X-Trace" to send trace information to my service.

Converting the JaegerHeader variable to a readonly and set it from the constructor. If it's not present there, use the default value "uber-trace-id".

-

Martinay:
Fixes #3836 

This PR allows setting a custom jaeger header name in the JaegerPropagator, e.g. _jaeger-trace-id_ instead of _uber-trace-id_.

The outdated C# Client for Jaeger supported a custom header names by providing a method [WithSpanContextKey](https://github.com/jaegertracing/jaeger-client-csharp/blob/a29d3fe4ae16d01d94fd57e4134f065fa5805cac/src/Jaeger.Core/Propagation/TextMapCodec.cs#L156). This PR will add support for that as well. [Link to a test in the C# Client for Jaeger](https://github.com/jaegertracing/jaeger-client-csharp/blob/a29d3fe4ae16d01d94fd57e4134f065fa5805cac/test/Jaeger.Core.Tests/Propagation/TextMapCodecTests.cs#L13) which checks the header "jaeger-trace-id".

## Changes
- Add second constructor which allows changing the jaeger header name
- Add unit tests

### DIFF
--- a/src/OpenTelemetry.Extensions.Propagators/.publicApi/net462/PublicAPI.Unshipped.txt
+++ b/src/OpenTelemetry.Extensions.Propagators/.publicApi/net462/PublicAPI.Unshipped.txt
@@ -1,0 +1,1 @@
+OpenTelemetry.Extensions.Propagators.JaegerPropagator.JaegerPropagator(string jaegerHeader) -> void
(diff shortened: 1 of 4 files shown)


utpilla:
@Martinay The propagator implementation logic is driven by the spec. Any such changes to the existing behavior should be coming from the spec.

You could try to create an issue in the spec repo for this to get the spec updated first. Or you could write you own custom propagator as suggested by @Kielek in https://github.com/open-telemetry/opentelemetry-dotnet/issues/3836#issuecomment-1294714805.

PR #4523, opened 2023-05-27
Add ToOtlpLog unit tests for scopes

### LINKED ISSUE #2873: OTLPLogExporter tests
https://github.com/open-telemetry/opentelemetry-dotnet/pull/2871 - A very basic test is added here. Opening an issue to track adding more tests to OTLP Log Exporter. The exporter itself is not fully ready and is tracked here - https://github.com/open-telemetry/opentelemetry-dotnet/issues/2482

mfogliatto:
Fixes #2873

## Changes

Add ToOtlpLog unit tests for scopes


test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/OtlpLogExporterTests.cs
```diff
@@ -607,6 +608,241 @@ public void Export_WhenExportIsSuccessful_ReturnsExportResultSuccess()
...
+                    options.AddInMemoryExporter(logRecords);
+                });
+            });
+            var logger = loggerFactory.CreateLogger("Some category");
+
+            const string expectedScopeKey = "Some scope key";
+            const string expectedScopeValue = "Some scope value";
+
+            // Act.
+            using (logger.BeginScope(new List<KeyValuePair<string, object>>
```
cijothomas, line 630: could you add the case where scope is something other than List<KVP> ? like a simple string, for example, or some other ds like dictionary, etc.

cijothomas (approved):
LGTM.
Not a blocker, but we need to add tests when scope is something other than List<KVP> as well.
Also worth adding is the case where the keys are duplicated in multiple scope levels and in the logger.log itself.

Issue #4525, opened 2023-05-30
Activity Display Name: Route Template instead of path


elvince:
# Bug Report

We have api versioning for our endpoints which .net supports so for example: [Route("api/v{apiVersion:apiVersion}/[controller]")] which is being reported as api/v{apiVersion:apiVersion}/users

## Symptom

**What is the expected behavior?**

api/v1/user
api/v2/user

**What is the actual behavior?**

api/v{apiVersion:apiVersion}/users

we have the partial route template and {apiVersion:apiVersion} is not transformed where as [Controller] is.

It seems we can hack this behavior by using this: 

```
.AddAspNetCoreInstrumentation(o =>
      {
          o.EnrichWithHttpResponse = (activity, response) =>
          {
                  var displayName = $"{response.HttpContext.Request.Path}";
                  activity.DisplayName = displayName;
          };
      })
```

But it would be more appropriate if this is fixed in the library itself.

If it is by design, could you please tell me why you choose to not have the version in the display name?

We are using opentelemetry and the displayname is mapped to trace group. At the end, we don't know what version of our APIs had bad performance or issues as everything is aggregated.

Thanks,

dhabierre:
Same behavior here with dotnet 7 and latest OpenTelemetry nugets.

```
http.flavor: 1.1 http.method: GET http.route: api/v{version:apiVersion}/Tiers/{userId} http.scheme: http http.status_code: 200 net.host.name: localhost net.host.port: 8080 Histogram
```

The `{version:apiVersion}` should be substituted because of multiple versions.
The `{userId}` shouldn't?

Controller:

```cs
namespace My.WebApi.Controllers.V1
{
    ...

    [ApiController]
    [Route("api/v{version:apiVersion}/[controller]")]
    [ApiVersion("1.0")]
    public sealed class TiersController : ControllerBase
    {
        ...

        [HttpGet("{userId}")]
        [ProducesResponseType(typeof(GetTiersListByUserIdResponse), StatusCodes.Status200OK)]
        [ProducesResponseType(StatusCodes.Status404NotFound)]
        public async Task<IActionResult> GetTiersByUserId(int userId)
        {
            ...
        }
    }
}
```

Thanks.

Want100Cookies:
This also makes it pretty hard to search traces in Jaeger. The displayName is used as operation name results in a very bloated index which can be prevented if we use the route template.

[...]

Issue #2967, opened 2022-03-03
HTTP OPTIONS request , transaction name(display name) is shown as request path instead of HTTP route ( template ) in asp.net core api app


madhub:
# Bug Report
HTTP OPTIONS request, transaction name(display name) is shown as **request path** instead of **HTTP route ( template )** in asp.net core api app

Expected transaction name(display name) : **Todo/{tenantid}/{version}**
What is shown is : **/todo/456-123-56abc-x134/2022-02-01**

It seems to happen for HTTP OPTIONS request ,like browser HTTP OPTIONS pre-flight request to check the CORS.

## List of all OpenTelemetry NuGet
packages](https://www.nuget.org/profiles/OpenTelemetry) and version that you are
using (e.g. `OpenTelemetry 1.0.2`):

ASP.NET Core 3.1 , Windows 10
Zipkin & Console exporter is configured 

OTel version
- OpenTelemetry.Exporter.Zipkin1.1.0
- OpenTelemetry.Instrumentation.AspNetCore 1.0.0-rc7
- OpenTelemetry.Extensions.Hosting 1.0.0-rc7
- System.Diagnostics.DiagnosticSource 5.0.1

## Runtime version :

ASP.NET Core 3.1 , Windows 10

## Symptom

For HTTP OPTIONS request,  transaction name(display name) is shown as **request path** instead of **HTTP route ( template )** in asp.net core api app

**What is the expected behavior?**

Expected transaction name (display name):  **Todo/{tenantid}/{version}**

**What is the actual behavior?**

What is shown is : **/todo/456-123-56abc-x134/2022-02-01**

## Reproduce
Any ASP.NET Core 3.1 API app with above packages & send HTTP OPTIONS request ,like browser HTTP OPTIONS pre-flight request to check the CORS.
POST request
```curl
curl -v -X POST http://localhost:5000/todo/456-123-56abc-x134/2022-02-01 -d {}
```
OPTIONS request
```curl
curl -v -X OPTIONS http://localhost:5000/todo/456-123-56abc-x134/2022-02-01 -H "Access-Control-Request-Method: POST"   -H "Access-Control-Request-Headers: content-type"   -H "Origin: https://reqbin.com" -d {}
```
C# Sample
```csharp
[...]

PR #4532, opened 2023-05-31
Add api-version support

dhabierre:
When using a versioned API with `Microsoft.AspNetCore.Mvc.Versioning` nuget package, the `{version:apiVersion}` is not replaced with the real controller version.

Discussion issue #2967 & #4525

## Changes

- Update `OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs` in order to replace `{version:apiVersion}` url segment with the real controller version.
- Add ApiVersioning support & ApiVersioningControllers in `TestApp.AspNetCore` project

This is my first PR here, hope all is fine :)
Any feedback about my proposal is welcome.

### DIFF
--- a/src/OpenTelemetry.Exporter.Prometheus.AspNetCore/CHANGELOG.md
+++ b/src/OpenTelemetry.Exporter.Prometheus.AspNetCore/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Added [api-versioning](https://github.com/dotnet/aspnet-api-versioning/wiki) support
+  [#2967](https://github.com/open-telemetry/opentelemetry-dotnet/issues/2967), [4525](https://github.com/open-telemetry/opentelemetry-dotnet/issues/4525)
+
 ## 1.5.0-rc.1
 
 Released 2023-May-25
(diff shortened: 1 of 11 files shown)


alanwest:
@dhabierre Thanks for the contribution! Including API version information in the Activity name (and/or the `http.route` attribute) sounds like it could be a really good idea. However, we're going to hold off on accepting this PR at the moment.

There is a strong push at the spec level going on at the moment to stabilize the HTTP semantic conventions and we're actively working to align our instrumentation to the conventions. Routing is complex as there are many scenarios to consider handling (API versioning being a potential scenario), so naming the Activity and including the route information is one of the trickier parts to hash out before we can release stable ASP.NET Core instrumentation.

@vishweshbankwar will be working with the ASP.NET Core team to sort out the optimal way to retrieve the route information we include in the Activity. We'll add API version to the list of things we're considering. Please bear with us as we work through this.

@vishweshbankwar suggested a potential solution in the interim could be to use `Enrich` to modify the Activity as you need it.

dhabierre:
Hi @alanwest and thanks for your post.
Hope that versionned APIs will be supported soon :)

Issue #4484, opened 2023-05-12
Http instrumentation libraries stable release plan


vishweshbankwar:
https://github.com/open-telemetry/opentelemetry-specification/pull/3443 lays out the detailed plan for stabilizing the http instrumentation libraries.

PRs that include major changes
https://github.com/open-telemetry/opentelemetry-specification/pull/3402
https://github.com/open-telemetry/opentelemetry-specification/pull/3426
https://github.com/open-telemetry/opentelemetry-specification/pull/3355

For .NET, we will proceed as follows

1) Update the following instrumentation libraries to look up `OTEL_SEMCONV_STABILITY_OPT_IN` environment variable and set the attributes as per the values specified in [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#:~:text=SHOULD%20introduce%20an,a%20seamless%20transition.).
* OpenTelemetry.Instrumentation.AspNetCore
* OpenTelemetry.Instrumentation.Http
* OpenTelemetry.Instrumentation.SqlClient
* OpenTelemetry.Instrumentation.GrpcNetClient

**Note**: The change impacts all the above libraries due to updates in networking attributes https://github.com/open-telemetry/opentelemetry-specification/pull/3402

2) Release the RC version with the above mentioned changes.

3) Drop `OTEL_SEMCONV_STABILITY_OPT_IN` environment variable from `OpenTelemetry.Instrumentation.AspNetCore` and `OpenTelemetry.Instrumentation.Http` in the final RC version.

**Note**: Other libraries should continue to look up the environment variable.

4) Release stable 1.0.0 version of `OpenTelemetry.Instrumentation.AspNetCore` and `OpenTelemetry.Instrumentation.Http`.

TimothyMothra:
FYI: this doc is useful for tracking changes to the conventions:
https://github.com/open-telemetry/opentelemetry-specification/blob/v1.22.0/schemas/1.21.0

TimothyMothra:
FYI.
Semantic-Conventions repo is readying their first release. The version will be v1.21.0. No ETA because they're building out the processes.
https://github.com/open-telemetry/semantic-conventions/blob/df8e53054147acf49918582bd446d0f524170071/CHANGELOG.md

Trask advised that we should link to "main" for now.

I will create a new PR to correct everything that merged already.

danelson:
[...]

PR #4537, opened 2023-06-01
update HttpSemanticConventions for Instrumentation.AspNetCore

TimothyMothra:
Design discussion issue #4484

Http Semantic Convention is introducing breaking changes.
This PR makes the change in the Instrumentation.AspNetCore library.

## Changes
- add new attributes to `SemanticConventions`
- update `Instrumentation.AspNetCore`
  - `HttpInListener`
  - `HttpInMetricsListener`

@vishweshbankwar please review.


src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs
```diff
@@ -197,27 +197,55 @@ public void OnStartActivity(Activity activity, object payload)
                 var path = (request.PathBase.HasValue || request.Path.HasValue) ? (request.PathBase + request.Path).ToString() : "/";
                 activity.DisplayName = path;
 
-                // see the spec https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md
-                if (request.Host.HasValue)
+                // see the spec https://github.com/open-telemetry/semantic-conventions/blob/main/specification/trace/semantic_conventions/http.md
```
vishweshbankwar, line 200: nit: Include https://github.com/open-telemetry/opentelemetry-specification/blob/v1.20.0/specification/trace/semantic_conventions/http.md
and
https://github.com/open-telemetry/semantic-conventions/blob/main/specification/trace/semantic_conventions/http.md separately for old and new.

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs
```diff
@@ -197,27 +197,55 @@ public void OnStartActivity(Activity activity, object payload)
...
+                        if (request.Host.Port is not null && request.Host.Port != 80 && request.Host.Port != 443)
+                        {
+                            activity.SetTag(SemanticConventions.AttributeServerPort, request.Host.Port);
+                        }
+                    }
+
+                    activity.SetTag(SemanticConventions.AttributeHttpRequestMethod, request.Method);
+                    activity.SetTag(SemanticConventions.AttributeUrlScheme, request.Scheme);
+                    activity.SetTag(SemanticConventions.AttributeUrlPath, path);
+                    activity.SetTag(SemanticConventions.AttributeUrlFull, GetUri(request));
```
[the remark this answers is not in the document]
TimothyMothra: Previously, this was `http.url`. was this a bug?

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs
```diff
@@ -197,29 +197,65 @@ public void OnStartActivity(Activity activity, object payload)
...
+                        if (request.Host.Port is not null && request.Host.Port != 80 && request.Host.Port != 443)
+                        {
+                            activity.SetTag(SemanticConventions.AttributeServerPort, request.Host.Port);
+                        }
+                    }
+
+                    if (request.QueryString.HasValue)
                     {
-                        activity.SetTag(SemanticConventions.AttributeHttpUserAgent, userAgent);
+                        activity.SetTag(SemanticConventions.AttributeUrlQuery, request.QueryString.Value);
```
vishweshbankwar, line 244: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#:~:text=%5B8%5D%3A%20Sensitive%20content%20provided%20in%20query%20string%20SHOULD%20be%20scrubbed%20when%20instrumentations%20can%20identify%20it.

Is there is a way to do this in .NET? need some clarity on this one.

TimothyMothra: That's a good callout. Would need to ask around and see if any other languages are doing this and how.

One point to note, the old code wasn't sanitizing query string: https://github.com/open-telemetry/opentelemetry-dotnet/blob/352f93942a0000743d8c151aca7f05787e33c779/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs#L432

TimothyMothra: I opened #4571 to track this.

src/OpenTelemetry.Api/Internal/SemanticConventions.cs
```diff
@@ -110,5 +110,19 @@ internal static class SemanticConventions
...
         public const string AttributeExceptionMessage = "exception.message";
         public const string AttributeExceptionStacktrace = "exception.stacktrace";
+
+        // NEW v1.21.0 Http Semantic Conventions
+        public const string AttributeClientSocketPort = "client.socket.port";
+        public const string AttributeHttpRequestMethod = "http.request.method";
+        public const string AttributeHttpResponseStatusCode = "http.response.status_code";
+        public const string AttributeNetworkProtocolVersion = "network.protocol.version";
+        public const string AttributeServerAddress = "server.address";
+        public const string AttributeServerName = "server.name";
```
vishweshbankwar, line 120: where is `server.name` mentioned in spec?

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs
vishweshbankwar: Thanks! - A todo comment with link to issue would be helpful

src/OpenTelemetry.Api/Internal/SemanticConventions.cs
TimothyMothra: good catch!
looks like a typo. net.host.name -> server.address
looks like i did not make the same mistake in the other pr.

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/IncomingRequestsCollectionsIsAccordingToTheSpecTests_New.cs
```diff
@@ -0,0 +1,197 @@
...
+            string urlPath,
+            string query,
+            string userAgent,
+            int statusCode,
+            string reasonPhrase,
+            bool recordException = false)
+        {
+            try
+            {
+                Environment.SetEnvironmentVariable("OTEL_SEMCONV_STABILITY_OPT_IN", "http");
```
vishweshbankwar, line 56: Could we include `http/dup` as well here?

TimothyMothra: done.

src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+* Updated [Http Semantic Conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/specification/trace/semantic_conventions/http.md).
```
utpilla, line 5: Probably better to point to a particular commit/tag so instead of main.

src/OpenTelemetry.Api/Internal/SemanticConventions.cs
```diff
@@ -110,5 +110,17 @@ internal static class SemanticConventions
...
         public const string AttributeExceptionMessage = "exception.message";
         public const string AttributeExceptionStacktrace = "exception.stacktrace";
+
+        // v1.21.0 Http Semantic Conventions
+        public const string AttributeClientSocketPort = "client.socket.port";
+        public const string AttributeHttpRequestMethod = "http.request.method";
+        public const string AttributeHttpResponseStatusCode = "http.response.status_code";
+        public const string AttributeNetworkProtocolVersion = "network.protocol.version";
+        public const string AttributeServerAddress = "server.address";
+        public const string AttributeServerPort = "server.port";
```
utpilla, line 120: nit:
It would be good to add comments to show what the old value for each of these were.

```csharp
public const string AttributeServerPort = "server.port"; // Used to be "net.host.port" (AttributeNetHostPort)
```

src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
TimothyMothra: The new repo doesn't have any tags yet! 😭
I'll get the commit url.

src/OpenTelemetry.Api/Internal/SemanticConventions.cs
TimothyMothra: great idea. this would have saved me so much time.

TimothyMothra: done

src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
TimothyMothra: updated

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/IncomingRequestsCollectionsIsAccordingToTheSpecTests_Dupe.cs
```diff
@@ -0,0 +1,204 @@
...
+        {
+            this.factory = factory;
+        }
+
+        [Theory]
+        [InlineData("/api/values", null, "user-agent", 503, "503")]
+        [InlineData("/api/values", "?query=1", null, 503, null)]
+        [InlineData("/api/exception", null, null, 503, null)]
+        [InlineData("/api/exception", null, null, 503, null, true)]
+        public async Task SuccessfulTemplateControllerCallGeneratesASpan_Dupe(
```
utpilla, line 46: Could we move these two new tests to the existing `IncomingRequestsCollectionsIsAccordingToTheSpecTests` class? That way they could also reuse `ValidateTagValue` and `TestCallbackMiddlewareImpl` from the existing class.

utpilla (approved):

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/IncomingRequestsCollectionsIsAccordingToTheSpecTests_Dupe.cs
TimothyMothra: I split these into separate classes because I was getting concurrency issues causing test failures.
Also having it this way will make it easier to clean up when this library goes stable (ie: delete the old & dupe).

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs
```diff
@@ -439,10 +484,19 @@ private static void AddGrpcAttributes(Activity activity, string grpcMethod, Http
...
 
-            activity.SetTag(SemanticConventions.AttributeNetPeerPort, context.Connection.RemotePort);
+            if (this.httpSemanticConvention.HasFlag(HttpSemanticConvention.Old))
+            {
+                activity.SetTag(SemanticConventions.AttributeNetPeerPort, context.Connection.RemotePort);
+            }
+
+            if (this.httpSemanticConvention.HasFlag(HttpSemanticConvention.New))
+            {
+                activity.SetTag(SemanticConventions.AttributeClientSocketPort, context.Connection.RemotePort);
```
utpilla, line 498: Are we clear on what this tag is meant for? And what does this value `context.Connection.RemotePort` represent?

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInMetricsListener.cs
```diff
@@ -82,20 +82,44 @@ public override void OnEventWritten(string name, object payload)
 
                 TagList tags = default;
 
-                tags.Add(new KeyValuePair<string, object>(SemanticConventions.AttributeHttpFlavor, HttpTagHelper.GetFlavorTagValueFromProtocol(context.Request.Protocol)));
-                tags.Add(new KeyValuePair<string, object>(SemanticConventions.AttributeHttpScheme, context.Request.Scheme));
-                tags.Add(new KeyValuePair<string, object>(SemanticConventions.AttributeHttpMethod, context.Request.Method));
-                tags.Add(new KeyValuePair<string, object>(SemanticConventions.AttributeHttpStatusCode, TelemetryHelper.GetBoxedStatusCode(context.Response.StatusCode)));
+                // see the spec https://github.com/open-telemetry/opentelemetry-specification/blob/v1.20.0/specification/trace/semantic_conventions/http.md
+                if (this.httpSemanticConvention.HasFlag(HttpSemanticConvention.Old))
```
CodeBlanch, line 86: Some thoughts...

* Just reading this work I found "New" and "Old" pretty confusing. Why not `Legacy` vs `Stable` or something along those lines?

* This is probably an issue for some users: https://github.com/open-telemetry/opentelemetry-dotnet/blob/5d321ce621a387b4cff226ecc1dbcc5f3b6611bb/src/OpenTelemetry.Api/Internal/HttpSemanticConventionHelper.cs#L53

   We should be using `IConfiguration` to get the switch.

* Shouldn't the environment key have HTTP in the name like `OTEL_HTTP_SEMCONV_STABILITY_OPT_IN`?

reyang: > * Just reading this work I found "New" and "Old" pretty confusing. Why not `Legacy` vs `Stable` or something along those lines?

+1, I had similar concern here https://github.com/open-telemetry/opentelemetry-dotnet/pull/4538#discussion_r1232783666

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs
TimothyMothra: I reviewed the spec and confirmed this was the wrong attribute. will change to `server.port`

- v1.20.0
  https://github.com/open-telemetry/opentelemetry-specification/blob/v1.20.0/specification/trace/semantic_conventions/http.md#http-client
  image

- v1.21.0
  https://github.com/open-telemetry/opentelemetry-specification/blob/v1.21.0/specification/trace/semantic_conventions/http.md#http-server-semantic-conventions
  image

TimothyMothra: Regarding the value of `context.Connection.RemotePort`, the documentation states:
> Gets or sets the port of the remote target
https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.connectioninfo.remoteport?view=aspnetcore-7.0

This is not changed, it was mapped to the former attribute.
https://github.com/open-telemetry/opentelemetry-dotnet/blob/2a6d88c35749c181f5c0816e994965c88228af49/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs#L491-L494

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInMetricsListener.cs
TimothyMothra: 1. > Just reading this work I found "New" and "Old" pretty confusing. Why not Legacy vs Stable or something along those lines?

How strongly do you feel about this? :)
We can change the enum, but `New` and `Old` shipped as part of OpenTelemetry.Api v1.5.0.
It's `internal` so we CAN change it.
But this would require re-releasing the Api library. (ie Instrumentation.AspNetCore would require an OpenTelemetry.Api v1.5.1).

Also, this is meant to be short lived. When these instrumentation libraries go stable this enum and helper class will be removed.

2. > We should be using IConfiguration to get the switch.

Can you please share an example of this?

3. > Shouldn't the environment key have HTTP in the name like OTEL_HTTP_SEMCONV_STABILITY_OPT_IN?

That makes sense to me, but the spec defines [OTEL_SEMCONV_STABILITY_OPT_IN](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#:~:text=OTEL_SEMCONV_STABILITY_OPT_IN)

src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+* Updated [Http Semantic Conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.21.0/specification/trace/semantic_conventions/http.md).
+  * This library can emit either old, new, or both attributes. Users can control
```
utpilla, line 6: We don't need a sub-bullet here. This change is simple enough to be put under the same bullet.

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInMetricsListener.cs
utpilla: > We can change the enum, but New and Old shipped as part of OpenTelemetry.Api v1.5.0.
It's internal so we CAN change it.
But this would require re-releasing the Api library. (ie Instrumentation.AspNetCore would require an OpenTelemetry.Api v1.5.1).

The instrumentation libraries do not use this `enum` from `OpenTelemetry.Api`. They link to the file so each of the libraries have their own copy of it. You should be able to make the necessary changes without worrying about version of `OpenTelemetry.Api`.

Note: This does not have to be done in this PR.

utpilla:
Merging this PR to unblock progress.

As discussed in the SIG meeting today. @TimothyMothra would work on using `IConfiguration` and updating the `enum` in follow-up PRs.

src/OpenTelemetry.Api/Internal/SemanticConventions.cs
```diff
@@ -110,5 +110,17 @@ internal static class SemanticConventions
...
         public const string AttributeExceptionMessage = "exception.message";
         public const string AttributeExceptionStacktrace = "exception.stacktrace";
+
+        // Http v1.21.0 https://github.com/open-telemetry/opentelemetry-specification/blob/v1.21.0/specification/trace/semantic_conventions/http.md
+        public const string AttributeClientSocketPort = "client.socket.port"; // replaces: "net.peer.port" (AttributeNetPeerPort)
+        public const string AttributeHttpRequestMethod = "http.request.method"; // replaces: "http.method" (AttributeHttpMethod)
+        public const string AttributeHttpResponseStatusCode = "http.response.status_code"; // replaces: "http.status_code" (AttributeHttpStatusCode)
+        public const string AttributeNetworkProtocolVersion = "network.protocol.version"; // replaces: "http.flavor" (AttributeHttpFlavor)
+        public const string AttributeServerAddress = "server.address"; // replaces: "net.host.name" (AttributeNetHostName)
+        public const string AttributeServerPort = "server.port"; // replaces: "net.host.port" (AttributeNetHostPort)
```
utpilla, line 120: // replaces: "net.peer.port" and "net.host.port"

PR #4542, opened 2023-06-02
Fix AOT warnings for RuntimeContext

utpilla:
Towards #3429 

This PR introduces a **Breaking change**. This would no longer support assigning custom types to `RuntimeContext.ContextSlotType`.

## Changes
- Fix AOT warnings for `RuntimeContext` class
- Drop support for custom implementations of `RuntimeContextSlot<T>`


reyang (approved):

Yun-Ting (approved):

eerhardt (approved):
Looks great! Thanks!

src/OpenTelemetry.Api/Context/RuntimeContext.cs
```diff
@@ -27,10 +27,40 @@ public static class RuntimeContext
...
+#if NETFRAMEWORK
+                else if (value == typeof(RemotingRuntimeContextSlot<>))
+                {
+                    contextSlotType = value;
+                }
+#endif
+                else
+                {
+                    throw new NotSupportedException($"{value} is not a supported type.");
+                }
```
Kielek, line 59: Should it really throw Exception? As I remember OTel SDK/API should rather silently ignore non-working parameters than throwing exception.

Just to double check if Event/Log is not a better option.

src/OpenTelemetry.Api/CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+* **Breaking change** `RuntimeContext.ContextSlotType` can only be assigned one
+  of the following types: `AsyncLocalRuntimeContextSlot<>`,
+  `ThreadLocalRuntimeContextSlot<>`, and `RemotingRuntimeContextSlot<>`. A
+  `System.NotSupportedException` will be thrown if you try to assign any type
+  other than the three types mentioned.
+  ([#4542](https://github.com/open-telemetry/opentelemetry-dotnet/pull/4542))
```
alanwest, line 10:
```suggestion
* **Breaking change** In order to make `RuntimeContext` compatible with
  ahead-of-time compilation (AOT),
  `RuntimeContext.ContextSlotType` can only be assigned one
  of the following types: `AsyncLocalRuntimeContextSlot<>`,
  `ThreadLocalRuntimeContextSlot<>`, and `RemotingRuntimeContextSlot<>`. A
  `System.NotSupportedException` will be thrown if you try to assign any type
  other than the three types mentioned.
  ([#4542](https://github.com/open-telemetry/opentelemetry-dotnet/pull/4542))
```

alanwest (approved):
Just a suggestion to add context for why this change is being made.

src/OpenTelemetry.Api/Context/RuntimeContext.cs
eerhardt: In this case, this isn't a "someone configured it wrong" or "something went wrong" situation. This is a "the developer used the API incorrectly" situation. So I think it is better to let the developer know early that what they are trying to do isn't possible. (i.e. throw an exception).

reyang: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/error-handling.md#basic-error-handling-principles

_The API or SDK MAY fail fast and cause the application to fail on initialization_

PR #4545, opened 2023-06-05
Update CHANGELOG for 1.5.0 (core)/ 1.5.0-beta.1 (non-core) release

utpilla:
## Changes
- Update CHANGELOG for 1.5.0 release
- Bumped the version of non-core package to keep their major and minor version in sync with the core packages


src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
```diff
@@ -2,6 +2,15 @@
 
 ## Unreleased
 
+## 1.5.0-rc.1
```
utpilla, line 5: @reyang @cijothomas We are bumping the package versions of non-core packages to keep their major and minor versions in sync.

reyang: Do we want rc or beta?

utpilla: I think we could use beta. The only concern with using beta is that users might feel a bit weird/uncomfortable when upgrading from an `rc` to a `beta` version (even though there is bump in the minor version).

reyang: Which one (rc or beta) communicates the actual intention?

alanwest: I'm good with beta... it's been a bit of a lie that we've been releasing `rc`s this whole time as we've shipped a number of breaking changes in the past. Though I agree, that some folks may wonder why we've effectively downgraded from `rc` to `beta`.

utpilla: I think that would be beta. The reason I chose rc was that for the last 2.5 years, we have been releasing rc and suddenly moving people to beta might come across as retrograding.

I don't have a strong opinion here. I could change this to beta if folks want.

src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
```diff
@@ -2,6 +2,15 @@
...
+
+Released 2023-Jun-05
+
+* Bumped the package version to `1.5.0-rc.1` to keep its major and minor version
+  in sync with that of the core packages. This would make it more intuitive for
+  users to figure out what version of core packages would work with a given
+  version of this package.
```
alanwest:
```suggestion
* Bumped the package version to `1.5.0-rc.1` to keep its major and minor version
  in sync with that of the core packages. This would make it more intuitive for
  users to figure out what version of core packages would work with a given
  version of this package. The pre-releases identifier has also been changed
  from `rc` to `beta` as we believe this more accurately reflects the status of
  this package. We believe the `rc` identifier will be more appropriate as
  semantic conventions reach stability. 
```

src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
alanwest: Just a suggestion in the event we decide to go with `beta` https://github.com/open-telemetry/opentelemetry-dotnet/pull/4545#discussion_r1218705425

reyang: I don't have strong opinions here, totally okay with whichever decision the maintainers agreed - 1) keep telling the lie if that's what most users want 2) stop telling the lie if the negative impact is reasonably small.

utpilla: Updated the version to `1.5.0-beta.1`

alanwest (approved):

vishweshbankwar (approved):

PR #4548, opened 2023-06-06
Add Console Activity Json Exporter

brunomartinspro:
## Changes

Added a Console Activity Json Exporter to output logs in JSON.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -26,8 +26,8 @@
     <PackageVersion Include="StyleCop.Analyzers" Version="[1.2.0-beta.435,2.0)" />
     <PackageVersion Include="System.Diagnostics.DiagnosticSource" Version="7.0.0" />
     <PackageVersion Include="System.Reflection.Emit.Lightweight" Version="4.7.0" />
-    <PackageVersion Include="System.Text.Encodings.Web" Version="4.7.2" />
-    <PackageVersion Include="System.Text.Json" Version="4.7.2" />
+    <PackageVersion Include="System.Text.Encodings.Web" Version="7.0.0" />
+    <PackageVersion Include="System.Text.Json" Version="7.0.2" />
     <PackageVersion Include="System.Threading.Tasks.Extensions" Version="4.5.4" />
   </ItemGroup>
   <!-- non-prod packages -->
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -2,7 +2,6 @@
   <PropertyGroup>
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
   </PropertyGroup>
-
   <!-- prod packages -->
   <ItemGroup>
     <PackageVersion Include="Google.Protobuf" Version="[3.19.4,4.0)" />
@@ -22,6 +21,7 @@
     <PackageVersion Include="Microsoft.NETFramework.ReferenceAssemblies" Version="[1.0.3,2.0)" />
     <PackageVersion Include="Microsoft.SourceLink.GitHub" Version="[1.1.1,2.0)" />
     <PackageVersion Include="MinVer" Version="[4.2.0,5.0)" />
+    <PackageVersion Include="Newtonsoft.Json" Version="13.0.1" />
     <PackageVersion Include="OpenTracing" Version="[0.12.1,0.13)" />
     <PackageVersion Include="StyleCop.Analyzers" Version="[1.2.0-beta.435,2.0)" />
     <PackageVersion Include="System.Diagnostics.DiagnosticSource" Version="7.0.0" />
@@ -30,7 +30,6 @@
     <PackageVersion Include="System.Text.Json" Version="4.7.2" />
     <PackageVersion Include="System.Threading.Tasks.Extensions" Version="4.5.4" />
   </ItemGroup>
-
   <!-- non-prod packages -->
   <!-- 'net7.0' is the default `TargetFramework`. Use `VersionOverride` in the project to override the package versions from a different `TargetFramework` -->
   <ItemGroup>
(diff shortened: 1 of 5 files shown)


src/OpenTelemetry.Exporter.Console/OpenTelemetry.Exporter.Console.csproj
```diff
@@ -16,6 +16,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Newtonsoft.Json" />
```
Kielek, line 19: There is an existing reference to `System.Text.Json`. Can you use it, instead of adding new dependency?

Directory.Packages.props
```diff
@@ -22,15 +21,15 @@
     <PackageVersion Include="Microsoft.NETFramework.ReferenceAssemblies" Version="[1.0.3,2.0)" />
     <PackageVersion Include="Microsoft.SourceLink.GitHub" Version="[1.1.1,2.0)" />
     <PackageVersion Include="MinVer" Version="[4.2.0,5.0)" />
+    <PackageVersion Include="Newtonsoft.Json" Version="13.0.1" />
```
Kielek, line 24: It is not used.

reyang:
What type of JSON do we use? I suggest that we start from https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/file-exporter.md#json-file-serialization.

src/OpenTelemetry.Exporter.Console/ConsoleActivityJsonExporter.cs
```diff
@@ -0,0 +1,160 @@
+using System.Diagnostics;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using OpenTelemetry.Resources;
+using OpenTelemetry.Trace;
+
+namespace OpenTelemetry.Exporter
+{
+    public class ConsoleActivityJsonExporter : ConsoleExporter<Activity>
```
utpilla, line 9: @brunomartinspro What's the motivation behind adding this new exporter to the repo? `ConsoleExporter` is a simple lightweight exporter meant for quick and naive testing/ troubleshooting for the developers. It was designed to be simple.

Directory.Packages.props
```diff
@@ -26,11 +25,10 @@
     <PackageVersion Include="StyleCop.Analyzers" Version="[1.2.0-beta.435,2.0)" />
     <PackageVersion Include="System.Diagnostics.DiagnosticSource" Version="7.0.0" />
     <PackageVersion Include="System.Reflection.Emit.Lightweight" Version="4.7.0" />
-    <PackageVersion Include="System.Text.Encodings.Web" Version="4.7.2" />
-    <PackageVersion Include="System.Text.Json" Version="4.7.2" />
+     <PackageVersion Include="System.Text.Encodings.Web" Version="7.0.0" />
+    <PackageVersion Include="System.Text.Json" Version="7.0.2" />
```
reyang, line 29: What's the impact of this version bump? How many breaking changes would this introduce?

E.g. https://blog.georgekosmidis.net/net-6-a-guide-for-the-high-impact-breaking-changes.html

reyang (changes requested):
https://github.com/open-telemetry/opentelemetry-dotnet/pull/4548/files#r1220139771

Directory.Packages.props
```diff
@@ -26,11 +25,10 @@
     <PackageVersion Include="StyleCop.Analyzers" Version="[1.2.0-beta.435,2.0)" />
     <PackageVersion Include="System.Diagnostics.DiagnosticSource" Version="7.0.0" />
     <PackageVersion Include="System.Reflection.Emit.Lightweight" Version="4.7.0" />
-    <PackageVersion Include="System.Text.Encodings.Web" Version="4.7.2" />
-    <PackageVersion Include="System.Text.Json" Version="4.7.2" />
+     <PackageVersion Include="System.Text.Encodings.Web" Version="7.0.0" />
```
reyang, line 28: nit, indent

src/OpenTelemetry.Exporter.Console/ConsoleActivityJsonExporter.cs
utpilla: @cijothomas Why do we want to add this to the existing exporter though? Users are allowed to create their own custom exporters and use them.

cijothomas: Its good to support a more standard/predictable output in the consoleexporter, without forcing everyone to author their own exporter.
The predictable/standard output could be the https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#json-protobuf-encoding OR could be the Prometheus exposition format for Metrics (there were issues asking for these earlier (this capability existed long back, but got removed), can't find it with my quick search)

utpilla: > Its good to support a more standard/predictable output in the consoleexporter, without forcing everyone to author their own exporter.

I definitely see value in standardizing the ConsoleExporter output. If we decide to go forward with this though, I think the implementation should meet certain requirements:
- We want to keep ConsoleExporter simple, so I don't think we should make this an optional feature. Instead, we should simply update the existing behavior to output data in the standardized format
- ConsoleExporter should not take any additional dependencies or bump the version of existing dependencies that might cause the user to run into version conflict issues with their application code

cijothomas: Agree with this.
It would also be good to add disclaimer that the output format could change (it has changed many times in the past) and people should not take dependency on it. (i.e use it for learning purposes, but not for production purposes to have telemetry output to stdout, and have some agent parse it to store telemetry)

PR #4550, opened 2023-06-06
Add comments to package dependency to clarify the direction

reyang:
I think this will help to clarify the direction and also help to provide context/clarity to contributors (e.g. https://github.com/open-telemetry/opentelemetry-dotnet/pull/4548/files#r1220139771).


cijothomas (approved):

Directory.Packages.props
```diff
@@ -24,10 +24,20 @@
     <PackageVersion Include="MinVer" Version="[4.2.0,5.0)" />
     <PackageVersion Include="OpenTracing" Version="[0.12.1,0.13)" />
     <PackageVersion Include="StyleCop.Analyzers" Version="[1.2.0-beta.435,2.0)" />
+
+    <!-- The latest stable version of System.Diagnostics.DiagnosticSource should be used here since the .NET runtime team provides extra backward compatibility guarantee even during major version bumps. -->
```
Kielek, line 28: It is not fully true, 7.0.2 is released.

utpilla (approved):

Directory.Packages.props
tarekgh: @Kielek what do you mean you that? What breaking did you run into when upgrading to 7.0.2?

tarekgh: Would there be any problem referencing 7.0.2 if it is available? Why do you want to restrict it to the major release version? just asking to get more context (not objecting to anything here).

tarekgh: `<PackageVersion Include="System.Diagnostics.DiagnosticSource" Version="7.0.0" />` can be `<PackageVersion Include="System.Diagnostics.DiagnosticSource" Version="7.0.2" />` if needed.

reyang: > No breaking changes. The first part of sentence is not true. In fact the latest major release should be referenced.

I think we should agree on which of the following is the intended one:
1) The latest stable version of System.Diagnostics.DiagnosticSource should be used here.
2) The latest stable major version of System.Diagnostics.DiagnosticSource should be used here.

And to me 1) is the right direction. Of course, if we agree on that, we should bump the version to 7.0.2 and that can/should be a separate PR.

alanwest: or

3. Use the similar guidance for System.Diagnostics.DiagnosticSource as you're suggesting for the other packages: "A conservative version of System.Diagnostics.DiagnosticSource must be used here even though there are strong backward compatibility guarantees during major version bumps."

I'm not opposed to updating DiagnosticSource ref when we need to, but Is there a good reason to always ensure we're on the latest DiagnosticSource even if the latest version does not have any new features we need?

Kielek: For the best Automatic Instrumentation user experience, my vote is for a conservative major version (probably included into the lowest supported .NET version). If we can do that, we will have really good user experience in November 2024 for all supported versions.

If we allow to bump version here to the latest minor it will increase range of the issue: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/55ec192a11aa3dac58d57f173118957a25f84b74/docs/troubleshooting.md#assembly-version-conflicts

reyang: > or
> 
> 3. Use the similar guidance for System.Diagnostics.DiagnosticSource as you're suggesting for the other packages: "A conservative version of System.Diagnostics.DiagnosticSource must be used here even though there are strong backward compatibility guarantees during major version bumps."
> 
> I'm not opposed to updating DiagnosticSource ref when we need to, but Is there a good reason to always ensure we're on the latest DiagnosticSource even if the latest version does not have any new features we need?

I think this is something very special about System.Diagnostics.DiagnosticSource:
1. Most of the new OpenTelemetry API features would go to the newer version of System.Diagnostics.DiagnosticSource (e.g. if we need hint/advice for metrics) - which means if there is major version bump, OpenTelemetry .NET will almost certainly take that dependency.
2. If there is a minor version bump, almost certainly it is some bug fix or security patch (otherwise the .NET release bar would push back the change), and OpenTelemetry .NET would also want to bump that version since we don't want the user to get an insecure/buggy version by default.

reyang: > For the best Automatic Instrumentation user experience, my vote is for a conservative major version (probably included into the lowest supported .NET version). If we can do that, we will have really good user experience in November 2024 for all supported versions.
> 
> If we allow to bump version here to the latest minor it will increase range of the issue: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/55ec192a11aa3dac58d57f173118957a25f84b74/docs/troubleshooting.md#assembly-version-conflicts

I agree this is a problem and we should solve it. I disagree that we should solve it here by keeping the lowest version of System.Diagnostics.DiagnosticSource.

alanwest: > OpenTelemetry .NET will almost certainly take that dependency.

Absolutely. Though, when we upgrade a dependency it's nice to explain why beyond just a policy that we always just preemptively stay on the latest.

It looks like we did not do this, but for example, when we upgraded to DS v7 we could have communicated better something like "Upgraded the dependency on DiagnosticSource to v7 in order to support the new UpDownCounter instrument".

When/if we update to v8 is there a more interesting statement we can make? I'll admit, I may not be entirely up to date with all the new features included in v8 that are relevant to us.

> If there is a minor version bump, almost certainly it is some bug fix or security patch

Good point. Maybe a reasonable policy is to remain conservative with major bumps, but preemptive with minor bumps because of DiagnosticSource's specialness.

alanwest: > I agree this is a problem and we should solve it. I disagree that we should solve it here by keeping the lowest version of System.Diagnostics.DiagnosticSource.

Also agree. Though just to raise another example where upgrading DS has caused issues: Azure Functions. See https://github.com/Azure/azure-functions-host/issues/8938.

This too is not something I think we should burden ourselves with solving here, but just want to acknowledge the friction.

reyang: > > OpenTelemetry .NET will almost certainly take that dependency.
> 
> Absolutely. Though, when we upgrade a dependency it's nice to explain why beyond just a policy that we always just preemptively stay on the latest.
> 
> It looks like we did not do this, but for example, when we upgraded to DS v7 we could have communicated better something like "Upgraded the dependency on DiagnosticSource to v7 in order to support the new UpDownCounter instrument".

💯 Yep, that's a great callout. Let me tweak the message here.

reyang: Updated the wording.

alanwest (approved):
Approving as I'm not actually all that concerned if we always decide to stay on latest DiagnosticSource, but separately, I would like to better understand what is coming in .NET 8 that necessitates a DS version bump.

> .NET 8 is improving dependency injection support for Meter/Instrument

It's my understanding that this support may actually be shipped in a new package [Microsoft.Extensions.Diagnostics](https://github.com/dotnet/runtime/tree/main/src/libraries/Microsoft.Extensions.Diagnostics).

I guess the more interesting thing that is coming in DS v8 for our purposes is support for [instrumentation scope attributes on `Meter`](https://github.com/dotnet/runtime/blob/a2656afabd9057dd0ccabcecd411589a4afaaafa/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Metrics/Meter.cs#L62). Would be nice to see this for `ActivitySource` too.

Directory.Packages.props
utpilla: Update the version to `7.0.2`?

reyang: > Update the version to `7.0.2`?

At least this should be a separate PR IMHO if we want to do it (I try to scope this PR to what the title suggests - just adding comments).

utpilla (approved):

cijothomas (approved):

reyang:
> I guess the more interesting thing that is coming in DS v8 for our purposes is support for [instrumentation scope attributes on `Meter`](https://github.com/dotnet/runtime/blob/a2656afabd9057dd0ccabcecd411589a4afaaafa/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Metrics/Meter.cs#L62).

Thanks! I agree with what you mentioned here, updated the PR to provider better/accurate examples.

reyang:
> I would like to better understand what is coming in .NET 8 that necessitates a DS version bump.

Here is a subset (scoped to metrics) https://github.com/dotnet/runtime/issues?q=is%3Aopen+label%3Aarea-System.Diagnostics.Metric+milestone%3A"8.0.0".

PR #4573, opened 2023-06-09
bump grpc client version to fix unobserved exception

### LINKED ISSUE #4570: Grpc.Net.Client unobserved task Exception
# Bug Report

Depencency Grpc.Net.Client 2.43.0 has an unobserved task exception bug:
https://github.com/grpc/grpc-dotnet/pull/1660

## Symptom

Task for sending spans has an unobserved exception if the agent/collector is unavailable.

## Reproduce

General reproduction provided in https://github.com/grpc/grpc-dotnet/issues/1656. Nothing is specific to the OTLP exporter; it's anything using Grpc.Net.Client

## Additional Context

This was fixed in 2.45.0+
https://github.com/grpc/grpc-dotnet/releases/tag/v2.45.0-pre1
but may consider updating to latest package.

masonyc:
Fixes #4570

## Changes
Bump grpc net client to 2.53


reyang (approved):

vishweshbankwar (approved):
@masonyc - Could you please also update the changelog for otlp exporter https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/CHANGELOG.md

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/CHANGELOG.md
```diff
@@ -14,6 +14,9 @@
   for instructions to enable exemplars.
   ([#4553](https://github.com/open-telemetry/opentelemetry-dotnet/pull/4553))
 
+* Updated Grpc.Net.Client to v2.53 to fix unobserved exception
+  from failed calls
+
```
reyang, line 19:
```suggestion
  from failed calls.
  ([#4573](https://github.com/open-telemetry/opentelemetry-dotnet/pull/4573))

```

Kielek:
Finally, I have find the time to review impact.

For Auto Instrumentation, it will leads to reduce set of application which can be auto-instrumented without adjustment for dependencies. Similar discussion was done for https://github.com/open-telemetry/opentelemetry-dotnet/pull/4407

Especially, we will have to increase minimal instrumented version Grpc.Client.Net from 2.43.0 to the upgraded one.

As 2.45 contains obvious bug-fix, I see the reason to do this, but I have doubts if it should go up to 2.53.0.

@alanwest, you were mainly checking what we can do for protobuf. I suppose that you have also good insight for this particulat package.

alanwest:
@masonyc thanks for your patience! We spoke about this PR today in our weekly meeting.

Regarding @Kielek's comment:

> As 2.45 contains obvious bug-fix, I see the reason to do this, but I have doubts if it should go up to 2.53.0.

We agreed a more conservative bump to 2.45 is preferable, but before we do that, I wanted to understand the issue a bit more.

I haven't studied this issue closely enough to understand in what circumstance this bug manifests itself with OpenTelemetry .NET. My thought is that if we understand the bug a bit more then maybe we could find a way to fix it by changing how we use gRPC client to get around the bug rather than bumping the package version.

I've briefly reviewed https://github.com/grpc/grpc-dotnet/issues/1656, but the repro it provides does not resemble how we currently invoke gRPC client, so it left us wondering how this bug is manifested in OpenTelemetry .NET.

@masonyc do you have a repro you could share generating this bug using OpenTelemetry .NET?

rwkarg:
It isn’t specific to how the gRPC client is being used. It’s a Task inside the internal client code that needs isn’t explicitly observed when there is an exception making a call (such as a transient network issue). If you’re making a call, then the internal code path that triggers the gRCP client behavior mentioned in the issue will be exercised.

alanwest:
> It’s a Task inside the internal client code

Gotcha. I've had a chance to look more closely and this makes sense now.

@masonyc please change to 2.45 and I think we're good to merge.

Directory.Packages.props
```diff
@@ -7,7 +7,7 @@
   <ItemGroup>
     <PackageVersion Include="Google.Protobuf" Version="[3.19.4,4.0)" />
     <PackageVersion Include="Grpc" Version="[2.44.0,3.0)" />
-    <PackageVersion Include="Grpc.Net.Client" Version="[2.43.0,3.0)" />
+    <PackageVersion Include="Grpc.Net.Client" Version="[2.53.0,3.0)" />
```
reyang, line 10:
```suggestion
    <PackageVersion Include="Grpc.Net.Client" Version="[2.45.0,3.0)" />
```

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/CHANGELOG.md
```diff
@@ -14,6 +14,10 @@
   for instructions to enable exemplars.
   ([#4553](https://github.com/open-telemetry/opentelemetry-dotnet/pull/4553))
 
+* Updated Grpc.Net.Client to v2.53 to fix unobserved exception
```
reyang, line 17:
```suggestion
* Updated Grpc.Net.Client to v2.45 to fix unobserved exception
```

utpilla (approved):

PR #4574, opened 2023-06-09
Fix dotnet-xunit version

reyang:
I haven't looked into why this happened and nobody noticed this previously, just got the warning while doing local build.

image

The problem is that `dotnet-xunit` is a `DotNetCliToolReference` rather than `PackageReference`, so the version specified in https://github.com/open-telemetry/opentelemetry-dotnet/blob/2e57e2daab812e49746a160174cb8afaffbdea79/Directory.Packages.props#L39 is not being used at all.

Also, the existing code has reference to `DotNetXUnitCliVer` while that variable is not even defined anywhere: https://github.com/open-telemetry/opentelemetry-dotnet/blob/2e57e2daab812e49746a160174cb8afaffbdea79/test/OpenTelemetry.Api.Tests/OpenTelemetry.Api.Tests.csproj#L31

### DIFF
--- a/build/Common.props
+++ b/build/Common.props
@@ -28,8 +28,6 @@
     -->
     <MicrosoftCodeCoveragePkgVer>[17.4.1]</MicrosoftCodeCoveragePkgVer>
     <OTelPreviousStableVer>1.5.0</OTelPreviousStableVer>
-    <SerilogPkgVer>[2.8.0,3.0)</SerilogPkgVer>
-    <SystemCollectionsImmutablePkgVer>1.4.0</SystemCollectionsImmutablePkgVer>
   </PropertyGroup>
 
   <ItemGroup>
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -36,7 +36,6 @@
   <ItemGroup>
     <PackageVersion Include="BenchmarkDotNet" Version="[0.13.5,0.14)" />
     <PackageVersion Include="CommandLineParser" Version="[2.9.1,3.0)" />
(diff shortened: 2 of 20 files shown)


build/Common.props
```diff
@@ -28,8 +28,6 @@
     -->
     <MicrosoftCodeCoveragePkgVer>[17.4.1]</MicrosoftCodeCoveragePkgVer>
     <OTelPreviousStableVer>1.5.0</OTelPreviousStableVer>
-    <SerilogPkgVer>[2.8.0,3.0)</SerilogPkgVer>
-    <SystemCollectionsImmutablePkgVer>1.4.0</SystemCollectionsImmutablePkgVer>
```
reyang, line 32: @CodeBlanch these are no longer needed I assume?

utpilla:
Is adding DotNetCliToolReference required? I removed the `<DotNetCliToolReference Include="dotnet-xunit" />` from the test projects and it still works (both on Visual Studio TestExplorer and dotnet test)

reyang:
> Is adding DotNetCliToolReference required? I removed the `<DotNetCliToolReference Include="dotnet-xunit" />` from the test projects and it still works (both on Visual Studio TestExplorer and dotnet test)

Probably not needed on modern tools https://xunit.net/docs/nuget-packages.

<html>
<body>


dotnet-xunit | This package was retired as of 2.4 Beta 2. .NET Core users should use xunit.runner.visualstudio and the dotnet test command line.
-- | --



</body>
</html>

utpilla (approved):

PR #4580, opened 2023-06-12
Adds trimming and AOT related attribute definitions for down-level TFMs

vitek-karas:
Related to https://github.com/open-telemetry/opentelemetry-dotnet/issues/3429

## Changes

Adds the definition of `RequiresUnreferencedCodeAttribute`, `RequiresDynamicCodeAttribute` and `DynamicallyAccessedMembersAttribute` when one is not provided by the framework (so for anything but .NET6+).

There's a complexity associated with this change. Normally the attributes are added as `internal` to any build target which doesn't have them in the framework. For example, if an assembly is built for `netstandard2.0` an `internal` definition of the attributes would be added to the compilation. All the tools consuming the attributes only look at the `Namespace.TypeName` of the attribute and they ignore which assembly the attribute came from.

OpenTelemetry builds some assemblies for `net6.0` while other assemblies are just `netstandard2.0`. For example, `OpenTelemetry` targets `net6.0`, but its dependency `OpenTelemetry.Api` only targets `netstandard2.0`. That would normally not be a problem since the added attribute definitions in `OpenTelemetry.Api` are `internal`. Unfortunately, `OpenTelemetry` has `InternalsVisibleTo` into `OpenTelemetry.Api`. When building the `net6.0` target of `OpenTelemetry` the compiler sees the attributes defined in the framework as well as the ones in `OpenTelemetry.Api`.

This situation has already been discussed in https://github.com/open-telemetry/opentelemetry-dotnet/pull/4460#discussion_r1195868228.

The solution implemented in this change is to add `net6.0` as a target for `OpenTelemetry.Api`. That this does is that when building `OpenTelemetry` for `net6.0` it builds against `OpenTelemetry.Api` for `net6.0` as well and neither assembly will have the attributes defined internally, they will all come from the framework.

This internal attribute definitions are in `OpenTelemetry.Api` project and are compiled into other assemblies as necessary. Part of this change moves the `UnconditionalSuppressedMessageAttribute` (which is used for trimming/AOT suppressions) to `OpenTelemetry.Api` as well to keep all of those attributes together.

It also adds these attributes to the compilation whenever the `PropertyFetcher.cs` is added. This is currently not needed in this PR, but it will be needed in the future, and it was cleanest to do it here. It also allows for testing that the ifdefs are correct across the projects.

The attribute definitions are copied from `dotnet/runtime` repo and I modified them to fit the code style in this repo - so I changed the file headers and some other small formatting changes. They are also defined as `internal` instead of `public`.


src/OpenTelemetry.Api/OpenTelemetry.Api.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <!-- OmniSharp/VS Code requires TargetFrameworks to be in descending order for IntelliSense and analysis. -->
```
eerhardt, line 3: What exactly does "descending order" mean here? Is this still in "descending order"?

src/OpenTelemetry.Api/Internal/Shims/DynamicallyAccessedMembersAttribute.cs
```diff
@@ -0,0 +1,68 @@
...
+    /// should only be used on instance methods of types assignable to System.Type (or string, but no methods
+    /// will use it there).
+    /// </remarks>
+    [AttributeUsage(
+        AttributeTargets.Field | AttributeTargets.ReturnValue | AttributeTargets.GenericParameter |
+        AttributeTargets.Parameter | AttributeTargets.Property | AttributeTargets.Method |
+        AttributeTargets.Class | AttributeTargets.Interface | AttributeTargets.Struct,
+        Inherited = false)]
+    internal
+    sealed class DynamicallyAccessedMembersAttribute : Attribute
```
eerhardt, line 49:
```suggestion
    internal sealed class DynamicallyAccessedMembersAttribute : Attribute
```

src/OpenTelemetry.Api/Internal/Shims/DynamicallyAccessedMemberTypes.cs
```diff
@@ -0,0 +1,112 @@
...
+{
+    /// <summary>
+    /// Specifies the types of members that are dynamically accessed.
+    ///
+    /// This enumeration has a <see cref="FlagsAttribute"/> attribute that allows a
+    /// bitwise combination of its member values.
+    /// </summary>
+    [Flags]
+    internal
+    enum DynamicallyAccessedMemberTypes
```
eerhardt, line 29:
```suggestion
    internal enum DynamicallyAccessedMemberTypes
```

build/Common.props
```diff
@@ -65,4 +65,12 @@
     <Compile Include="$(RepoRoot)\src\OpenTelemetry.Api\Internal\StatusHelper.cs" Link="Includes\StatusHelper.cs" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(IncludeCodeAnalysisAttributes)'=='true' Or '$(IncludeInstrumentationHelpers)'=='true' Or '$(IncludeDiagnosticSourceInstrumentationHelpers)'=='true'">
```
eerhardt, line 68: Who sets `IncludeCodeAnalysisAttributes`? I'm not seeing it used.

eerhardt, line 68: Is it weird to include `Or '$(IncludeInstrumentationHelpers)'=='true' Or '$(IncludeDiagnosticSourceInstrumentationHelpers)'=='true'` here? Why are they needed?

vitek-karas: Currently no one. So maybe I should remove it - it was meant to be used if a project wants to include the attributes. But I guess this is typically done one by one here. What do you think?

vitek-karas: Currently this is not needed, but once I annotate PropertyFetcher this will be necessary because both of these properties will include PropertyFetcher.

Alternatively I could include these files in the same group as PropertyFetcher.

src/OpenTelemetry.Api/OpenTelemetry.Api.csproj
vitek-karas: Thanks for pointing this out. This seems to originate here: https://github.com/open-telemetry/opentelemetry-dotnet/pull/2015#discussion_r622391733

And that in turn says that VS Code only uses the first TFM in the list. So this would change that selection to .NET 6.
Personally I would argue that's a good thing. And it's also consistent with `OpenTelemetry.csproj` which also starts with `net6.0`.

build/Common.props
eerhardt: I kind of like the model of just including what you need. But others may have different opinions.

If no project is using the setting, I would drop it until someone uses it.

eerhardt:
This approach looks good to me. But it looks like the build is broken:

https://github.com/open-telemetry/opentelemetry-dotnet/actions/runs/5250186245/jobs/9483858384

```
Build FAILED.

D:\a\opentelemetry-dotnet\opentelemetry-dotnet\src\OpenTelemetry.Api\Logs\IDeferredLoggerProviderBuilder.cs(26,18): error RS0016: Symbol 'IDeferredLoggerProviderBuilder' is not part of the declared API [D:\a\opentelemetry-dotnet\opentelemetry-dotnet\src\OpenTelemetry.Api\OpenTelemetry.Api.csproj::TargetFramework=net6.0]
D:\a\opentelemetry-dotnet\opentelemetry-dotnet\src\OpenTelemetry.Api\Logs\Logger.cs(24,23): error RS0016: Symbol 'Logger' is not part of the declared API [D:\a\opentelemetry-dotnet\opentelemetry-dotnet\src\OpenTelemetry.Api\OpenTelemetry.Api.csproj::TargetFramework=net6.0]
D:\a\opentelemetry-dotnet\opentelemetry-dotnet\src\OpenTelemetry.Api\Logs\LoggerProvider.cs(28,14): error RS0016: Symbol 'LoggerProvider' is not part of the declared API [D:\a\opentelemetry-dotnet\opentelemetry-dotnet\src\OpenTelemetry.Api\OpenTelemetry.Api.csproj::TargetFramework=net6.0]
D:\a\opentelemetry-dotnet\opentelemetry-dotnet\src\OpenTelemetry.Api\Logs\LoggerProvider.cs(28,14): error RS0016: Symbol 'implicit constructor for 'LoggerProvider'' is not part of the declared API [D:\a\opentelemetry-dotnet\opentelemetry-dotnet\src\OpenTelemetry.Api\OpenTelemetry.Api.csproj::TargetFramework=net6.0]
D:\a\opentelemetry-dotnet\opentelemetry-dotnet\src\OpenTelemetry.Api\Logs\LoggerProviderBuilder.cs(24,23): error RS0016: Symbol 'LoggerProviderBuilder' is not part of the declared API [D:\a\opentelemetry-dotnet\opentelemetry-dotnet\src\OpenTelemetry.Api\OpenTelemetry.Api.csproj::TargetFramework=net6.0]
D:\a\opentelemetry-dotnet\opentelemetry-dotnet\src\OpenTelemetry.Api\Logs\LogRecordAttributeList.cs(30,15): error RS0016: Symbol 'LogRecordAttributeList' is not part of the declared API [D:\a\opentelemetry-dotnet\opentelemetry-dotnet\src\OpenTelemetry.Api\OpenTelemetry.Api.csproj::TargetFramework=net6.0]
D:\a\opentelemetry-dotnet\opentelemetry-dotnet\src\OpenTelemetry.Api\Logs\LogRecordAttributeList.cs(30,15): error RS0016: Symbol 'implicit constructor for 'LogRecordAttributeList'' is not part of the declared API [D:\a\opentelemetry-dotnet\opentelemetry-dotnet\src\OpenTelemetry.Api\OpenTelemetry.Api.csproj::TargetFramework=net6.0]
D:\a\opentelemetry-dotnet\opentelemetry-dotnet\src\OpenTelemetry.Api\Logs\LogRecordAttributeList.cs(299,19): error RS0016: Symbol 'Enumerator' is not part of the declared API [D:\a\opentelemetry-dotnet\opentelemetry-dotnet\src\OpenTelemetry.Api\OpenTelemetry.Api.csproj::TargetFramework=net6.0]
D:\a\opentelemetry-dotnet\opentelemetry-dotnet\src\OpenTelemetry.Api\Logs\LogRecordAttributeList.cs(299,19): error RS0016: Symbol 'implicit constructor for 'Enumerator'' is not part of the declared API [D:\a\opentelemetry-dotnet\opentelemetry-dotnet\src\OpenTelemetry.Api\OpenTelemetry.Api.csproj::TargetFramework=net6.0]
```

eerhardt:
> Throwing out potential alternatives for discussion.
use #if preprocessors for lower .NET frameworks with AOT attributes as internal rather than public.

Can you prototype what this change would look like? Are you suggesting to just use `#if NETCOREAPP` around the application of the attributes in the source code?

If that is what you are proposing, you will still need to target `net6.0` for OpenTelemetry.Api, assuming that assembly needs to apply these attributes to its code. It only targets `netstandard2.0;net461` today, which means the `#if NETCOREAPP` will never be true and the attributes will never be applied in that assembly.

vitek-karas:
As of right now, only the `OpenTelemetry.Api.ProviderBuilderExtensions` needs the attributes, `OpenTelemetry.Api` itself doesn't need them. So as @eerhardt points out we would have to target `OpenTelemetry.Api.ProviderBuilderExtensions` to `net6.0`. And any other assembly which needs the attributes (though I think all the other ones are already targeting `net6.0`)

Yun-Ting:
@vitek-karas and @eerhardt - 
Yes to we will need to add net6.0 as the target to the projects with the preprocessors.
Basically this https://github.com/open-telemetry/opentelemetry-dotnet/pull/4370/commits change (instead of net7.0, use net6.0) + `#if NETCOREAPP`.

vitek-karas:
I modified this to reduce the impact of the change.

I kept the attribute definition files in `OpenTelemetry.Api` directory, since there's no "common" and it makes sense to me to have them in the lowest level assembly, as they're going to be used by other assemblies all over the place.
But `OpenTelemetry.Api` doesn't compile them, and so I removed retargeting it to `net6.0` which felt like the most contentious issue.

Instead I retargeted `OpenTelemetry.Api.ProviderBuilderExtensions` as that one makes use of the new attributes. I added the attributes to this assembly as necessary to make it trim compatible (note that it doesn't reduce the number of warnings in the AOT test because the AOT test uses 7.0 AOT compiler which had some bugs in it).
For consideration - also add a pure trimming test (`PublishTrimmed=true`), to make sure that both AOT and trimming work correctly.

src/OpenTelemetry.Api/OpenTelemetry.Api.csproj
```diff
@@ -15,4 +15,8 @@
   <ItemGroup>
     <PackageReference Include="System.Diagnostics.DiagnosticSource" />
   </ItemGroup>
+
+  <ItemGroup>
+    <Compile Remove="Internal\Shims\*.*" />
```
eerhardt, line 20: Would it make more sense to put these in https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/src/Shared instead? I think it is a bit confusing to put them in this folder, but not actually used by this project.

eerhardt:
> For consideration - also add a pure trimming test (PublishTrimmed=true), to make sure that both AOT and trimming work correctly.

Can you elaborate more on this? I know that AOT and trimming run different tools. But from a .NET library's perspective, shouldn't AOT warnings be a superset of trimming warnings? (assuming the underlying tools have no bugs).

src/OpenTelemetry.Api.ProviderBuilderExtensions/OpenTelemetry.Api.ProviderBuilderExtensions.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <!-- OmniSharp/VS Code requires TargetFrameworks to be in descending order for IntelliSense and analysis. -->
-    <TargetFrameworks>netstandard2.0;net462</TargetFrameworks>
+    <TargetFrameworks>net6.0;netstandard2.0;net462</TargetFrameworks>
```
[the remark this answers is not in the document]
vitek-karas: Yes - it has IVT to `OpenTelemetry`, so basically the same problem as `OpenTelemetry.Api`.
https://github.com/open-telemetry/opentelemetry-dotnet/blob/5f825cb8fcadd3f5689467e314bb25fa9f1f8419/src/OpenTelemetry.Api.ProviderBuilderExtensions/AssemblyInfo.cs#L19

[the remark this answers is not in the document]
vitek-karas: It's also in-line with the result of a discussion around retargeting to .NET 6. The conclusion was that retargeting projects which actually make use of the new attributes is a good solution.

PR #4587, opened 2023-06-12
Retry policy for OTLP/gRPC exporter

alanwest:
Implements the retry behavior [described by the specification](https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#failures) for the OTLP/gRPC exporter.

### DIFF
--- a/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/CHANGELOG.md
+++ b/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/CHANGELOG.md
@@ -16,7 +16,7 @@
   [specification](https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#failures).
   Retry support is not available in the `netstandard2.0` or `net462`
   targets of the OTLP exporter.
-  ([#TBD](https://github.com/open-telemetry/opentelemetry-dotnet/pull/TBD))
+  ([#4587](https://github.com/open-telemetry/opentelemetry-dotnet/pull/4587))
 
 * Add back support for Exemplars. See [exemplars](../../docs/metrics/customizing-the-sdk/README.md#exemplars)
   for instructions to enable exemplars.
--- a/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/CHANGELOG.md
+++ b/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/CHANGELOG.md
@@ -10,6 +10,14 @@
   are now included in this package.
   ([#4556](https://github.com/open-telemetry/opentelemetry-dotnet/pull/4556))
 
+* Added support for the OTLP gRPC exporter to retry failed requests due to
+  transient network faults. The status codes that are
+  retried are defined in the
+  [specification](https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#failures).
+  Retry support is not available in the `netstandard2.0` or `net462`
+  targets of the OTLP exporter.
+  ([#TBD](https://github.com/open-telemetry/opentelemetry-dotnet/pull/TBD))
+
 * Add back support for Exemplars. See [exemplars](../../docs/metrics/customizing-the-sdk/README.md#exemplars)
   for instructions to enable exemplars.
   ([#4553](https://github.com/open-telemetry/opentelemetry-dotnet/pull/4553))
--- a/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpExporterOptionsExtensions.cs
+++ b/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpExporterOptionsExtensions.cs
@@ -22,6 +22,7 @@ using Grpc.Core;
 using OpenTelemetry.Exporter.OpenTelemetryProtocol.Implementation.ExportClient;
 #if NETSTANDARD2_1 || NET6_0_OR_GREATER
 using Grpc.Net.Client;
+using Grpc.Net.Client.Configuration;
 #endif
 using LogOtlpCollector = OpenTelemetry.Proto.Collector.Logs.V1;
 using MetricsOtlpCollector = OpenTelemetry.Proto.Collector.Metrics.V1;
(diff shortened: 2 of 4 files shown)


src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpExporterOptionsExtensions.cs
```diff
@@ -43,7 +44,41 @@ public static Channel CreateChannel(this OtlpExporterOptions options)
...
+                InitialBackoff = TimeSpan.FromSeconds(1),
+                MaxBackoff = TimeSpan.FromSeconds(5),
+                BackoffMultiplier = 1.5,
+
+                // https://github.com/open-telemetry/opentelemetry-java/commit/6f755cc8128f8077643b4990327c22b293c46e45#diff-2d25b543bc2c5a13e495895b6f6a2db80e1167bbe3387024cb917efda35a99acR24-R34
+                RetryableStatusCodes =
+                {
+                    StatusCode.Cancelled,
+                    StatusCode.DeadlineExceeded,
+                    StatusCode.ResourceExhausted,
```
alanwest, line 62: TODO: add a test validating that the Grpc.Net.Client library honors the `retry_delay` sent from the server.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/CHANGELOG.md
```diff
@@ -10,6 +10,14 @@
   are now included in this package.
   ([#4556](https://github.com/open-telemetry/opentelemetry-dotnet/pull/4556))
 
+* Added support for the OTLP gRPC exporter to retry failed requests due to
```
utpilla, line 13: This doesn't cover `PartialSuccess`, right? We wouldn't be retrying any of the rejected items from a partially successful response.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpExporterOptionsExtensions.cs
utpilla: The [server ](https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#failures:~:text=To%20signal%20backpressure%20when%20using%20gRPC%20transport%2C%20the%20server%20MUST%20return%20an%20error%20with%20code%20Unavailable%20and%20MAY%20supply%20additional%20details%20via%20status%20using%20RetryInfo.)could be sending a `retry_delay` even with `StatusCode.Unavailable`. Client should be respecting that as well?

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpExporterOptionsExtensions.cs
```diff
@@ -43,7 +44,41 @@ public static Channel CreateChannel(this OtlpExporterOptions options)
             }
 
 #if NETSTANDARD2_1 || NET6_0_OR_GREATER
-            return GrpcChannel.ForAddress(options.Endpoint);
+            // The specification does not dictate the parameters of the default retry policy.
+            // These defaults are borrowed from opentelemetry-java.
+            var retryPolicy = new RetryPolicy
```
vishweshbankwar, line 49: how does this impact the default [deadline](https://github.com/open-telemetry/opentelemetry-dotnet/blob/b3ffb46e3220f3751defec3ed6bfa973292e9bfb/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/OtlpGrpcMetricsExportClient.cs#L44) we set?

Is it possible that while retrying the deadline is exceeded and the retry does not complete or not performed(if retryAfter is more than default deadline)

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/CHANGELOG.md
alanwest: No, the spec does not support retrying rejected items from a partially successful response. However, you bring up a good point, the spec does say

> The client MUST NOT retry the request when it receives a partial success response where the partial_success is populated.

I need to fix this.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpExporterOptionsExtensions.cs
alanwest: @JamesNK I'm curious if you have a moment to share guidance here...

This PR implements retry for our gRPC exporter using a `RetryPolicy`. There are a number of status codes which can be retried, however, we have a [requirement](https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#failures) that cannot be expressed using a `RetryPolicy` alone when status is `Unavailable` or `Resource_Exhausted`:

> The server MUST indicate retryable errors using code [Unavailable](https://godoc.org/google.golang.org/grpc/codes) and MAY supply additional [details via status](https://godoc.org/google.golang.org/grpc/status#Status.WithDetails) using [RetryInfo](https://github.com/googleapis/googleapis/blob/6a8c7914d1b79bd832b5157a09a9332e8cbd16d4/google/rpc/error_details.proto#L40) containing 0 value of RetryDelay.

My hope was to avoid rolling our own retry mechanism, but from what I can tell Grpc.Net.Client does not support handling [RetryInfo](https://github.com/googleapis/googleapis/blob/6a8c7914d1b79bd832b5157a09a9332e8cbd16d4/google/rpc/error_details.proto#L40).

Options I can think of
1. Roll our own retry mechanism for all status codes conforming to our requirements.
2. Use a RetryPolicy for retryable codes which do require handling RetryInfo and only roll our own logic for Unavailable and Resource_Exhaused.

Are there benefits to one option over the other? Are there better options?

alanwest: > Do other gRPC client libraries support RetryInfo?

No, not from what I've seen. This RetryInfo seems to be something specific to some Google APIs. It's unfortunate that the OTel spec adopted this over the standard gRPC retry spec.

OTLP supports both gRPC and HTTP and since HTTP has a [`Retry-After`](https://www.rfc-editor.org/rfc/rfc9110.html#section-10.2.3) response header I suspect the decision to adopt this RetryInfo thing for gRPC was inspired by the desire to keep parity with the retry behavior between both HTTP and gRPC.

JamesNK: gRPC retry supports the [`grpc-retry-pushback-ms`](https://github.com/grpc/proposal/blob/master/A6-client-retries.md#pushback) response header which seems similar.

Grpc.Net.Client retries doesn't support RetryInfo. I suggest you look at other implementations to see what they do. If they have a clever technique then maybe that is something that could also be done in .NET. If they are all doing retries manually then you'll have to do it manually.

alanwest:
It seems that using the built in retry mechanism of Grpc.Net.Client may not work for our purposes since the OTel spec has some unique non-standard requirements. I've got another solution in the works that is spec compliant. Closing this PR for now and will open another soon.

PR #4592, opened 2023-06-14
Move shared utility files outside of Sdk project

utpilla:
Related to #3429 

## Changes
- Move the following files which are not used by the Sdk outside of the project to a shared folder:
  - `DiagnosticSourceListener.cs`
  - `DiagnosticSourceSubscriber.cs`
  - `InstrumentationEventSource.cs`
  - `ListenerHandler.cs`
  - `PropertyFetcher.cs`
  - `ActivityInstrumentationHelper.cs`
  
  
There is scope to move more such files from the API and SDK project. That could be explored further after this PR.

@Yun-Ting This reduces three more warnings in the AOT test because `PropertyFetcher` is being moved outside of Sdk. However, this PR is merely transferring the issue from Sdk to other components that use `PropertyFetcher`.


OpenTelemetry.sln
```diff
@@ -268,6 +268,20 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "OpenTelemetry.AotCompatibil
...
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "links-sampler", "docs\trace\links-based-sampler\links-sampler.csproj", "{62AF4BD3-DCAE-4D44-AA5B-991C1071166B}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Shared", "Shared", "{A49299FB-C5CD-4E0E-B7E1-B7867BBD67CC}"
+	ProjectSection(SolutionItems) = preProject
+		src\Shared\ActivityInstrumentationHelper.cs = src\Shared\ActivityInstrumentationHelper.cs
+	EndProjectSection
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "DiagnosticSourceInstrumentation", "DiagnosticSourceInstrumentation", "{28F3EC79-660C-4659-8B73-F90DC1173316}"
+	ProjectSection(SolutionItems) = preProject
+		src\Shared\DiagnosticSourceInstrumentation\DiagnosticSourceListener.cs = src\Shared\DiagnosticSourceInstrumentation\DiagnosticSourceListener.cs
```
reyang, line 278: How are these files organized under the `Shared` folder? (e.g. do we want the relative path and the namespace to align)

e.g.

* Shared\Instrumentation\ListenerHandler.cs (since the namespace is OpenTelemetry.Instrumentation)

reyang (approved):
LGTM.

OpenTelemetry.sln
utpilla: > do we want the relative path and the namespace to align

Good point! That could be our eventual target. Changing the namespace, is a more intrusive change and I intend to keep the PR changes small for simplicity.

We could probably work on aligning the namespace once we move all the required files outside of API and SDK?

reyang: Sounds good to me (not a blocking issue).

Yun-Ting (approved):

vishweshbankwar (approved):

PR #4604, opened 2023-06-19
Remove zpages

reyang:
Remove the ZPages component:
* The total download number is very low (9.4k in the past 3+ years) https://www.nuget.org/packages/OpenTelemetry.Exporter.ZPages.
* There hasn't been any real change for years https://github.com/open-telemetry/opentelemetry-dotnet/commits/main/src/OpenTelemetry.Exporter.ZPages.
* There is no plan to make this official from the specification perspective.


CodeBlanch (approved):
LGTM

Kielek (approved):
AutoInstrumentation is not using this package.

Please remember to deprecate it also on nuget.org side (worth to make a bit bigger cleanup https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/1100).

vishweshbankwar (approved):

cijothomas:
> There is no plan to make this official from the specification perspective.

https://github.com/open-telemetry/opentelemetry-specification/blob/main/experimental/trace/zpages.md
Would this doc be updated to indicate so?

src/OpenTelemetry.Exporter.ZPages/README.md
```diff
@@ -1,26 +0,0 @@
-# ZPages Exporter for OpenTelemetry .NET
```
cijothomas, line 1: maybe keep this page for sometime, with a description saying this is decommissioned. Or we think a nuget.org deprecation is fine?

cijothomas (approved):
Agree with this.
Maintainers may either keep a readme.md page with this info or update the nuget package to indicate its deprecaed.

src/OpenTelemetry.Exporter.ZPages/README.md
reyang: Both the package and the readme file have very low number of visitors so I don't feel there is a need to keep the readme.

alanwest (approved):

PR #4605, opened 2023-06-20
[AOT] Make PropertyFetcher AOT compatible.

Yun-Ting:
Fixes #
Design discussion issue #

## Changes

### DIFF
--- a/src/Shared/DiagnosticSourceInstrumentation/PropertyFetcher.cs
+++ b/src/Shared/DiagnosticSourceInstrumentation/PropertyFetcher.cs
@@ -14,8 +14,12 @@
 // limitations under the License.
 // </copyright>
 
+#if NET6_0_OR_GREATER
+using System.Diagnostics.CodeAnalysis;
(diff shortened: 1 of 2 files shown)


src/Shared/DiagnosticSourceInstrumentation/PropertyFetcher.cs
```diff
@@ -85,6 +78,9 @@ public bool TryFetch(object obj, out T value, bool skipObjNullCheck = false)
         }
 
         // see https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/DiagnosticSourceEventSource.cs
+#if NET6_0_OR_GREATER
```
vitek-karas, line 81: As discussed in other places, I think it would be cleaner to use the attribute always, regardless of TFM, but it requires solving the other problems with adding the internal definition of the attributes. I just had to put it here :-)

PR #4616, opened 2023-06-26
Implement OTLP retry mechanism

alanwest:
#4587 implemented a retry mechanism using Grpc.Net.Client's built in retry mechanism. However, the requirements of the OTLP specification differ from the standard gRPC retry mechanism.

This PR contains a spec-compliant OTLP retry mechanism. It can be used by both the gRPC and HTTP exporter.

This PR does not yet wire up the exporters to use this retry mechanism. This will be in a follow up.

**Example usage**

```csharp
public ExportResult Export(...)
{
    var nextRetryDelayMilliseconds = OtlpRetry.InitialBackoffMilliseconds;

    while (true)
    {
        try
        {
            // Do the export
        }
        catch (RpcException ex)
        {
            if (OtlpRetry.TryGetGrpcRetryResult(ex.StatusCode, deadline, ex.Trailers, nextRetryDelayMilliseconds, out var retryResult))
            {
                Thread.Sleep(retryResult.RetryDelay);
                nextRetryDelayMilliseconds = retryResult.NextRetryDelayMilliseconds;
            }
            else
            {
                throw;
            }
        }
    }
}
```

Follow up PRs:

* Expand tests to cover HTTP
* Integrate retry with gRPC exporters
* Integrate retry with HTTP exporters


src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/OtlpRetry.cs
```diff
@@ -0,0 +1,248 @@
...
+        // Without a max number of retry attempts, retries would continue until the deadline.
+        // Maybe this is ok? However, it may lead to an unexpected behavioral change. For example:
+        //    1) When using a batch processor, a longer delay due to repeated
+        //       retries up to the deadline may lead to a higher chance that the queue will be exhausted.
+        //    2) When using the simple processor, a longer delay due to repeated
+        //       retries up to the deadline will lead to a prolonged blocking call.
+        // if (attemptCount >= MaxAttempts)
+        // {
+        //     return false
+        // }
```
alanwest, line 93: I'm inclined to set a maximum cap of 5 retries, but I'd like to get other people's thoughts.

vishweshbankwar: 5 sounds good. 
`1)`, users can adjust the queue length if they frequently see dropped items. for `2)` there is no mitigation given the retries are not configurable, correct?

vishweshbankwar:
Having this on by default for Http exporter may impact certain users who have a custom HttpClient set up with their own retry mechanism. I have not seen this but something we might run into this so bringing it up.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/OtlpRetry.cs
```diff
@@ -0,0 +1,248 @@
...
+#else
+        return statusCode == (HttpStatusCode)429 || statusCode == HttpStatusCode.ServiceUnavailable
+#endif
+            ? headers.RetryAfter?.Delta
+            : null;
+    }
+
+    private static bool IsGrpcStatusCodeRetryable(StatusCode statusCode, bool hasRetryDelay)
+    {
+        switch (statusCode)
```
vishweshbankwar, line 187: @alanwest Is it possible to do this custom retry for `Unavailable` and `ResourceExhausted` only? And rest with https://learn.microsoft.com/en-us/aspnet/core/grpc/retries?view=aspnetcore-7.0#configure-a-grpc-retry-policy

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/OtlpRetry.cs
alanwest: > 5 sounds good.

Yes, I agree. I plan to leave this as a follow up for now. Put it in place when I actually wire this up with the exporter.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/OtlpRetry.cs
alanwest: Probably not, especially if we introduce a max number of retries. There's no way to know how many attempts the client library attempted vs. special handling for unavailable/resource exhausted.

Though, even if it were possible, I'd be hard pressed to say the complexity would be buying us anything. I think it's pretty nice that gRPC and HTTP uses the exact same retry logic.

alanwest:
> Having this on by default for Http exporter may impact certain users who have a custom HttpClient set up with their own retry mechanism. I have not seen this but something we might run into this so bringing it up.

I'd like to think on this a bit more to gauge how concerned we need to be about this. Worst case, we could disable retry if someone is using the HttpClientFactory option. This is one reason I think I'd like to hold on introducing support for the HttpClientFactory for the gRPC exporter. That is, I'd like to land retry first and then consider adding HttpClientFactory support proposed in #4625.

dhhoang:
> I'd like to think on this a bit more to gauge how concerned we need to be about this. Worst case, we could disable retry if someone is using the HttpClientFactory option - though I think this would be a bit of a bummer.
> 
> This is one reason I think I'd like to hold on introducing support for the HttpClientFactory for the gRPC exporter. That is, I'd like to land retry first and then consider adding HttpClientFactory support proposed in #4625.

Would it be better if retry is *not* enabled by default but offered as an option?
* If users upgrade to new versions -> no breaking change. 
* Make explicit documentation to say that this retry mechanism is separate from any custom retry from `HttpClientFactory` so that users are aware when they opt in.

reyang:
> **Example usage**
> 
> ```cs
> public ExportResult Export(...)
> {
>     var nextRetryDelayMilliseconds = OtlpRetry.InitialBackoffMilliseconds;
> 
>     while (true)
>     {
>         try
>         {
>             // Do the export
>         }
>         catch (RpcException ex)
>         {
>             if (OtlpRetry.TryGetGrpcRetryResult(ex.StatusCode, context.Options.Deadline, ex.Trailers, nextRetryDelayMilliseconds, out var retryResult))
>             {
>                 Thread.Sleep(retryResult.RetryDelay);
>                 nextRetryDelayMilliseconds = retryResult.NextRetryDelayMilliseconds;
>             }
>             else
>             {
>                 throw;
>             }
>         }
>     }
> }
> ```

I want to understand the direction here, the above code snippet has two parts that are concerning:
1. `Thread.Sleep` based on an external input without policy, this would block the exporting thread indefinitely.
2. What's the overall strategy - if something failed to export, do we prioritize to exporter the latest data or we'll keep retrying the old data (while the new data might fill up the buffer and get dropped).

reyang:
> * If users upgrade to new versions -> no breaking change.

What's breaking?

vishweshbankwar:
> > Having this on by default for Http exporter may impact certain users who have a custom HttpClient set up with their own retry mechanism. I have not seen this but something we might run into this so bringing it up.
> 
> I'd like to think on this a bit more to gauge how concerned we need to be about this. Worst case, we could disable retry if someone is using the HttpClientFactory option - though I think this would be a bit of a bummer.
> 
> This is one reason I think I'd like to hold on introducing support for the HttpClientFactory for the gRPC exporter. That is, I'd like to land retry first and then consider adding HttpClientFactory support proposed in #4625.

Looked into this bit more. Just sharing some points for consideration:

I think we should consider this as an opt-in feature **or** on by default with ability to turn off. More for http protobuf as we allow custom HttpClient.

a) Retrying with this approach has some overhead involved for recreating request messages for each retry during failures. 
b) There can be other scenarios such as using Circuit breaker which users might want to implement?
c) Could there be other [response codes](https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#failures:~:text=All%20other%20HTTP%20responses%20that%20are%20not%20explicitly%20listed%20in%20this%20document%20should%20be%20treated%20according%20to%20HTTP%20specifications.) based on the server's implementation?.

alanwest:
@reyang 

> Thread.Sleep based on an external input without policy, this would block the exporting thread indefinitely.

The call to Export is a blocking call. Task.Delay could be used, but probably to the same effect. I studied the retry mechanism for Grpc.Net.Client extensively to guide this implementation. It uses Task.Delay as the delay is make from an asynchronous method https://github.com/grpc/grpc-dotnet/blob/16e01b9369a1c2b1440bbe7f77c1520b11dc6db3/src/Grpc.Net.Client/Internal/Retry/RetryCall.cs#L219.

> What's the overall strategy - if something failed to export, do we prioritize to exporter the latest data or we'll keep retrying the old data (while the new data might fill up the buffer and get dropped).

The strategy is guided by the spec. I'm not looking to make anything up here, so I think our goal should be that we mutually understand the specification. Here's my interpretation with regards to your question:

If export fails with a retryable status code, then the OTLP exporter should retry the same data. Per the spec, retries attempts will continue up until the configured timeout which defaults to 10 seconds. If using the batch export processor this could mean the buffer fills up and new data is dropped.

That said, I am proposing one additional requirement not  in the spec and that is a cap of 5 retry attempts - this is not currently implemented in this PR.

reyang:
> The strategy is guided by the spec. I'm not looking to make anything up here, so I think our goal should be that we mutually understand the specification. Here's my interpretation with regards to your question:
> 
> If export fails with a retryable status code, then the OTLP exporter should retry the same data. Per the spec, retries attempts will continue up until the configured timeout which defaults to 10 seconds. If using the batch export processor this could mean the buffer fills up and new data is dropped.

Where are these coming from? Is this the one we're trying to follow? https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#retry

alanwest:
https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#failures

reyang:
> https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#failures

I don't see how this section in the spec is related to the code snippet from https://github.com/open-telemetry/opentelemetry-dotnet/pull/4616#issuecomment-1614087957.

I also couldn't find where "per the spec, retries attempts will continue up until the configured timeout which defaults to 10 seconds. If using the batch export processor this could mean the buffer fills up and new data is dropped" is coming from.

vishweshbankwar (approved):
LGTM considering this will be enabled via feature flag allowing us to clarify things with spec and gather feedback.

Feature flag can be added as a separate PR while wiring things up.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/google/rpc/status.proto
```diff
@@ -0,0 +1,49 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package google.rpc;
+
+import "google/protobuf/any.proto";
+
+option cc_enable_arenas = true;
```
utpilla, line 21: Could we remove these option statements for other languages?

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/google/rpc/error_details.proto
```diff
@@ -0,0 +1,285 @@
...
+  // it. For example, it can be used to identify requests in the service's logs.
+  string request_id = 1;
+
+  // Any data that was used to serve this request. For example, an encrypted
+  // stack trace that can be sent back to the service provider for debugging.
+  string serving_data = 2;
+}
+
+// Describes the resource that is being accessed.
+message ResourceInfo {
```
utpilla, line 233: Could we remove the unused `message` definitions from this file?

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/OtlpRetry.cs
```diff
@@ -0,0 +1,248 @@
...
+        }
+    }
+
+    private static int GetRandomNumber(int min, int max)
+    {
+#if NET6_0_OR_GREATER
+        return Random.Shared.Next(min, max);
+#else
+        // TODO: Implement this better to minimize lock contention.
+        // Consider pulling in Random.Shared implementation.
```
utpilla, line 227: I think we could simply use a Random instance per thread.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/OtlpRetry.cs
```diff
@@ -0,0 +1,248 @@
...
+        lock (Random)
+        {
+            return Random.Next(min, max);
+        }
+#endif
+    }
+
+    public struct RetryResult
```
utpilla, line 235: There might be some benefit to making this a `readonly` struct?

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/OtlpRetry.cs
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    private static bool IsHttpStatusCodeRetryable(HttpStatusCode statusCode, bool hasRetryDelay)
+    {
+        switch (statusCode)
+        {
+#if NETSTANDARD2_1_OR_GREATER || NET6_0_OR_GREATER
+            case HttpStatusCode.TooManyRequests:
+#else
+            case (HttpStatusCode)429:
```
utpilla, line 210: Is this a valid case for `net462` or `netstandard2.0`?

This status code in not defined in `HttpStatusCode` enum for the above targets:
- https://learn.microsoft.com/en-us/dotnet/api/system.net.httpstatuscode?view=netframework-4.6.2
- https://learn.microsoft.com/en-us/dotnet/api/system.net.httpstatuscode?view=netstandard-2.0

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/OtlpRetry.cs
```diff
@@ -0,0 +1,248 @@
...
+        if (deadline.HasValue && IsDeadlineExceeded(deadline + delayDuration))
+        {
+            return false;
+        }
+
+        if (throttleDelay.HasValue)
+        {
+            try
+            {
+                nextRetryDelayMilliseconds = Convert.ToInt32(throttleDelay.Value.TotalMilliseconds);
```
utpilla, line 120: Could we just use `double` as the type for `InitialBackoffMilliseconds` and `nextRetryDelayMilliseconds` to keep things simple?

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/OtlpRetry.cs
```diff
@@ -0,0 +1,248 @@
...
+    private static int GetRandomNumber(int min, int max)
+    {
+#if NET6_0_OR_GREATER
+        return Random.Shared.Next(min, max);
+#else
+        // TODO: Implement this better to minimize lock contention.
+        // Consider pulling in Random.Shared implementation.
+        lock (Random)
+        {
+            return Random.Next(min, max);
```
utpilla, line 230: Do we want to use `max + 1`? The second argument for `Random.Next` is the exclusive upper bound of the random number returned.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/OtlpRetry.cs
```diff
@@ -0,0 +1,248 @@
...
+            case StatusCode.DataLoss:
+                return true;
+            case StatusCode.ResourceExhausted:
+                return hasRetryDelay;
+            default:
+                return false;
+        }
+    }
+
+    private static bool IsHttpStatusCodeRetryable(HttpStatusCode statusCode, bool hasRetryDelay)
```
utpilla, line 203: nit: 
```suggestion
    private static bool IsHttpStatusCodeRetryable(HttpStatusCode statusCode, bool _)
```

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/OtlpRetry.cs
```diff
@@ -0,0 +1,248 @@
...
+    private static TimeSpan? TryGetGrpcRetryDelay(StatusCode statusCode, Metadata trailers)
+    {
+        Debug.Assert(trailers != null, "trailers was null");
+
+        if (statusCode != StatusCode.ResourceExhausted && statusCode != StatusCode.Unavailable)
+        {
+            return null;
+        }
+
+        var statusDetails = trailers.Get(GrpcStatusDetailsHeader);
```
utpilla, line 155: Is every Grpc service expected to set this trailer?

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/google/rpc/error_details.proto
alanwest: Sure, often when I forklift stuff I tend to keep it all in tact for ease of maintenance, but likely we'll never need to touch these files again.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/OtlpRetry.cs
alanwest: Agreed, I was just punting in hopes of working towards a full solution and then coming back and cleaning up TODOs.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/OtlpRetry.cs
alanwest: Yes, a 429 HTTP status code is valid irrespective of version of .NET. It's just that the HttpStatusCode enum did not include an explicit value for it in older versions.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/OtlpRetry.cs
alanwest: I'd considered this but landed on converting for now thinking it might simplify the consuming logic. Though, I might become convinced using a double is the better way to go. I'm happy to come back and reconsider this later if that works for you. Added a TODO.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/OtlpRetry.cs
alanwest: 🤷 Not sure it matters a whole lot. I'm happy to go either way. Just for some additional context, the behavior of the retry logic in this PR is very much inspired by grpc-dotnet's implementation of the gRPC retry specification including the calculation for random jitter.

Using `Random.Next(min, max)` is equivalent to grpc-dotnet's implementation

https://github.com/grpc/grpc-dotnet/blob/01049834beb8342d1cf93bc5902a7fb2d3a310e1/src/Grpc.Net.Client/Internal/Retry/RetryCall.cs#L216

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/OtlpRetry.cs
alanwest: Metadata.Get returns null if the header does not exist.

https://grpc.github.io/grpc/csharp/api/Grpc.Core.Metadata.html#Grpc_Core_Metadata_Get_System_String_

utpilla: My question was more around the general availability of this header for our use. Is this left to the service implementation? Could we expect OTLP Collector to be setting this header?

alanwest: > Could we expect OTLP Collector to be setting this header?

Yes, per the spec, any OTLP receiver like the one for the collector should support this header for throttling requests. https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#otlpgrpc-throttling

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/OtlpRetry.cs
utpilla: I was actually concerned about the `enum` casting for an undefined value but looks like it would be fine.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/OtlpRetry.cs
alanwest: Ah, but yea I think I see the confusion... the spec does not make it obvious that it's a header by the name `grpc-status-details-bin`. This is the header the collector responds with. Similarly it's the header that New Relic's OTLP ingest responds with.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/OtlpRetry.cs
utpilla: > Not sure it matters a whole lot.

Probably not. It's more of a code readability thing. It might seem a bit odd as the variable names suggest that the max value could indeed be a possible `delayDuration` value.

I'm okay with the code as-is.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/OtlpRetry.cs
utpilla: > This is the header the collector responds with.

I think it would be good to add a code comment about this.

utpilla (approved):

PR #4637, opened 2023-07-06
[Instrumentation.AspNetCore] Fix issue causing Activity is not exported when used with custom propagator/sampler.

### LINKED ISSUE #4626: [Instrumentation.AspNetCore] [.NET7.0] Activity is not exported when used with custom propagator/sampler
Runtime version `.NET7.0`

## Symptom

**Background**
In case of custom propagation a new activity is created with the context extracted from the propagator. Starting `.NET7.0` we use the `ActivitySource` on the activity created by ASP.NET Core to create and start this new [activity](https://github.com/open-telemetry/opentelemetry-dotnet/blob/104cf32d97b2aba7319635fc23d84d5f7663fe9b/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs#L150). The activity is then later identified and stopped by checking the [first tag](https://github.com/open-telemetry/opentelemetry-dotnet/blob/104cf32d97b2aba7319635fc23d84d5f7663fe9b/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs#L323).

**Issue**
As the activity is created using `ActivitySource`, the assumption that the tag `IsCreatedByInstrumentation` will be the first one is not correct. Custom samplers can add tags to activity during sampling which is done before the activity is started. In such cases the added tag `IsCreatedByInstrumentation`  by instrumentation will not be the first one preventing the activity to get stopped. We need to update the logic to check the first occurrence of that tag for `.NET7.0`+.

## Reproduce

Set the following sampler on [this](https://github.com/open-telemetry/opentelemetry-dotnet/blob/104cf32d97b2aba7319635fc23d84d5f7663fe9b/test/OpenTelemetry.Instrumentation.AspNetCore.Tests/BasicTests.cs#L198) test and run it for `.NET7.0`

```csharp
public class MyCustomSampler : Sampler
{
    public override SamplingResult ShouldSample(in SamplingParameters samplingParameters)
    {
        return new SamplingResult(
        SamplingDecision.RecordAndSample,
        new Dictionary<string, object>
            {
                { "SomeTag", "SomeKey" },
            });
    }
}
```

TimothyMothra:
Fixes #4626

When a custom propagator is used, the Instrumentation.AspNetCore library will create a new activity that is identified by a specific tag ("IsCreatedByInstrumentation").
For performance, only the first tag was inspected.

However, if a custom sampler is also used and adds tags that will affect the ordering and the instrumentation library will not identify the activity.
This causes activities to be dropped.

## Changes
- Added new extension method `ActivityHelperExtensions.TryGetTagValue()`
- Fix `HttpInListener` to inspect all tags, not just first.


src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
   which attributes are emitted by setting the environment variable
   `OTEL_SEMCONV_STABILITY_OPT_IN`.
 
+* Fixed an issue affecting NET 7.0+. If custom propagation is being used
+  and a custom sampler adds tags to an Activity that Activity would be dropped.
```
vishweshbankwar, line 11:
```suggestion
  and tags are added to an Activity during sampling then that Activity would be dropped.
```

vishweshbankwar (approved):

src/OpenTelemetry.Api/Internal/ActivityHelperExtensions.cs
```diff
@@ -95,6 +95,24 @@ public static object GetTagValue(this Activity activity, string tagName)
             return null;
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static bool TryGetTagValue(this Activity activity, string tagName, out object tagValue)
```
utpilla, line 99: We don't have to add a new method. We could use the existing method `GetTagValue`: https://github.com/open-telemetry/opentelemetry-dotnet/pull/4637/files#diff-343f92e68a24f0d2d6ba9f91024a52f4f4b7085b5d30597dfec9303a07c14fc2R83

TimothyMothra: How strongly do you feel about using the existing method?

I think the TryGet is cleaner and saves an unnecessary assertion (`if returnValue != null`).

reyang: I think the min bar is:
1. don't duplicate code (e.g. if we want to add this method, at least the existing one `GetTagValue` should be updated so it'll leverage the `TryGetTagValue`).
2. keep it consistent - examine the existing usage and make proper update.

utpilla: > saves an unnecessary assertion (if returnValue != null)

Even with the new method you have to check `(if returnValue == true)`.

reyang: > I would be in favor of reusing the existing method in this case and avoid having to introduce more code (and related tests) just to achieve a `boolean` check instead of a `null` check.

+1

TimothyMothra: removed

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/BasicTests.cs
```diff
@@ -219,7 +219,11 @@ public async Task CustomPropagator()
                             builder.ConfigureTestServices(services =>
                             {
                                 Sdk.SetDefaultTextMapPropagator(propagator.Object);
-                                this.tracerProvider = Sdk.CreateTracerProviderBuilder().AddAspNetCoreInstrumentation().AddInMemoryExporter(exportedItems).Build();
+                                this.tracerProvider = Sdk.CreateTracerProviderBuilder()
+                                                        .SetSampler(new TestSampler(SamplingDecision.RecordAndSample, new Dictionary<string, object> { { "SomeTag", "SomeKey" }, }))
```
utpilla, line 223: Would we still have a test for `CustomPropagator` which does not use a Sampler that adds tags?

If not, then make this test a `Theory` to which depending on a boolean value adds such a `Sampler` to the `CustomPropagator` test to cover both the scenarios.

utpilla (approved):

PR #4538, opened 2023-06-01
update HttpSemanticConventions for Instrumentation.Http

TimothyMothra:
Design discussion issue #4484

## Changes
- add new attributes to `SemanticConventions`
- update `Instrumentation.Http`
  - `HttpHandlerDiagnosticListener`
  - `HttpHandlerMetricsDiagnosticListener`

Please provide a brief description of the changes here.

## Merge requirement checklist

* [ ] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/CONTRIBUTING.md) guidelines followed (nullable enabled, static analysis, etc.)
* [ ] Unit tests added/updated
* [ ] Appropriate `CHANGELOG.md` files updated for non-trivial changes
* [ ] Changes in public API reviewed (if applicable)

@vishweshbankwar please review

PR #4639, opened 2023-07-06
update HttpSemanticConventions for Instrumentation.Http (part2)

TimothyMothra:
Design discussion issue #4484
Follow up to #4538

Previous PR updated Net Core implementation.
This PR makes the same changes for Net Fx.

## Changes

- update `Instrumentation.Http` `HttpWebRequestActivitySource.netfx.cs` to emit new attributes
  - http.method -> htttp.request.method
  - net.peer.name -> server.address
  - net.peer.port ->server.port
  - http.scheme -> REMOVED
  - http.url -> url.full
  - http.flavor -> network.protocol.version
  - http.status.code -> http.response.status.code
- remove "url.scheme" from other classes. 
  This was introduced in the previous PR. This is for server spans only, not client.
- Updated tests. This class was large, so I copied the original test class to make it easier to separate the scenarios.
  For review, pay attention to the helper methods at the bottom of the class near Line 200 `VerifyActivityStartTags()` and `VerifyActivityStopTags()`
  - Added `HttpWebRequestActivitySourceTestsNew.netfx.cs` to test the new attributes.
    This class can replace `HttpWebRequestActivitySourceTests.netfx.cs` when this library is GA.
  - Added `HttpWebRequestActivitySourceTestsDupe.netfx.cs` to test emitting both new and old attributes.
    This class can be deleted when this library is GA.


vishweshbankwar (approved):
LGTM - needs test https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/test/OpenTelemetry.Instrumentation.Http.Tests/HttpWebRequestTests.cs

test/OpenTelemetry.Instrumentation.Http.Tests/HttpWebRequestActivitySourceTestsDupe.netfx.cs
```diff
@@ -0,0 +1,927 @@
...
+        }
+
+        private static void VerifyActivityStopTags(int statusCode, Activity activity)
+        {
+            // New
+            Assert.Equal(statusCode, activity.GetTagValue(SemanticConventions.AttributeHttpResponseStatusCode));
+
+            // Old
+            Assert.Equal(statusCode, activity.GetTagValue(SemanticConventions.AttributeHttpStatusCode));
+        }
```
TimothyMothra, line 830: These are the new validation methods

test/OpenTelemetry.Instrumentation.Http.Tests/HttpWebRequestActivitySourceTestsNew.netfx.cs
```diff
@@ -0,0 +1,910 @@
...
+
+            Assert.Equal(netPeerName, activity.GetTagValue(SemanticConventions.AttributeServerAddress));
+
+            Assert.Equal(url, activity.GetTagValue(SemanticConventions.AttributeUrlFull));
+        }
+
+        private static void VerifyActivityStopTags(int statusCode, Activity activity)
+        {
+            Assert.Equal(statusCode, activity.GetTagValue(SemanticConventions.AttributeHttpResponseStatusCode));
+        }
```
TimothyMothra, line 813: These are the new validation methods

test/OpenTelemetry.Instrumentation.Http.Tests/HttpWebRequestActivitySourceTestsDupe.netfx.cs
utpilla, line 1: Copying the entire test file doesn't look right. We really only need to test the new/dupe attributes for just the happy-path scenario. Could you simply add a new test/theory in the existing test file to check the attributes for happy-path case?

src/OpenTelemetry.Instrumentation.Http/Implementation/HttpWebRequestActivitySource.netfx.cs
```diff
@@ -96,16 +97,34 @@ private static void AddRequestTagsAndInstrumentRequest(HttpWebRequest request, A
 
             if (activity.IsAllDataRequested)
             {
-                activity.SetTag(SemanticConventions.AttributeHttpMethod, request.Method);
-                activity.SetTag(SemanticConventions.AttributeNetPeerName, request.RequestUri.Host);
-                if (!request.RequestUri.IsDefaultPort)
+                // see the spec https://github.com/open-telemetry/opentelemetry-specification/blob/v1.20.0/specification/trace/semantic_conventions/http.md
+                if (Options.HttpSemanticConvention.HasFlag(HttpSemanticConvention.Old))
```
utpilla, line 101: These could be extracted into boolean variables to avoid checking them repeatedly. This can be done in another PR.

utpilla:
Need to update the existing CHANGELOG entry with this PR. It looks like we missed to add the previous PR in the CHANGELOG entry: https://github.com/open-telemetry/opentelemetry-dotnet/blob/f95c689d25fd9c635fa62fe9c70db0b3ac9dad4f/src/OpenTelemetry.Instrumentation.Http/CHANGELOG.md?plain=1#L5-L8

src/OpenTelemetry.Instrumentation.Http/Implementation/HttpWebRequestActivitySource.netfx.cs
TimothyMothra: The pattern used in the other libraries won't work here because this is a static class.

utpilla: Interesting. I'm not sure why this class uses a static instance of the options.

However, you should still be able to extract them into `static bool` variables in this case and use them instead of checking if the `enum` has a flag repeatedly.

test/OpenTelemetry.Instrumentation.Http.Tests/HttpWebRequestActivitySourceTestsDupe.netfx.cs
TimothyMothra: This is the same approach we took for the AspNetCore library.

I can work on a new test that just evaluates the happy-path.

src/OpenTelemetry.Instrumentation.Http/Implementation/HttpWebRequestActivitySource.netfx.cs
TimothyMothra: Not just a static instance of the options, the entire class is static.

I changed the Options field to a property so I can use the setter to evaluate the booleans in this commit: https://github.com/open-telemetry/opentelemetry-dotnet/pull/4639/commits/b91509ef80a791c45ed1cd0a88e6c8f39a2254ca

test/OpenTelemetry.Instrumentation.Http.Tests/HttpWebRequestActivitySourceTestsDupe.netfx.cs
TimothyMothra: I need more time for this. The attribute changes affect 3 different methods and there's not one test scenario that calls all three. I should have new tests by the end of today (Wed).

TimothyMothra: A separate class is going to be necessary.
I tried adding a new test to the existing class, but because this is working with static objects, it breaks the other tests in this class.

TimothyMothra: Discussed with Utkarsh, going to keep the separate test classes to avoid impacting the existing tests.
New classes will scope down to a single test.

TimothyMothra:
I've corrected the links to the spec (new repo) in all comments and changelog.
Please re-review.

utpilla (approved):

Issue #4641, opened 2023-07-06
Allow suppressing nested activities in native instrumentations without dependency on OTel SDK


lmolkova:
# Feature Request

Native instrumentations cannot suppress underlying HTTP spans without taking a dependency on OTel. They also cannot enrich/improve generic HTTP spans.

Higher-level (than HTTP) Instrumentations such as Azure SDK, [AWS](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/13c14fcc24ffe9f3fd0cbe3485d6755037cb2d37/src/OpenTelemetry.Contrib.Instrumentation.AWS/Implementation/AWSTracingPipelineHandler.cs#L226), [Elasticsearch](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/13c14fcc24ffe9f3fd0cbe3485d6755037cb2d37/src/OpenTelemetry.Instrumentation.ElasticsearchClient/Implementation/ElasticsearchRequestPipelineDiagnosticListener.cs#L87) clients may want to suppress default generic HTTP instrumentations. Here're the reasons they might not want default HTTP spans:
- too verbose. E.g. client library may want to report logical calls and supply all information there.
- add custom details:
  - in the case of Azure SDKs, we want to report `x-ms-*request-id` headers to correlate with server-side events
  - report better request duration - e.g. we know when stream reading is complete (see https://github.com/open-telemetry/opentelemetry-specification/issues/3520 discussion)
  - report better status: e.g. we may know when 404 or 409 represents/does not represent an error

Currently suppressing generic instrumentation is only possible with `SuppressInstrumentationScope`:
- Azure SDK native instrumentations cannot do it
-  [AWS](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/13c14fcc24ffe9f3fd0cbe3485d6755037cb2d37/src/OpenTelemetry.Contrib.Instrumentation.AWS/Implementation/AWSTracingPipelineHandler.cs#L226), [Elasticsearch](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/13c14fcc24ffe9f3fd0cbe3485d6755037cb2d37/src/OpenTelemetry.Instrumentation.ElasticsearchClient/Implementation/ElasticsearchRequestPipelineDiagnosticListener.cs#L87)  instrumentations do it in instr libraries that depend on OTel.

**Describe the solution you'd like:**

[...]

PR #1049, opened 2020-08-12
Add benchmark for SuppressInstrumentation

alanwest:
Related to #1079 and #1098.

|                            Method |       Mean |    Error |   StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|---------------------------------- |-----------:|---------:|---------:|-------:|------:|------:|----------:|
|      SuppressInstrumentationFalse | 1,349.4 ns | 24.13 ns | 22.58 ns | 0.0267 |     - |     - |     744 B |
|       SuppressInstrumentationTrue |   303.2 ns |  6.01 ns |  6.44 ns | 0.0091 |     - |     - |     240 B |
| SuppressInstrumentationTrueNested |   425.1 ns |  8.52 ns |  9.12 ns | 0.0172 |     - |     - |     456 B |
|                      NoneListener |   147.7 ns |  2.26 ns |  1.89 ns | 0.0007 |     - |     - |      24 B |
|       AllDataAndRecordedListener1 | 1,236.0 ns | 23.93 ns | 27.55 ns | 0.0248 |     - |     - |     656 B |
|       AllDataAndRecordedListener2 | 1,315.1 ns | 25.98 ns | 31.91 ns | 0.0267 |     - |     - |     744 B |

@reyang

PR #4642, opened 2023-07-06
Allow suppressing nested activities without dependency on OTel

lmolkova:
Demonstrates approach proposed in #4641

### DIFF
--- a/src/OpenTelemetry/SuppressInstrumentationScope.cs
+++ b/src/OpenTelemetry/SuppressInstrumentationScope.cs
@@ -16,6 +16,7 @@
 
 #nullable enable
 
+using System.Diagnostics;
 using System.Runtime.CompilerServices;
 using OpenTelemetry.Context;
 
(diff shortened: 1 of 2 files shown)


src/OpenTelemetry/SuppressInstrumentationScope.cs
```diff
@@ -29,17 +30,18 @@ public sealed class SuppressInstrumentationScope : IDisposable
...
 
+        private static readonly string SuppressActivityProperty = "otel.suppress_instrumentation";
         private readonly SuppressInstrumentationScope? previousScope;
         private bool disposed;
 
         internal SuppressInstrumentationScope(bool value = true)
         {
             this.previousScope = Slot.Get();
-            this.Depth = value ? -1 : 0;
+            this.Depth = (value || (this.previousScope == null && IsSuppressedWithActivityProperty())) ? -1 : 0;
```
reyang, line 40: We'll need perf numbers like #1049

reyang (changes requested):
How to do it efficiently is hard, we might need some primitives from .NET team.

Issue #4648, opened 2023-07-11
AOT: Make serialization of JSON arrays AOT compatible


vitek-karas:
# Feature Request

Several exporters use `System.Text.Json.JsonSerializer.Serialize(array)` in their implementation of `TagTransformer`. This API is not AOT compatible because it uses reflection at runtime to write the array values into JSON.

There are 3 options how to solve this. Note that all of them rely on the fact that `TagTransformer` will only try to serialize arrays of primitive types as listed here: https://github.com/open-telemetry/opentelemetry-dotnet/blob/b23b1460e96efb5ecd78d1b36c2e00e84de7086b/src/OpenTelemetry/Internal/TagTransformer.cs#L112-L131

## Option 1 - Source generator

The best solution (by the book solution) is to use the `System.Text.Json` source generator. This feature was introduced in version 6.0 of the package. This solution would require upgrading the `System.Text.Json` dependency to 6.0 (and also `System.Text.Encodings.Web` to 6.0).

After the upgrade the code change is relatively simple: https://github.com/open-telemetry/opentelemetry-dotnet/compare/main...vitek-karas:opentelemetry-dotnet:JsonFix_SourceGenerator?expand=1

The change introduces a helper which can serialize all of the array types from `TagTransformer` to JSON and it uses the source generator to do so.

The pros of this approach are that it's the recommended solution and thus fully supported and tested by the owners of `System.Text.Json`.
The cons is the need to upgrade the package reference to 6.0.

## Option 2 - Using low lever JSON writer

If upgrading the package reference is not an option, then an alternative solution is to manually implement serialization of the several primitive array types. The solution still uses `System.Text.Json` but only the low-level `Utf8JsonWriter` API which is fully AOT compatible.

I prepared another branch which shows how such a modification could look: https://github.com/open-telemetry/opentelemetry-dotnet/compare/main...vitek-karas:opentelemetry-dotnet:JsonFix_RawWriter?expand=1

The change is similar to the source generator in that it introduces a helper but the helper is manually implemented to support serialization of the limitted set of primitive array types necessary.

The pros of this is the fact that there's no need to upgrade package references.
The cons is that it's more code and it's not as easily extensible if there are more JSON serialization needs in the future.

## Option 3 - Use suppressions and hint to make it AOT compatible

[...]

PR #4656, opened 2023-07-14
Make serialization of JSON arrays AOT compatible

vitek-karas:
Design discussion issue #4648

## Changes

Adds a `TagTransformerJsonHelper` which provides a `JsonSerializeArrayTag` method to be used by exporters which need to serialize tag array values into JSON. The helper is only meant to be used in this specific case as it makes assumptions about which types of arrays it needs to support.

The implementation doesn't use normal JSON serialization API, instead it directly uses the low-level `Utf8JsonWriter` and implements the serialization logic on its own.
It also includes a simple implementation of `IBufferWriter` which uses ArrayPool to avoid too many allocations (it still has to allocate the returned string though).

This contributes to #3429 


vitek-karas:
@Yun-Ting @eerhardt 

@eerhardt could you please take a look?

src/Shared/TagTransformerJsonHelper.cs
```diff
@@ -0,0 +1,190 @@
...
+
+    private static void WriteLongValue(long v, Utf8JsonWriter writer) => writer.WriteNumberValue(v);
+
+    private static void WriteFloatValue(float v, Utf8JsonWriter writer) => writer.WriteNumberValue(v);
+
+    private static void WriteDoubleValue(double v, Utf8JsonWriter writer) => writer.WriteNumberValue(v);
+
+    private static string WriteToString<T>(T[] data, Action<T, Utf8JsonWriter> writeAction)
+    {
+        // Some rough estimate of the necessary size. We estimate each element to be 10 characters long (plus the separator)
```
eerhardt, line 80: (minor question): would it make sense for the type-specific code to pass this "hint" in to the method? For example `ushort` can only be 5 chars, right? 0-65535.

eerhardt, line 80: > (plus the separator)

What about the the brackets?

src/Shared/TagTransformerJsonHelper.cs
```diff
@@ -0,0 +1,190 @@
...
+        {
+            // This switch needs to support the same set of types as those in TagTransformer.TransformArrayTagInternal
+            char[] charArray => WriteToString(charArray, WriteCharValue),
+            string[] stringArray => WriteToString(stringArray, WriteStringValue),
+            bool[] booleanArray => WriteToString(booleanArray, WriteBooleanValue),
+
+            // Runtime allows casting byte[] to sbyte[] and vice versa, so pattern match to sbyte[] doesn't work since it goes through byte[]
+            // Similarly (array is sbyte[]) doesn't help either. Only real type comparison works.
+            // This is true for byte/sbyte, short/ushort, int/uint and so on
+            byte[] byteArray => array.GetType() == typeof(sbyte[]) ? WriteToString((sbyte[])array, WriteSByteValue) : WriteToString(byteArray, WriteByteValue),
```
eerhardt, line 38: `byte[]` serialize to base64 strings with JsonSerializer.

image

src/Shared/TagTransformerJsonHelper.cs
```diff
@@ -0,0 +1,190 @@
...
+            short[] shortArray => array.GetType() == typeof(ushort[]) ? WriteToString((ushort[])array, WriteUShortValue) : WriteToString(shortArray, WriteShortValue),
+            int[] intArray => array.GetType() == typeof(uint[]) ? WriteToString((uint[])array, WriteUIntValue) : WriteToString(intArray, WriteIntValue),
+
+            long[] longArray => WriteToString(longArray, WriteLongValue),
```
eerhardt, line 42: does the above casting issue exist for `ulong`?

src/Shared/TagTransformerJsonHelper.cs
vitek-karas: I thought so, but look at the tests - it uses the reflection based serializer as the "expected" value. Originally I actually coded it to produce a Base64 string, but the tests failed, because the reflection based serializer didn't do it.

I'll play with it a bit more - probably will try to debug into the existing OTel code to see what it will end up doing in real.

vitek-karas: image

I think OTel ended up always calling the method with `Array` typed parameter, so it got the second behavior (which is to write it as an array of numbers).

src/Shared/TagTransformerJsonHelper.cs
vitek-karas: Good suggestion - added

vitek-karas: good point - added

src/Shared/TagTransformerJsonHelper.cs
vitek-karas: Yes - it does - the original code didn't handle ulong specifically but due to the default casting behavior it probably worked regardless. I added ulong handling as well.

src/Shared/TagTransformerJsonHelper.cs
eiriktsarpalis: This is a side-effect of hardcoding Base64 encoding specifically for values of declared type `byte[]`. It results in different outputs if the same value is upcast to a different type:
```C#
byte[] data = new byte[] { 1 };

Console.WriteLine(JsonSerializer.Serialize(data)); // "AQ=="
Console.WriteLine(JsonSerializer.Serialize<IEnumerable<byte>>(data)); // [1]
Console.WriteLine(JsonSerializer.Serialize<Array>(data)); // [1]
```

Yun-Ting:
Implementation LGTM. Shall we add perf benchmarks? 
Comparing the implementation in this PR as opposed to using Source generator (option 1 from this issue: https://github.com/open-telemetry/opentelemetry-dotnet/issues/4648) ?

vitek-karas:
I added benchmarks to compare the new implementation, the source generator based approach and the current situation (using reflection):
```
|                                         Method |     Value |     Mean |   Error |  StdDev |
|----------------------------------------------- |---------- |---------:|--------:|--------:|
|            TagTransformerJsonSerializeArrayTag |   Char[1] | 163.6 ns | 3.14 ns | 2.79 ns |
|  TagTransformerJsonSerializeArrayTag_SourceGen |   Char[1] | 146.4 ns | 1.69 ns | 1.41 ns |
| TagTransformerJsonSerializeArrayTag_Reflection |   Char[1] | 230.6 ns | 1.97 ns | 1.84 ns |
|            TagTransformerJsonSerializeArrayTag |  Int32[5] | 189.2 ns | 2.44 ns | 2.16 ns |
|  TagTransformerJsonSerializeArrayTag_SourceGen |  Int32[5] | 155.1 ns | 1.54 ns | 1.29 ns |
| TagTransformerJsonSerializeArrayTag_Reflection |  Int32[5] | 478.4 ns | 7.29 ns | 6.82 ns |
|            TagTransformerJsonSerializeArrayTag | String[5] | 265.9 ns | 4.14 ns | 3.87 ns |
|  TagTransformerJsonSerializeArrayTag_SourceGen | String[5] | 248.5 ns | 3.45 ns | 2.88 ns |
| TagTransformerJsonSerializeArrayTag_Reflection | String[5] | 399.3 ns | 5.48 ns | 4.57 ns |
```

The source generator is slightly faster than the hand-made solution, I suspect this is because of better handling of pooled buffers (System.Text.Json uses preallocated pooled buffers for simple situations) and due to it using the latest perf trick (test in .NET 7), while the new code is netstandard2.0 compatible. We could probably make it closer, but it would likely increase complexity and amount of code. Not sure how perf-critical this is.

The currently used reflection based solution is the slowest as expected.

src/Shared/TagTransformerJsonHelper.cs
```diff
@@ -0,0 +1,192 @@
...
+namespace OpenTelemetry.Internal;
+
+#nullable enable
+
+internal static class TagTransformerJsonHelper
+{
+    internal static string JsonSerializeArrayTag(Array array)
+        => array switch
+        {
+            // This switch needs to support the same set of types as those in TagTransformer.TransformArrayTagInternal
```
eiriktsarpalis, line 30: I think it should be possible to avoid all this switching if you just use `JsonSerializer.Serialize` with the array value upcast to `object` (since `object` types are aways serialized polymorphically). Assuming you wanted to avoid `JsonSerializer` in order to customize what gets written for each array type, you could replace your `writeAction` delegate with a custom converter that you register in your serializer configuration.

vitek-karas: See for the discussion which led to this: https://github.com/open-telemetry/opentelemetry-dotnet/issues/4648
Ideally this would use source generators, but the goal is to avoid upgrading the `System.Text.Json` package reference to 6.0. Thus we can't use source generator. That means that I can't use `JsonSerializer` without it being AOT incompatible, so I coded this which avoids it completely.

eiriktsarpalis: Assuming you only need to serialize arrays of specific types, registering custom converters for the relevant types should ensure that everything works in AOT even with the reflecton-based APIs. For example, the console app only uses pre-net6.0 STJ APIs but works just fine in AOT:

```C#
var options = new JsonSerializerOptions
{
    Converters = 
    { 
        new MyArrayConverter<int>(static (writer, i) => writer.WriteNumberValue(i))
    }
};

Console.WriteLine(JsonSerializer.Serialize(new int[] { 1, 2, 3 }, options));

public class MyArrayConverter<T> : JsonConverter<T[]>
{
    private readonly Action<Utf8JsonWriter, T> _writeAction;

    public MyArrayConverter(Action<Utf8JsonWriter, T> writeAction)
        => _writeAction = writeAction;

    public override T[]? Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
        => throw new NotImplementedException();

    public override void Write(Utf8JsonWriter writer, T[] value, JsonSerializerOptions options)
    {
        writer.WriteStartArray();
        foreach (T item in value)
        {
            _writeAction(writer, item);
        }
        writer.WriteEndArray();
    }
}
```

eiriktsarpalis: The `JsonSerializer.Serialize` method will still warn but with the above configuration it's a false positive that can be suppressed.

vitek-karas: That would fit option 3 from https://github.com/open-telemetry/opentelemetry-dotnet/issues/4648. I agree it would work, but it's fragile and effectively relies on internal workings of `System.Text.Json`. The main advantage would be to reuse the buffer pooling logic in System.Text.Json, which would be nice, but I'm not sure it's worth the cost of possible future risks.

test/Benchmarks/TagTransformerJsonHelperBenchmarks.cs
```diff
@@ -0,0 +1,65 @@
...
+using System.Text.Json.Serialization;
+using BenchmarkDotNet.Attributes;
+using OpenTelemetry.Internal;
+
+namespace Benchmarks
+{
+    /// <summary>
+    /// Compares the three different ways to perform JSON array serialization.
+    /// </summary>
+    public partial class TagTransformerJsonHelperBenchmarks
```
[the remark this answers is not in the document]
Yun-Ting: Updated benchmark with heap allocation:

// * Summary *

BenchmarkDotNet=v0.13.5, OS=Windows 10 (10.0.19045.3208/22H2/2022Update)
Intel Xeon CPU E5-1650 v4 3.60GHz, 1 CPU, 12 logical and 6 physical cores
.NET SDK=7.0.306
  [Host]     : .NET 7.0.9 (7.0.923.32018), X64 RyuJIT AVX2
  DefaultJob : .NET 7.0.9 (7.0.923.32018), X64 RyuJIT AVX2


|                                         Method |     Value |     Mean |    Error |   StdDev |   Gen0 | Allocated |
|----------------------------------------------- |---------- |---------:|---------:|---------:|-------:|----------:|
|            TagTransformerJsonSerializeArrayTag |   Char[1] | 191.6 ns |  1.18 ns |  0.98 ns | 0.0315 |     248 B |
|  TagTransformerJsonSerializeArrayTag_SourceGen |   Char[1] | 203.0 ns |  3.31 ns |  3.10 ns | 0.0069 |      56 B |
| TagTransformerJsonSerializeArrayTag_Reflection |   Char[1] | 289.3 ns |  5.56 ns |  5.20 ns | 0.0067 |      56 B |
|            TagTransformerJsonSerializeArrayTag |  Int32[5] | 230.2 ns |  1.00 ns |  0.83 ns | 0.0336 |     264 B |
|  TagTransformerJsonSerializeArrayTag_SourceGen |  Int32[5] | 213.7 ns |  4.20 ns |  3.93 ns | 0.0060 |      48 B |
| TagTransformerJsonSerializeArrayTag_Reflection |  Int32[5] | 607.5 ns | 11.81 ns | 13.13 ns | 0.0210 |     168 B |
|            TagTransformerJsonSerializeArrayTag | String[5] | 329.0 ns |  5.75 ns |  5.38 ns | 0.0458 |     360 B |
|  TagTransformerJsonSerializeArrayTag_SourceGen | String[5] | 322.6 ns |  6.18 ns |  6.07 ns | 0.0181 |     144 B |
| TagTransformerJsonSerializeArrayTag_Reflection | String[5] | 522.4 ns |  8.87 ns |  7.87 ns | 0.0181 |     144 B |

src/Shared/TagTransformerJsonHelper.cs
```diff
@@ -0,0 +1,192 @@
...
+    private static void WriteFloatValue(float v, Utf8JsonWriter writer) => writer.WriteNumberValue(v);
+
+    private static void WriteDoubleValue(double v, Utf8JsonWriter writer) => writer.WriteNumberValue(v);
+
+    private static string WriteToString<T>(T[] data, Action<T, Utf8JsonWriter> writeAction, int elementSizeHint)
+    {
+        // Some rough estimate of the necessary size. We multiple the length of the array (+1 to be on the safe side)
+        // by the estimated size of an element and add 2 for the brackets.
+        using PooledByteBufferWriter bufferWriter = new PooledByteBufferWriter(((data.Length + 1) * (elementSizeHint + 1)) + 2);
+        using (Utf8JsonWriter writer = new Utf8JsonWriter(bufferWriter))
```
Yun-Ting, line 85: I'm curious to learn the reasons on choosing `PooledByteBufferWriter` over the `Stream` overload?
https://github.com/open-telemetry/opentelemetry-dotnet/blob/e7f7cd68759d81aa3a34cb88d1b2590e09c21e89/src/OpenTelemetry.Exporter.Zipkin/ZipkinExporter.cs#L241C21-L241C62

vitek-karas: Well `Stream` itself is just an interface (abstract class, but that's the same thing basically) - the interesting bit is where does the implementation of it gets the memory to store the results in. In this case we need to store the results for only a brief moment, then convert it to string and drop it.
The typical implementation would use `MemoryStream`, but that will allocate twice for each serialization (the `MemoryStream` itself has to be allocated and then we have to allocate the buffer within it). The `PooledByteBufferWriter` avoids the second allocation (the buffer is taken from an `ArrayPool`). So less stress on GC.
We could also implemented something like `PooledByBufferStream` and use the `Stream` overload, but there would be basically no difference in behavior - and the `Stream` is more complex to implement (more methods to implement and the writing is more complex) and probably less performant as the JSON writer can't write directly to the internal buffer - so it would require additional memory copy.

Unfortunately we can't use the same trick in the Zipkin exporter as it has to implement `HttpContent` APIs which are all `Stream` based.

test/Benchmarks/TagTransformerJsonHelperBenchmarks.cs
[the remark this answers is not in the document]
vitek-karas: Thanks a lot for adding that. Obviously we could optimize this code some more to decrease the allocations even more. But honestly it feels like diminishing returns. If we really want the best perf, then we should just use source gen.

utpilla:
This work was completed in #4679

PR #4675, opened 2023-07-20
[AOT] Make `PropertyFetcher` AOT compatible.

Yun-Ting:

## Changes

https://github.com/open-telemetry/opentelemetry-dotnet/blob/457925baea95a05db006f1362c2d4985ad3d93b5/src/Shared/DiagnosticSourceInstrumentation/PropertyFetcher.cs#L135C1-L153C100


src/Shared/DiagnosticSourceInstrumentation/PropertyFetcher.cs
```diff
@@ -38,30 +43,16 @@ public PropertyFetcher(string propertyName)
         this.propertyName = propertyName;
     }
 
-    /// <summary>
```
Yun-Ting, line 41: dead code even before this code change

test/OpenTelemetry.AotCompatibility.Tests/AotCompatibilityTests.cs
```diff
@@ -85,7 +85,7 @@ public void EnsureAotCompatibility()
             Assert.True(process.ExitCode == 0, "Publishing the AotCompatibility app failed. See test output for more details.");
 
             var warnings = expectedOutput.ToString().Split('\n', '\r').Where(line => line.Contains("warning IL"));
-            Assert.Equal(36, warnings.Count());
+            Assert.Equal(42, warnings.Count());
```
CodeBlanch, line 88: Interesting this is going up with these changes. I would expect it to be going down always. Should we add suppressions in the test for where we know this code is being called?

src/Shared/DiagnosticSourceInstrumentation/PropertyFetcher.cs
```diff
@@ -26,6 +27,9 @@ namespace OpenTelemetry.Instrumentation;
 /// <typeparam name="T">The type of the property being fetched.</typeparam>
 internal sealed class PropertyFetcher<T>
 {
+#if NET6_0_OR_GREATER
+    private const string TrimCompatibilityMessage = "PropertyFetcher access property on the object type; therefore is not trim compatible.";
```
CodeBlanch, line 31: nit: Feels kind of awkward. Here is my attempt at making it better:

`PropertyFetcher is used to access properties on objects dynamically by design and cannot be made trim compatible`

src/Shared/DiagnosticSourceInstrumentation/PropertyFetcher.cs
```diff
@@ -105,10 +93,28 @@ static PropertyFetch CreateFetcherForProperty(PropertyInfo propertyInfo)
                     return null;
                 }
 
-                var typedPropertyFetcher = typeof(TypedPropertyFetch<,>);
-                var instantiatedTypedPropertyFetcher = typedPropertyFetcher.MakeGenericType(
-                    typeof(T), propertyInfo.DeclaringType, propertyInfo.PropertyType);
-                return (PropertyFetch)Activator.CreateInstance(instantiatedTypedPropertyFetcher, propertyInfo);
+                if (propertyInfo.DeclaringType!.IsValueType || propertyInfo.PropertyType.IsValueType || typeof(T).IsValueType)
+                {
+                    return new BoxedValueTypePropertyFetch(propertyInfo);
```
CodeBlanch, line 98: @Yun-Ting Does any of the existing code using `PropertyFetcher` today hit this branch? If not, could we just change this into a `throw NotSupportedException` and mark it all as AoT-safe?

test/OpenTelemetry.AotCompatibility.Tests/AotCompatibilityTests.cs
eerhardt: This is typical for trim/AOT warnings. If you have a centralized method that is doing something incompatible, and you annotate it as such, now you start getting warnings for all the callsites of that method, and you need to fix them.

src/Shared/DiagnosticSourceInstrumentation/PropertyFetcher.cs
```diff
@@ -105,10 +93,28 @@ static PropertyFetch CreateFetcherForProperty(PropertyInfo propertyInfo)
                     return null;
                 }
 
-                var typedPropertyFetcher = typeof(TypedPropertyFetch<,>);
-                var instantiatedTypedPropertyFetcher = typedPropertyFetcher.MakeGenericType(
-                    typeof(T), propertyInfo.DeclaringType, propertyInfo.PropertyType);
-                return (PropertyFetch)Activator.CreateInstance(instantiatedTypedPropertyFetcher, propertyInfo);
+                if (propertyInfo.DeclaringType!.IsValueType || propertyInfo.PropertyType.IsValueType || typeof(T).IsValueType)
```
eerhardt, line 96: We should also add a check for:

```suggestion
                if (!RuntimeFeature.IsDynamicCodeSupported && (propertyInfo.DeclaringType!.IsValueType || propertyInfo.PropertyType.IsValueType || typeof(T).IsValueType))
```

This way we use the current approach when not on the NativeAOT runtime.

utpilla:
@vitek-karas I discussed this idea with @eerhardt and I would like to hear your thoughts on it.

For the instrumentation libraries using `PropertyFetcher`, we already know the types (and their properties) logged by `DiagnosticSource` events that we subscribe to. I was thinking if there's a way we could use this information to our benefit.

What do you thinking about individual libraries using type-targeted AOT-safe reflection APIs instead of using the general purpose `PropertyFetcher`?

For example:
This [code](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Instrumentation.Http/Implementation/HttpHandlerDiagnosticListener.cs#L115) could be updated to use something like this instead of `PropertyFetcher`:
```csharp
var assembly = typeof(HttpRequestMessage).Assembly;
var propertyInfo = assembly.GetType("System.Net.Http.DiagnosticsHandler+ActivityStartData").GetProperty("Request", typeof(HttpRequestMessage));
var request = propertyInfo.GetValue(payload) as HttpRequestMessage;
```

- There is a risk of the runtime team changing the type name but it's unlikely that they would do it in the minor/patch version releases? For any new major version release, we would be testing our libraries against them anyway before they are released.
- If the `DiagnosticSource` event logs an anonymous type, then the runtime code should be updated to add `DynamicallyAccessedMembers` attribute just like AspNetCore has added it [here](https://github.com/dotnet/aspnetcore/blob/1bda10b33b6cc6f3bbaceabbadb4ddd18ca6e68e/src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddlewareImpl.cs#L174) for logging [exception ](https://github.com/dotnet/aspnetcore/blob/1bda10b33b6cc6f3bbaceabbadb4ddd18ca6e68e/src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddlewareImpl.cs#L156)in an anonymous type.

Yun-Ting:
> * There is a risk of the runtime team changing the type name but it's unlikely that they would do it in the minor/patch version releases? For any new major version release, we would be testing our libraries against them anyway before they are released.

There had been discussions on making the properties public: 
https://github.com/dotnet/aspnetcore/issues/48616, https://github.com/dotnet/runtime/issues/82910, https://github.com/grpc/grpc-dotnet/issues/2154.

> * If the `DiagnosticSource` event logs an anonymous type, then the runtime code should be updated to add `DynamicallyAccessedMembers` attribute just like AspNetCore has added it [here](https://github.com/dotnet/aspnetcore/blob/1bda10b33b6cc6f3bbaceabbadb4ddd18ca6e68e/src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddlewareImpl.cs#L174) for logging [exception ](https://github.com/dotnet/aspnetcore/blob/1bda10b33b6cc6f3bbaceabbadb4ddd18ca6e68e/src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddlewareImpl.cs#L156)in an anonymous type.

Discussed in 7/25 SIG meeting that current Design for PropertyFetcher is mainly because of perf reasons, https://github.com/open-telemetry/opentelemetry-dotnet/pull/4675/files#:~:text=//%201.%20ReferenceTypePropertyFetch%20is,a%20value%20type.

vitek-karas:
I think there are multiple angles to this:
* Trim correctness - who makes sure that the properties are kept if they're accessed via reflection
* Performance - reflection APIs are slow
* AOT correctness - some tricks can't be done with AOT
* Current point in time versus what it should look like in the future

### Trim correctness
For types which are non-anonymous I agree that using `Type.GetType("System.Net.Http.DiagnosticsHandler+ActivityStartData").GetProperty("Request")` is slightly better than the current approach. It doesn't rely on proper annotations in the runtime repo, but it does introduce coupling to internal type names. You trade one dependency for another.

For anonymous types we still have to rely on the source to make sure to preserve the properties necessary. 

Note that currently Http, ASP.NET and Grpc all make sure that the properties are kept, so it's not a functional issue right now. And they're likely to keep doing that because there are other tools than just OTel which make this a requirement (dotnet-monitor for example).

Also of note is that for this new approach to be effective it has to be done carefully as not all code patterns are recognized by the trim tools and thus won't work. For example, the pattern you wrote will not work (it's basically a missing feature that the trimmer doesn't track `Assembly` values, so it won't know what `assembly.GetType` actually does). This is fixable, but requires special care - on the other hand trim warnings will be very clear what will work.

One other fact to note: Currently `PropertyFetcher` will work on multiple different payload types. This can be nicely seen in this test: https://github.com/open-telemetry/opentelemetry-dotnet/blob/1fa50fb5386624737def51f221ed8069ae076aa3/test/OpenTelemetry.Tests/Instrumentation/PropertyFetcherTest.cs#L55-L68. The same instance of `PropertyFetcher` is used on multiple different payload types and it always works.
I know that the Http instrumentation makes use of this in one case. The proposed approach would not work like this and the `PropertyFetcher` would only work on one payload type. This is obviously fixable, just wanted to point out that it would be a slightly bigger change because of this.

### Performance

Always calling `PropertyInfo.GetValue` is probably a performance problem - I can only assume this because I don't know why OTel started to use `PropertyFetcher`. The current design of `PropertyFetcher` serves two purposes - it isolates knowledge of reflection APIs so that the callsite is "simple" and it improves perf by caching reflection API results. 

To mitigate the perf problem the current implementation uses the `PropertyInfo.GetMethod.CreateDelegate` to cache a delegate to the target property getter, effectively avoiding any reflection calls on all but the first invocations.

The same approach could be used with the hardcoded type names you suggested, but it would still mean we need to keep something similar to `PropertyFetcher` around (for both the "hiding details" and "perf" reasons).

### AOT correctness

Unfortunately the same property trick with `CreateDelegate` doesn't always work with AOT. This is partially due to missing data flow tracking in the compiler, but that would be pretty complex to fix fully. The good news is that it will work if all types around the property are reference types, in that case the trick can be used even with AOT. Except for one event in HTTP all events use reference types everywhere, so this is not a big problem right now, but the code needs to be able to handle both cases. (This PR falls back to going through reflection every time for value types). Another good reason to have some abstraction like `PropertyFetcher` to hide these details from the callsites.

### Future

The ideal solution to this is to make the payload types public. ASP.NET did this partially in .NET 7 and OTel makes use of it (notice that in that case there's no `PropertyFetcher` used at all). I filed issue for HTTP, ASP.NET and Grpc to make the payload types public consistently. This will not happen in .NET 8 fully, but maybe there will be some progress already.

The value is clear because it helps all targets. Having a public type allows the consumer (OTel and others) to have a cast and direct call which is faster than anything based on reflection and 100% trim and AOT compatible by definition.
https://github.com/dotnet/runtime/issues/82910
https://github.com/dotnet/aspnetcore/issues/48616
https://github.com/grpc/grpc-dotnet/issues/2154

Current solution is necessary for "now", but we should push for public types going forward (still have to keep the current solution though to support downlevel targets).

### Conclusion

I think we could change `PropertyFetcher` to have a trim-compatible path which takes `PropertyInfo` instead of just a string name. In that case there would be no trim warnings anywhere. The caller would have to do the work to lookup the type and property on it - using code similar to your proposal. This would shift some of the reflection calls from `PropertyFetcher` to the caller, but most of the value of having `PropertyFetcher` would still remain.
We would still have to keep the AOT side of this PR, since those can't currently be fixed any other way as far as I know.
I would still keep `PropertyFetcher` as the thing which acts like the "cache" to make accesses fast (still use the `CreateDelegate` trick).
We would still have to keep the existing functionality for `SqlClient` at the very least because of its use of anonymous types - but that won't be trim compatible right now no matter what we do here.

I can try to take a shot at this to see what it would look like with this change. That said the value is basically the tradeoff between hardcoding internal type names versus relying on correct annotations in the source repo. Personally I don't have enough library experience to be able to judge which is better.

eerhardt:
> What do you thinking about individual libraries using type-targeted AOT-safe reflection APIs instead of using the general purpose PropertyFetcher?

Giving this idea some thought, my opinion is that don't hard-code these private type names into OpenTelemetry. Instead, we depend on the fact that the properties are annotated in the underlying libraries - http, aspnet, grpc - to preserve the properties.

My reasoning is because it is much more likely that we will move/rename/expose these private types in the future (ex. https://github.com/dotnet/runtime/issues/82910) than it is we would remove the annotations to preserve the properties.

So if we hard-code the type names now, and then in the future expose the types publicly with a different name, the OpenTelemetry code would need to be updated.

vitek-karas:
I just realized one other thing. If the type name changes in any of the sources, open telemetry would have to probe for both the old and new one (as long as the pre-change TFM is supported). The annotations don't have this problem since the consumer never sees the actual type name.

test/OpenTelemetry.AotCompatibility.Tests/AotCompatibilityTests.cs
Yun-Ting: @CodeBlanch:
I originally was thinking to isolate changes of the callers into the PropertyFetcher, and create follow-up PRs for that.

But SqlClient code changes are very straightforward; I just pushed a commit for that:
https://github.com/open-telemetry/opentelemetry-dotnet/pull/4675/commits/4a7534cb642930d8a3e62e140c7727bf802c445f

src/Shared/DiagnosticSourceInstrumentation/PropertyFetcher.cs
vitek-karas: Unfortunately it is used: https://github.com/open-telemetry/opentelemetry-dotnet/blob/0343116e7781779975a1a0fe269be1778a484eeb/src/OpenTelemetry.Instrumentation.Http/Implementation/HttpHandlerDiagnosticListener.cs#L46

The `TaskStatus` is a value type (enum) and so we can't use the `MakeGenericType` in AOT for it.

test/OpenTelemetry.AotCompatibility.Tests/AotCompatibilityTests.cs
utpilla: @Yun-Ting Could you move the Sql changes to a separate PR? It would be to see its impact alone on the number of warnings instead of mixing up with PropertyFetcher's effect on the warnings. (as PropertyFetcher had an increasing effect on it)

utpilla:
> My reasoning is because it is much more likely that we will move/rename/expose these private types in the future (ex. [dotnet/runtime#82910](https://github.com/dotnet/runtime/issues/82910)) than it is we would remove the annotations to preserve the properties.
> 
> So if we hard-code the type names now, and then in the future expose the types publicly with a different name, the OpenTelemetry code would need to be updated.

@eerhardt Regardless of the approach we take, we would always have to update the OpenTelemetry code, whenever runtime team makes these types public as we would have to switch from using reflection to direct access.

utpilla:
> I just realized one other thing. If the type name changes in any of the sources, open telemetry would have to probe for both the old and new one (as long as the pre-change TFM is supported). The annotations don't have this problem since the consumer never sees the actual type name.

@tarekgh @noahfalk We are trying to make the instrumentation libraries of this repo AOT safe. These libraries subscribe to `DiagnosticsSource` events of AspNetCore, HttpClient etc. and use reflection to get hold of certain properties from the payload that gets logged by those `DiagnosticSource` events. We would have a simpler/cleaner way of making these libraries AOT safe if we can bet on the runtime team not changing the type of the payload logged. For example:
- [ActivityStartData](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Net.Http/src/System/Net/Http/DiagnosticsHandler.cs#L214-L229) logged [here](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Net.Http/src/System/Net/Http/DiagnosticsHandler.cs#L122)
- [ActivityStopData](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Net.Http/src/System/Net/Http/DiagnosticsHandler.cs#L231-L245) logged [here](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Net.Http/src/System/Net/Http/DiagnosticsHandler.cs#L182)
- [BeforeActionEventData](https://github.com/dotnet/aspnetcore/blob/1bda10b33b6cc6f3bbaceabbadb4ddd18ca6e68e/src/Mvc/Mvc.Core/src/Diagnostics/MvcDiagnostics.cs#L14-L60) logged [here](https://github.com/dotnet/aspnetcore/blob/main/src/Mvc/Mvc.Core/src/MvcCoreDiagnosticListenerExtensions.cs#L42-L44)
- There are some anonymous types as well - for which we currently use reflection to get hold of their property

Would it be unwise to bet on the runtime not changing the payload types? (I'm kind of hoping that the runtime would definitely not change these types for minor/patch version releases. For major version releases, it would be great if they simply make these types public instead of changing them while keeping them internal.)

tarekgh:
> Would it be unwise to bet on the runtime not changing the payload types? (I'm kind of hoping that the runtime would definitely not change these types for minor/patch version releases. For major version releases, 

I am not expecting these types to change soon. by the way [BeforeActionEventData](https://github.com/dotnet/aspnetcore/blob/1bda10b33b6cc6f3bbaceabbadb4ddd18ca6e68e/src/Mvc/Mvc.Core/src/Diagnostics/MvcDiagnostics.cs#L14-L60) is already public. 

> it would be great if they simply make these types public instead of changing them while keeping them internal.)

This is something we can consider if there is not any other better solution.

eerhardt:
> @eerhardt Regardless of the approach we take, we would always have to update the OpenTelemetry code whenever runtime team makes these types public as we would have to switch from using reflection to direct access.

If you kept the same approach as you are doing now, and relied on the annotations in the runtime, OpenTelemetry code would continue to work even if the runtime team makes these types public. You wouldn't "have to switch" from using reflection to direct access.

The property names of these types can never change because they are a public contract - the names get hard-coded in `dotnet monitor` and other diagnostic tools. The property names couldn't be changed without breaking a lot of things. Since the property names can never change, your code would never "have to" change.

That being said, of course we would **want** to change to the direct access approach when it happens.

eerhardt:
> Taking the current approach means that even if someone is using the older version of OpenTelemetry (which uses Property Name based reflection on the object.GetType()) and uses the new .NET runtime with the public type names changed would continue to work.

The concrete example here is:

Say in .NET 9 the runtime team fixed https://github.com/dotnet/runtime/issues/82910 to make `System.Net.Http.DiagnosticsHandler.ActivityStartData` public. But this is a nested class inside an internal class. It is very unlikely API review would approve of making the outer class public just to make a nested class public (when we have guidelines to not use public nested types). So it is very likely the public class would have a different type name.

If we took the approach of hard-coding `Type.GetType("System.Net.Http.DiagnosticsHandler.ActivityStartData")` in OpenTelemetry `v1.6.0`, and then .NET 9 ships with the public `HttpActivityStartData` class renamed/moved from a nested class, anyone using `v1.6.0` of OpenTelemetry on .NET 9 would break. They would be forced to update to the updated OpenTelemetry (which hopefully ships at the same time as .NET 9). Breaking changes like this cause friction in upgrading.

test/OpenTelemetry.AotCompatibility.Tests/AotCompatibilityTests.cs
Yun-Ting: I've reverted the SqlClientInstrumentation suppression changes.

test/OpenTelemetry.AotCompatibility.TestApp/Program.cs
```diff
@@ -14,4 +14,17 @@
...
+    PropertyFetcherAotTest.Test();
+}
+catch (Exception ex)
+{
+    Console.WriteLine(ex);
+    return -1;
+}
+
+Console.WriteLine("Passed.");
+return 0;
```
eerhardt, line 30: When/how does this get executed and checked to ensure that `0` was returned? Should we run the executable as part of "assert-publish-aot-testApp-warningCount.ps1"?

vitek-karas: Yes we should.
My bad - when I was adding this code we didn't have the CI leg yet and I didn't want to mess with the test knowing it will change.

@Yun-Ting could you please add this - just run the produced binary and check that it's exit code is 0. It should print out any failures to output already.

src/Shared/DiagnosticSourceInstrumentation/PropertyFetcher.cs
```diff
@@ -77,61 +68,119 @@ public bool TryFetch(object obj, out T value, bool skipObjNullCheck = false)
...
+        //    Compiler supports it even if it is a value type; the type is known statically during compilation
+        //    since PropertyFetcher is used with it.
+        // 2. Then, the declared object type was passed as a generic parameter to a generic method on PropertyFetcher<T> (or nested type.)
+        //    Therefore, calling into MakeGenericMethod will only require specifying one parameter - the declared object type.
+        //    The declared object type is guaranteed to be a reference type (throw on value type.) Thus, MakeGenericMethod is AOT compatible.
+        private static PropertyFetch CreateInstantiated<TDeclaredObject>(PropertyInfo propertyInfo)
+            where TDeclaredObject : class
+            => new PropertyFetchInstantiated<TDeclaredObject>(propertyInfo);
+
+        // ReferenceTypePropertyFetch is the optimized version because it uses CreateDelegate to get a Delegate directly to get the property.
```
vitek-karas, line 164: I think we can remove this comment now - since there's no "Second" implementation

src/Shared/DiagnosticSourceInstrumentation/PropertyFetcher.cs
```diff
@@ -77,61 +68,119 @@ public bool TryFetch(object obj, out T value, bool skipObjNullCheck = false)
...
+                {
+                    return DynamicInstantiationHelper(declaringType, propertyInfo);
+                }
+
+                // Separated as local function to be able to target the suppression to just this call
+                // IL3050 was generated here because of the call to MakeGenericType, which is problematic in AOT if one of the type parameters is a value type;
+                // because the compiler might need to generate code specific to that type.
+                // If the type parameter is reference type, there will be no problem; because the generated code can be shared among all reference type instantiations.
+#if NET6_0_OR_GREATER
+                [RequiresUnreferencedCode(TrimCompatibilityMessage)]
```
vitek-karas, line 123: Why do we need `RequiresUnreferencedCode` here? It should not be needed. Only the AOT suppression should be needed here.

src/Shared/DiagnosticSourceInstrumentation/PropertyFetcher.cs
```diff
@@ -77,61 +68,119 @@ public bool TryFetch(object obj, out T value, bool skipObjNullCheck = false)
...
+        // 2. Then, the declared object type was passed as a generic parameter to a generic method on PropertyFetcher<T> (or nested type.)
+        //    Therefore, calling into MakeGenericMethod will only require specifying one parameter - the declared object type.
+        //    The declared object type is guaranteed to be a reference type (throw on value type.) Thus, MakeGenericMethod is AOT compatible.
+        private static PropertyFetch CreateInstantiated<TDeclaredObject>(PropertyInfo propertyInfo)
+            where TDeclaredObject : class
+            => new PropertyFetchInstantiated<TDeclaredObject>(propertyInfo);
+
+        // ReferenceTypePropertyFetch is the optimized version because it uses CreateDelegate to get a Delegate directly to get the property.
+#if NET6_0_OR_GREATER
+        [UnconditionalSuppressMessage("AOT", "IL2109", Justification = "The code guarantees that TDeclaredObject is a reference type.")]
```
vitek-karas, line 166: Cleaner solution would be to put `RequiresUnreferencedCode` on this type as well.

src/Shared/DiagnosticSourceInstrumentation/PropertyFetcher.cs
```diff
@@ -77,61 +68,119 @@ public bool TryFetch(object obj, out T value, bool skipObjNullCheck = false)
...
-            public TypedPropertyFetch(PropertyInfo property)
+            public PropertyFetchInstantiated(PropertyInfo property)
             {
                 this.propertyName = property.Name;
-                this.propertyFetch = (Func<TDeclaredObject, TDeclaredProperty>)property.GetMethod.CreateDelegate(typeof(Func<TDeclaredObject, TDeclaredProperty>));
+                this.propertyFetch = (Func<TDeclaredObject, T>)property.GetMethod!.CreateDelegate(typeof(Func<TDeclaredObject, T>));
             }
 
+#if NET6_0_OR_GREATER
+            [RequiresUnreferencedCode(TrimCompatibilityMessage)]
```
vitek-karas, line 182: Once the attribute is on the type this can be removed

src/Shared/DiagnosticSourceInstrumentation/PropertyFetcher.cs
```diff
@@ -77,61 +68,119 @@ public bool TryFetch(object obj, out T value, bool skipObjNullCheck = false)
...
+                        .MakeGenericMethod(declaringType) // This is validated in the earlier call chain to be a reference type.
+                        .Invoke(null, new object[] { propertyInfo })!;
+                }
             }
         }
 
         public virtual bool TryFetch(object obj, out T value)
         {
-            value = default;
+            value = default!;
```
vitek-karas, line 138: I think for the nullable annotations to be correct, it will have to be `out T? value` - there's no guarantee that the value will not be null - even if the fetch succeeds.

test/OpenTelemetry.AotCompatibility.TestApp/Program.cs
Yun-Ting: I've created a side PR for this:
https://github.com/open-telemetry/opentelemetry-dotnet/pull/4743

src/Shared/DiagnosticSourceInstrumentation/PropertyFetcher.cs
```diff
@@ -85,54 +76,109 @@ public bool TryFetch(object obj, out T value, bool skipObjNullCheck = false)
...
     {
         public static PropertyFetch Create(TypeInfo type, string propertyName)
         {
-            var property = type.DeclaredProperties.FirstOrDefault(p => string.Equals(p.Name, propertyName, StringComparison.OrdinalIgnoreCase));
-            if (property == null)
-            {
-                property = type.GetProperty(propertyName);
-            }
+            var property = type.DeclaredProperties.FirstOrDefault(p => string.Equals(p.Name, propertyName, StringComparison.OrdinalIgnoreCase)) ?? type.GetProperty(propertyName);
+            return CreateFetcherForProperty(property!)!;
```
eerhardt, line 87: This isn't correct. Instead, you should make this:

```suggestion
            return CreateFetcherForProperty(property!);
```

And then update the signature of this method. See the other suggestion.

src/Shared/DiagnosticSourceInstrumentation/PropertyFetcher.cs
```diff
@@ -85,54 +76,109 @@ public bool TryFetch(object obj, out T value, bool skipObjNullCheck = false)
     }
 
     // see https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/DiagnosticSourceEventSource.cs
+#if NET6_0_OR_GREATER
+    [RequiresUnreferencedCode(TrimCompatibilityMessage)]
+#endif
     private class PropertyFetch
     {
         public static PropertyFetch Create(TypeInfo type, string propertyName)
```
eerhardt, line 84:
```suggestion
        public static PropertyFetch? Create(TypeInfo type, string propertyName)
```

src/Shared/DiagnosticSourceInstrumentation/PropertyFetcher.cs
```diff
@@ -85,54 +73,109 @@ public bool TryFetch(object obj, out T value, bool skipObjNullCheck = false)
...
+        // Create generics with the declared object type as a generic parameter is OK, but we need the return type
+        // of the property to be a value type (on top of reference types.)
+        // Normally, we would have a helper class like `PropertyFetchInstantiated` that takes 2 generic parameters,
+        // the declared object type, and the type of the property value.
+        // But that would mean calling MakeGenericType, with value type parameters which AOT won't support.
+        //
+        // As a workaround, Generic instantiation was split into:
+        // 1. The property value type comes from the PropertyFetcher generic parameter.
+        //    Compiler supports it even if it is a value type; the type is known statically during compilation
+        //    since PropertyFetcher is used with it.
```
CodeBlanch, line 153: This is a bit confusing because "value type" means two different things (I think) in this sentence. Maybe change the first one?

```csharp
        // 1. The object type comes from the PropertyFetcher generic parameter.
        //    Compiler supports it even if it is a value type; the type is known statically during compilation
        //    since PropertyFetcher is used with it.
```

CodeBlanch: Instead of saying "Then, the declared object type **was** passed as a generic parameter" maybe say "Then, the declared object type **is** passed as a generic parameter" (the past tense "was" made it confusing for me)

src/Shared/DiagnosticSourceInstrumentation/PropertyFetcher.cs
```diff
@@ -85,54 +73,109 @@ public bool TryFetch(object obj, out T value, bool skipObjNullCheck = false)
...
+                if (declaringType!.IsValueType)
+                {
+                    throw new NotSupportedException(
+                        $"Type: {declaringType.FullName} is a value type. PropertyFetcher can only operate on reference payload types.");
+                }
+
+                if (declaringType == typeof(object))
+                {
+                    // This is only necessary on .NET 7.
+                    // It was used to force the AOT compiler to create an instantiation of the method with a reference type.
```
CodeBlanch, line 104: nit: Could just be me but the past tense seems confusing 😄 

```diff
- It was used to force the AOT compiler to create an instantiation of the method with a reference type.
+ It is used to force the AOT compiler to create an instantiation of the method with a reference type.
```

test/OpenTelemetry.Tests/Instrumentation/PropertyFetcherTest.cs
```diff
@@ -78,6 +78,24 @@ public void FetchPropertyMultiplePayloadTypes_IgnoreTypesWithoutExpectedProperty
         Assert.Equal("A", propertyValue);
     }
 
+    [Fact]
```
CodeBlanch, line 81: Possible gaps in tests:

* A test to verify the throw behavior when a root value type is handed to property fetcher.
* Does one of these tests fall into the `if (declaringType == typeof(object))` branch? Something like this (I think 😄)...

   ```csharp
      private class PayloadTypeWithObjectType
       {
           public object Property { get; set; } = new DerivedType();
       }
   ```

src/Shared/DiagnosticSourceInstrumentation/PropertyFetcher.cs
```diff
@@ -85,54 +73,109 @@ public bool TryFetch(object obj, out T value, bool skipObjNullCheck = false)
     }
 
     // see https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/DiagnosticSourceEventSource.cs
+#if NET6_0_OR_GREATER
+    [RequiresUnreferencedCode(TrimCompatibilityMessage)]
```
utpilla, line 77: If you have marked the class `PropertyFetcher` with `RequiresUnreferencedCode` attribute, do you still need to individually add this attribute to other nested types and methods in this class?

utpilla: Or do you even have to add the annotation to the class when you have already targeted the suppression to the specific methods and properties?

Yun-Ting: Yes, we will still need this attribute in this class. We'll need to annotate the call chain.

utpilla: > We'll need to annotate the call chain.

If we just need to annotate the call chain, then we should add annotations ONLY to the specific methods of the type which are not trim safe. Why do we need to annotate the type?

What's the guiding principle here for adding annotations?
- Do we want to add annotations ONLY to the specific methods of a type to clearly demonstrate which code is unsafe?
- Do we want to add the annotation to the type? Does this mean that every method of the type would then be considered unsafe?
- Is there some value to annotating both the type and its methods?

src/Shared/DiagnosticSourceInstrumentation/PropertyFetcher.cs
```diff
@@ -85,54 +73,109 @@ public bool TryFetch(object obj, out T value, bool skipObjNullCheck = false)
...
+        //    Compiler supports it even if it is a value type; the type is known statically during compilation
+        //    since PropertyFetcher is used with it.
+        // 2. Then, the declared object type is passed as a generic parameter to a generic method on PropertyFetcher<T> (or nested type.)
+        //    Therefore, calling into MakeGenericMethod will only require specifying one parameter - the declared object type.
+        //    The declared object type is guaranteed to be a reference type (throw on value type.) Thus, MakeGenericMethod is AOT compatible.
+        private static PropertyFetch CreateInstantiated<TDeclaredObject>(PropertyInfo propertyInfo)
+            where TDeclaredObject : class
+            => new PropertyFetchInstantiated<TDeclaredObject>(propertyInfo);
+
+        // ReferenceTypePropertyFetch is the optimized version because it uses CreateDelegate to get a Delegate directly to get the property.
```
utpilla, line 161: Remove this comment? We are not using `ReferenceTypePropertyFetch` now.

src/Shared/DiagnosticSourceInstrumentation/PropertyFetcher.cs
```diff
@@ -85,54 +73,109 @@ public bool TryFetch(object obj, out T value, bool skipObjNullCheck = false)
...
         {
-            var property = type.DeclaredProperties.FirstOrDefault(p => string.Equals(p.Name, propertyName, StringComparison.OrdinalIgnoreCase));
-            if (property == null)
-            {
-                property = type.GetProperty(propertyName);
-            }
-
-            return CreateFetcherForProperty(property);
+            var property = type.DeclaredProperties.FirstOrDefault(p => string.Equals(p.Name, propertyName, StringComparison.OrdinalIgnoreCase)) ?? type.GetProperty(propertyName);
+            return CreateFetcherForProperty(property!);
```
eerhardt, line 84:
```suggestion
            return CreateFetcherForProperty(property);
```

See below for how you can remove this.

src/Shared/DiagnosticSourceInstrumentation/PropertyFetcher.cs
```diff
@@ -85,54 +73,109 @@ public bool TryFetch(object obj, out T value, bool skipObjNullCheck = false)
...
 
-            static PropertyFetch CreateFetcherForProperty(PropertyInfo propertyInfo)
+            static PropertyFetch? CreateFetcherForProperty(PropertyInfo propertyInfo)
```
eerhardt, line 86:
```suggestion
            static PropertyFetch? CreateFetcherForProperty(PropertyInfo? propertyInfo)
```

src/Shared/DiagnosticSourceInstrumentation/PropertyFetcher.cs
```diff
@@ -85,54 +73,109 @@ public bool TryFetch(object obj, out T value, bool skipObjNullCheck = false)
...
+                    return (PropertyFetch?)typeof(PropertyFetch)
+                        .GetMethod(nameof(CreateInstantiated), BindingFlags.NonPublic | BindingFlags.Static)!
+                        .MakeGenericMethod(declaringType) // This is validated in the earlier call chain to be a reference type.
+                        .Invoke(null, new object[] { propertyInfo })!;
+                }
             }
         }
 
-        public virtual bool TryFetch(object obj, out T value)
+        public virtual bool TryFetch(object obj, out T? value)
```
eerhardt, line 133: Since we don't ever create instances of the concrete `PropertyFetch` class, we could just make the class and this method `abstract`.

```suggestion
        public abstract bool TryFetch(object obj, out T? value);
```

eerhardt (approved):

src/Shared/DiagnosticSourceInstrumentation/PropertyFetcher.cs
eerhardt: > If you have marked the class PropertyFetcher with RequiresUnreferencedCode attribute, do you still need to individually add this attribute to other nested types and methods in this class?

Yes, you need to add the attribute to nested types as well. The `RequiresUnreferencedCode` attribute on an outer class doesn't apply to nested classes. So they both are necessary.

> What's the guiding principle here for adding annotations?

If there are many methods in a class (especially a public class), and only a subset of the methods are incompatible, then just annotating the incompatible methods is preferred.

Annotating the whole class as `RequiresUnreferencedCode` is just a convenience. It means "this whole class can be considered incompatible with trimming. Don't warn inside this class, but just warn all the callers of this class (any constructor calls, and any static method invocations)." You can accomplish the same thing manually by adding the attribute to every method.

In this specific case, `PropertyFetch` only has 2 methods - both of which are incompatible. It wouldn't make sense to add the attribute to both methods when just applying the attribute to the whole class will suffice.

Another situation where using `RequiresUnreferencedCode` at the class level has an advantage is when you have an inheritance hierarchy - like a strategy pattern - and only one of the implementations are incompatible.

```C#

abstract class Abstraction { abstract void Method(); }

class SafeImpl { override void Method() {} }

class UnsafeImpl { override void Method() { // something not compatible with trimming} }
```

In this case, you can't annotate `UnsafeImpl.Method()` as `RequiresUnreferencedCode` because the base method isn't annotated as `RequiresUnreferencedCode`. The whole hierarchy needs to be annotated the same way. And you don't want to make `SafeImpl` raise warnings, because there are no issues with that implementation.

So to solve this, you annotate the whole `UnsafeImpl` class as `RequiresUnreferencedCode`. This warns at all the places that create an `UnsafeImpl` object. And it also resolves the warnings inside `UnsafeImpl.Method()` because the whole class is annotated.

src/Shared/DiagnosticSourceInstrumentation/PropertyFetcher.cs
```diff
@@ -85,54 +73,104 @@ public bool TryFetch(object obj, out T value, bool skipObjNullCheck = false)
...
+                {
+                    throw new NotSupportedException(
+                        $"Type: {declaringType.FullName} is a value type. PropertyFetcher can only operate on reference payload types.");
+                }
 
-        public virtual bool TryFetch(object obj, out T value)
-        {
-            value = default;
-            return false;
+                if (declaringType == typeof(object))
```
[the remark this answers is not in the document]
Yun-Ting: Update on this: 
The branch will never be executed. It's here for the AOT-compiler (net7.0) so the trimmer does not remove this code.
Will update the comment accordingly.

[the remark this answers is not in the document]
eerhardt: @agocke - can you confirm this?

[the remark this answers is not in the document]
agocke: In general, the bar is pretty high for me to backport AOT changes to .NET 7 since it was intentionally a limited release. But even more importantly, .NET 7 is going out of support this year, so we almost certainly won't be backporting anything.

test/OpenTelemetry.Tests/Instrumentation/PropertyFetcherTest.cs
Yun-Ting: > * A test to verify the throw behavior when a root value type is handed to property fetcher.

-> this was resolved in https://github.com/open-telemetry/opentelemetry-dotnet/pull/4675/commits/cabf49f5f6e87a2e401b0fdd72621d10d9e1ffb7

> * Does one of these tests fall into the `if (declaringType == typeof(object))` branch? Something like this (I think 😄)...
>   ```cs
>      private class PayloadTypeWithObjectType
>       {
>           public object Property { get; set; } = new DerivedType();
>       }
>   ```
-> Check out this discussion: 
https://github.com/open-telemetry/opentelemetry-dotnet/pull/4675#discussion_r1290495522
and the updated comment on this specific if branch: https://github.com/open-telemetry/opentelemetry-dotnet/pull/4675/commits/b7811147e5f9b5da3a77bcad38207601a5ccfed0

utpilla (approved):

src/Shared/DiagnosticSourceInstrumentation/PropertyFetcher.cs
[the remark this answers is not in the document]
vitek-karas: Just to close on this discussion - as is, the code is unreachable and only there to hint the compiler to keep the instantiation for ref types (canon).
I think in theory this can be changed to use any other ref type - so for example `if (declatingType == typeof(Type))` woudl work as well - I picked `Type` since it's a public ref type which has properties - so this is potentially reachable. But I think using object is more "self explanatory".
I agree with @agocke that backporting the fix to the AOT compiler to 7 would be a hard sell - it's likely not a simple backport (the compiler changed a lot between 7 and 8) and this code is the only place where we ran into this (this code in OTel actually made us fix the compiler in 8).

PR #4688, opened 2023-07-24
[AOT] Annotate APIs which call `IServiceCollection.TryAddSingleton` for trim compatibility

vitek-karas:

`IServiceCollection.TryAddSingleton<T>` has trim annotation on `T`, this needs to be propagated to the caller to make it trim compatible.

This change just propagates the annotation to the signature of the method calling the `TryAddSingleton`. The trim tooling will make sure that the annotation is fulfilled by the caller of the OTel public methods with this change.

Note that these changes don't affect the outcome of the `EnsureAotCompatibility` test. This is because that test uses .NET 7 AOT compiler which has a bug and it doesn't report the warning it should have in these cases. Running the same test as a Trim compatibility test instead (`/p:PublishAot=false /p:PublishTrimmed=true`) will show warnings for each of the change in this PR. The .NET 8 version of the AOT compiler is fixed and will also report these warnings, so we need to fix these for both trimming and AOT compatibility.

I'd be open to adding a `EnsureTrimCompatibility` test, but it will make CI slower as it takes similar amount of time to the AOT compatibility test which is already causing CI slowdowns. @Yun-Ting has been discussing possibility of moving the AOT test into its own CI leg, adding the trim test there would be much less disruptive.

@Yun-Ting @eerhardt

### DIFF
--- a/src/OpenTelemetry.Api.ProviderBuilderExtensions/Logs/OpenTelemetryDependencyInjectionLoggerProviderBuilderExtensions.cs
+++ b/src/OpenTelemetry.Api.ProviderBuilderExtensions/Logs/OpenTelemetryDependencyInjectionLoggerProviderBuilderExtensions.cs
@@ -14,6 +14,9 @@
 // limitations under the License.
 // </copyright>
 
+#if NET6_0_OR_GREATER
+using System.Diagnostics.CodeAnalysis;
+#endif
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection.Extensions;
 using OpenTelemetry.Internal;
@@ -35,7 +38,11 @@ public static class OpenTelemetryDependencyInjectionLoggerProviderBuilderExtensi
     /// <typeparam name="T">Instrumentation type.</typeparam>
     /// <param name="loggerProviderBuilder"><see cref="LoggerProviderBuilder"/>.</param>
     /// <returns>The supplied <see cref="LoggerProviderBuilder"/> for chaining.</returns>
-    public static LoggerProviderBuilder AddInstrumentation<T>(this LoggerProviderBuilder loggerProviderBuilder)
+    public static LoggerProviderBuilder AddInstrumentation<
+#if NET6_0_OR_GREATER
+        [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors)]
+#endif
+        T>(this LoggerProviderBuilder loggerProviderBuilder)
         where T : class
     {
         loggerProviderBuilder.ConfigureServices(services => services.TryAddSingleton<T>());
--- a/src/OpenTelemetry.Api.ProviderBuilderExtensions/Metrics/OpenTelemetryDependencyInjectionMeterProviderBuilderExtensions.cs
+++ b/src/OpenTelemetry.Api.ProviderBuilderExtensions/Metrics/OpenTelemetryDependencyInjectionMeterProviderBuilderExtensions.cs
@@ -14,6 +14,9 @@
 // limitations under the License.
 // </copyright>
 
+#if NET6_0_OR_GREATER
+using System.Diagnostics.CodeAnalysis;
+#endif
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection.Extensions;
 using OpenTelemetry.Internal;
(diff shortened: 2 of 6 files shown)


src/OpenTelemetry.Api.ProviderBuilderExtensions/Logs/OpenTelemetryDependencyInjectionLoggerProviderBuilderExtensions.cs
```diff
@@ -14,6 +14,9 @@
 // limitations under the License.
 // </copyright>
 
+#if NET6_0_OR_GREATER
```
eerhardt, line 17: We have UnconditionalSuppressMessageAttribute shim checked into the repo:

https://github.com/open-telemetry/opentelemetry-dotnet/blob/a808276efaf7e7a13420d993cea903cbff262996/src/Shared/Shims/UnconditionalSuppressMessageAttribute.cs#L30

Should we continue down that path here? Using both approaches feels wrong.

vitek-karas: The team decided to unify TFMs across the projects and include `net6.0` in all of them and then use ifdefs to only add the annotations to the `net6.0` versions.
So to make things consistent we should probably remove the shim for `UnconditionalSuppressMessageAttribute` and use ifdefs there as well.

@Yun-Ting as fyi

Yun-Ting: > The team decided to unify TFMs across the projects and include `net6.0` in all of them and then use ifdefs to only add the annotations to the `net6.0` versions. So to make things consistent we should probably remove the shim for `UnconditionalSuppressMessageAttribute` and use ifdefs there as well.
> 
I Agree. Please see the issue: https://github.com/open-telemetry/opentelemetry-dotnet/issues/4591

utpilla:
> @Yun-Ting has been discussing possibility of moving the AOT test into its own CI leg, adding the trim test there would be much less disruptive.

+1

vitek-karas:
I created https://github.com/open-telemetry/opentelemetry-dotnet/issues/4690 to track the work make a new CI leg for AOT/trim compatibility tests.

utpilla:
> The trim tooling will make sure that the annotation is fulfilled by the caller of the OTel public methods with this change.

@vitek-karas Could you explain a bit more as to what this means for the users of our APIs?

vitek-karas:
> > The trim tooling will make sure that the annotation is fulfilled by the caller of the OTel public methods with this change.

> @vitek-karas Could you explain a bit more as to what this means for the users of our APIs?

For users which don't trim/AOT their app there's no effect at all (the attribute is completely ignored by compiler and runtime).
For users which do trim/AOT the trimmer/AOT-compiler will check what is the type passed to the annotated `T`. If it's a specific type, the tool will preserve the necessary parts and there's nothing for the user to do - it just works. If it's another generic type (so somebody has a generic helper) they would get a warning and would need to propagate the annotation same way I did in this change. For example:

```C#
void RegisterInstrumentation<TInstrumentation>(LoggerProviderBuilder builder)
{
    builder.AddInstrumentation<TInstrumentation>(); // IL#### - the annotation on TInstrumentation doesn't match those required by AddInstrumentation
}
```

Fixed:

```C#
void RegisterInstrumentation<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructor)] TInstrumentation>(LoggerProviderBuilder builder)
{
    builder.AddInstrumentation<TInstrumentation>(); // no warning
}
```

This is a standard behavior of many other APIs which are annotated with the `DynamicallyAccessedMembers` attribute, so it's likely that users which do trim/AOT would already be familiar with how it works.

vitek-karas:
The test failure is a timed out AOT publish test. I ran it locally multiple times and it passes just fine. It also passed on Linux (in 1.5 minutes, on Windows it timed out after 4 minutes).

src/OpenTelemetry.Api.ProviderBuilderExtensions/Logs/OpenTelemetryDependencyInjectionLoggerProviderBuilderExtensions.cs
Yun-Ting: https://github.com/open-telemetry/opentelemetry-dotnet/pull/4696

utpilla (approved):

reyang (approved):

PR #4695, opened 2023-07-26
[AOT] Remove usage of Linq.Expressions and especially lambda compilation

vitek-karas:

In AOT `Expression.Lambda.Compile` method is implemented via an interpreter. So it does work, but it's relatively slow. Additionally it's not fully NativeAOT compatible yet.

There are two places where it's used in the source code:
* `ExceptionProcessor` uses it to access `Marshal.GetExceptionPointers` method which is not part of netstandard API. So in this case simply ifdef and on net6.0+ access the method directly (as it's visible in the net6.0 APIs) and fallback to the Expression.Lambda solution on down-level platforms. Since AOT/trimming is net6.0+ this fixes the problem.
* `ActivityInstrumentationHelper` uses it to access two setters on the `Activity` class which are internal. Using reflection and then `CreateDelegate` will provide basically the same performance (a delegate which directly calls the setter) without a need for `Expression.Lambda` - so the code is simpler. Since the reflection is statically analyzable (all types are known, the property names are known) this code is fully trim and AOT compatible.

There's no impact on warning count because the .NET 7 ASP.NET has Linq.Expression dependency as well. On .NET 8 though this removes all warnings from Linq.Expressions.

@Yun-Ting @eerhardt 


src/Shared/ActivityInstrumentationHelper.cs
```diff
@@ -28,19 +26,13 @@ internal static class ActivityInstrumentationHelper
 
     private static Action<Activity, ActivitySource> CreateActivitySourceSetter()
     {
-        ParameterExpression instance = Expression.Parameter(typeof(Activity), "instance");
-        ParameterExpression propertyValue = Expression.Parameter(typeof(ActivitySource), "propertyValue");
-        PropertyInfo sourcePropertyInfo = typeof(Activity).GetProperty("Source");
-        var body = Expression.Assign(Expression.Property(instance, sourcePropertyInfo), propertyValue);
-        return Expression.Lambda<Action<Activity, ActivitySource>>(body, instance, propertyValue).Compile();
+        return (Action<Activity, ActivitySource>)typeof(Activity).GetProperty("Source")
```
eerhardt, line 29: When I spoke to @tarekgh about this, and asked why these setters aren't exposed publicly (since OpenTelemetry needs to call them), he told me that these setters only need to be called on `< net7.0` runtimes. If you look at the callsites, you'll see this is true for 2 of the 3 callsites:

https://github.com/open-telemetry/opentelemetry-dotnet/blob/0343116e7781779975a1a0fe269be1778a484eeb/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs#L195-L198

https://github.com/open-telemetry/opentelemetry-dotnet/blob/0343116e7781779975a1a0fe269be1778a484eeb/src/OpenTelemetry.Instrumentation.Http/Implementation/HttpHandlerDiagnosticListener.cs#L161-L165

And in Grpc, it always calls it:

https://github.com/open-telemetry/opentelemetry-dotnet/blob/0343116e7781779975a1a0fe269be1778a484eeb/src/OpenTelemetry.Instrumentation.GrpcNetClient/Implementation/GrpcClientDiagnosticListener.cs#L124-L127

I think the correct thing to do here is to not compile this code on `net7.0+` TFMs, and then fix the 2 places above to have `#if !NET7_0_OR_GREATER` checks around them that don't already.

vitek-karas: @utpilla do you agree to change GRPC to not call these on net7.0+?

vitek-karas: There's an additional complication - it would require adding `net7.0` as a TFM to Http and Grcp instrumentation assemblies. Http currently uses a boolean which is determined by looking at version of the `System.Net.Http` assembly, the trimmer would not be able to figure this out. I don't think we have a runtime property which the trimmer will know a constant value of and can be used to determine the version.

tarekgh: CC @cijothomas @reyang

vishweshbankwar: FYI- For Http and ASP.NET Core we are subscribing to ActivitySource from the libraries e.g. https://github.com/open-telemetry/opentelemetry-dotnet/blob/0343116e7781779975a1a0fe269be1778a484eeb/src/OpenTelemetry.Instrumentation.Http/TracerProviderBuilderExtensions.cs#L107-L110. In Grpc, we have not tested/done that change, IF it works in Grpc we should be able to do that. All of them will rely on detecting the targetframework.

src/OpenTelemetry/Trace/ExceptionProcessor.cs
```diff
@@ -31,6 +33,11 @@ internal sealed class ExceptionProcessor : BaseProcessor<Activity>
 
     public ExceptionProcessor()
     {
+#if NET6_0_OR_GREATER
+        this.fnGetExceptionPointers = Marshal.GetExceptionPointers;
+#else
+        // When running on netstandard and similar the Marshal class it no part of the netstandard API
```
utpilla, line 39:
```suggestion
        // When running on netstandard or similar the Marshal class is not a part of the netstandard API
```

src/OpenTelemetry/Trace/ExceptionProcessor.cs
```diff
@@ -31,6 +33,11 @@ internal sealed class ExceptionProcessor : BaseProcessor<Activity>
...
+        // but it still most likely available in the underlying framework, so use reflection to get to it.
```
utpilla, line 40:
```suggestion
        // but it would still most likely be available in the underlying framework, so use reflection to retrieve it.
```

src/Shared/ActivityInstrumentationHelper.cs
utpilla: Agree with @vishweshbankwar. We need to test the Grpc instrumentation library to ensure that we can introduce such changes to it.

utpilla: > Additionally, Grpc package is currently netstandard2.0/netstandard2.1 only...

@Yun-Ting Did we miss adding the default set of TFMs for Grpc instrumentation?

Yun-Ting: @utpilla : we did it on-purpose: https://github.com/open-telemetry/opentelemetry-dotnet/pull/4595
Not updating TFM for this OpenTelemetry.Exporter.OpenTelemetryProtocol.Grpc as the project is subject to change. https://github.com/open-telemetry/opentelemetry-dotnet/issues/3421#issuecomment-1314630702

utpilla: We are referring to Grpc instrumentation project here: https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/src/OpenTelemetry.Instrumentation.GrpcNetClient

Yun-Ting: @utpilla understood.
I'll look at what @vishweshbankwar mentioned, whether it's possible to subscribe to the ActivitySource from the instrumentation library.
Not sure what the complexities would be now as unlike HttpNetClient, GrpcNetClient is created from the channel.

eerhardt: Note: I don't necessarily think we need to block this PR for this - especially if everyone is OK with the change from Linq.Expressions to Reflection .CreateDelegate.

I just wanted to point out that for `net7.0+` this code shouldn't be necessary at all (at least from my understanding from @tarekgh).

tarekgh: I want to clarify one thing here. If the code creating the `Activity` object is using ActivitySource, then it is correct you can get rid of this code which is setting the `Kind` or `Source` on the activity. If the Activity object is created using legacy way `new Activity(...)` then maybe this code still be need even running on .NET 7.0+.

eerhardt: > If the Activity object is created using legacy way new Activity(...) then maybe this code still be need even running on .NET 7.0+.

If that is a supported/valid scenario - why would we not expose the setters publicly in Activity?

tarekgh: `Source` means pointing to the `ActivitySource` that created this `Activity`. It doesn't make sense to have a setter for it. 
`Kind` is immutable value inside the activity that shouldn't be mutated after the Activity is created. No setter for it. This is per OpenTelemetry specs.

eerhardt: I'm not following your reasoning here. You're saying these properties shouldn't be set. Yet, here is code that is setting these properties.

And then above you say "then maybe this code still be need even running on .NET 7.0+.".  So it feels like a contradiction/inconsistency.

tarekgh: Older versions of aspnet and http client used to create the `Activity` objects the legacy way using `new Activity`. OpenTelemetry wanted to support these old versions of such platforms which needed to set `Kind` properly on such activity. This was the need for reflection. Aspnet and http client then moved to create the Activity object using `ActivitySource` which set the `Kind` correctly on the created `Activity` object so there was no need to do anything more with such objects or use the reflection. 

I don't know gRPC scenarios. If it is same as aspnet then there is no need to call reflection on .NET 7.0+.

@cijothomas you were the one implementing that as I recall if you have more input here.

eerhardt: > I don't know gRPC scenarios.

It looks like it isn't following this guidance of using `ActivitySource.CreateActivity`. 😢 For either it's own Activity:

https://github.com/grpc/grpc-dotnet/blob/0ab3ada34aa76740b436b3b1d320f2f406f82f50/src/Grpc.Net.Client/Internal/GrpcCall.cs#L823

nor its copy of the HTTP `System.Net.Http.HttpRequestOut` Activity:

https://github.com/grpc/grpc-dotnet/blob/0ab3ada34aa76740b436b3b1d320f2f406f82f50/src/Shared/TelemetryHeaderHandler.cs#L70

Maybe someone should log an issue on the gRPC client to move away from the "legacy way `new Activity(...)`" and use an ActivitySource. Then OpenTelemetry wouldn't need to use reflection here at all.

cc @JamesNK

utpilla (approved):
Looks good.

Things to do outside of this PR:
- Update targets for GrpcNetClient instrumentation library
- Check if GrpcNetClient instrumentation library can be updated to listen to ActivitySource on net7+ targets

src/Shared/ActivityInstrumentationHelper.cs
JamesNK: grpc-dotnet was written for .NET Core 3, and `ActivitySource` didn't exist then. grpc-dotnet still targets .NET Core 3 + .NET 5, but I'm going to drop the unsupported targets for .NET 8.

Grpc.Net.Client targets netstandard2.0, but ActivitySource is in System.Diagnostics.DiagnosticSource.

src/OpenTelemetry/Trace/ExceptionProcessor.cs
```diff
@@ -31,6 +33,11 @@ internal sealed class ExceptionProcessor : BaseProcessor<Activity>
 
     public ExceptionProcessor()
     {
+#if NET6_0_OR_GREATER
```
utpilla, line 36: @vitek-karas Could we use the public API for net462 as well?

https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.marshal.getexceptionpointers?view=netframework-4.6.2

src/Shared/ActivityInstrumentationHelper.cs
JamesNK: Issue: https://github.com/grpc/grpc-dotnet/issues/2180#issuecomment-1654809816

vitek-karas:
I don't understand the test failure - it passes locally just fine.

vitek-karas:
> Why doesn't this change decrease the AOT warning count?

It does on .NET 8. But the test runs on .NET 7 and in .NET 7 ASP.NET uses Linq.Expression is certain places which pull it in -> causing the same warnings.

Yun-Ting (approved):

utpilla (approved):

src/Shared/ActivityInstrumentationHelper.cs
JamesNK: PR: https://github.com/grpc/grpc-dotnet/pull/2244

Issue #4640, opened 2023-07-06
Use file scoped namespaces


utpilla:
Update the code files in the repo to use file scoped namespaces: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-10.0/file-scoped-namespaces

This should be done in smaller PRs each preferably targeting three to four `src` projects and their corresponding `test` projects in one PR.

utpilla:
@kyle-lt Thanks for offering your help! Feel free to send the PRs.

kyle-lt:
Great, thanks @utpilla , can you pls assign me and I'll start to chip away?

kyle-lt:
PR #4654 only covered the core OpenTelemetry src and test project, so we may want to keep this issue open until all of the projects are covered. I’ll knock that out in a few more PRs. --Kyle TullyOn Jul 13, 2023, at 5:50 PM, Utkarsh Umesan Pillai ***@***.***> wrote:﻿
Closed #4640 as completed via #4654.

utpilla:
@kyle-lt This issue got closed because the PR description had "Fixes https://github.com/open-telemetry/opentelemetry-dotnet/issues/4640" in it. Could you use another term instead of "Fixes" such as "Towards https://github.com/open-telemetry/opentelemetry-dotnet/issues/4640" in the PRs going forward.

kyle-lt:
I changed it to “Partially” Fixes…But, clearly that didn’t do the trick, apologies for the nuisance. I’ll make sure to use your suggested wording in the subsequent PRs. Thanks!--Kyle TullyOn Jul 13, 2023, at 6:12 PM, Utkarsh Umesan Pillai ***@***.***> wrote:﻿
@kyle-lt This issue got closed because the PR description had "Fixes #4640" in it. Could you use another term instead of "Fixes" such as "Towards #4640" in the PRs going forward.

jarosal:
@utpilla first PR! #4680

jarosal:
Hi @utpilla,
I've created PRs for all of the remaining projects/folders in this repo. After they are merged I think you should merge #4716 so all of the future files use this convention.

utpilla:
Thank you @kyle-lt and @jarosal for working on this!

PR #4714, opened 2023-07-27
Update Shared to use filescoped namespaces

jarosal:
Towards #4640 

## Changes

Update Shared to use file scoped namespaces.

### DIFF
--- a/src/Shared/Guard.cs
+++ b/src/Shared/Guard.cs
@@ -55,161 +55,160 @@ namespace System.Diagnostics.CodeAnalysis
 }
 #endif
 
-namespace OpenTelemetry.Internal
+namespace OpenTelemetry.Internal;
+
+/// <summary>
+/// Methods for guarding against exception throwing values.
+/// </summary>
+internal static class Guard
 {
     /// <summary>
-    /// Methods for guarding against exception throwing values.
+    /// Throw an exception if the value is null.
     /// </summary>
-    internal static class Guard
+    /// <param name="value">The value to check.</param>
+    /// <param name="paramName">The parameter name to use in the thrown exception.</param>
+    [DebuggerHidden]
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static void ThrowIfNull([NotNull] object? value, [CallerArgumentExpression("value")] string? paramName = null)
     {
-        /// <summary>
-        /// Throw an exception if the value is null.
-        /// </summary>
-        /// <param name="value">The value to check.</param>
-        /// <param name="paramName">The parameter name to use in the thrown exception.</param>
-        [DebuggerHidden]
-        [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static void ThrowIfNull([NotNull] object? value, [CallerArgumentExpression("value")] string? paramName = null)
+        if (value is null)
         {
-            if (value is null)
-            {
-                throw new ArgumentNullException(paramName, "Must not be null");
-            }
+            throw new ArgumentNullException(paramName, "Must not be null");
         }
+    }
 
-        /// <summary>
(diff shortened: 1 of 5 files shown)


reyang (approved):

src/Shared/Guard.cs
```diff
@@ -55,161 +55,160 @@ internal sealed class NotNullAttribute : Attribute
 }
 #endif
 
-namespace OpenTelemetry.Internal
+namespace OpenTelemetry.Internal;
```
utpilla, line 58: This is erroring out with: `error CS8956: File-scoped namespace must precede all other members in a file.` 

I think we should move `CallerArgumentExpressionAttribute` and `NotNullAttribute` to their own classes.

jarosal: I will sort this out in an hour or two

jarosal: @utpilla I'm having an issue with moving those attributes to their own classes as this is code is kinda unusual, mind guiding me?

utpilla: You could add a new file named `CallerArgumentExpressionAttribute.cs` under `Shared/Shims/` and move this code into that file:

```csharp
#if !NET6_0_OR_GREATER
namespace System.Runtime.CompilerServices
{
    /// <summary>Allows capturing of the expressions passed to a method.</summary>
    [AttributeUsage(AttributeTargets.Parameter, AllowMultiple = false, Inherited = false)]
    internal sealed class CallerArgumentExpressionAttribute : Attribute
    {
        public CallerArgumentExpressionAttribute(string parameterName)
        {
            this.ParameterName = parameterName;
        }

        public string ParameterName { get; }
    }
}
#endif
```

jarosal: Yeah but will Guard.cs pick up those new classes same as when they are in one file?

utpilla: Similarly, you could create a file named `NotNullAttribute.cs` under `Shared/Shims/` and move this code into that file:

```csharp
#if !NET6_0_OR_GREATER && !NETSTANDARD2_1_OR_GREATER
namespace System.Diagnostics.CodeAnalysis
{
    /// <summary>Specifies that an output is not <see langword="null"/> even if
    /// the corresponding type allows it. Specifies that an input argument was
    /// not <see langword="null"/> when the call returns.</summary>
    [AttributeUsage(AttributeTargets.Property | AttributeTargets.Field | AttributeTargets.Parameter | AttributeTargets.ReturnValue, Inherited = false)]
    internal sealed class NotNullAttribute : Attribute
    {
    }
}
#endif
```

utpilla: > Yeah but will Guard.cs pick up those new classes same as when they are in one file?

`Guard` class can access them because it imports their namespaces:

```csharp
using System.Diagnostics.CodeAnalysis;
```

jarosal: Doesn't seem to work this way, see the test results after my last push

Kielek: @jarosal, you should probably find all occurrences if Guard.cs file and include new files in the similar way.

See https://github.com/open-telemetry/opentelemetry-dotnet/blob/8f38400a142df32259bd78d3cbeea8fd3c8e867a/src/OpenTelemetry.Exporter.Console/OpenTelemetry.Exporter.Console.csproj#L33

utpilla: @jarosal On second thoughts, could you suppress the warning for using file-scoped namespace in this class?

utpilla:
Addressed this in https://github.com/open-telemetry/opentelemetry-dotnet/pull/4716/commits/04854c7f73ed55c81372ebaf4980480aff99e54a

Issue #4690, opened 2023-07-25
Create new CI leg for AOT/trim compatibility tests


vitek-karas:
The recently added test https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/test/OpenTelemetry.AotCompatibility.Tests is relatively slow, it can take several minutes to run (depending on the hardware). Since currently it's part of the main test CI leg it makes the entire CI run much longer than before. It would be better to create a new CI leg to run just this test.

Once this is separated it would be good to also add a trimming test. Basically the same test but run the publish with `PublishAot=false` and `PublishTrimmed=true` instead and maybe target net6.0 in that case. This would validate that OTel can be used in trimmed apps (non-AOT) as well.

Eventually it would also be great if this leg could include net8.0 sdk (maybe RC1 once it's available) and run the AOT test targeting net8.0 instead. This is because:
* There are known bugs in the net7.0 AOT compiler which are fixed in net8.0 and which have effect on OTel (for example: https://github.com/open-telemetry/opentelemetry-dotnet/pull/4688)
* OTel includes ASP.NET dependencies which were only made AOT compatible in net8.0

@Yun-Ting

Yun-Ting:
@open-telemetry/dotnet-approvers this issue can be closed.

PR #4717, opened 2023-07-27
[AOT] Added new CI for AOTCompatibility tests and removed unit test

Yun-Ting:
Towards https://github.com/open-telemetry/opentelemetry-dotnet/issues/4690
cc: @vitek-karas @eerhardt 

followed (nullable enabled, static analysis, etc.)


.github/workflows/ci-aot.yml
```diff
@@ -0,0 +1,33 @@
...
+    paths-ignore:
+    - '**.md'
+
+jobs:
+  aot-test:
+    strategy:
+      fail-fast: false  # ensures the entire test matrix is run, even if one permutation fails
+      matrix:
+        os: [ windows-latest ]
+        version: [ net6.0, net7.0 ]
```
utpilla, line 19: Why are we adding `net6.0` now?

build/assert-publish-aot-testApp-warning-count.ps1
```diff
@@ -0,0 +1,36 @@
+Set-ExecutionPolicy -Scope Process -ExecutionPolicy Bypass
```
utpilla, line 1: Do you need this?

build/assert-publish-aot-testApp-warning-count.ps1
```diff
@@ -0,0 +1,36 @@
...
+if ($trimmingWarningCount -ne $expectedTrimmingWarningCount ) 
+{
+    $testPassed = 1
+    Write-Host "trimmingWarningCount: ", $trimmingWarningCount, "is not as expected. Expected count is:", $expectedTrimmingWarningCount
+}
+
+if ($aotWarningCount -ne $expectedaotWarningCount ) 
+{
+    $testPassed = 1
+    Write-Host "aotWarningCount: ", $aotWarningCount, "is not as expected. Expected count is:", $expectedaotWarningCount
```
utpilla, line 33:
```suggestion
    Write-Host "aotWarningCount: ", $aotWarningCount, " is not as expected. Expected count is: ", $expectedaotWarningCount 

```

build/assert-publish-aot-testApp-warning-count.ps1
```diff
@@ -0,0 +1,36 @@
...
+$expectedaotWarningCount = 19
+
+Write-Host "trimmingWarningCount: ", $trimmingWarningCount
+Write-Host "aotWarningCount: ", $aotWarningCount
+
+$testPassed = 0
+if ($trimmingWarningCount -ne $expectedTrimmingWarningCount ) 
+{
+    $testPassed = 1
+    Write-Host "trimmingWarningCount: ", $trimmingWarningCount, "is not as expected. Expected count is:", $expectedTrimmingWarningCount
```
utpilla, line 27:
```suggestion
    Write-Host "trimmingWarningCount: ", $trimmingWarningCount, " is not as expected. Expected count is: ", $expectedTrimmingWarningCount

```

.github/workflows/ci-aot.yml
```diff
@@ -0,0 +1,33 @@
...
+
+    runs-on: ${{ matrix.os }}
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0 # fetching all
+
+    # powershell does not support AOT compiler yet
```
vitek-karas, line 27: We should figure out what the problem is - I don't believe that powershell doesn't support AOT compiler, the problem is very likely elsewhere.

build/assert-publish-aot-testApp-warning-count.ps1
```diff
@@ -0,0 +1,36 @@
...
+    { 
+        $trimmingWarningCount += 1
+    } elseif ($line -like "*AOT analysis warning IL*") 
+    {
+        $aotWarningCount += 1
+    }
+}
+
+$expectedTrimmingWarningCount = 11
+$expectedaotWarningCount = 19
```
vitek-karas, line 18: I don't think it makes much sense to track the different warnings separately. Additionally we could potentially miss other warnings (like single-file related warnings which also prevent AOT apps from working).
We should simply count all warnings - the publish of the test app should be warning free.

.github/workflows/ci-aot.yml
vitek-karas: I don't think we need it now - AOT will run on 7.0 only anyway.
As a subsequent change adding a trim test should target 6.0 though (since that's where we introduced trimming), but we might run into similar issues as with AOT on 7, that not all of the dependent libraries are correctly annotated for trimming in 6 (specifically ASP.NET comes in mind).

build/assert-publish-aot-testApp-warning-count.ps1
eerhardt: > the publish of the test app should be warning free.

Unfortunately, the publish test app also includes the warnings from the "product" assemblies (as far as I can tell).

.github/workflows/ci-aot.yml
```diff
@@ -0,0 +1,33 @@
...
+    - name: Publich AOT testApp
+      run: dotnet publish ./test/OpenTelemetry.AotCompatibility.TestApp/OpenTelemetry.AotCompatibility.TestApp.csproj --self-contained -nodeReuse:false /p:UseSharedCompilation=false > ./build/publishAotTestAppOutput.log
```
eerhardt, line 29: We should be able to remove `--self-contained` in this command. And instead make the .csproj so you can just `dotnet publish` it. For .NET 7, you need to put:

```xml
<SelfContained>true</SelfContained>
```

into the .csproj.

When we move to the .NET 8 SDK, we can remove that line because `<PublishAot>` implies `<SelfContained>` in the new version.

.github/workflows/ci-aot.yml
utpilla: > As a subsequent change adding a trim test should target 6.0 though (since that's where we introduced trimming)

For now, could we just test both trimming and AOT on net7.0? Or would trimming be better tested on net6.0?

build/assert-publish-aot-testApp-warning-count.ps1
utpilla: It looks like the commonly followed convention is to add a space after the colon:

https://github.com/open-telemetry/opentelemetry-dotnet/actions/runs/5697194997/job/15443511912?pr=4717#step:5:2306

.github/workflows/ci-aot.yml
Yun-Ting: @eerhardt this is what I saw the other day from PS.
image

@vitek-karas mentioned that it might be because of publishing the app twice in a row. But I'm not able to repro this behavior now.

.github/workflows/ci-aot.yml
vitek-karas: I think it's a great start to run on .NET 7 always - in fact once we can I would even say run on .NET 8 only.
Technically it would be nice to support trimming at least for a subset of OTel (probably without ASP.NET) on .NET 6, but it's much lower priority, so we could easily wait for feedback on that one.

build/assert-publish-aot-testApp-warning-count.ps1
vitek-karas: If that's the case - could we maybe detect all "IL####" warnings instead - that's a good approximation of anything even related to AOT.

build/assert-publish-aot-testApp-warning-count.ps1
```diff
@@ -0,0 +1,24 @@
...
+}
+
+Write-Host "Actual warning count is:", $actualWarningCount
+$expectedWarningCount = 28
+
+$testPassed = 0
+if ($actualWarningCount -ne $expectedWarningCount)
+{
+    $testPassed = 1
+    Write-Host "Actual warning count:", actualWarningCount, "is not as expected. Expected warning count is:", $expectedWarningCount
```
xiang17, line 21: If the warning count doesn't match, do we need to reveal some details about what's the actual output or what's the mismatch?

.github/workflows/ci-aot.yml
```diff
@@ -0,0 +1,29 @@
+name: Publish AOTCompatibility testApp
```
eerhardt, line 1: How does this action get hooked up into the CI system? Do we have to wait for it to merge before we can run it? Or is it already running on this PR?

build/assert-publish-aot-testApp-warning-count.ps1
Yun-Ting: fixed format!

build/assert-publish-aot-testApp-warning-count.ps1
Yun-Ting: nope and removed

.github/workflows/ci-aot.yml
Yun-Ting: removed net6.0 from the target for now as suggested

.github/workflows/ci-aot.yml
reyang: It's already running and can be seen from the "Checks" tab
 
image

reyang (approved):

.github/workflows/ci-aot.yml
Yun-Ting: It's already running with this PR: https://github.com/open-telemetry/opentelemetry-dotnet/actions/runs/5719889408/job/15498727666?pr=4717
This workflow will be triggered automatically on push: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#on

https://github.com/open-telemetry/opentelemetry-dotnet/pull/4717/files#diff-65a128c6ce6f6956d1ff4c87ff6d27540e6faa4bcd3a2e59343a6947d8e83c15:~:text=on%3A,push%3A

build/assert-publish-aot-testApp-warning-count.ps1
utpilla: That's a good point. We should print out the warnings in any case.

utpilla (approved):

build/assert-publish-aot-testApp-warning-count.ps1
utpilla: @Yun-Ting Consider doing this in a follow-up PR.

Yun-Ting: https://github.com/open-telemetry/opentelemetry-dotnet/pull/4734

PR #4721, opened 2023-07-28
[AOT] Trim compatibility of OpenTelemetryLoggerOptions with configuration binder

vitek-karas:
Configuration binder is currently (.NET 7) incompatible with trimming and AOT. It's essentially a deserializer which uses reflection to access property setters on the options type.

There's no way to make it verifiably correct in .NET 7, but for simple cases it can be validated externally. `OpenTelemetryLoggerOptions` currently only contains boolean properties, so there are actually no incompatibilities. See the comment in the code for detailed explanation.

The change suppresses the warnings produced by the usage of configuration binder. It also adds a test which verifies that the options type only contains primitive properties. If there will be changes to the options type which would make it incompatible, the test should fail.

In .NET 8 there will be a source generator for configuration binder which should fix these cases. But currently the specific use case in OTel is not planned to be covered yet by the source generator. It might be possible to rewrite the usage here to make use of the source generator, but it would mean retargeting to `net8.0` and it would not help with trimming in `net6.0` or AOT in `net7.0` anyway. Might be worth revisiting once OTel retargets to `net8.0` in the future.

@Yun-Ting @eerhardt 


src/OpenTelemetry/Logs/ILogger/OpenTelemetryLoggingExtensions.cs
```diff
@@ -79,6 +82,25 @@ public static class OpenTelemetryLoggingExtensions
...
+        // But it would not work recursively into complex property values;
+        // This should be fully fixed with the introduction of Configuration binder source generator in .NET 8
+        // and then there should be a way to do this without any warnings.
+        // The correctness of these suppressions is verified by a test which validates that all properties of OpenTelemetryLoggerOptions
+        // are of a primitive type.
+#if NET6_0_OR_GREATER
+        [UnconditionalSuppressMessage("Trimming", "IL2026", Justification = "OpenTelemetryLoggerOptions only contains boolean properties.")]
+        [UnconditionalSuppressMessage("AOT", "IL3050", Justification = "OpenTelemetryLoggerOptions only contains boolean properties.")]
+#endif
+        static void RegisterLoggerProviderOptions(ILoggingBuilder builder)
```
vitek-karas, line 100: The reason to separate this into a local function is so that the suppressions can be applied to just the one statement. Attributes can only apply to entire methods, so putting these on the outer method would potentially suppress these warnings everywhere in the method - using a local function allows precise targeting of the suppressions.

eerhardt (approved):
Looks good!

Agreed that the best long-term solution here is the source generator. But that will require a dependency on the 8.0 version of Configuration.Binder, which hasn't shipped yet.

test/OpenTelemetry.Tests/Logs/OpenTelemetryLoggingExtensionsTests.cs
```diff
@@ -114,4 +115,18 @@ void OptionsCallback(OpenTelemetryLoggerOptions options)
...
+
+    // This test validates that the OpenTelemetryLoggerOptions contains only primitive type properties.
+    // This is necessary to ensure trim correctness since that class is effectively deserialized from
+    // configuration. The top level properties are ensured via annotation on the RegisterProviderOptions API
+    // but if there was any complex type property, members of the complex type would not be preserved
+    // and could lead to incompatibilities with trimming.
+    [Fact]
+    public void TestTrimmingCorrectnessOfOpenTelemetryLoggerOptions()
+    {
+        foreach (var prop in typeof(OpenTelemetryLoggerOptions).GetProperties(BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance))
```
Yun-Ting, line 127: just curious - why test only these 3 BindingFlags?

src/OpenTelemetry/Logs/ILogger/OpenTelemetryLoggingExtensions.cs
```diff
@@ -79,6 +82,25 @@ public static class OpenTelemetryLoggingExtensions
                     disposeProvider: false)));
 
         return builder;
+
+        // The warning here is about the fact that the OpenTelemetryLoggerOptions will be bound to configuration using ConfigurationBinder
+        // That uses reflection a lot - so if any of the properties on that class were complex types reflection would be used on them
+        // and nothing could guarantee its correctness.
+        // Since currently this class only contains primitive (boolean) properties this is OK. The top level properties are kept
+        // because the first generic argument is annotated with DAM(All) so will preserve everything on the OpenTelemetryLoggerOptions.
```
Yun-Ting, line 90: For my learning - what does DAM stands for here?

Yun-Ting (approved):

Yun-Ting:
For tracking: https://github.com/open-telemetry/opentelemetry-dotnet/issues/4725

src/OpenTelemetry/Logs/ILogger/OpenTelemetryLoggingExtensions.cs
reyang: https://github.com/search?q=repo%3Adotnet%2Fruntime%20DAM&type=code

src/OpenTelemetry/Logs/ILogger/OpenTelemetryLoggingExtensions.cs
utpilla, line 100: This code only needs the `ServiceCollection` of the `builder`. Could you update the method signature to take in `ServiceCollection` instead?

src/OpenTelemetry/Logs/ILogger/OpenTelemetryLoggingExtensions.cs
```diff
@@ -79,6 +82,25 @@ public static class OpenTelemetryLoggingExtensions
                     disposeProvider: false)));
 
         return builder;
+
+        // The warning here is about the fact that the OpenTelemetryLoggerOptions will be bound to configuration using ConfigurationBinder
+        // That uses reflection a lot - so if any of the properties on that class were complex types reflection would be used on them
+        // and nothing could guarantee its correctness.
+        // Since currently this class only contains primitive (boolean) properties this is OK. The top level properties are kept
```
utpilla, line 89: It would be good to add some more details here. Something like this:

The top-level properties of `OpenTelemetryLoggerOptions` are preserved to provide reflection access to the `ConfigurationBinder`. This is due to the fact that the first generic argument of `RegisterProviderOptions` in the runtime code is annotated with DAM(All).

src/OpenTelemetry/Logs/ILogger/OpenTelemetryLoggingExtensions.cs
```diff
@@ -79,6 +82,25 @@ public static class OpenTelemetryLoggingExtensions
...
+        // because the first generic argument is annotated with DAM(All) so will preserve everything on the OpenTelemetryLoggerOptions.
+        // But it would not work recursively into complex property values;
+        // This should be fully fixed with the introduction of Configuration binder source generator in .NET 8
+        // and then there should be a way to do this without any warnings.
+        // The correctness of these suppressions is verified by a test which validates that all properties of OpenTelemetryLoggerOptions
+        // are of a primitive type.
+#if NET6_0_OR_GREATER
+        [UnconditionalSuppressMessage("Trimming", "IL2026", Justification = "OpenTelemetryLoggerOptions only contains boolean properties.")]
```
utpilla, line 97: Could we update the message to say ""OpenTelemetryLoggerOptions only contains primitive typed properties." instead?

The current wording gives the impression that `boolean` type is some special-case and that we should not add any non-boolean property to `OpenTelemetryOptions` in future to remain AOT-safe.

test/OpenTelemetry.Tests/Logs/OpenTelemetryLoggingExtensionsTests.cs
vitek-karas: These 3 flags basically mean "everything but static properties". Configuration binder doesn't bind static members, so we can ignore those.

src/OpenTelemetry/Logs/ILogger/OpenTelemetryLoggingExtensions.cs
vitek-karas: Sorry - I'm so used to shortening `DynamicallyAccessedMembers` as `DAM` (the full attribute name is just too long :-)) - I'll fix this.

src/OpenTelemetry/Logs/ILogger/OpenTelemetryLoggingExtensions.cs
vitek-karas: I changed the wording on this to hopefully make it clearer.

src/OpenTelemetry/Logs/ILogger/OpenTelemetryLoggingExtensions.cs
```diff
@@ -79,6 +82,26 @@ public static class OpenTelemetryLoggingExtensions
...
+        // Since currently this class only contains primitive properties this is OK. The top level properties are kept
+        // because the first generic argument of RegisterProviderOptions below is annotated with
+        // DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.All) so it will preserve everything on the OpenTelemetryLoggerOptions.
+        // But it would not work recursively into complex property values;
+        // This should be fully fixed with the introduction of Configuration binder source generator in .NET 8
+        // and then there should be a way to do this without any warnings.
+        // The correctness of these suppressions is verified by a test which validates that all properties of OpenTelemetryLoggerOptions
+        // are of a primitive type.
+#if NET6_0_OR_GREATER
+        [UnconditionalSuppressMessage("Trimming", "IL2026", Justification = "OpenTelemetryLoggerOptions only contains primitive properties.")]
```
utpilla, line 98:
```suggestion
        [UnconditionalSuppressMessage("Trimming", "IL2026", Justification = "OpenTelemetryLoggerOptions contains only primitive properties.")]
```

src/OpenTelemetry/Logs/ILogger/OpenTelemetryLoggingExtensions.cs
```diff
@@ -79,6 +82,26 @@ public static class OpenTelemetryLoggingExtensions
...
+        [UnconditionalSuppressMessage("AOT", "IL3050", Justification = "OpenTelemetryLoggerOptions only contains primitive properties.")]
```
utpilla, line 99:
```suggestion
        [UnconditionalSuppressMessage("AOT", "IL3050", Justification = "OpenTelemetryLoggerOptions contains only primitive properties.")]
```

utpilla (approved):

PR #4731, opened 2023-07-29
Otlp grpc tls certificates

### LINKED ISSUE #2009: Add OTLP Exporter TLS/mTLS configuration options

dhhoang:
Fixes #2009
Design discussion issue #2009

## Changes

Implement the configurations according to specs:
* Certificate file. `OTEL_EXPORTER_OTLP_CERTIFICATE`
* Client key file. `OTEL_EXPORTER_OTLP_CLIENT_KEY`
* Client certificate file. `OTEL_EXPORTER_OTLP_CLIENT_CERTIFICATE`

## Notes
Some known issues:
* When using `HTTP/Protobuf` protocol, the options only work for `NET6_0_OR_GREATER` targets. Older .NET lacks the required APIs to load PEM certificates. Implementing this would require essentially porting those APIs back from .NET 6. Given that `HttpClientFactory` can be used for `HTTP/Protobuf` I don't think it's worth the effort.
* **Testing**: I'm not currently satisfied especially regarding `GRPC flow`. I'd love to have some end-to-end tests where the test case can communicate with a GRPC server to verify the mTLS functionalities. That's the only way to know whether the code works, because some of the options here passed to native C bindings (for GRPC in older .NET) and hence not possible to do any meaningful unit test. I'd like to hear recommendations on how these kinds of test could be done.


test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests.csproj
```diff
@@ -22,6 +22,12 @@
     <PackageReference Include="Microsoft.Extensions.Http" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Content Include="Certificates\**">
```
alanwest, line 26: In #4732 I tried to avoid adding test certs/keys to the repo and generated them on the fly. See: https://github.com/open-telemetry/opentelemetry-dotnet/pull/4732/files#diff-5b358d0b47462dc9c3919270476736612be64b9db3362961d5df3d8b6defc257R4-R6.

The tests you have in this PR aren't run in docker of course, but I'm wondering if we can do something similar to avoid adding certs and keys.

alanwest:
> * I'd love to have some end-to-end tests where the test case can communicate with a GRPC server to verify the mTLS functionalities.

💯 yes this would be great!

Once we can get #4732 landed, then I think you could expand on it and test mTLS. You could probably add an additional OTLP receiver with the necessary mTLS config. See: https://github.com/open-telemetry/opentelemetry-dotnet/pull/4732/files#diff-836987be2117c46a9d22eb1a86fe30cee3eee98299b46abceb7fcc6e9e84e1c9R14

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpExporterOptions.cs
```diff
@@ -157,6 +176,48 @@ public Uri Endpoint
...
+#if NET6_0_OR_GREATER
+    /// <summary>
+    /// Gets or sets the trusted certificate to use when verifying a server's TLS credentials.
+    /// </summary>
+#else
+    /// <summary>
+    /// Gets or sets the trusted certificate to use when verifying a server's TLS credentials.
+    /// </summary>
+    /// <remarks>
+    /// This option does not affect <see cref="OtlpExportProtocol.HttpProtobuf"/> protocol.
```
alanwest, line 188: For now, maybe we should consider throwing a PlatformNotSupported exception. Ideally we avoid taking a dependency on something like BouncyCastle as you note https://github.com/open-telemetry/opentelemetry-dotnet/issues/2009#issuecomment-1657221586

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpExporterOptions.cs
```diff
@@ -84,6 +88,21 @@ public OtlpExporterOptions()
             this.TimeoutMilliseconds = timeout;
         }
 
+        if (configuration.TryGetStringValue(ClientCertificateFileEnvVarName, out var clientCertificateFile))
```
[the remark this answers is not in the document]
dhhoang: After trying this I think a `TryGetFilePath()` method doesn't make much sense here. The `TryGet...` methods simply check if the value is syntactically valid and not whether the resource actually exists (e.g. URI).   
If the certificate file with specified path doesn't exist then `TryGetFilePath()` would return `false` and the option would be set to `null` and treated as if no option has been set by the developer.
I think from a developer PoV, if the file doesn't exist I'd prefer to be notified with a `FileNotFoundException`.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpExporterOptions.cs
dhhoang: Yes, after much fruitless effort I've concluded that it's not worth implementing the support for older .NET, especially when much of the functionalities can be manually wired using `HttpClientFactory`.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpExporterOptions.cs
```diff
@@ -203,6 +264,47 @@ internal static void RegisterOtlpExporterOptionsFactory(IServiceCollection servi
...
     }
 
+    internal HttpClientHandler GetDefaultHttpMessageHandler()
+    {
+        var handler = new HttpClientHandler();
+
+#if NET6_0_OR_GREATER
+        if (this.CertificateFile != null)
+        {
+            var trustedCA = LoadCertificateFromPemFile(this.CertificateFile, null);
```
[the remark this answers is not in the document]
dhhoang: Done in the latest commit.

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests.csproj
alanwest: Hey @dhhoang, apologies, I haven't had a chance yet to circle back and give this PR another pass. I will look at it more closely again tomorrow.

Though, quick thought about generating these certs, another idea may be to perform these tests using the integration test framework I landed in #4732. It's a minor bummer to have the tests you have so far depend on docker-compose to run them, but I figure you'll want to expand the integration tests anyways to test mTLS end to end.

alanwest: It just occurred to be a readme for how to run the integration test locally might be useful. See #4741.

dhhoang: Giving this a try right now, I'll mark the PR as draft in the mean time.

dhhoang:
Finally got the time to work on this! I've finished the test cases:
* Added scripts to generate test certificates for unit tests on Windows (pwsh) and Unix (bash)
* Added the logic needed for integration test (Linux/Docker):
  * Trusted certificates with TLS/mTLS.
  * Untrusted certificates with custom trust store.
  * Negative validation for cases above.
* Manual testing on Windows for `net6.0` `net7.0` and `net462` (i.e. gRPC using C binding).

Marking this PR as ready for review.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpExporterOptions.cs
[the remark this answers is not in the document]
alanwest: Syntactic validation is what I had in mind - using something like `Path.GetInvalidPathChars` or something. Though file paths may be tricky to validate syntactically in a meaningful/helpful way, so I'm fine with not doing anything more in this PR.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpExporterOptions.cs
```diff
@@ -205,6 +268,54 @@ internal static void RegisterOtlpExporterOptionsFactory(IServiceCollection servi
...
+#if NET6_0_OR_GREATER
+            var trustedCA = LoadCertificateFromPemFile(this.CertificateFile, null);
+
+            handler.ServerCertificateCustomValidationCallback = (message, serverCert, chain, policyErrs) =>
+            {
+                chain.ChainPolicy.TrustMode = X509ChainTrustMode.CustomRootTrust;
+                chain.ChainPolicy.CustomTrustStore.Add(trustedCA);
+                var buildResult = chain.Build(serverCert);
+                return buildResult;
+            };
```
alanwest, line 286: I'm not an expert on these things, so still need to do some of my own research to understand if this is sufficient.

One thing that leaves me a bit unsure here is that `policyErrs` is ignored. My understanding is the "out-of-the-box" validations would be reflected in `policyErrs` (e.g., is the cert signed by a trusted CA, is the cert expired, etc). I suspect we want the custom validation here to be in addition to the out-of-the-box validations.

dhhoang: You're correct that `policyErrs` would contain the error that the "default" validation method. The default validator would by default trust the certificates in the root trust store (aka certificates "installed" on the machine).  
If user is manually trusting a certificate, it means the certificate might not be in the machine root store and we need to at least ignore that error.

The other validation criteria (e.g. expiration, revocation) is encoded in the [`X509Chain` parameter](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509chain?view=net-7.0). In this implementation I keep them as default setting. This means that other failure criteria would still work (e.g. the trusted cert still fails if it's expired).

alanwest:
Hey @dhhoang, thanks for your patience. I've had limited time to give this PR a thorough review, so I appreciate you bearing with me 😄 - I am looking forward to getting this support landed!

Since this is your first PR to this repo, would you mind landing it in a few iterations? If you land a small PR first, then your "first-time contributor" status will go away and then on subsequent PRs the CI will just run automatically.

Here's a suggestion for breaking this PR up into multiple PRs:

1. **First PR** I really like what you're doing by generating the necessary certificates/keys via MSBuild. First PR could be to improve the existing integration test to use this new method. I like that it stores the certs in the build output directory. We should just be able to configure the docker compose volume to point to the output directory. Also, I think we could then do away with the `create-cert` container that spins up via docker compose since now you have the cert generated at build time.
2. **Second PR** mTLS support. This seems to be the more straight-forward piece to implement from the standpoint of the APIs used to add client certificates to the handler.
3. **Third PR** Server certificate support. I think the validation you're doing is pretty close to what we need, but I'm keen on splitting this into its own PR because I think we'll be able to solicit a review from a security expert at Microsoft. I've researched the validation a bit and found a few examples that I believe are closer to what we'll want. Here's just one example https://github.com/dotnet/runtime/issues/39835#issuecomment-663106476.

sandy2008:
@alanwest Hi! https://github.com/open-telemetry/opentelemetry-dotnet/pull/5818 We made the first PR accordingly, would you mind taking a look?

PR #4734, opened 2023-08-02
Run AOT CI on Ubuntu.

Yun-Ting:
## Changes
Updated platform to let AOT CI run on Ubuntu and added debug print.

### DIFF
--- a/.github/workflows/ci-aot.yml
+++ b/.github/workflows/ci-aot.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false  # ensures the entire test matrix is run, even if one permutation fails
       matrix:
-        os: [ windows-latest ]
+        os: [ windows-latest, ubuntu-latest ]
         version: [ net7.0 ]
 
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-aot.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false  # ensures the entire test matrix is run, even if one permutation fails
       matrix:
-        os: [ windows-latest ]
+        os: [ windows-latest, ubuntu-latest ]
```
utpilla, line 18: @vitek-karas Would you recommend we test this on both Windows and Linux? Is there any value in doing that?

vitek-karas: I don't think there's much value right now in running this on both platforms. I would probably go with just Linux (it's cheaper to run).

Later on if/when we add more "meat" to the AOT test app (so that it actually runs and exercises some interesting code) it might make sense to test on both platforms.

utpilla (approved):

vitek-karas (approved):

utpilla:
@Yun-Ting Please add a no-op workflow for this for when a PR only has markdown file changes, the expected AOT check would not run. For example: https://github.com/open-telemetry/opentelemetry-dotnet/pull/4741

PR #4737, opened 2023-08-02
Add opt-in support for metric overflow attribute

utpilla:
Add opt-in support for metric overflow attribute: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#overflow-attribute

For example:

 If we have an SDK configuration where the max MetricPoint limit is four and the application emits the following measurements using a Counter:

- red, apple: 6
- yellow, lemon: 7
- green, apple: 2
- red, cherries: 4
- green, kiwi: 6

As of today, the SDK would:
- Reserve one MetricPoint for measurements with zero tags
- Use the other three MetricPoints for every other measurement with a unique combination of tags

Therefore, we would only pick the first three unique measurements and drop (red, cherries) and (green, kiwi). Note that there is no way to know the total number of fruits sold as we have dropped the last two MetricPoints.

With the changes in this PR, the SDK would:
- Reserve one MetricPoint for measurements with zero tags
- Reserve one MetricPoint for the overflow attribute
- Use the other two MetricPoints for every other measurement with a unique combination of tags

In this case, we would only pick the first two unique measurements, however we would not drop the other three measurements. (green,apple), (red, cherries) and (green, kiwi) would all be aggregated using the new MetricPoint with the overflow attribute. 

There are two advantages of doing this:

1. It provides an easy way to check if the SDK dropped any measurements. The presence of `otel.metric.overflow: true` when querying a metric indicates that the SDK could not aggregate some number of measurements under their own MetricPoint. As of today, one would have to check the SDK internal logs for this information.
2. It preserves partial information about the measurements that would have otherwise been dropped. In this particular example, if a user queries for the metric sum across all tag combinations, they will have the true count of the total number of fruits sold: 6 + 7+ 12 (sum of overflow measurements) = 40. In case of Histogram, they would also know the count of measurements that could not be aggregated under their own MetricPoint.

## Changes
- Use `metric.overflow.attribute` to record metric points that are emitted after the max MetricPoints limit is reached. One MetricPoint in the MetricPoint array would be reserved for this.
- Users would enable this behavior by setting the environment variable `OTEL_DOTNET_EXPERIMENTAL_METRICS_EMIT_OVERFLOW_ATTRIBUTE` to `true`
- Updated `MetricAPITests`, `MetricSnapshotTests`, and `AggregatorTests` to also run with this behavior enabled


src/OpenTelemetry/Metrics/AggregatorStore.cs
```diff
@@ -329,12 +351,20 @@ private void UpdateLong(long value, ReadOnlySpan<KeyValuePair<string, object>> t
             var index = this.FindMetricAggregatorsDefault(tags);
             if (index < 0)
             {
-                if (Interlocked.CompareExchange(ref this.metricCapHitMessageLogged, 1, 0) == 0)
+                if (emitOverflowAttribute)
```
vishweshbankwar, line 354: Any particular reason to skip logging here?

utpilla: We are not dropping the metric now.

vishweshbankwar: we could change the message in logging. Ok to leave without logging as well as it is opt in so users would know what it means if they see `metric.overflow.attribute`

vishweshbankwar (approved):

vishweshbankwar:
Need changelog and readme update (can be a separate PR)

src/OpenTelemetry/Metrics/AggregatorStore.cs
cijothomas: the metric itself has the overflow attribute, dont think additional logging is required.

test/OpenTelemetry.Tests/Metrics/MetricOverflowAttributeTests.cs
```diff
@@ -0,0 +1,240 @@
...
+            }
+
+            MetricPoint overflowMetricPoint;
+
+            // We still have not exceeded the max MetricPoint limit
+            if (temporalityPreference == MetricReaderTemporalityPreference.Delta)
+            {
+                Assert.DoesNotContain(metricPoints, mp => mp.Tags.KeyAndValues[0].Key == "otel.metric.overflow");
+            }
+            else
```
cijothomas, line 76: add a comment why we have this for cumulative from the very beginning..

src/OpenTelemetry/Metrics/AggregatorStore.cs
```diff
@@ -81,6 +84,25 @@ internal sealed class AggregatorStore
...
+        {
+            // We need at least two metric points. One is reserved for zero tags and the other one for overflow attribute
+
+            emitOverflowAttribute = true;
+
+            // Setting this to as we would reserve the metricPoints[1] for overflow attribute.
+            // Newer attributes should be added starting at the index: 2
+            this.metricPointIndex = 1;
+
+            this.metricPoints[1] = new MetricPoint(this, this.aggType, new KeyValuePair<string, object>[] { new("otel.metric.overflow", true) }, this.histogramBounds, this.exponentialHistogramMaxSize, this.exponentialHistogramMaxScale);
```
cijothomas, line 104: Should we do this when actually needed (i.e the very first time?), similar to how we initialize zero tag
https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/Metrics/AggregatorStore.cs#L185

utpilla: Great suggestion! Thanks for that. I have updated the code.

src/OpenTelemetry/CHANGELOG.md
```diff
@@ -2,6 +2,22 @@
 
 ## Unreleased
 
+* **Experimental Feature** Added an opt-in feature to aggregate any metric
+  measurements that were dropped due to reaching the max MetricPoints limit.
+  When this feature is enabled, SDK would aggregate such measurements using a
+  reserved MetricPoint with a single tag with key as `otel.metric.overflow` amd
```
cijothomas, line 8:
```suggestion
  reserved MetricPoint with a single tag with key as `otel.metric.overflow` and
```

src/OpenTelemetry/Metrics/AggregatorStore.cs
```diff
@@ -81,6 +86,21 @@ internal sealed class AggregatorStore
             this.tagKeysInteresting = hs;
             this.tagsKeysInterestingCount = hs.Count;
         }
+
+        // If the switch is not set at all or if it's explicitly set to false
+        if (AppContext.TryGetSwitch("OTel.Dotnet.EmitMetricOverflowAttribute", out bool shouldEmitOverflowAttribute) == false ||
```
cijothomas, line 91: nit: maybe move the logic to check if this feature is enabled/diabled to MeterProvider, and have it pass a simple bool to AggregatorStore.

src/OpenTelemetry/Metrics/AggregatorStore.cs
```diff
@@ -329,12 +365,21 @@ private void UpdateLong(long value, ReadOnlySpan<KeyValuePair<string, object>> t
...
-                    OpenTelemetrySdkEventSource.Log.MeasurementDropped(this.name, this.metricPointCapHitMessage, MetricPointCapHitFixMessage);
+                    this.InitializeOverflowTagPointIfNotInitialized();
+                    this.metricPoints[1].Update(value);
+                    return;
                 }
+                else
+                {
+                    if (Interlocked.CompareExchange(ref this.metricCapHitMessageLogged, 1, 0) == 0)
+                    {
+                        OpenTelemetrySdkEventSource.Log.MeasurementDropped(this.name, this.metricPointCapHitMessage, MetricPointCapHitFixMessage);
```
cijothomas, line 378: Its good to improve this message giving use a hint about the new feature to avoid this.

src/OpenTelemetry/CHANGELOG.md
```diff
@@ -2,6 +2,22 @@
 
 ## Unreleased
 
+* **Experimental Feature** Added an opt-in feature to aggregate any metric
+  measurements that were dropped due to reaching the max MetricPoints limit.
```
vishweshbankwar, line 6: a ref link to this might be helpful https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/docs/metrics/customizing-the-sdk#changing-maximum-metricpoints-per-metricstream

alanwest (commented):
@utpilla I think this is a really interesting proposal! Are you considering proposing this in the specification? I think it might be a good "official" addition and enable us to remove the feature flag in the future.

Given this PR is also proposing a new way for us to feature flag experimental stuff... I have some thoughts about this `AppContext` mechanic.

src/OpenTelemetry/Metrics/AggregatorStore.cs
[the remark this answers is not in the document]
utpilla: > Are you considering proposing this in the specification?

This is coming from the [spec ](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#overflow-attribute). It's marked Experimental.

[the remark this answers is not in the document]
utpilla: I'll update the code to use the simpler environment variable approach. Your comment got me thinking if there's any additional benefit to using `AppContext` switch over environment variables. I would try to get more information on this.

utpilla: I have updated the code.

src/OpenTelemetry/Metrics/AggregatorStore.cs
```diff
@@ -22,9 +22,11 @@ namespace OpenTelemetry.Metrics;
 
 internal sealed class AggregatorStore
 {
-    private static readonly string MetricPointCapHitFixMessage = "Modify instrumentation to reduce the number of unique key/value pair combinations. Or use Views to drop unwanted tags. Or use MeterProviderBuilder.SetMaxMetricPointsPerMetricStream to set higher limit.";
+    private static readonly string MetricPointCapHitFixMessage = "Consider opting in for the experimental SDK feature to emit all the throttled metrics under the overflow attribute. You could also modify instrumentation to reduce the number of unique key/value pair combinations. Or use Views to drop unwanted tags. Or use MeterProviderBuilder.SetMaxMetricPointsPerMetricStream to set higher limit.";
```
cijothomas, line 25:
```suggestion
    private static readonly string MetricPointCapHitFixMessage = "Consider opting in for the experimental SDK feature to emit all the throttled metrics under the overflow attribute by setting env variable OTEL_DOTNET_EXPERIMENTAL_METRICS_EMIT_OVERFLOW_ATTRIBUTE = true. You could also modify instrumentation to reduce the number of unique key/value pair combinations. Or use Views to drop unwanted tags. Or use MeterProviderBuilder.SetMaxMetricPointsPerMetricStream to set higher limit.";
```

src/OpenTelemetry/Metrics/AggregatorStore.cs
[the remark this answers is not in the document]
utpilla: @alanwest I have updated the code to use environment variables.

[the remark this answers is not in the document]
alanwest: > This is coming from the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#overflow-attribute). It's marked Experimental.

You coulda just lied and told me this brilliant idea was your own 😆.

alanwest (approved):

reyang (approved):

src/OpenTelemetry/Metrics/AggregatorStore.cs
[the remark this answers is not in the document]
utpilla: > You coulda just lied and told me this brilliant idea was your own 😆.

What can I say? When it comes to Metrics, I strive to be an "Exemplar" 😄

PR #4751, opened 2023-08-07
[AOT] Marked SqlClient Instrumentation package as trim incompatible.

Yun-Ting:

cc: @eerhardt, @vitek-karas 

## Changes

`OpenTelemetry.Instrumentation.SqlClient` uses PropertyFetcher to fetch the following properties:
https://github.com/open-telemetry/opentelemetry-dotnet/blob/64df5e3df8db55e4da482e577c3524b53213f4b2/src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientDiagnosticListener.cs#L35-L41

Due to the fact that `SqlClient` library uses anonymous method to write commands, i.e.:
https://github.com/dotnet/SqlClient/blob/f478be5bf3597f5c0f0940ac471c6b481c38416f/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/SqlClient/SqlClientDiagnosticListenerExtensions.cs#L47

and there is currently no plan for SqlClient library to make it AOT-compatible, `OpenTelemetry.Instrumentation.SqlClient` annotates methods/classes with
[[RequiresUnreferencedCode(SqlClientInstrumentation.SqlClientTrimmingUnsupportedMessage)]](https://github.com/open-telemetry/opentelemetry-dotnet/pull/4751/files#diff-be932e6f59b00b1dbf2d82cb7d4384cc8a93f8d52e906c71aab7594c4f616359R31:~:text=internal%20const%20string%20SqlClientTrimmingUnsupportedMessage%20%3D%20%22Trimming%20is%20not%20yet%20supported%20with%20SqlClient%20instrumentation.%22%3B)
to notify the caller of the instrumentation package of such fact.


build/test-aot-compatibility.ps1
```diff
@@ -10,6 +10,14 @@ foreach ($line in $($publishOutput -split "`r`n"))
     if ($line -like "*analysis warning IL*") 
     {
         Write-Host $line
+
+        # OpenTelemetry.Instrumentation.SqlClient is properly AOT-annotated.
+        # The caller of OpenTelemetry.Instrumentation.SqlClient will be notified.
+        if ($line -like "*Trimming is not yet supported with SqlClient*")
```
eerhardt, line 16: Should we just remove this from the TestApp?

https://github.com/open-telemetry/opentelemetry-dotnet/blob/64df5e3df8db55e4da482e577c3524b53213f4b2/test/OpenTelemetry.AotCompatibility.TestApp/OpenTelemetry.AotCompatibility.TestApp.csproj#L25

eerhardt: Actually, these warnings are coming from callers of the SqlClient APIs, not from inside the SqlClient library. Since the library's APIs are annotated with `RequiresUnreferencedCode` you shouldn't be getting warnings.

We should remove this code from the script and annotate any callers as `RequiresUnreferencedCode` as well.

Yun-Ting: I've properly annotated the callers; thank you.

src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientDiagnosticListener.cs
```diff
@@ -40,6 +46,9 @@ internal sealed class SqlClientDiagnosticListener : ListenerHandler
     private readonly PropertyFetcher<Exception> exceptionFetcher = new("Exception");
     private readonly SqlClientInstrumentationOptions options;
 
+#if NET6_0_OR_GREATER
+    [RequiresUnreferencedCode(SqlClientInstrumentation.SqlClientTrimmingUnsupportedMessage)]
```
eerhardt, line 50: You don't need this on both the class and the constructor. You can remove this on the constructor.

eerhardt (approved):

reyang (approved):

src/OpenTelemetry.Instrumentation.SqlClient/SqlClientInstrumentation.cs
```diff
@@ -24,7 +27,9 @@ namespace OpenTelemetry.Instrumentation.SqlClient;
 internal sealed class SqlClientInstrumentation : IDisposable
 {
     internal const string SqlClientDiagnosticListenerName = "SqlClientDiagnosticListener";
-
+#if NET6_0_OR_GREATER
+    internal const string SqlClientTrimmingUnsupportedMessage = "Trimming is not yet supported with SqlClient";
```
utpilla, line 31:
```suggestion
    internal const string SqlClientTrimmingUnsupportedMessage = "Trimming is not yet supported with SqlClient instrumentation";
```

CodeBlanch (approved):
LGTM

PR #4765, opened 2023-08-11
HttpSemanticConventions - update helper class

TimothyMothra:
Towards #4484

When I started on this, I was working off the older version of the spec in the old repo.
The current version of the spec in the new repo changed the environment variable to a comma-separated list.

> SHOULD introduce an environment variable OTEL_SEMCONV_STABILITY_OPT_IN in the existing major version which is a comma-separated list of values. 
https://github.com/open-telemetry/semantic-conventions/tree/v1.21.0/docs/http

## Changes
- refactor `HttpSemanticConventionsHelper`
  - rewrite to parse a CSV
- updated several projects to include `Shared\Shims\NullableAttributes`
- refactor unit tests. 
  - replaced `InlineData` with `MemberData` so each test can use the same collection of test cases.
  - added additional csv test cases

## Note about duplicate values in the Environment Variable.
Edit: 8/14: The spec has been updated to specify if both `http` and `http/dup` are specified, `http/dup` will take precedence.
https://github.com/open-telemetry/semantic-conventions/pull/249

As of today, the only valid values are `http` (emit new attributes) and `http/dup` (emit both new and old attributes).
The spec doesn't define how to handle a scenario where a user provides both values in the CSV.
I opened an issue to get clarity: https://github.com/open-telemetry/semantic-conventions/issues/240

In this scenario, Java takes a user-friendly approach of the approach of enabling both new and old attributes. I think we should do the same here.
https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/32c5d4c26708ccaa20221bb6d687f56c9efde979/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/internal/SemconvStability.java#L26-L39

### DIFF
--- a/src/OpenTelemetry/OpenTelemetry.csproj
+++ b/src/OpenTelemetry/OpenTelemetry.csproj
@@ -25,6 +25,7 @@
     <Compile Include="$(RepoRoot)\src\Shared\MathHelper.cs" Link="Includes\MathHelper.cs" />
     <Compile Include="$(RepoRoot)\src\Shared\Options\*.cs" Link="Includes\Options\%(Filename).cs" />
     <Compile Include="$(RepoRoot)\src\Shared\ResourceSemanticConventions.cs" Link="Includes\ResourceSemanticConventions.cs" />
+    <Compile Remove="C:\REPOS\opentelemetry-dotnet-fork\src\Shared\Shims\NullableAttributes.cs" />
     <Compile Include="$(RepoRoot)\src\Shared\Shims\NullableAttributes.cs" />
   </ItemGroup>
 
--- a/src/OpenTelemetry/OpenTelemetry.csproj
+++ b/src/OpenTelemetry/OpenTelemetry.csproj
@@ -25,7 +25,6 @@
     <Compile Include="$(RepoRoot)\src\Shared\MathHelper.cs" Link="Includes\MathHelper.cs" />
     <Compile Include="$(RepoRoot)\src\Shared\Options\*.cs" Link="Includes\Options\%(Filename).cs" />
     <Compile Include="$(RepoRoot)\src\Shared\ResourceSemanticConventions.cs" Link="Includes\ResourceSemanticConventions.cs" />
-    <Compile Remove="C:\REPOS\opentelemetry-dotnet-fork\src\Shared\Shims\NullableAttributes.cs" />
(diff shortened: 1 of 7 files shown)


src/OpenTelemetry/OpenTelemetry.csproj
```diff
@@ -25,6 +25,7 @@
     <Compile Include="$(RepoRoot)\src\Shared\MathHelper.cs" Link="Includes\MathHelper.cs" />
     <Compile Include="$(RepoRoot)\src\Shared\Options\*.cs" Link="Includes\Options\%(Filename).cs" />
     <Compile Include="$(RepoRoot)\src\Shared\ResourceSemanticConventions.cs" Link="Includes\ResourceSemanticConventions.cs" />
+    <Compile Remove="C:\REPOS\opentelemetry-dotnet-fork\src\Shared\Shims\NullableAttributes.cs" />
```
utpilla, line 28: Added by mistake?

TimothyMothra: yes, this is a mistake. removing now.

test/OpenTelemetry.Tests/Shared/HttpSemanticConventionHelperTest.cs
```diff
@@ -22,6 +22,29 @@ namespace OpenTelemetry.Tests.Shared;
...
+            new object[] { "HTTP/DUP",  HttpSemanticConvention.Dupe },
+            new object[] { "junk,,junk",  HttpSemanticConvention.Old },
+            new object[] { "junk,JUNK",  HttpSemanticConvention.Old },
+            new object[] { "junk1,junk2",  HttpSemanticConvention.Old },
+            new object[] { "junk,http",  HttpSemanticConvention.New },
+            new object[] { "junk,http , http ,junk",  HttpSemanticConvention.New },
+            new object[] { "junk,http/dup",  HttpSemanticConvention.Dupe },
+            new object[] { "junk, http/dup ",  HttpSemanticConvention.Dupe },
+            new object[] { "junk, http/dup ",  HttpSemanticConvention.Dupe },
+            new object[] { "http,http/dup",  HttpSemanticConvention.Dupe },
```
utpilla, line 45: Add another test case for `http/dup,http` to test the reversed order as well.

utpilla (approved):

vishweshbankwar (approved):

PR #4766, opened 2023-08-11
HttpSemanticConventions - update AspNetCore histogram

TimothyMothra:
Towards #4484.

**CHANGED THIS PR TO DRAFT.
This PR is blocked until we have support for new default histogram bounds.**

## Changes
- all the following changes are made behind the environment variable `OTEL_SEMCONV_STABILITY_OPT_IN`.
- update Instrumentation.AspNetCore class `HttpInMetricsListener` to emit new metric.
  - `http.server.duration` (milliseconds) replaced with `http.server.request.duration` (seconds)
    - This is defined in the as-of-yet unreleased v1.22.0 https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-metrics.md
    - This requires custom histogram bounds.
      As of today (Aug 22) this is implemented using the Views API. 
      This is a blocker for this PR while we explore alternatives.
- Update unit test `MetricTests.RequestMetricIsCaptured` to evaluate all states of the environment variable (old, new, both).


src/OpenTelemetry.Instrumentation.AspNetCore/MeterProviderBuilderExtensions.cs
```diff
@@ -78,6 +78,10 @@ public static class MeterProviderBuilderExtensions
 
         builder.AddMeter(AspNetCoreMetrics.InstrumentationName);
 
+        builder.AddView(
+            instrumentName: HttpInMetricsListener.HttpServerDurationMetricName,
```
utpilla, line 82: Add a check for both Meter name and Instrument name.

src/OpenTelemetry.Instrumentation.AspNetCore/MeterProviderBuilderExtensions.cs
```diff
@@ -78,6 +78,10 @@ public static class MeterProviderBuilderExtensions
 
         builder.AddMeter(AspNetCoreMetrics.InstrumentationName);
 
+        builder.AddView(
```
cijothomas, line 81: This should be done by the application user, not the instrumentation library. Instrumentation Library can set Hint (once that feature is available), but it should not try to Configure View. View is a SDK concept, and instrumentation library should not have sdk dependency at all. (its an issue that this library has an undesired dependency on sdk, thats tracked separately)...

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInMetricsListener.cs
```diff
@@ -42,7 +46,7 @@ internal HttpInMetricsListener(string name, Meter meter, AspNetCoreMetricsInstru
     {
         this.meter = meter;
         this.options = options;
-        this.httpServerDuration = meter.CreateHistogram<double>(HttpServerDurationMetricName, "ms", "Measures the duration of inbound HTTP requests.");
+        this.httpServerDuration = meter.CreateHistogram<double>(HttpServerDurationMetricName, "s", "Measures the duration of inbound HTTP requests.");
```
cijothomas, line 49: [the ](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInMetricsListener.cs#L147) must be changed, right? Its still doing ms, not seconds..

TimothyMothra: Thanks for catching that. I was focused on figuring out the View and the unit test working that I missed this.

src/OpenTelemetry.Instrumentation.AspNetCore/MeterProviderBuilderExtensions.cs
TimothyMothra: Experimental spec for Hint API ("instrument advice") is here:
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument-advice

Seems that Trask is ready to move it to Stable:
https://github.com/open-telemetry/opentelemetry-specification/issues/3391

I had a brief chat with Utkarsh about this. 
Seems the Hint API will be a blocker for this PR.

cijothomas: Hint API require DiagnosticSource changes, so not coming this year.
But i don't know if Hint API support is required before making this change. As users always have View API to adjust bounds.

TimothyMothra: IMO, that seems like a bad user experience.
If the Instrumentation library ships with the wrong buckets and users need to manually adjust.

utpilla: @cijothomas On thinking a bit more about this, I think it should be okay for the instrumentation library to use `View` until the Hint API is offered from DiagnosticSource. When the Hint API is available, the instrumentation library could get rid of `View` and switch to using the Hint API.

As of today, anyone using AspNetCoreInstrumentation library is going to take a dependency on the SDK anyway.

TimothyMothra: I can include a TODO here to call out that this needs to be changed to Hint API when available.

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInMetricsListener.cs
```diff
@@ -144,7 +148,7 @@ public override void OnEventWritten(string name, object payload)
             // We are relying here on ASP.NET Core to set duration before writing the stop event.
             // https://github.com/dotnet/aspnetcore/blob/d6fa351048617ae1c8b47493ba1abbe94c3a24cf/src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs#L449
             // TODO: Follow up with .NET team if we can continue to rely on this behavior.
-            this.httpServerDuration.Record(Activity.Current.Duration.TotalMilliseconds, tags);
+            this.httpServerDuration.Record(Activity.Current.Duration.TotalSeconds, tags);
```
alanwest, line 151: Java gates this change with the `OTEL_SEMCONV_STABILITY_OPT_IN` setting. Maybe we should too since this is potentially disruptive to end users and backends.

Unfortunately, it's not explicitly stated that this change should be gated, but maybe it should be... @trask what do you think?

trask: within the scope of HTTP semconv, I think(?) this is implied by the spec, since unit `seconds` is part of the frozen http semconv for `http.server.request.duration`

src/OpenTelemetry.Instrumentation.AspNetCore/MeterProviderBuilderExtensions.cs
trask: > I think it should be okay for the instrumentation library to use `View` until the Hint API is offered from DiagnosticSource. When the Hint API is available, the instrumentation library could get rid of `View` and switch to using the Hint API.
>
> As of today, anyone using AspNetCoreInstrumentation library is going to take a dependency on the SDK anyway.

I would open a dedicated issue to discuss and record this decision. It's a pretty fundamental principle that instrumentations should only depend on the API and not on the SDK, and so you'll want to document the rationale clearly in this case and make sure your entire community is onboard and understands the ramifications.

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInMetricsListener.cs
alanwest: > I think(?) this is implied by the spec

This was my original interpretation... but I also can understand how someone might have a different interpretation since the words from the [warning text](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/README.md#semantic-conventions-for-http) are very attribute-centric.

I'd be happy to take a stab at some clarifying verbiage to the text if folks would find that helpful.

trask: > the words from the [warning text](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/README.md#semantic-conventions-for-http) are very attribute-centric.

ah, thx, I see the confusion now

> I'd be happy to take a stab at some clarifying verbiage to the text if folks would find that helpful.

that would be great

src/OpenTelemetry.Instrumentation.AspNetCore/MeterProviderBuilderExtensions.cs
cijothomas: > 

It can still cause other issues as Views are additive. If user adds another view matching the same metric, to drop an unwanted attribute, it'll result in 2 metrics stream. 

Its a bug that the instrumentation has a SDK dependency. It is fixable (once we move the things the instrumentation need from SDK to API itself.). Its even worse if the instrumentation leverages the SDK dependency to define behavior, reserved for application owner to define.

There are other instrumentations where this trick/hack of using views from SDK cannot be used  - like asp.net framework.

At the cost of exposing yet another public API, an option like `RecordDurationInMillisecond` could work, and is probably more desirable than misusing SDK dependency.

PR #4767, opened 2023-08-12
[repo] Uptake .NET Package Validation

### LINKED ISSUE #2081: ApiCompat improvements
Copied from https://github.com/open-telemetry/opentelemetry-dotnet/issues/1711#issuecomment-860815199

.NET Team is moving this tool into a separate package, along with other compact check tools. We'll need to adjust accordingly as it happens. - @utpilla

Only enabled for core packages now, as only core packages have ever released stable package. Need to update this once non-core components also release stable. - @cijothomas

The check is part of a non-mandatory CI now, as the tool itself is new and we don't want to break existing CI if the tool has issues. Once the tool is stable, (expected by .NET 6 release time), we can make this CI check mandatory. @cijothomas

The CI check is failing today. @utpilla has investigated this and opened dotnet/arcade#7511, as this is likely a tooling issue. @utpilla

Need to update the release process, so that every stable release updates the "base or contract" version to check compat against. @cijothomas

mfogliatto:
Fixes #2081

## Changes

Uptake .NET Package Validation according to the documentation/suggestions in https://github.com/Azure/azure-sdk-for-net/pull/36965#issuecomment-1665855629

- Add PropertyGroup in Directory.Build.targets
- Add package-validation.yml.workflow


ViktorHofer (commented):
You want to remove the old APICompat (that comes from dotnet/arcade) entirely and enable the new APICompat to run during the build as well.

src/Directory.Build.targets
```diff
@@ -1,4 +1,9 @@
 <Project>
+  <PropertyGroup Condition="'$(MinVerTagPrefix)' == 'core-' AND '$(RunPackageValidation)' == 'true' AND '$(RunApiCompat)' == 'true'">
```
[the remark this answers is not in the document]
mfogliatto: > Rather than introducing a new property RunPackageValidation, consider just using the existing one: EnablePackageValidation

Are you suggesting to set the value of `EnablePackageValidation` in the workflow as it is now, and then here do like this?

```xml
<PropertyGroup Condition="'$(MinVerTagPrefix)' == 'core-' AND '$(EnablePackageValidation)' == 'true' AND '$(RunApiCompat)' == 'true'">
    <PackageValidationBaselineVersion>$(OTelLatestStableVer)</PackageValidationBaselineVersion>
</PropertyGroup>
```

.github/workflows/package-validation.yml
```diff
@@ -0,0 +1,27 @@
...
+        fetch-depth: 0 # fetching all
+
+    - name: Setup dotnet
+      uses: actions/setup-dotnet@v3
+
+    - name: Install dependencies
+      run: dotnet restore
+
+    - name: Pack
+      run: dotnet pack --configuration Release --no-restore
```
utpilla, line 27: Could you add two jobs for this workflow similar to the existing CI workflow? One for [stable releases](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/.github/workflows/ci.yml#L14) and one for [experimental features](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/.github/workflows/ci.yml#L42). You would have to split the build and pack steps for that.

src/Directory.Build.targets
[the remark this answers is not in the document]
utpilla: I see some value in introducing the property `RunPackageValidation`. This would let us skip package validation check when actually [publishing packages](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/.github/workflows/publish-packages-1.0.yml). If we run package validation check for every PR then we can be assured that the workflow for actually publishing the package doesn't have to run it again.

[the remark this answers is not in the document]
ViktorHofer: When doing your publishing package step, you could just pass in `/p:EnablePackageEvalidation=false` to make sure it doesn't run.

.github/workflows/package-validation.yml
mfogliatto: Thanks @utpilla! I just splitted the job as suggested.

Please let me know in case there's anything else.

src/Directory.Build.targets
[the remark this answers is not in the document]
mfogliatto: Yes that's also an option. Another nice perk about having the additional property is that the package validation configuration stuff remains all in one place.

@utpilla what do you think? Shall we leave it as is or remove the introduced property and do like Viktor is suggesting?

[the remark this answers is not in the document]
utpilla: @mfogliatto I think we could just use the `EnablePackageValidation` property instead of introducing one ourselves.

@ViktorHofer Does the .NET Package Validation tool print out any output? It would be great if we could look at the build output in the CI workflow to check if the tool actually ran. Currently, it doesn't seem to log anything related to validation at all.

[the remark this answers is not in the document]
ViktorHofer: It doesn't as msbuild by default runs in `minimal` verbosity mode. If you want it to print a confirmation message, you would need to change the msbuild verbosity level globally.

Note that many other tools run as part of the build, not just Roslyn and NuGet. You don't see confirmation messages for their run either. The default for modern .NET is to not emit a high priority message unless necessary.

If you set the `EnablePackageValidation` property to true, PackageValidation/APICompat will always run after the Pack target.

.github/workflows/package-validation.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       run: dotnet restore
 
     - name: Pack
-      run: dotnet pack --configuration Release --no-restore --property:EnablePackageValidation=true;ExposeExperimentalFeatures=false
+      run: dotnet pack --configuration Release --no-restore --property:EnablePackageValidation=true --property:ExposeExperimentalFeatures=false
```
ViktorHofer, line 25: just a nit comment as I saw this. I think you are one of the few customers that use the --property syntax to pass properties to msbuild. Nearly everyone else uses `/p:` which is shorter.

mfogliatto: Thanks Viktor, sure no need to keep unnecessarily long :)

src/OpenTelemetry.Api.ProviderBuilderExtensions/OpenTelemetry.Api.ProviderBuilderExtensions.csproj
```diff
@@ -8,10 +8,11 @@
     <AnalysisLevel>latest-all</AnalysisLevel>
   </PropertyGroup>
 
-  <!--Do not run ApiCompat as this package has never released a stable version.
+  <!--Do not run ApiCompat nor PackageValidation as this package has never released a stable version.
   Remove this property once we have released a stable version.-->
   <PropertyGroup Condition="'$(TargetFramework)' == 'net6.0'">
     <RunApiCompat>false</RunApiCompat>
+    <SkipPackageValidation>true</SkipPackageValidation>
```
ViktorHofer, line 15: By setting this property, you actually don't skip package validation as it still runs. You are skipping package baseline validation by not setting the `PackageValidationBaselineVersion` property.

ViktorHofer: I would recommend to not add any extra properties, if not strictly necessary. I.e. here you could just unset the `PackageValidationBaselineVersion` property: `<PackageValidationBaselineVersion />` and move the src/Directory.Build.targets changes into the Directory.Build.props file.

That said, I'm not 100% sure if that provides a better UX.

mfogliatto: My attempt to disable the validation by setting `EnablePackageValidation` to false in the same `PropertyGroup` didn't seem to work. This is why I opted to add a new `SkipPackageValidation` property instead.

ViktorHofer: That's because you pass `EnablePackageValidation=true` in as a global property in the YML file. Global properties can't be overwritten. What's the intent here? Do you want to disable package validation completely or just the package baseline validation?

mfogliatto: The intent is to not perform package validation for some of the packages, specifically for those that do not yet have a stable version to compare with. I'm assuming this is a requirement as this is how it is working for the existing `ApiCompat` workflow. @utpilla could you please comment?

ViktorHofer: If that's the case then we've gone full circle and you need to pass in a custom property in the YML and then set EnablePackageValidation=true based on that custom property. Kind of what you already had. I didn't know about your requirement of being able to disable validation for certain projects.

mfogliatto: Actually you are right. Since the requirement to have a stable version to compare with only makes sense for package baseline validation, I guess we can keep the other validations in place even in such cases.

mfogliatto: I just applied your suggestion to use `<PackageValidationBaselineVersion />` instead of a new prop. @ViktorHofer Could you please take another look?

mfogliatto:
@ViktorHofer for some reason package validation is failing to pull the packages to perform the baseline version comparison:

https://github.com/open-telemetry/opentelemetry-dotnet/actions/runs/6057759836/job/16439245926?pr=4767

Running the same dotnet pack command in my dev box works just fine.

Do you have any insights on why that could happen?

ViktorHofer:
I'm super sorry, I misspoke previously. To disable package baseline validation, you want to set `<DisablePackageBaselineValidation>true</DisablePackageBaselineValidation>` in the project files:

https://github.com/dotnet/sdk/blob/4f3276dcda47e9248b36d13474e0afabca674bd6/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.ApiCompat.ValidatePackage.targets#L30C53-L30C85

mfogliatto:
Thanks Viktor! No worries at all! I just replaced all the instances with
DisablePackageBaselineValidation.

src/OpenTelemetry.Api.ProviderBuilderExtensions/OpenTelemetry.Api.ProviderBuilderExtensions.csproj
```diff
@@ -8,10 +8,11 @@
     <AnalysisLevel>latest-all</AnalysisLevel>
   </PropertyGroup>
 
-  <!--Do not run ApiCompat as this package has never released a stable version.
+  <!--Do not run ApiCompat nor Package Baseline Validation as this package has never released a stable version.
   Remove this property once we have released a stable version.-->
   <PropertyGroup Condition="'$(TargetFramework)' == 'net6.0'">
     <RunApiCompat>false</RunApiCompat>
+    <DisablePackageBaselineValidation>true</DisablePackageBaselineValidation>
```
ViktorHofer, line 15: Unfortunately this won't work as you are setting the property in the inner build (`net6.0`) while Package Validation runs as part of the `Pack` target which runs in the outer build. This might sound complicated but it essentially means that Package Validation looks at the entire package which requires all inner builds to be completed. An inner build is a build for one TFM for multi-targeting projects.

That `DisablePackageBaselineValidation` switch can't be set in an inner build. Move it out of that property group.

src/OpenTelemetry.Api.ProviderBuilderExtensions/OpenTelemetry.Api.ProviderBuilderExtensions.csproj
```diff
@@ -8,10 +8,11 @@
     <AnalysisLevel>latest-all</AnalysisLevel>
   </PropertyGroup>
 
-  <!--Do not run ApiCompat as this package has never released a stable version.
+  <!--Do not run ApiCompat nor Package Baseline Validation as this package has never released a stable version.
   Remove this property once we have released a stable version.-->
   <PropertyGroup Condition="'$(TargetFramework)' == 'net6.0'">
     <RunApiCompat>false</RunApiCompat>
```
ViktorHofer, line 14: One more thing: You are mixing the old APICompat with the new Package Validation which isn't really supported. Unfortunately, the `RunApiCompat` property still exists in the new Package Validation (we should delete in our code - I'll file an issue for that). If you keep that set, you might see unexpected behavior in the new Package Validation tooling.

That goes back to my initial comment in this thread about better not using both the old and the new APICompat (PackageValidation) together.

ViktorHofer:
cc @ericstj so that you are aware of this work.

src/OpenTelemetry.Api.ProviderBuilderExtensions/OpenTelemetry.Api.ProviderBuilderExtensions.csproj
mfogliatto: Got it, thanks! Moved it to a separate `PropertyGroup`.

src/OpenTelemetry.Api.ProviderBuilderExtensions/OpenTelemetry.Api.ProviderBuilderExtensions.csproj
mfogliatto: Are you saying we cannot keep both for the time being without them interfering with each other?

ViktorHofer: You can keep the property as long as it is only set in an inner build (meaning only in multi-targeting projects conditioned on a TFM).

src/OpenTelemetry.Api.ProviderBuilderExtensions/OpenTelemetry.Api.ProviderBuilderExtensions.csproj
mfogliatto: Resolving as this was achieved by adding `DisablePackageBaselineValidation` as suggested in another comment.

src/Directory.Build.targets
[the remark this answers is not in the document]
mfogliatto: Agreed to use the existing property instead of adding a new one. Resolving the comment.

src/OpenTelemetry.Api.ProviderBuilderExtensions/OpenTelemetry.Api.ProviderBuilderExtensions.csproj
mfogliatto: After checking this with @utpilla , we decided to remove the old APICompat to avoid any unexpected behavior. Could you please have a look at the latest iteration?

ViktorHofer (approved):

mfogliatto:
@ViktorHofer the workflow is failing because of the pack command failing with the following error:

`error MSB4018: System.IO.FileNotFoundException: Package 'C:\Users\runneradmin\.nuget\packages\opentelemetry.extensions.hosting\1.6.0\opentelemetry.extensions.hosting.1.6.0.nupkg' not found. Please provide a valid package path.`

Is there any step missing here?

.github/workflows/package-validation.yml
```diff
@@ -0,0 +1,42 @@
...
+        fetch-depth: 0 # fetching all
+
+    - name: Setup dotnet
+      uses: actions/setup-dotnet@v3
+
+    - name: Install dependencies
+      run: dotnet restore
+
+    - name: Pack
+      run: dotnet pack --configuration Release --no-restore /p:EnablePackageValidation=true /p:ExposeExperimentalFeatures=false
```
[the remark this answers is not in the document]
mfogliatto: Thanks Viktor! Just applied your suggestion and now the pack command and validation work like a charm :)

src/Directory.Build.targets
```diff
@@ -1,4 +1,7 @@
 <Project>
+  <PropertyGroup Condition="'$(MinVerTagPrefix)' == 'core-'">
+    <PackageValidationBaselineVersion>$(OTelLatestStableVer)</PackageValidationBaselineVersion>
+  </PropertyGroup>
```
CodeBlanch, line 4: Is there a reason this went into this file and not `Common.prod.props`? Most of the settings are there so I would like to move it if possible.

mfogliatto: Thanks for the comment! If I place this in `Common.prod.props` the package validation will simply not run. That's why it is here.

I believe this is because props are imported earlier than targets. Therefore, when the `PropertyGroup` is imported and used by the PackageValidation target, both `MinVerTagPrefix` and `OTelLatestStableVer` are empty. 

@ViktorHofer feel free to expand on this if you have any additional insights or suggestion.

utpilla (approved):

CodeBlanch (approved):

utpilla:
Thanks @mfogliatto for moving us to the new package validation tool!

src/Directory.Build.targets
utpilla: @mfogliatto Check this: https://github.com/open-telemetry/opentelemetry-dotnet/pull/4923/files#diff-5baf5f9e448ad54ab25a091adee0da05d4d228481c9200518fcb1b53a65d4156R7-R12

CodeBlanch: Just to follow up on this, yes there is an ordering issue. The `.props` files in the repo apply before the `csproj` so `MinVerTagPrefix` is not available to them. I just learned `msbuild` has a preprocess command you can use to spit out an XML to see everything in order:

`msbuild /pp:preprocess.xml src/OpenTelemetry/OpenTelemetry.csproj`

Kind of neat!

To get around that I just defined it as a target so we can control when it runs/applies the properties.

utpilla:
@ViktorHofer @ericstj Does enabling package validation eliminate the need for a separate [public API analyzers](https://github.com/dotnet/roslyn-analyzers/tree/main#microsoftcodeanalysispublicapianalyzers) check?

src/Directory.Build.targets
ViktorHofer: > @mfogliatto Check this: https://github.com/open-telemetry/opentelemetry-dotnet/pull/4923/files#diff-5baf5f9e448ad54ab25a091adee0da05d4d228481c9200518fcb1b53a65d4156R7-R12

Note that this might not work with static graph restore (which is a NuGet feature). Those properties are kind of expected to be set at evaluation time, not inside a target. IMO where they are right now is fine.

ViktorHofer:
I answered that question offline.

PR #4768, opened 2023-08-14
Add HttpClient metrics for .NET Framework

### LINKED ISSUE #4764: Implement support for metrics in HttpClient for .NET framework
# Feature Request

Currently, HttpClient instrumentation supports metrics for .NET (Core), but not for .NET framework. This is mentioned in a [warning in the documentation](https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/src/OpenTelemetry.Instrumentation.Http#metrics).

This is unexpected and seen as a bug by users I've interacted with. Users of e. g. the auto-instrumentation packages or other instrumentation that packages that use HttpClient under the hood will likely not find the warning.

Adding support for metrics also for .NET framework would be a desired solution.

Users who currently face this issue have to either provide their own instrumentation library, add manual instrumentation, or create metrics from fully sampled traces.

## Additional Context

[Related Slack discussion](https://cloud-native.slack.com/archives/C01N3BC2W7Q/p1678833977411999)
[Warning in the documentation](https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/src/OpenTelemetry.Instrumentation.Http#metrics)

matt-hensley:
Fixes #4764

## Changes

Extends the existing tracing implementation to also generate metrics. This requires tracing to be enabled - while not ideal, this significantly limits the scope of changes to get metrics working.

**To enable metrics, you must call `AddHttpClientInstrumentation()` on `TracerProviderBuilder`.**

A single metric is published: `http.client.duration`, same as the .NET Core/.NET metrics implementation.

Unit tests have been modified to cover HttpClient metrics on .NET Framework. All unit test are passing.


src/OpenTelemetry.Instrumentation.Http/Implementation/HttpWebRequestActivitySource.netfx.cs
```diff
@@ -395,7 +401,49 @@ private static void ProcessResult(IAsyncResult asyncResult, AsyncCallback asyncC
             HttpInstrumentationEventSource.Log.FailedProcessResult(ex);
         }
 
+        activity.SetEndTime(DateTime.UtcNow);
```
cijothomas, line 404: what is the need for this change?

src/OpenTelemetry.Instrumentation.Http/Implementation/HttpWebRequestActivitySource.netfx.cs
```diff
@@ -395,7 +401,49 @@ private static void ProcessResult(IAsyncResult asyncResult, AsyncCallback asyncC
...
         activity.Stop();
+
+        TagList tags = default;
+
+        // see the spec https://github.com/open-telemetry/opentelemetry-specification/blob/v1.20.0/specification/trace/semantic_conventions/http.md
+        if (emitOldAttributes)
+        {
+            foreach (var tag in activity.TagObjects)
```
cijothomas, line 412: this is quite different from how metrics are created for the .NET, .NET Core flavor. In those cases, even if a user does not enable AddHttpClientInstrumentation() on TracerProviderBuilder, metrics would flow fine.

This PR will require AddingHttpClientInstrumentation() done on TracerProviderBuilder, else no metrics would flow. Might be okay, but require an explicit callout in the documentation.

src/OpenTelemetry.Instrumentation.Http/Implementation/HttpWebRequestActivitySource.netfx.cs
matt-hensley: Cribbed from the .NET/.NET Core implementation that calls out relying on HttpClient setting the end time:
https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Instrumentation.Http/Implementation/HttpHandlerMetricsDiagnosticListener.cs#L100

src/OpenTelemetry.Instrumentation.Http/Implementation/HttpWebRequestActivitySource.netfx.cs
matt-hensley: 100% agree, it'll need a prominent notice that tracing is required. I couldn't see a clean way to just enable metrics without a decent amount of rework and risk. I'd consider this an intermediate step towards fully separate metrics and traces.

FWIW, the auto-instrumentation requires ASP.NET tracing to be enabled to get metrics on .NET Framework, so this would not be a unique situation for those on .NET Framework still.

src/OpenTelemetry.Instrumentation.Http/Implementation/HttpWebRequestActivitySource.netfx.cs
cijothomas: Don't think its needed here. activity is stopped before we derive metrics from it, so duration is already set.

src/OpenTelemetry.Instrumentation.Http/README.md
```diff
@@ -68,7 +68,7 @@ public class Program
 #### Metrics
 
 > **Note**
-> Metrics are not available for .NET Framework.
+> Metrics are only available for .NET Framework when tracing is enabled.
```
cijothomas, line 71: I think we need to add more info here. Its not just sufficient to enable Tracing, but the sampler should also be returning Recorded+Sampled for each span in-order for metrics to work. As Metrics rely on Activity to populate tags, and Activity itself won't be populated unless Sampler decision is favorable.

src/OpenTelemetry.Instrumentation.Http/Implementation/HttpWebRequestActivitySource.netfx.cs
```diff
@@ -396,6 +402,47 @@ private static void ProcessResult(IAsyncResult asyncResult, AsyncCallback asyncC
         }
 
         activity.Stop();
+
+        TagList tags = default;
+
+        // see the spec https://github.com/open-telemetry/opentelemetry-specification/blob/v1.20.0/specification/trace/semantic_conventions/http.md
+        if (emitOldAttributes)
+        {
+            foreach (var tag in activity.TagObjects)
```
vishweshbankwar, line 411: This adds the overhead of looping through the tags even if someone is not using metrics. Could we add a check here before looping?

src/OpenTelemetry.Instrumentation.Http/MeterProviderBuilderExtensions.cs
```diff
@@ -54,7 +68,22 @@ public static class MeterProviderBuilderExtensions
         //   RecordException - probably doesn't make sense for metric instrumentation
 
         builder.AddMeter(HttpClientMetrics.InstrumentationName);
```
vishweshbankwar, line 70: this should go to #else?

src/OpenTelemetry.Instrumentation.Http/MeterProviderBuilderExtensions.cs
```diff
@@ -34,12 +34,26 @@ public static class MeterProviderBuilderExtensions
...
+        return AddHttpClientInstrumentation(builder, name: null);
+    }
+
+    /// <summary>
+    /// Enables HttpClient instrumentation.
+    /// </summary>
+    /// <param name="builder"><see cref="MeterProviderBuilder"/> being configured.</param>
+    /// <param name="name">Name which is used when retrieving options.</param>
+    /// <returns>The instance of <see cref="MeterProviderBuilder"/> to chain the calls.</returns>
+    public static MeterProviderBuilder AddHttpClientInstrumentation(
```
TimothyMothra, line 47: Hi, you're getting some public API warnings here.
It looks like you only updated the net462 files.
You'll need to update the files for the other frameworks as well (net6.0 and netstandard2.0).

The tooling here is not obvious :)
You can either do this manually, or remove your changes and let Intellisense apply the changes for the full project.

src/OpenTelemetry.Instrumentation.Http/Implementation/HttpWebRequestActivitySource.netfx.cs
TimothyMothra, line 411: I think you'll want to use this pattern here:
```suggestion
            foreach (ref readonly var tag in activity.EnumerateTagObjects())
```

src/OpenTelemetry.Instrumentation.Http/Implementation/HttpWebRequestActivitySource.netfx.cs
```diff
@@ -396,6 +402,47 @@ private static void ProcessResult(IAsyncResult asyncResult, AsyncCallback asyncC
...
+                        tags.Add(new KeyValuePair<string, object>(tag.Key, tag.Value));
+                        break;
+                }
+            }
+        }
+
+        // see the spec https://github.com/open-telemetry/semantic-conventions/blob/v1.21.0/docs/http/http-spans.md
+        if (emitNewAttributes)
+        {
+            foreach (var tag in activity.TagObjects)
```
TimothyMothra, line 430: same as above:
```suggestion
            foreach (ref readonly var tag in activity.EnumerateTagObjects())
```

src/OpenTelemetry.Instrumentation.Http/Implementation/HttpWebRequestActivitySource.netfx.cs
matt-hensley: TIL https://github.com/dotnet/runtime/issues/68056

src/OpenTelemetry.Instrumentation.Http/README.md
matt-hensley: @cijothomas please check my latest revision, better clarified what is needed to enable tracing

src/OpenTelemetry.Instrumentation.Http/README.md
```diff
@@ -68,7 +68,9 @@ public class Program
 #### Metrics
 
 > **Note**
-> Metrics are not available for .NET Framework.
+> Metrics are only available for .NET Framework when traces are recorded. This requires:
+> 1. Tracing to be enabled by calling `.AddHttpClientInstrumentation()` on `TracerProviderBuilder`
+> 2. [If a sampler is in place, return `SamplingDecision.RecordAndSampled` for `ShouldSample`](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/docs/trace/extending-the-sdk/README.md#filtering-processor)
```
cijothomas, line 73: nits:
_"If a sampler"_ is not right, as there is **always** a sampler! Suggestion:
The sampler must return `RecordAndSample` for *ever* `Activity` in order for the Metrics to be accurate. If some spans (Activity) are sampled, but others not, it'll affect Metrics accuracy, as Metric Tags (Attributes) are generated based on the Activity.

cijothomas: also, the user provided Filters can also affect this, https://github.com/open-telemetry/opentelemetry-dotnet/blob/06c704b99f8c101299e0f2b90f6f11a0f15da065/src/OpenTelemetry.Instrumentation.Http/Implementation/HttpWebRequestActivitySource.netfx.cs#L274.

1. Enable tracing.
2. Sampler returns RecordAndSample
3. No user defined filters.

cijothomas: I suggest revisit the design of relying on Activity for generating metrics, and instead produce Metrics independently.
The approach done in this PR is more like a "Span2Metrics" processor, and can done as a regular `ActivityProcessor` instead, so it'll be easy to convey that Metrics are only as good as the underlying Activity.

Could you check how big of a change it'd to produce metrics independently? (It may not be as bad to rely on Activity for duration alone...but not for the whole tags.)

PR #4772, opened 2023-08-15
Remove unnecessary `RequiresUnreferencedCode` attribute.

vitek-karas:
The entire type is marked with `RequiresUnreferencedCode` so it is implicitly inherited by all code inside that type (except for nested types).

Also - this method doesn't do anything which would make it produce trim warnings, it only produces AOT warning (which is explicitly suppressed).


vitek-karas:
@Yun-Ting FYI

Yun-Ting (approved):
LGTM! Thanks for following up on this.

reyang (approved):

utpilla (approved):

Issue #4324, opened 2023-03-22
Duplicate Log Attribute Keys


c4brei:
# Question

Use [Github Discussions](https://github.com/open-telemetry/opentelemetry-dotnet/discussions/new).

Currently when using Logging attributes in Scope and LogMessage I observed, that the dotnet exporter exports every attribute not just the latest occurrence.
```
using (logger.BeginScope(new[] { new KeyValuePair<string, object>("fooAttr", "abc - scope0") }))    // scope 0
using (logger.BeginScope(new[] { new KeyValuePair<string, object>("fooAttr", "def - scope1") }))    // scope 1
{
    logger.LogInformation("Hello Log Message: '{fooAttr}' '{barAttr}'", "ghi - log msg", "barAttrValue");
}
```
In this example `fooAttr` is added 3 times to the attribute list

Output of ConsoleLogExporter:
> ```
> LogRecord.Timestamp:               2023-03-22T12:44:02.5096056Z
> LogRecord.CategoryName:            Program
> LogRecord.LogLevel:                Information
> LogRecord.FormattedMessage:        Hello Log Message: 'ghi - log msg' 'barAttrValue'
> LogRecord.StateValues (Key:Value):
>     fooAttr: ghi - log msg
>     barAttr: barAttrValue
>     OriginalFormat (a.k.a Body): Hello Log Message: '{fooAttr}' '{barAttr}'
> LogRecord.ScopeValues (Key:Value):
> [Scope.0]:fooAttr: abc - scope0
> [Scope.1]:fooAttr: def - scope1
> ```

OpenTelemetry Collector receives list with duplicate attribute keys:
> ```
> Timestamp: 2023-03-22 12:44:02.5096056 +0000 UTC
> SeverityText: Information
> SeverityNumber: Info(9)
> Body: Str(Hello Log Message: 'ghi - log msg' 'barAttrValue')
> Attributes:
>      -> dotnet.ilogger.category: Str(Program)
>      -> fooAttr: Str(ghi - log msg)
>      -> barAttr: Str(barAttrValue)
>      -> {OriginalFormat}: Str(Hello Log Message: '{fooAttr}' '{barAttr}')
>      -> fooAttr: Str(abc - scope0)
>      -> fooAttr: Str(def - scope1)
> Trace ID:
> Span ID:
> Flags: 0
> ```

This could be an unintentional side effect from the removal of the scope depth prefix (#3843).
[...]

PR #4781, opened 2023-08-17
[otlp] Remove ilogger and exception attributes

vishweshbankwar:
Fixes #
Design discussion issue #

## Changes
This PR excludes the following fields on `LogRecord` from being exported by the OTLP exporter:

* `EventId`
* `CategoryName`
* `Exception`

### DIFF
--- a/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/CHANGELOG.md
+++ b/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/CHANGELOG.md
@@ -3,7 +3,8 @@
 ## Unreleased
 
 * Excluded attributes corresponding to `LogRecord.EventId`,
-`LogRecord.CategoryName` and `LogRecord.Exception` from the exported data.
+`LogRecord.CategoryName` and `LogRecord.Exception` from the exported data. This
+is done as the semantic conventions for these attributes are not yet stable.
 ([#4781](https://github.com/open-telemetry/opentelemetry-dotnet/pull/4781))
 
 * Added extension method for configuring export processor options for otlp log
--- a/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/CHANGELOG.md
+++ b/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 * Excluded attributes corresponding to `LogRecord.EventId`,
 `LogRecord.CategoryName` and `LogRecord.Exception` from the exported data. This
-is done in order to move ahead with stable release of exporter. THe attributes
+is done in order to move ahead with stable release of exporter. The attributes
 will be brought back in the pre-release version post stable release.
 ([#4781](https://github.com/open-telemetry/opentelemetry-dotnet/pull/4781))
 
--- a/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/CHANGELOG.md
+++ b/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+* Excluded attributes corresponding to `LogRecord.EventId`,
+`LogRecord.CategoryName` and `LogRecord.Exception` from the exported data. This
+is done in order to move ahead with stable release of exporter. THe attributes
+will be brought back in the pre-release version post stable release.
+([#4781](https://github.com/open-telemetry/opentelemetry-dotnet/pull/4781))
+
 * Added extension method for configuring export processor options for otlp log
 exporter.
 ([#4733](https://github.com/open-telemetry/opentelemetry-dotnet/pull/4733))
--- a/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/CHANGELOG.md
+++ b/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/CHANGELOG.md
@@ -3,9 +3,7 @@
 ## Unreleased
 
 * Excluded attributes corresponding to `LogRecord.EventId`,
-`LogRecord.CategoryName` and `LogRecord.Exception` from the exported data. This
-is done in order to move ahead with stable release of exporter. The attributes
-will be brought back in the pre-release version post stable release.
+`LogRecord.CategoryName` and `LogRecord.Exception` from the exported data.
 ([#4781](https://github.com/open-telemetry/opentelemetry-dotnet/pull/4781))
 
 * Added extension method for configuring export processor options for otlp log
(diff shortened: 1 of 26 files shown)


src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/LogRecordExtensions.cs
```diff
@@ -80,6 +78,10 @@ internal static OtlpLogs.LogRecord ToOtlpLog(this LogRecord logRecord, SdkLimitO
             var attributeValueLengthLimit = sdkLimitOptions.AttributeValueLengthLimit;
             var attributeCountLimit = sdkLimitOptions.AttributeCountLimit ?? int.MaxValue;
 
+            /*
+            // Removing this temporarily for stable release
+            // https://github.com/open-telemetry/opentelemetry-dotnet/issues/4776
+            // https://github.com/open-telemetry/opentelemetry-dotnet/issues/3491
```
alanwest, line 84: We should also consider removing scope attributes for the initial stable release

https://github.com/open-telemetry/opentelemetry-dotnet/blob/79295b4e5489d952dd81ce2ccd12b54b94c31beb/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/LogRecordExtensions.cs#L154

There is an existing issue related to ILogger's scope attributes we have not yet resolved #4324.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+* Excluded attributes corresponding to `LogRecord.EventId`,
+`LogRecord.CategoryName` and `LogRecord.Exception` from the exported data. This
+is done in order to move ahead with stable release of exporter. The attributes
```
utpilla, line 7: > This is done in order to move ahead with stable release of exporter.

I don't think we need to have this on the CHANGELOG. It's better to cite the reason as the convention for exporting these attributes not being stable.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/LogRecordExtensions.cs
utpilla: I think there's more value to exporting scope attributes as majority of the users who use scopes wouldn't run into the duplicate attributes issue and they would benefit from OTLPExporter exporting these attributes.

For #4324, I don't think the SDK should handle it. We can just put the onus on the user to not provide duplicate values or just have the backend systems deal with it.

alanwest: I haven't thought this through enough to provide a well-informed opinion. Though, if others share this stance and you're confident it's the right path, then I suggest closing #4324 as "by design". Perhaps this behavior is worth documenting?

cijothomas: de-dup is going to be expensive, so we should not do it by default (we are not doing it for Metrics also). Post stable, we could consider (based on actual demand) to offer a Exporter specific feature to dedup, so the dedup cost is occurring in the dedicated thread, not in the main user threads.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+* Excluded attributes corresponding to `LogRecord.EventId`,
+`LogRecord.CategoryName` and `LogRecord.Exception` from the exported data. This
```
cijothomas, line 6: Something to consider:
Removing Exception could be undone. Agree its experimental, but 1.0 (of Trace) onwards, we have done this already: https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Api/Trace/ActivityExtensions.cs#L86 despite the convention being experimental! 

Would be good at checking if Java is doing this in stable or as experimental only! (Either way, this is lower priority than the goal is shipping stable OTLP, so this is no way a blocking comment.)

alanwest: It is doubtful that Java has this to deal with in their exporter given their data model does not include an exception field.

With instrumentation Java is well organized in managing expectations utilizing semver appropriately with regards to experimental conventions. We definitely do not want to couple the ongoing stabilization of conventions to the potential need of a major version bump of our OTLP exporter.

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/OtlpLogExporterTests.cs
```diff
@@ -557,6 +562,51 @@ public void CheckToOtlpLogRecordRespectsAttributeLimits()
         var exceptionStackTraceAtt = TryGetAttribute(otlpLogRecord, SemanticConventions.AttributeExceptionStacktrace);
         Assert.Null(exceptionStackTraceAtt);
     }
+    */
+
+    // Remove this when adding back the category, eventid and exception attributes
+    [Fact]
+    public void CheckToOtlpLogRecordRespectsAttributeLimits()
```
alanwest, line 569: I think we should keep this test long term. I see you pulled in this test from the refactor I did on #4762.

I think you should refactor the existing test in place like I did on #4762 because it better reflects what the name of the test suggests it does - i.e., validates attributes are truncated. It was weird that this test was conflating testing attribute truncation only for the exception attributes.

alanwest (approved):
LGTM. One small thing to address regarding a test.

utpilla (approved):

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/OtlpLogExporterTests.cs
vishweshbankwar: Done

CodeBlanch (approved):
LGTM

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+* Excluded attributes corresponding to `LogRecord.EventId`,
```
cijothomas, line 5: its better if we add a stronger message here, as this is a significant breaking change for users who are relying on these features.

*Breaking*/*Please Note* or something to catch attention.

samsp-msft: Not having the EventId & Category properties makes it harder to filter log messages to the ones that you care about. They could go in the attributes as they are SDK specific.

amccool: > 

+1

amccool: > Not having the EventId & Category properties makes it harder to filter log messages to the ones that you care about. They could go in the attributes as they are SDK specific.

+1

austindrenski: > Not having the EventId & Category properties makes it harder to filter log messages to the ones that you care about. They could go in the attributes as they are SDK specific.

Indeed. Missed this in last week's release notes and ran into this today. Was a real head-scratcher for a few hours because this is the last place I expected these (fairly ubiquitous) dotnet attrs to be getting dropped ☹️

PR #4787, opened 2023-08-18
Add logging for metric overflow attribute

utpilla:
## Changes
- Add logging for metric overflow attribute

### DIFF
--- a/src/OpenTelemetry/Metrics/MeterProviderSdk.cs
+++ b/src/OpenTelemetry/Metrics/MeterProviderSdk.cs
@@ -64,6 +64,8 @@ internal sealed class MeterProviderSdk : MeterProvider
 
         OpenTelemetrySdkEventSource.Log.MeterProviderSdkEvent("Building MeterProvider.");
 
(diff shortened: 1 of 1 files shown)


TimothyMothra (approved):
LGTM

vishweshbankwar (approved):

src/OpenTelemetry/Metrics/MeterProviderSdk.cs
```diff
@@ -64,6 +64,8 @@ internal sealed class MeterProviderSdk : MeterProvider
 
         OpenTelemetrySdkEventSource.Log.MeterProviderSdkEvent("Building MeterProvider.");
 
+        OpenTelemetrySdkEventSource.Log.MeterProviderSdkEvent($"Metric overflow attribute key set to: {isEmitOverflowAttributeKeySet}");
```
cijothomas, line 67: a more user friendly message would "Experimental feature "metric overflow attribute" is enabled/disabled".

The current msg could be interpreted as "key" used for overflow is set to true/false instead of "otel.metric.overflow". 🤭

utpilla: The intended audience for verbose logging is us so it should be fine 😀 I agree that we could still make it more user-friendly.

PR #4802, opened 2023-08-24
HttpSemanticConventions - new metric in AspNetCore Instrumentation. 

TimothyMothra:
Towards #4484
Supersedes #4766

This PR focus on introducing the new `http.server.request.duration` metrics in the AspNetCore Instrumentation library.

## Changes
- all the following changes are made behind the environment variable `OTEL_SEMCONV_STABILITY_OPT_IN`
- update Instrumentation.AspNetCore class HttpInMetricsListener to emit new metric.
  - `http.server.duration` (milliseconds) replaced with `http.server.request.duration` (~seconds~ ~milliseconds~ seconds)
    This is defined in the as-of-yet unreleased v1.22.0 https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-metrics.md
- remove opt-in attributes
  - `server.port`
  - `server.address`

## Future work
- ~The unit of this metric needs to be changed to seconds. This requires a different default histogram.~ DONE! 


src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInMetricsListener.cs
```diff
@@ -44,6 +47,10 @@ internal HttpInMetricsListener(string name, Meter meter, AspNetCoreMetricsInstru
         this.options = options;
         this.httpServerDuration = meter.CreateHistogram<double>(HttpServerDurationMetricName, "ms", "Measures the duration of inbound HTTP requests.");
 
+        // TODO: This needs to be changed to  "s" (seconds). This is blocked until we can change the default histogram.
+        // See: https://github.com/open-telemetry/opentelemetry-dotnet/issues/4797
+        this.httpServerRequestDuration = meter.CreateHistogram<double>(HttpServerRequestDurationMetricName, "ms", "Measures the duration of inbound HTTP requests.");
```
vishweshbankwar, line 52: Need to check for flag here as well.

cijothomas, line 52: do we create 2 instruments now?

vishweshbankwar: https://github.com/open-telemetry/opentelemetry-dotnet/pull/4802#discussion_r1304988605 We should only create 2 if user has opted in for both old and new(feature-freeze) conventions. @TimothyMothra - Could you please check if Java is following the same approach.

TimothyMothra: Looks like Java is emiting one or the other, but not both.
https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/c2a3ba02d3ad1d89e332932f0af3665b0afec362/instrumentation-api-semconv/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpServerMetrics.java#L61-L64

This conflicts with my understanding. Everywhere else we've used the environment variable, we've had to consider the both scenario. I'll touch base with @Trask and get his take.

cijothomas: https://github.com/open-telemetry/opentelemetry-specification/blob/59455fc1662415c9b4c1a5aea10f4aac0b9eddc7/specification/metrics/semantic_conventions/http-metrics.md?plain=1#L51

trask: > This conflicts with my understanding. Everywhere else we've used the environment variable, we've had to consider the both scenario. I'll touch base with @trask and get his take.

I agree 👍, I've sent this on the Java side: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/9320

(also note https://github.com/open-telemetry/semantic-conventions/pull/278)

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInMetricsListener.cs
```diff
@@ -144,7 +148,17 @@ public override void OnEventWritten(string name, object payload)
             // We are relying here on ASP.NET Core to set duration before writing the stop event.
             // https://github.com/dotnet/aspnetcore/blob/d6fa351048617ae1c8b47493ba1abbe94c3a24cf/src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs#L449
             // TODO: Follow up with .NET team if we can continue to rely on this behavior.
-            this.httpServerDuration.Record(Activity.Current.Duration.TotalMilliseconds, tags);
+            if (this.emitNewAttributes)
+            {
+                // TODO: This needs to be changed to TotalSeconds. This is blocked until we can change the default histogram.
+                // See: https://github.com/open-telemetry/opentelemetry-dotnet/issues/4797
+                this.httpServerRequestDuration.Record(Activity.Current.Duration.TotalMilliseconds, tags);
```
vishweshbankwar, line 155: same tags should not be used for both old and new.

TimothyMothra: good catch. updated.

src/OpenTelemetry.Instrumentation.AspNetCore/README.md
```diff
@@ -90,13 +90,22 @@ public void ConfigureServices(IServiceCollection services)
...
-The instrumentation is implemented based on [metrics semantic
-conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/http-metrics.md#metric-httpserverduration).
-Currently, the instrumentation supports the following metric.
+The instrumentation is implemented based on
+[metrics semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/4bbb8c907402caa90bc077214e8a2c78807c1ab9/docs/http/http-metrics.md).
 
-| Name  | Instrument Type | Unit | Description |
-|-------|-----------------|------|-------------|
-| `http.server.duration` | Histogram | `ms` | Measures the duration of inbound HTTP requests. |
+Have you opt-ed into the new Http Semantic Conventions?
```
vishweshbankwar, line 96: nit: Mention the environment variable name `OTEL_SEMCONV_STABILITY_OPT_IN` here for clarity.

TimothyMothra: done.

src/OpenTelemetry.Instrumentation.AspNetCore/README.md
```diff
@@ -90,13 +90,22 @@ public void ConfigureServices(IServiceCollection services)
...
-| Name  | Instrument Type | Unit | Description |
-|-------|-----------------|------|-------------|
-| `http.server.duration` | Histogram | `ms` | Measures the duration of inbound HTTP requests. |
+Have you opt-ed into the new Http Semantic Conventions using `OTEL_SEMCONV_STABILITY_OPT_IN`?
+
+* If yes, the instrumentation supports the following metric.
+
+    | Name  | Instrument Type | Unit | Description |
+    |-------|-----------------|------|-------------|
+    | `http.server.request.duration` | Histogram | `s` | Measures the duration of inbound HTTP requests. |
```
cijothomas, line 102: it'd be good to callout here that SDK uses a special treatment for this metric by using a suitable buckets.
https://github.com/open-telemetry/opentelemetry-dotnet/pull/4820/files

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInMetricsListener.cs
```diff
@@ -84,43 +109,87 @@ public override void OnEventWritten(string name, object payload)
...
+            // We are relying here on ASP.NET Core to set duration before writing the stop event.
+            // https://github.com/dotnet/aspnetcore/blob/d6fa351048617ae1c8b47493ba1abbe94c3a24cf/src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs#L449
+            // TODO: Follow up with .NET team if we can continue to rely on this behavior.
+            this.httpServerDuration.Record(Activity.Current.Duration.TotalMilliseconds, tags);
+        }
+    }
+
+    public void OnEventWritten_New(string name, object payload)
+    {
+        if (name == OnStopEvent)
```
vishweshbankwar, line 155: nit: you can move this check to `OnEventWritten`

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInMetricsListener.cs
```diff
@@ -84,43 +109,87 @@ public override void OnEventWritten(string name, object payload)
...
+            // If we want to suppress activity from Prometheus then we should use SuppressInstrumentationScope.
+            if (context.Request.Path.HasValue && context.Request.Path.Value.Contains("metrics"))
+            {
+                return;
+            }
+
+            TagList tags = default;
+
+            // see the spec https://github.com/open-telemetry/semantic-conventions/blob/v1.21.0/docs/http/http-spans.md
+            tags.Add(new KeyValuePair<string, object>(SemanticConventions.AttributeNetworkProtocolVersion, HttpTagHelper.GetFlavorTagValueFromProtocol(context.Request.Protocol)));
```
vishweshbankwar, line 189: There is [network.protocol.name](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/attributes.md#:~:text=network.protocol.name) attribute as well. The value would be same as `url.scheme` in this case. @TimothyMothra Please follow up on this to see if we need to add that as well. Can be done in follow up PR.

src/OpenTelemetry.Instrumentation.AspNetCore/README.md
TimothyMothra: How much information are you looking for?
Is it sufficient to say "http.server.request.duration uses unique buckets suitable for seconds" ?

I could include a link to the spec which shows the buckets:
https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-metrics.md#metric-httpserverrequestduration

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/MetricTests.cs
```diff
@@ -47,47 +49,85 @@ public void AddAspNetCoreInstrumentation_BadArgs()
...
-        this.meterProvider.Dispose();
+            using (var client = this.factory
+                .WithWebHostBuilder(builder =>
+                {
+                    builder.ConfigureLogging(loggingBuilder => loggingBuilder.ClearProviders());
+                })
+                .CreateClient())
+            {
+                using var response1 = await client.GetAsync("/api/values").ConfigureAwait(false);
+                using var response2 = await client.GetAsync("/api/values/2").ConfigureAwait(false);
```
vishweshbankwar, line 77: not sure why we need two calls here. Could be simplified by making just one call

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/MetricTests.cs
```diff
@@ -47,47 +49,85 @@ public void AddAspNetCoreInstrumentation_BadArgs()
         Assert.Throws<ArgumentNullException>(() => builder.AddAspNetCoreInstrumentation());
     }
 
-    [Fact]
-    public async Task RequestMetricIsCaptured()
+    [Theory]
+    [InlineData(null, true, false, 6, 0)] // emits old metric & attributes
+    [InlineData("http", false, true, 0, 5)] // emits new metric & attributes
+    [InlineData("http/dup", true, true, 6, 5)] // emits both old & new
+    public async Task RequestMetricIsCaptured(string environmentVarValue, bool validateOldSemConv, bool validateNewSemConv, int expectedOldTagsCount, int expectedNewTagsCount)
```
vishweshbankwar, line 56: I think it will be easier to do separate 3 tests. Will be easier to simply remove 2 of them later when we get rid of env variable.

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/MetricTests.cs
TimothyMothra: This was here before my changes. It appears to be testing with and without paramaters.
`expectedRoutes: new List<string> { "api/Values", "api/Values/{id}" },`

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/MetricTests.cs
TimothyMothra: No strong opinion here. In previous PRs where I split tests out, I got pushback.

I do see that these tests are failing. I assume there's a concurrency issue with the Environment Variable.

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInMetricsListener.cs
TimothyMothra: done

src/OpenTelemetry.Instrumentation.AspNetCore/README.md
cijothomas: something along the lines of
"This metric is emitted in "s" as unit, as per the semantic convention guide. The convention also recommends to provide a custom buckets (link), but that capability is not yet available in OTel .NET. A workaround is already included in the OTel SDK (link), which specially treats this metric, and uses the recommended buckets"

^ feel free to edit/modify!

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/MetricTests.cs
vishweshbankwar: Its bit tricky to review the tests when they are combined in to one. When we remove them later it will require more work to readjust these tests rather than simply deleting the unwanted ones. @utpilla do you have any strong preference here?

>  I assume there's a concurrency issue with the Environment Variable.
Let's use IConfiguration.

TimothyMothra: > Let's use IConfiguration.

100% agree. I looked into this earlier. The Options class has a ctor overload to take IConfiguration.
But I don't see an option to provide an instance of Options to the AddAspNetCoreInstrumentation() method.

Do you have any examples of how to do this?

vishweshbankwar: https://github.com/open-telemetry/opentelemetry-dotnet/blob/459783f700b87325e057493d0715338d93e2de63/test/OpenTelemetry.Instrumentation.Grpc.Tests/GrpcTests.server.cs#L126-L132

utpilla: > I think it will be easier to do separate 3 tests. Will be easier to simply remove 2 of them later when we get rid of env variable.

I agree with that. It would make it easier to remove the unwanted tests.

> No strong opinion here. In previous PRs where I split tests out, I got pushback.

@TimothyMothra Was that me? 😀. IIRC you had convinced me with the same rationale back then that it would make it easier to remove unwanted tests in future. We have dedicated unit tests for traces instrumentation, right? One for each possible value of the environment variable?

TimothyMothra: I think it was you 😄 Yeah, we've been trying to separate tests methods. I can make that change here

src/OpenTelemetry.Instrumentation.AspNetCore/README.md
TimothyMothra: updated. please re-review

src/OpenTelemetry.Instrumentation.AspNetCore/README.md
```diff
@@ -90,13 +90,28 @@ public void ConfigureServices(IServiceCollection services)
...
+
+    This metric is emitted in "seconds" as per the semantic convention. While
+    the convention [recommends using custom histogram buckets](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-metrics.md#metric-httpserverrequestduration)
+    , this feature is not yet available in OTel .NET. However, the OTel SDK
```
vishweshbankwar, line 106:
```suggestion
    , this feature is not yet available via .NET metrics API.  A workaround has been included in OTel SDK starting version `1.6.0` which applies recommended buckets by default for this metric.
```

vishweshbankwar: Also, this needs to be included in changelog: No action is needed from users targeting latest version of instrumentation library as that will bring in 1.6.0 SDK version by default.

src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+* Added new metric `http.server.request.duration` which replaces
+  `http.server.duration` if a user opted into the new semantic convention
+  by setting the `OTEL_SEMCONV_STABILITY_OPT_IN` environment variable.
+  This new metric is emitted in seconds and has unique histogram buckets.
```
vishweshbankwar, line 8: Link to spec for buckets

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/MetricTests.cs
```diff
@@ -212,10 +347,46 @@ private static List<MetricPoint> GetMetricPoints(Metric metric)
...
+            }
+        }
+    }
+
     private static KeyValuePair<string, object>[] AssertMetricPoint(
         MetricPoint metricPoint,
         string expectedRoute = "api/Values",
-        int expectedTagsCount = StandardTagsCount)
+        int expectedTagsCount = StandardTagsCount,
+        bool validateNewSemConv = false,
```
vishweshbankwar, line 388: This still won't allow us to just delete code without refactoring :)

Could we split this in to two? I do not see code reuse anyways in this method.

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/MetricTests.cs
```diff
@@ -212,10 +347,46 @@ private static List<MetricPoint> GetMetricPoints(Metric metric)
         return metricPoints;
     }
 
+    private static void AssertMetricPoints(
+        List<MetricPoint> metricPoints,
+        List<string> expectedRoutes,
+        int expectedTagsCount,
+        bool validateNewSemConv = false,
```
vishweshbankwar, line 354: same as https://github.com/open-telemetry/opentelemetry-dotnet/pull/4802/files#r1323337701

src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
TimothyMothra: done

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/MetricTests.cs
TimothyMothra: done

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/MetricTests.cs
TimothyMothra: done

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/MetricTests.cs
```diff
@@ -83,11 +91,134 @@ public async Task RequestMetricIsCaptured()
...
+        var requestMetrics = metricItems
+            .Where(item => item.Name == "http.server.duration")
+            .ToArray();
+
+        var metric = Assert.Single(requestMetrics);
+        Assert.Equal("ms", metric.Unit);
+        var metricPoints = GetMetricPoints(metric);
+        Assert.Equal(2, metricPoints.Count);
+
+        AssertMetricPoints_Old(
```
vishweshbankwar, line 202: to simplify this, we should remove multiple route values as that is not what we are testing here. This can be done as a follow up as its not something that is introduced in this PR.

src/OpenTelemetry.Instrumentation.AspNetCore/README.md
```diff
@@ -90,13 +90,29 @@ public void ConfigureServices(IServiceCollection services)
...
-| `http.server.duration` | Histogram | `ms` | Measures the duration of inbound HTTP requests. |
+Have you opt-ed into the new Http Semantic Conventions using `OTEL_SEMCONV_STABILITY_OPT_IN`?
+
+* If yes, the instrumentation supports the following metric.
+
+    | Name  | Instrument Type | Unit | Description |
+    |-------|-----------------|------|-------------|
+    | `http.server.request.duration` | Histogram | `s` | Measures the duration of inbound HTTP requests. |
+
+    This metric is emitted in "seconds" as per the semantic convention. While
```
vishweshbankwar, line 104:
```suggestion
    This metric is emitted in `seconds` as per the semantic convention. While
```

src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+* Added new metric `http.server.request.duration` which replaces
+  `http.server.duration` if a user opted into the new semantic convention
+  by setting the `OTEL_SEMCONV_STABILITY_OPT_IN` environment variable.
+  This new metric is emitted as seconds and has unique
```
vishweshbankwar, line 8: **Note** : The unit of `http.server.request.duration` is `seconds` which has changed from `milliseconds` in `http.server.duration`.  The histogram buckets are adjusted accordingly as per the [spec](link)

vishweshbankwar (approved):
LGTM: with some suggestions on changelog.

src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
```diff
@@ -2,6 +2,18 @@
 
 ## Unreleased
 
+* Introduced a new metric, `http.server.request.duration`, for users who opt
+  into the new semantic convention by configuring the `OTEL_SEMCONV_STABILITY_OPT_IN`
```
utpilla, line 6:
```suggestion
  in to the new semantic convention by configuring the `OTEL_SEMCONV_STABILITY_OPT_IN`
```

src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
```diff
@@ -2,6 +2,18 @@
...
+  environment variable. This metric measures time in seconds and offers unique
+  histogram buckets as per the
+  [spec](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-metrics.md#metric-httpserverrequestduration)
+  , replacing the previous `http.server.duration metric`, which measured time in
```
utpilla, line 10:
```suggestion
  , replacing the previous `http.server.duration` metric, which measured time in
```

src/OpenTelemetry.Instrumentation.AspNetCore/README.md
```diff
@@ -90,13 +90,29 @@ public void ConfigureServices(IServiceCollection services)
...
-The instrumentation is implemented based on [metrics semantic
-conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/http-metrics.md#metric-httpserverduration).
-Currently, the instrumentation supports the following metric.
+The instrumentation is implemented based on
+[metrics semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/4bbb8c907402caa90bc077214e8a2c78807c1ab9/docs/http/http-metrics.md).
 
-| Name  | Instrument Type | Unit | Description |
-|-------|-----------------|------|-------------|
-| `http.server.duration` | Histogram | `ms` | Measures the duration of inbound HTTP requests. |
+Have you opt-ed into the new Http Semantic Conventions using `OTEL_SEMCONV_STABILITY_OPT_IN`?
```
utpilla, line 96:
```suggestion
Have you opted in to the new Http Semantic Conventions using `OTEL_SEMCONV_STABILITY_OPT_IN`?
```

src/OpenTelemetry.Instrumentation.AspNetCore/README.md
```diff
@@ -90,13 +90,29 @@ public void ConfigureServices(IServiceCollection services)
...
+
+* If yes, the instrumentation supports the following metric.
+
+    | Name  | Instrument Type | Unit | Description |
+    |-------|-----------------|------|-------------|
+    | `http.server.request.duration` | Histogram | `s` | Measures the duration of inbound HTTP requests. |
+
+    This metric is emitted in `seconds` as per the semantic convention. While
```
utpilla, line 104:
```suggestion
    This metric value is emitted in `seconds` as per the semantic convention. While
```

src/OpenTelemetry.Instrumentation.AspNetCore/README.md
```diff
@@ -90,13 +90,29 @@ public void ConfigureServices(IServiceCollection services)
...
+    the convention [recommends using custom histogram buckets](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-metrics.md#metric-httpserverrequestduration)
```
utpilla, line 105: Use the commit tag?

src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
```diff
@@ -2,6 +2,18 @@
 
 ## Unreleased
 
+* Introduced a new metric, `http.server.request.duration`, for users who opt
+  into the new semantic convention by configuring the `OTEL_SEMCONV_STABILITY_OPT_IN`
+  environment variable. This metric measures time in seconds and offers unique
+  histogram buckets as per the
+  [spec](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-metrics.md#metric-httpserverrequestduration)
```
utpilla, line 9: Same as https://github.com/open-telemetry/opentelemetry-dotnet/pull/4802/files#r1323869142

utpilla (approved):
Left a few suggestions for CHANGELOG and README.

src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
```diff
@@ -2,6 +2,18 @@
 
 ## Unreleased
 
+* Introduced a new metric, `http.server.request.duration`, for users who opt
+  into the new semantic convention by configuring the `OTEL_SEMCONV_STABILITY_OPT_IN`
+  environment variable. This metric measures time in seconds and offers unique
+  histogram buckets as per the
```
cijothomas, line 8: > This metric measures time in seconds and offers unique
>   histogram buckets

Not sure how to interpret "offers unique histogram buckets"? What does it mean?

src/OpenTelemetry.Instrumentation.AspNetCore/README.md
```diff
@@ -90,13 +90,29 @@ public void ConfigureServices(IServiceCollection services)
...
-Currently, the instrumentation supports the following metric.
+The instrumentation is implemented based on
+[metrics semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/4bbb8c907402caa90bc077214e8a2c78807c1ab9/docs/http/http-metrics.md).
 
-| Name  | Instrument Type | Unit | Description |
-|-------|-----------------|------|-------------|
-| `http.server.duration` | Histogram | `ms` | Measures the duration of inbound HTTP requests. |
+Have you opt-ed into the new Http Semantic Conventions using `OTEL_SEMCONV_STABILITY_OPT_IN`?
+
+* If yes, the instrumentation supports the following metric.
```
cijothomas, line 98: this is not clear. The env variable can take none,http,http/dup as values, This doc should also be structured in that way. i.e what if the env variable is set to "none", (also the default)
what is the env varaible is set to "http" 
what is the env varaible is set to "http/dup" - both metric are emitted

src/OpenTelemetry.Instrumentation.AspNetCore/README.md
cijothomas: I'd highlight/make it clear that the workaround is "OpenTelemetry SDK specially treats this metric and uses a diff. default for histogram buckets, matching the recommended buckets" /similar wording. This will ensure most folks do not have any need to go and read the linked PR.

cijothomas: the linked PR is very light on details. That must be fixed separately. Specifically, 
1. it should list the exact name of the metrics for which SDK uses a diff. default.
2. make it clear that the special treatment is only for the default case. If user overrides the buckets, then that'll always be respected. 

(Not part of this PR, but this is essential to get the whole transition smoother for end users)

src/OpenTelemetry.Instrumentation.AspNetCore/README.md
TimothyMothra: I'm conflicted here. 
I would rather link to an explicit tag. There changes we're implementing now are post v1.21.0, but this repo hasn't made a new release yet.

I can change links in both the Readme and Changelog to a commit as you've suggested.
When the next release happens (presumably v1.22.0) I think we should update these links.

src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
TimothyMothra: "offers" is the wrong word here, "uses" would be more appropriate.
I'm trying to call attention to the fact that the buckets are different than the default.
4 lines below this I list the exact buckets.


I'm trying to find a succinct way to describe that this metric is different than what came before.
@cijothomas, do you think this would be better as a bulleted list?

ie:
```
- new metric: `http.server.request.duration`
- unit: seconds
- buckets: 0, 0.005, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1,
  2.5, 5, 7.5, 10
```

src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
```diff
@@ -2,6 +2,18 @@
...
+  [spec](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-metrics.md#metric-httpserverrequestduration)
+  , replacing the previous `http.server.duration metric`, which measured time in
+  milliseconds.
+  * Former buckets: `0, 5, 10, 25, 50, 75, 100, 250, 500, 750, 1000, 2500,
```
rajkumar-rangaraj, line 12: Would this qualify as a breaking change?

TimothyMothra: This histogram is not a breaking change because it's only applied to the new metric.

rajkumar-rangaraj (approved):

src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
rajkumar-rangaraj: It's a new and a replacement metrics. If spec does not treat it as breaking change we are fine.

src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
cijothomas: the list is good. Hint that the buckets are so, because SDK specially handles this metric.

TimothyMothra:
@utpilla, @cijothomas, @vishweshbankwar 
I've rewritten both the Readme and Changelog. Please re-review.

src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
cijothomas: Its not breaking change **now**, as it requires explicit opt-in.

src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
```diff
@@ -2,6 +2,26 @@
 
 ## Unreleased
 
+* Introduced a new metric, `http.server.request.duration`, for users who opt-in
```
cijothomas, line 5: Introduced a new metric, `http.server.request.duration`, for users who opt-in
to the new semantic convention by configuring the `OTEL_SEMCONV_STABILITY_OPT_IN`
environment variable to value "fillthis". This metric measures time in **seconds**, but histogram buckets are adjusted
automatically for this metric by OTel SDK which [specially treat this](https://github.com/open-telemetry/opentelemetry-dotnet/pull/4820) metric. This metric follows the OTel Semantic Conventions from (add link).
  * Metric Name: `http.server.request.duration`
  * Unit: `seconds`
  * Histogram Buckets: `0, 0.005, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5,
    0.75, 1,  2.5, 5, 7.5, 10` (due to OpenTelemetry SDK's special treatment)
  * Replaces old metric: `http.server.duration`
    * Unit: `miliseconds`
    * Histogram Buckets: `0, 5, 10, 25, 50, 75, 100, 250, 500, 750, 1000, 2500,
    5000, 7500, 10000` (the general default of Histogram buckets)

`OTEL_SEMCONV_STABILITY_OPT_IN` may be set to "fillthis", to help an easier transition, at the cost of emitting both old and new metrics.

^did some refactoring. Feel free to take it with modifications. 
A user who do not have any experience with OTel sem. conventions and the new vs old, should find it easy to follow the changelog. Key things I'd prefer to convey here are:
1. No breaking changes if you just update to this new version.
2. Opt-in feature to emit a new metric that follows a new convention, which despite using "s", still allows good histogram precision due to SDK treating it specially.
3. Hint at emitting both to smooth transition.
4. Hint that the new metric is the long term bet, as it'll be the one getting stabilizied by Otel sem convention.

TimothyMothra: Thanks for the refactor. I'll take another pass at this today.

Do we typically put instructions in the changelog? I was trying to keep the changelog succinct, and use the readme for the full information. The challenge is that this change has a LOT of details :)

TimothyMothra: just pushed my rewrite. I think I've addressed all your points. please re-review.

src/OpenTelemetry.Instrumentation.AspNetCore/README.md
```diff
@@ -90,13 +90,31 @@ public void ConfigureServices(IServiceCollection services)
 
 #### List of metrics produced
 
-The instrumentation is implemented based on [metrics semantic
-conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/http-metrics.md#metric-httpserverduration).
-Currently, the instrumentation supports the following metric.
+A different metric is emitted depending if a user opts-in to the new Http
```
utpilla, line 93:
```suggestion
A different metric is emitted depending on whether a user opts-in to the new Http
```

utpilla (approved):

utpilla:
@TimothyMothra Thanks for your patience on working on this PR!

TimothyMothra:
> @TimothyMothra Thanks for your patience on working on this PR!

This one was tricky and touched a lot of parts! Glad we could get this merged.

PR #4812, opened 2023-08-30
[OpenTracing.Shim] TracerShim constructor

### LINKED ISSUE #4727: OpenTracing shim constructor doesn't match spec
Per spec, the operation to create the OpenTracing Tracer shim should receive a `TracerProvider` and use it to create a `Tracer` with the name `opentracing-shim`. Currently, it receives a Tracer and lets the user select the name of the `Tracer`:

https://github.com/open-telemetry/opentelemetry-dotnet/blob/ce1e196b49b4a817b5814f141e177c5d9edbd948/examples/Console/TestOpenTracingShim.cs#L40-L42

Of course, removing the current constructor will be a breaking change. Since no stable versions of the OT shim package were released so far it seems that this should be done rather soon. The current constructor could be marked as obsolete for one or two releases before removing it completely.

Kielek:
Fixes #4727

## Changes

TracerShim constructor met https://github.com/open-telemetry/opentelemetry-specification/blob/90eec0ffd8e52dddbd308e346cd5098f611164dd/specification/compatibility/opentracing.md
Old constructor marked as obsolete. It can be removed now/after this release.
2 new constructors to met criteria. If there is no propagator, the default one is used.
Unit tests covers only non-obsolete constructor.

### DIFF
--- a/examples/Console/TestOpenTracingShim.cs
+++ b/examples/Console/TestOpenTracingShim.cs
@@ -27,18 +27,18 @@ internal class TestOpenTracingShim
 {
     internal static object Run(OpenTracingShimOptions options)
     {
-        // Enable OpenTelemetry for the source "MyCompany.MyProduct.MyWebServer"
+        // Enable OpenTelemetry for the source "opentracing-shim"
         // and use Console exporter.
         using var tracerProvider = Sdk.CreateTracerProviderBuilder()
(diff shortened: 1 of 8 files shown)


pjanotti (approved):
LGTM. Thanks @Kielek

utpilla (approved):

PR #4821, opened 2023-09-01
Add ASP.NET Core logging example

reyang:
Still need to update the *.sln file and some README files.


docs/logs/getting-started-aspnetcore/README.md
```diff
@@ -0,0 +1,73 @@
...
+telemetry.sdk.name: opentelemetry
+telemetry.sdk.language: dotnet
+telemetry.sdk.version: 1.6.0-rc.1.15
+```
+
+Congratulations! You are now collecting logs using OpenTelemetry.
+
+What does the above program do?
+
+The program has configured the ASP.NET Core logging pipeline by enabling
```
cijothomas, line 64: more like "Added OpenTelemetry as a logging provider to the existing logging pipeline.. OpenTelemetry logger provider is paired with its ConsoleExporter that prints the logs in the console."..

Also, this as good place to disambiguate/explain why this vs the ConsoleLoggerProvider is different.

docs/logs/getting-started-aspnetcore/HelloLogs.cs
```diff
@@ -0,0 +1,24 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+
+public static partial class HelloLogs
+{
+    [LoggerMessage(
+        EventId = 1,
+        Level = LogLevel.Information,
+        Message = "Hello, world!")]
```
cijothomas, line 22: stick with fruit name and price, so it can show how attributes can come?

docs/logs/getting-started-aspnetcore/README.md
```diff
@@ -0,0 +1,73 @@
...
+LogRecord.CategoryName:            Program
+LogRecord.Severity:                Info
+LogRecord.SeverityText:            Information
+LogRecord.Body:                    Hello, world!
+LogRecord.Attributes (Key:Value):
+    OriginalFormat (a.k.a Body): Hello, world!
+LogRecord.EventId:                 1
+LogRecord.EventName:               SayHello
+LogRecord.ScopeValues (Key:Value):
+[Scope.0]:SpanId: b888feebe74e9685
```
cijothomas, line 45: lets put an explanation here saying this scope is coming from asp.net core itself. and warn that some of the fields (like TraceId etc) are duplicate.

Separate:
(Its a good feature to provide feature to control scopes like option to limit the number of scope level/depth and which direction. A lot of users end up with all scopes from asp.net core itself, though their goal was to merely add a common userid/transactionid to whole logs themselves...)

docs/logs/getting-started-aspnetcore/Program.cs
```diff
@@ -0,0 +1,44 @@
...
+        .CreateDefault()
+        .AddService(builder.Environment.ApplicationName);
+
+    logging.SetResourceBuilder(resourceBuilder)
+        .AddConsoleExporter();
+});
+
+var app = builder.Build();
+
+app.MapGet("/", (ILogger<Program> logger) =>
```
reyang, line 37: I don't know what's the actual recommendation from .NET team, do we want to use the DI logger or something different (e.g. https://learn.microsoft.com/en-us/aspnet/core/fundamentals/logging#log-in-programcs)?

reyang: @davidfowl @noahfalk @tarekgh FYI

docs/logs/getting-started-aspnetcore/README.md
reyang: > more like "Added OpenTelemetry as a logging provider to the existing logging pipeline.. OpenTelemetry logger provider is paired with its ConsoleExporter that prints the logs in the console."..

Good point, I've updated the wording.

> Also, this as good place to disambiguate/explain why this vs the ConsoleLoggerProvider is different.

I feel this is out of scope, should be explained somewhere else (e.g. ConsoleExporter perhaps).

docs/logs/getting-started-aspnetcore/README.md
```diff
@@ -0,0 +1,98 @@
...
+and
+[OpenTelemetry.Extensions.Hosting](../../../src/OpenTelemetry.Extensions.Hosting/README.md)
+packages:
+
+```sh
+dotnet add package OpenTelemetry.Exporter.Console
+dotnet add package OpenTelemetry.Extensions.Hosting
+```
+
+Copy the [HelloLogs.cs](./HelloLogs.cs) and [Program.cs](./Program.cs) files to
```
cijothomas, line 24: FoodSupplyLogs.cs

docs/logs/getting-started-aspnetcore/README.md
```diff
@@ -0,0 +1,98 @@
...
+    contentRoot: D:\repo\opentelemetry-dotnet\docs\logs\getting-started-aspnetcore\
+    OriginalFormat (a.k.a Body): Content root path: {contentRoot}
+
+Resource associated with LogRecord:
+service.name: getting-started-aspnetcore
+service.instance.id: d4f09b09-6613-4d09-8a2d-1508e4ce9067
+telemetry.sdk.name: opentelemetry
+telemetry.sdk.language: dotnet
+telemetry.sdk.version: 1.6.0-rc.1.16
+
```
cijothomas, line 47: would be nice to modify consoleexporter separately to put some marker between each LogRecord. Like "======"/similar.

docs/logs/getting-started-aspnetcore/README.md
```diff
@@ -0,0 +1,98 @@
...
+LogRecord.SeverityText:            Information
+LogRecord.Body:                    Food `{name}` price changed to `{price}`.
+LogRecord.Attributes (Key:Value):
+    name: artichoke
+    price: 9.99
+    OriginalFormat (a.k.a Body): Food `{name}` price changed to `{price}`.
+LogRecord.EventId:                 1
+LogRecord.EventName:               FoodPriceChanged
+LogRecord.ScopeValues (Key:Value):
+[Scope.0]:SpanId: 46e671d30d10df22
```
cijothomas, line 63: neutral about this part... given this is beginner/getting-started guide, maybe don't enable scopes at all here, to keep it really simple.

cijothomas: especially since each exporter handles scope differently. (OTLP does not prefix "Scope.depth" to scopes, but console does....)

docs/logs/getting-started-aspnetcore/Program.cs
```diff
@@ -0,0 +1,44 @@
...
+builder.Logging.AddOpenTelemetry(logging =>
+{
+    logging.IncludeScopes = true;
+    logging.ParseStateValues = true;
+
+    var resourceBuilder = ResourceBuilder
+        .CreateDefault()
+        .AddService(builder.Environment.ApplicationName);
+
+    logging.SetResourceBuilder(resourceBuilder)
```
cijothomas, line 31: most of the code is now showing `ConfigureResource` pattern, so its best to do that for consistency:
https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/docs/trace/getting-started-aspnetcore/Program.cs#L25

docs/logs/getting-started-aspnetcore/Program.cs
cijothomas: that example from asp.net core docs looks confusing as it shows the approach of using `LoggerFactory` directly, but no other Asp.Net Core examples shows that approach!

cijothomas: https://github.com/dotnet/AspNetCore.Docs/issues/30239 reported issue to asp.net core doc.

docs/logs/getting-started-aspnetcore/FoodSupplyLogs.cs
```diff
@@ -0,0 +1,24 @@
...
+// limitations under the License.
+// </copyright>
+
+public static partial class FoodSupplyLogs
+{
+    [LoggerMessage(
+        EventId = 1,
+        Level = LogLevel.Information,
+        Message = "Food `{name}` price changed to `{price}`.")]
+    public static partial void FoodPriceChanged(this ILogger logger, string name, double price);
```
noahfalk, line 23: Creating a new 25 line file for this one method feels very verbose given the ASP.NET Core minimal APIs were designed to embrace minimalism. I'd suggest defining this class in Program.cs to avoid the extra file. Other doc samples already do these types of inline class definitions ([example](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/minimal-apis/responses?view=aspnetcore-7.0#configure-json-serialization-options-globally)). Starting in Microsoft.Extensions.Logging.Abstractions v8 EventId can be auto-generated from a hash of the method name and more stream-lined constructor overloads are available so you might reformat this as:

```C#
public static partial class FoodSupplyLogs
{
    [LoggerMessage(LogLevel.Information, "Food `{name}` price changed to `{price}`.")]
    public static partial void FoodPriceChanged(this ILogger logger, string name, double price);
}
```

If your sample isn't yet using version 8 (reasonable given its not quite out yet) then the current formatting is fine and I'd suggest putting it in Program.cs.

docs/logs/getting-started-aspnetcore/README.md
noahfalk: I would leave the scopes enabled in the sample code under the assumption that many developers will want them and discovering them when they are missing from the sample is much harder. I do think it would be fine to abbreviate the output more.

docs/logs/getting-started-aspnetcore/FoodSupplyLogs.cs
davidfowl: Right, for hello world, we want to show simple logging, then we want to show how to get to this point when you "mature". I wouldn't start with this in the minimal API docs.

noahfalk: @davidfowl are you suggesting that the OTel docs include multiple samples describing this progression and the reasoning behind when you choose to "mature"? I worry that will make the docs 2x longer than they are now and many devs are going to stop reading as soon as they have the first example that works. IMO the increment in complexity is small enough that if we had their attention long enough for only one example, I'd want them to see an example that uses LoggerMessage.

docs/logs/getting-started-aspnetcore/HelloLogs.cs
reyang: Updated.

docs/logs/getting-started-aspnetcore/README.md
reyang: Out of the scope of this PR, I think we should really output YAML by default.

docs/logs/getting-started-aspnetcore/Program.cs
reyang: There is OpenTelemetry feature/issue which we need to solve before making the examples consistent across logs/metrics/traces.

docs/logs/getting-started-aspnetcore/README.md
```diff
@@ -0,0 +1,132 @@
...
+and
+[OpenTelemetry.Extensions.Hosting](../../../src/OpenTelemetry.Extensions.Hosting/README.md)
+packages:
+
+```sh
+dotnet add package OpenTelemetry.Exporter.Console
+dotnet add package OpenTelemetry.Extensions.Hosting
+```
+
+Copy the [FoodSupplyLogs.cs](./FoodSupplyLogs.cs) and [Program.cs](./Program.cs)
```
cijothomas, line 24:
```suggestion
Copy the [Program.cs](./Program.cs)
```

docs/logs/getting-started-aspnetcore/README.md
```diff
@@ -0,0 +1,132 @@
...
+
+What does the above program do?
+
+The program has added OpenTelemetry as a [logging
+provider](https://learn.microsoft.com/dotnet/core/extensions/logging-providers)
+to the existing logging pipeline. OpenTelemetry SDK is then configured with a
+`ConsoleExporter` to export the logs to the console. In addition,
+`OpenTelemetryLoggerOptions.IncludeScopes` is enabled so the logs will include
+the [log
+scopes](https://learn.microsoft.com/aspnet/core/fundamentals/logging/#log-scopes).
```
cijothomas, line 88: https://learn.microsoft.com/en-us/dotnet/core/extensions/logging?tabs=command-line#log-scopes maybe used, as we use .NET logging docs elsewhere, and not the asp.net core docs,

docs/logs/getting-started-aspnetcore/appsettings.Development.json
```diff
@@ -0,0 +1,9 @@
+{
+  "DetailedErrors": true,
```
cijothomas, line 2: what does this do?

docs/logs/getting-started-aspnetcore/README.md
```diff
@@ -0,0 +1,132 @@
...
+```
+
+Copy the [FoodSupplyLogs.cs](./FoodSupplyLogs.cs) and [Program.cs](./Program.cs)
+files to the project folder.
+
+Run the application again (using `dotnet run`) and then browse to the URL shown
+in the console for your application (e.g. `http://localhost:5000`). You should see
+the logs output from the console.
+
+```text
```
cijothomas, line 31: update the output with latest run.

docs/logs/getting-started-aspnetcore/appsettings.Development.json
reyang: I think this is the default experience for development (e.g. showing callstacks).

docs/logs/getting-started-aspnetcore/README.md
cijothomas: We have traditionally kept getting started as the minimal thing to get started. (https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/docs/trace/getting-started-aspnetcore/Program.cs#L27-L29 - it does not even customize sampler which most people do in their real apps).
Scopes is a common use case, but not necessarily a requirement for getting-started.

cijothomas (approved):
LGTM. 
I personally don't prefer showing Scopes, ParseStateValues in the getting-started guide, but no blockers.

docs/logs/getting-started-aspnetcore/FoodSupplyLogs.cs
reyang: I've made the updates based on our discussion, PTAL.

docs/logs/getting-started-aspnetcore/Program.cs
```diff
@@ -0,0 +1,55 @@
...
+
+using OpenTelemetry.Logs;
+using OpenTelemetry.Resources;
+
+var builder = WebApplication.CreateBuilder(args);
+
+builder.Logging.AddOpenTelemetry(logging =>
+{
+    logging.IncludeScopes = true;
+    logging.ParseStateValues = true;
```
cijothomas, line 25: @CodeBlanch Whats your thoughts on `ParseStateValues` in this getting started example? Its non-trivial to explain, and probably not required as well?

docs/logs/getting-started-aspnetcore/FoodSupplyLogs.cs
davidfowl: I don't think it needs to be 2x bigger to show these examples

docs/logs/getting-started-aspnetcore/Program.cs
noahfalk: As one anecdotal data point, I saw this and had no idea what it does or why the sample sets it. That isn't to say its wrong, but I've yet to go research what this does.

docs/logs/getting-started-aspnetcore/README.md
```diff
@@ -0,0 +1,150 @@
...
+```csharp
+app.MapGet("/", (ILogger<Program> logger) =>
+{
+    logger.FoodPriceChanged("artichoke", 9.99);
+
+    return "Hello from OpenTelemetry Logs!";
+});
+```
+
+To achieve structured logging and high performance, [compile-time logging source
```
noahfalk, line 123: You still get structured logging via other APIs as well. What you don't get in the other APIs are type checked parameters.

docs/logs/getting-started-aspnetcore/README.md
```diff
@@ -0,0 +1,150 @@
...
+public static partial class ApplicationLogs
+{
+    [LoggerMessage(EventId = 1, Level = LogLevel.Information, Message = "Starting the app...")]
+    public static partial void StartingApp(this ILogger logger);
+
+    [LoggerMessage(EventId = 2, Level = LogLevel.Information, Message = "Food `{name}` price changed to `{price}`.")]
+    public static partial void FoodPriceChanged(this ILogger logger, string name, double price);
+}
+```
+
```
noahfalk, line 137: I think it would be fine to mention somewhere around here that there are also less verbose alternatives like logger.LogInformation(...) but you lose out on some performance and type checked parameters if you choose to use them. Personally I'd also be fine not mentioning it, but I think both Sam and Fowler were in favor of covering the option. I'm happy as long as anyone whose attention expired before reaching this point is going to use LoggerMessage, which I consider the option with higher likelihood of a good outcome in the long run.

docs/logs/getting-started-aspnetcore/FoodSupplyLogs.cs
noahfalk: You are right it can be done more succinctly if we use only small snippets rather than replicating the full example. I made a [comment](https://github.com/open-telemetry/opentelemetry-dotnet/pull/4821/files#r1318516277) where I think that stuff could go.

docs/logs/getting-started-aspnetcore/README.md
davidfowl: We should show the inverse. Show log information and then mention logger message.

cijothomas: Those are better covered in ILogger tutorial/docs. This doc is intended to show how to use OpenTelemetry with ILogger, so its best to just show *the recommended*  way of logging, instead of showing all/more-than-one possible ways of ILogger usage itself.

noahfalk: > We should show the inverse. Show log information and then mention logger message.

As I understand it, @reyang's team has concrete past experience that if we do that it leads to questions and support tickets. A subset of devs copy whatever code is in the first snippet, don't read the following text, then scale up that usage. Its a deliberate choice that if we only have their attention span long enough to copy/paste one thing, lets have that thing be the performant and scalable option rather than the succinct option.

reyang: > > We should show the inverse. Show log information and then mention logger message.
> 
> As I understand it, @reyang's team has concrete past experience that if we do that it leads to questions and support tickets. A subset of devs copy whatever code is in the first snippet, don't read the following text, then scale up that usage.

+100, based on the support tickets and questions we've observed during the past 2 years, I think we should only provide pit of success, and we must not show any negative example that are not recommended for real production development. In addition, with the trend of LLM/ChatGPT/CoPilot, I think we should consider how good/bad examples could influence AI.

docs/logs/getting-started-aspnetcore/Program.cs
```diff
@@ -0,0 +1,55 @@
...
+
+    var resourceBuilder = ResourceBuilder
+        .CreateDefault()
+        .AddService(builder.Environment.ApplicationName);
+
+    logging.SetResourceBuilder(resourceBuilder)
+        .AddConsoleExporter();
```
davidfowl, line 32: I don't see code disabling the built in console logger.

docs/logs/getting-started-aspnetcore/README.md
noahfalk: > we must not show any negative example that are not recommended for real production development

I wouldn't go as far as saying the other APIs are unsuitable or not recommended for production, they just have a suitable scale threshold that is lower than LoggerMessage.

reyang: > > we must not show any negative example that are not recommended for real production development
> 
> I wouldn't go as far as saying the other APIs are unsuitable or not recommended for production, they just have a suitable scale threshold that is lower than LoggerMessage.

Fair, I think we should also try to avoid slippery slopes.

docs/logs/getting-started-aspnetcore/Program.cs
reyang: I've made the update.

docs/logs/getting-started-aspnetcore/Program.cs
reyang: Good point. I've removed it. Probably something we should cover in a different example (e.g. customizing the SDK).

docs/logs/getting-started-aspnetcore/README.md
```diff
@@ -0,0 +1,128 @@
...
+```sh
+dotnet add package OpenTelemetry.Exporter.Console
+dotnet add package OpenTelemetry.Extensions.Hosting
+```
+
+Update the `Program.cs` file with the code from [Program.cs](./Program.cs).
+
+Run the application (using `dotnet run`) and then browse to the URL shown in the
+console for your application (e.g. `http://localhost:5000`). You should see the
```
cijothomas, line 27: since we removed built-in ConsoleProvider, do we still see the url in console?

docs/logs/getting-started-aspnetcore/README.md
```diff
@@ -0,0 +1,128 @@
...
+The program has added OpenTelemetry as a [logging
+provider](https://learn.microsoft.com/dotnet/core/extensions/logging-providers)
+to the existing logging pipeline. OpenTelemetry SDK is then configured with a
+`ConsoleExporter` to export the logs to the console. In addition,
+`OpenTelemetryLoggerOptions.IncludeScopes` is enabled so the logs will include
+the [log
+scopes](https://learn.microsoft.com/dotnet/core/extensions/logging#log-scopes).
+From the console output we can see the log scopes that are coming from the
+ASP.NET Core framework, and we can see logs from both our logger and the ASP.NET
+Core framework loggers.
```
cijothomas, line 87:
```suggestion
Core framework loggers, as indicated by the LogRecord.CategoryName.
```

maybe?

docs/logs/getting-started-aspnetcore/README.md
reyang: Yes, from the OTel console exporter.

docs/logs/getting-started-aspnetcore/README.md
reyang: I've reworded this section, PTAL.

docs/logs/getting-started-aspnetcore/README.md
```diff
@@ -0,0 +1,150 @@
...
+console for your application (e.g. `http://localhost:5000`). You should see the
+logs output from the console:
+
+```text
+LogRecord.Timestamp:               2023-09-06T22:59:17.9787564Z
+LogRecord.CategoryName:            getting-started-aspnetcore
+LogRecord.Severity:                Info
+LogRecord.SeverityText:            Information
+LogRecord.Body:                    Starting the app...
+LogRecord.Attributes (Key:Value):
```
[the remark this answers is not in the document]
reyang: I've simplified the output a bit, still decided to keep ~30 lines to cover the following points for folks who simply want to read through the article without having to run the actual code:
1. There are different categories of logs coming from user code and ASP.NET Core.
2. The actual binding URL (which the user needs to access via a browser).
3. The logs from DI and app.Logger (different category names).

cijothomas (approved):

docs/logs/getting-started-aspnetcore/README.md
```diff
@@ -0,0 +1,129 @@
...
+LogRecord.EventName:               FoodPriceChanged
+
+...
+```
+
+Congratulations! You are now collecting logs using OpenTelemetry.
+
+What does the above program do?
+
+The program has added OpenTelemetry as a [logging
```
cijothomas, line 78: nit: mention we cleared all existing providers?

Kielek:
@reyang, not directly related to this PR, but it will be great to propagate your example (with some description) also to https://opentelemetry.io/docs/instrumentation/net/manual/

cijothomas:
> @reyang, not directly related to this PR, but it will be great to propagate your example (with some description) also to https://opentelemetry.io/docs/instrumentation/net/manual/

that particular page need a lot of [other fixes too](https://github.com/open-telemetry/opentelemetry.io/issues/3252)!

reyang:
> @reyang, not directly related to this PR, but it will be great to propagate your example (with some description) also to https://opentelemetry.io/docs/instrumentation/net/manual/

I know, I think the 1st step is that we agree on a good example (current we don't have a good example) and that's what this PR is about.

docs/logs/getting-started-aspnetcore/Program.cs
```diff
@@ -0,0 +1,55 @@
...
+builder.Logging.ClearProviders().AddOpenTelemetry(logging =>
+{
+    logging.IncludeScopes = true;
+
+    var resourceBuilder = ResourceBuilder
+        .CreateDefault()
+        .AddService(builder.Environment.ApplicationName);
+
+    logging.SetResourceBuilder(resourceBuilder)
+        .AddConsoleExporter();
```
reyang, line 32: Sam suggested that we should consider OTLP + some local viewer tool.

reyang: For this document - keep OTel ConsoleExporter, remove Console Provider, leave a comment with OTLP exporter, and mention in the README. Later when a backend tool (e.g. something like Jaeger) is ready, revisit the example and see if we should switch to OTLP + backend tool.

For the .NET official doc, use the default (Console Provider), link to OpenTelemetry doc (which will be added to .NET docs) for folks who want to export in other ways.

docs/logs/getting-started-aspnetcore/README.md
reyang: Decided to keep the scopes.

docs/logs/getting-started-aspnetcore/Program.cs
reyang: I've made the update.

docs/logs/getting-started-aspnetcore/Program.cs
```diff
@@ -0,0 +1,57 @@
...
+    return "Hello from OpenTelemetry Logs!";
+});
+
+app.Logger.StartingApp();
+
+app.Run();
+
+public static partial class ApplicationLogs
+{
+    [LoggerMessage(EventId = 1, Level = LogLevel.Information, Message = "Starting the app...")]
```
tarekgh, line 52: What do you think to have a small comment pointing at the link https://learn.microsoft.com/en-us/dotnet/core/extensions/logger-message-generator?

tarekgh: nevermind, I am seeing you have some text in the readme about it.

tarekgh (approved):
LGTM. 

CC @noahfalk @samsp-msft if having any more comments.

noahfalk (approved):
LGTM modulo one suggestion inline

docs/logs/getting-started-aspnetcore/getting-started-aspnetcore.csproj
```diff
@@ -0,0 +1,14 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
```
utpilla, line 3: This property group is not required. All of these properties are already defined in `Common.props` and `Directory.Build.props`.

utpilla (approved):

docs/logs/getting-started-aspnetcore/getting-started-aspnetcore.csproj
utpilla: We can remove that in a separate PR.

utpilla: #4583

davidfowl:
Maybe this is by design but clearing all of the loggers clears:
- The event log logger on windows (there's no console on IIS)
- The console logger (but we add a replacement one)
- The event source logger (so logs can't be collected with etw/event pipe)

PR #4838, opened 2023-09-07
Fix grpc exporter connecting error port when use insecure request on …

### LINKED ISSUE #4837: Grpc exporter is not compatible with insecure requests on port 80.
# Bug Report

* OpenTelemetry.Exporter.OpenTelemetryProtocol 1.5.1

* net48

## Symptom

``` C#
var endpoint = new Uri("http://172.17.12.73");
#if NETFRAMEWORK
// If replace `endpoint.Authority` with `endpoint.Host, endpoint.Port`, will not get a failure.
var channel = new Channel(endpoint.Authority, endpoint.Scheme == Uri.UriSchemeHttps ? ChannelCredentials.SecureSsl : ChannelCredentials.Insecure);
#else
var channel = GrpcChannel.ForAddress(endpoint);
#endif

var client = new Greeter.GreeterClient(channel);

try
{
    await client.SayHelloAsync(new());
}
catch (RpcException ex)
{
    Console.WriteLine(ex.Status);
}

Console.WriteLine(channel.State);
```

`ChannelState: Ready`

```
Status(StatusCode="Unavailable", Detail="failed to connect to all addresses", DebugException="Grpc.Core.Internal.CoreErrorDetailException: {"created":"@1694568219.677000000","description":"Failed to pick subchannel","file":"..\..\..\src\core\ext\filters\client_channel\client_channel.cc","file_line":3218,"referenced_errors":[{"created":"@1694568219.677000000","description":"failed to connect to all addresses","file":"..\..\..\src\core\lib\transport\error_utils.cc","file_line":165,"grpc_status":14}]}")
ChannelState: TransientFailure
```

> Use wireshark,  I didn't see any TCP connections while the code was running.

## Additional Context
.net 6.0+ have no error.

pengweiqhca:
…port 80 (net462 and netstandard2.0 only).

Fixes #4837
Design discussion issue #

## Changes


src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpExporterOptionsExtensions.cs
```diff
@@ -55,7 +55,7 @@ public static Channel CreateChannel(this OtlpExporterOptions options)
             channelCredentials = ChannelCredentials.Insecure;
         }
 
-        return new Channel(options.Endpoint.Authority, channelCredentials);
+        return new Channel(options.Endpoint.Host, options.Endpoint.Port, channelCredentials);
```
alanwest, line 58: I think this change is otherwise ok, however, we are discussing dropping our dependency on the old gRPC library. If we do, this code will go away. See #4860

PR #4855, opened 2023-09-15
Align console logging example with ASP.NET Core logging example

reyang:
Essentially stealing good stuff from #4821.


docs/logs/getting-started-console/Program.cs
```diff
@@ -17,33 +17,38 @@
 using Microsoft.Extensions.Logging;
 using OpenTelemetry.Logs;
 
-namespace SourceGeneration;
-
-public class Program
+using var loggerFactory = LoggerFactory.Create(builder =>
 {
-    public static void Main()
+    builder.AddOpenTelemetry(options =>
```
reyang, line 22: Do we want to demonstrate resources here?

utpilla: I think it's fine to not include `Resources` here. We could show that in customizing the SDK exmaple.

utpilla, line 22: Rename `options` to `logging` similar to #4821? We could use this style for the other examples as well.

utpilla (approved):

docs/logs/getting-started-console/Program.cs
vishweshbankwar: its there -> https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/docs/logs/customizing-the-sdk#setresourcebuilder

vishweshbankwar (approved):

cijothomas (approved):

docs/logs/getting-started-console/Program.cs
```diff
@@ -17,33 +17,38 @@
 using Microsoft.Extensions.Logging;
 using OpenTelemetry.Logs;
 
-namespace SourceGeneration;
-
-public class Program
+using var loggerFactory = LoggerFactory.Create(builder =>
```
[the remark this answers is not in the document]
reyang: What options do we have and which one do you like?
1. Remove `using`, leave Dispose to GC finalizer (if there is a finalizer).
2. Add a comment to tell the user.
3. Remove `using`, add an explicit `Dispose` at the end of the application.

I'll give 1) -100, 2) +1 3) +2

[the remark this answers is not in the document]
cijothomas: A combination of 2, 3 i.e remove using, and add explicit Dispose at the end. And comments saying
1. Loggerfactory must be kept active for the logging to work. (to warn about disposing too early)
2. Loggerfactory must be disposed at the end/shutdown to make sure any in-buffers telemetry is pushed out as well. (to warn about not disposing at all)

Not something to be fixed for this PR, as this issue is there for other getting-started docs as well

PR #4874, opened 2023-09-21
Use OTLP push mode in Prometheus tutorial

reyang:
Prometheus has added OTLP support https://prometheus.io/docs/prometheus/latest/feature_flags/#otlp-receiver.
Similar to #3940, this PR switches the tutorial from PrometheusExporter (PrometheusHttpListener to be precise) to OtlpExporter (which is stable).

### DIFF
--- a/docs/metrics/getting-started-prometheus-grafana/Program.cs
+++ b/docs/metrics/getting-started-prometheus-grafana/Program.cs
@@ -29,7 +29,7 @@ public class Program
     {
         using var meterProvider = Sdk.CreateMeterProviderBuilder()
             .AddMeter("MyCompany.MyProduct.MyLibrary")
-            .AddPrometheusHttpListener()
+            .AddOtlpExporter(options => { options.Endpoint = new Uri("http://localhost:9090/api/v1/otlp/v1/metrics"); })
             .Build();
 
         Console.WriteLine("Press any key to exit");
--- a/docs/metrics/getting-started-prometheus-grafana/Program.cs
+++ b/docs/metrics/getting-started-prometheus-grafana/Program.cs
@@ -29,10 +29,11 @@ public class Program
     {
         using var meterProvider = Sdk.CreateMeterProviderBuilder()
             .AddMeter("MyCompany.MyProduct.MyLibrary")
+            .AddConsoleExporter()
             .AddOtlpExporter(options =>
             {
-                options.Protocol = OpenTelemetry.Exporter.OtlpExportProtocol.HttpProtobuf;
                 options.Endpoint = new Uri("http://localhost:9090/api/v1/otlp/v1/metrics");
+                options.Protocol = OpenTelemetry.Exporter.OtlpExportProtocol.HttpProtobuf;
             })
             .Build();
 
--- a/docs/metrics/getting-started-prometheus-grafana/Program.cs
+++ b/docs/metrics/getting-started-prometheus-grafana/Program.cs
@@ -29,7 +29,11 @@ public class Program
     {
         using var meterProvider = Sdk.CreateMeterProviderBuilder()
             .AddMeter("MyCompany.MyProduct.MyLibrary")
-            .AddOtlpExporter(options => { options.Endpoint = new Uri("http://localhost:9090/api/v1/otlp/v1/metrics"); })
+            .AddOtlpExporter(options =>
+            {
+                options.Protocol = OpenTelemetry.Exporter.OtlpExportProtocol.HttpProtobuf;
+                options.Endpoint = new Uri("http://localhost:9090/api/v1/otlp/v1/metrics");
+            })
             .Build();
 
         Console.WriteLine("Press any key to exit");
--- a/docs/metrics/getting-started-prometheus-grafana/README.md
+++ b/docs/metrics/getting-started-prometheus-grafana/README.md
@@ -1,12 +1,12 @@
 # Getting Started with Prometheus and Grafana
 
 - [Export metrics from the application](#export-metrics-from-the-application)
-  - [Check results in the browser](#check-results-in-the-browser)
+  - [Check results in the console](#check-results-in-the-console)
 - [Collect metrics using Prometheus](#collect-metrics-using-prometheus)
-  - [Configuration](#configuration)
-  - [Start Prometheus](#start-prometheus)
+  - [Install and run Prometheus](#install-and-run-prometheus)
   - [View results in Prometheus](#view-results-in-prometheus)
 - [Explore metrics using Grafana](#explore-metrics-using-grafana)
+- [Final cleanup](#final-cleanup)
 - [Learn more](#learn-more)
 
 ## Export metrics from the application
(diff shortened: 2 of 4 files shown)


reyang:
I hit an error saying OTLP Exporter is using `snappy` compression while Prometheus only supports `gzip`

image

cijothomas:
> I hit an error saying OTLP Exporter is using `snappy` compression while Prometheus only supports `gzip`
> 
> image

Let me try another language OTLP Exporter to see if it works...

vishweshbankwar:
I have pushed a change to update protocol to http protobuf. With that I can see the metrics in prometheus
image

**However**, I still see that error message on console
> ts=2023-09-22T16:42:44.685Z caller=write_handler.go:205 level=error component=web msg="Error decoding remote write request" err="unsupported compression: snappy. Only \"gzip\" or no compression supported"
ts=2023-09-22T16:42:44.686Z caller=dedupe.go:112 component=remote level=error remote_name=5653f3 url=http://localhost:9090/api/v1/otlp/v1/metrics msg="non-recoverable error" count=4 exemplarCount=0 err="server returned HTTP status 400 Bad Request: unsupported compression: snappy. Only \"gzip\" or no compression supported"

I have tried myapp > otel-collector > prometheus(otlp endpoint) as well. That shows the same error. Looking in to this further to find the source of that error.

docs/metrics/getting-started-prometheus-grafana/README.md
```diff
@@ -91,54 +120,20 @@ while (!Console.KeyAvailable)
...
-
-Please note that we will need pass in `otel.yml` file as the argument:
-
-```console
-./prometheus --config.file=otel.yml
+```sh
+./prometheus --enable-feature=otlp-write-receiver
 ```
 
 ### View results in Prometheus
```
vishweshbankwar, line 139: @reyang Something I noticed (not related to a change in this PR). This instruction says 
> type `MyFruitCounter` in the expression bar of the UI; finally, click the execute

But when i run it, it is `MyFruitCounter_total` . see [screenshot](https://github.com/open-telemetry/opentelemetry-dotnet/pull/4874#issuecomment-1731748476)

docs/metrics/getting-started-prometheus-grafana/README.md
```diff
@@ -91,54 +120,20 @@ while (!Console.KeyAvailable)
...
-global:
-  scrape_interval: 10s
-  evaluation_interval: 10s
-scrape_configs:
-  - job_name: "otel"
-    static_configs:
-      - targets: ["localhost:9464"]
-```
+access. Run the `prometheus(.exe)` server executable with feature flag
+[otlp-receiver](https://prometheus.io/docs/prometheus/latest/feature_flags/#otlp-receiver)
```
vishweshbankwar, line 132:
```suggestion
[otlp-write-receive](https://prometheus.io/docs/prometheus/latest/feature_flags/#otlp-receiver)
```

reyang: The name is actually coming from the officlal doc / link.

vishweshbankwar: got it. That `-` confused me a bit.

docs/metrics/getting-started-prometheus-grafana/README.md
reyang: You're right, I've updated the docs (I didn't change the screenshot though).

vishweshbankwar (approved):
LGTM. We can update other examples too (Separately) 🎉

docs/metrics/getting-started-prometheus-grafana/README.md
```diff
@@ -23,46 +23,73 @@ cd getting-started-prometheus
...
+    })
+    .Build();
 ```
 
-`PrometheusHttpListener` is a wrapper that contains `PrometheusExporter`. With
-`AddPrometheusHttpListener()`, OpenTelemetry `PrometheusExporter` will export
-data via the endpoint defined by
-[PrometheusHttpListenerOptions.UriPrefixes](../../../src/OpenTelemetry.Exporter.Prometheus.HttpListener/README.md#uriprefixes),
-which is `http://localhost:9464/` by default.
+When we run the application, the `ConsoleExporter` was printing the metrics on
```
Yun-Ting, line 76: minor: inconsistent verb tense

Yun-Ting (approved):
LGTM. 
Kudos specifically to using console/OTLP exporters to explain the concept and good visual aids.

docs/metrics/getting-started-prometheus-grafana/README.md
reyang: Good catch, fixed (also fixed the Jaeger doc).

docs/metrics/getting-started-prometheus-grafana/Program.cs
```diff
@@ -29,7 +29,12 @@ public static void Main()
     {
         using var meterProvider = Sdk.CreateMeterProviderBuilder()
             .AddMeter("MyCompany.MyProduct.MyLibrary")
-            .AddPrometheusHttpListener()
+            .AddConsoleExporter()
```
noahfalk, line 32: How do you feel about eliminating the console exporter, and the steps of the tutorial that check the console. Partly I worry that people doing quick copy/paste will have an extra exporter they don't need, and partly checking the console makes the tutorial longer. It didn't feel like seeing the data in the console was adding much value vs. seeing it in Prometheus.

docs/metrics/getting-started-prometheus-grafana/Program.cs
```diff
@@ -29,7 +29,12 @@ public static void Main()
...
+            .AddOtlpExporter(options =>
+            {
+                options.Endpoint = new Uri("http://localhost:9090/api/v1/otlp/v1/metrics");
+                options.Protocol = OpenTelemetry.Exporter.OtlpExportProtocol.HttpProtobuf;
```
noahfalk, line 36: Nit: A using statement above and not including the full namespace here would be a more typical style.

docs/metrics/getting-started-prometheus-grafana/README.md
```diff
@@ -71,7 +98,9 @@ end
 
 Instrument --> | Measurements | MeterProvider
 
-MeterProvider --> | Metrics | MetricReader --> | Pull | PrometheusHttpListener
+MeterProvider --> | Metrics | MetricReader --> | Push | OtlpExporter
```
noahfalk, line 101: Talking about MetricReader feels like a level of detail unneeded for the tutorial. I expect user's mental model is something like:
Recorded measurements -> OpenTelemetry -> OtlpExporter -> Remote Prometheus server

docs/metrics/getting-started-prometheus-grafana/Program.cs
reyang: The default export interval is 60 seconds https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#periodic-exporting-metricreader.

I was considering the following scenarios:
1. A user tried the example, had the application running for 20 seconds without seeing data, got impatient and gave up.
2. A user ran into some issue while trying to tweak the code, metrics stopped flowing to Prometheus and the user ran out of idea how to troubleshoot.

I think the 1st one can be easily solved by explicitly specifying a shorter export period (e.g. 3 seconds). The problem is that we're recommending something that's against production practice.
The 2nd one is something I'm not sure, we can tell the user to enable internal troubleshooting logs, it's just much harder.

reyang: > Partly I worry that people doing quick copy/paste will have an extra exporter they don't need, and partly checking the console makes the tutorial longer.

Now that you've pointed it out, I've been thinking about this - how about that we just deprecate the stable version of Console Exporter from NuGet, and keep releasing it as alpha version? Or completely disable it if the application is running in Release mode, unless a special environment variable is specified?

docs/metrics/getting-started-prometheus-grafana/Program.cs
reyang: https://github.com/open-telemetry/opentelemetry-dotnet/pull/4878

docs/metrics/getting-started-prometheus-grafana/README.md
reyang: See if this is better https://github.com/open-telemetry/opentelemetry-dotnet/pull/4878.

docs/metrics/getting-started-prometheus-grafana/Program.cs
reyang: A simpler version is here https://github.com/open-telemetry/opentelemetry-dotnet/pull/4878.
(I convinced myself that users can follow the console tutorial so we don't need to duplicate the content here)

noahfalk: Thoughts for issue (1) above
- Short term: I'd suggest leaving the update interval at 60 seconds in the initial snippet, calling out in the text that the updates will take some time, and then demonstrating in a separate tiny code snippet how the code could be changed to make it update faster.
- Long term: I feel like there is a chronic problem in telemetry systems to use high latency by default that is a difficult or impossible setting to change. High latency is presumably more cost-efficient for long-term high scale monitoring, but it is awful for anything interactive such as demos, testing telemetry data flow, or seeing the effect of changes during deployments and incident response. Ideally telemetry protocols, ingestion mechanisms and tools would collaborate so that some initial signal of connectivity occurs with extremely low latency. Even better, though I imagine technically more involved, would be allowing the end-user/tool to request lower latency transmission on-demand while interactive work is being done.

For (2) its probably useful for OTel to have a troubleshooting guide for how to handle telemetry that isn't reaching the intended destination. The tutorial could link to that guide.

> how about that we just deprecate the stable version of Console Exporter from NuGet, and keep releasing it as alpha version? Or completely disable it if the application is running in Release mode (making it no-op), unless a special environment variable is specified?

Are you hearing from users that accidentally enabling the console exporter is a common problem? My guess is that as long as our samples don't encourage people to turn it on by default that would be sufficient protection and nothing more drastic would be needed. But if feedback showed that wasn't sufficient I'm not opposed to taking a stronger measure.

reyang: > For (2) its probably useful for OTel to have a troubleshooting guide for how to handle telemetry that isn't reaching the intended destination. The tutorial could link to that guide.

There is already a doc https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/src/OpenTelemetry#troubleshooting, I think the SDK troubleshooting story still has lots of room for improvements.

> Are you hearing from users that accidentally enabling the console exporter is a common problem?

Not very often, but I've seen couple times developers left the Console Exporter enabled in production and I asked them to remove it.
Stepping back a little bit, I hope that we have one console exporter (including the Console logging provider) that is production ready, and it follows certain standards (so it can better integrate with existing systems such as Fluentbit / OTel Collector) instead of using our own taste while formatting the output.

noahfalk: > I think the SDK troubleshooting story still has lots of room for improvements.

Something is better than nothing, but agreed plenty of room for improvement. What I was envisioning would be a topic within the trouble-shooting guide that gets a little more specific such as: "I configured sending telemetry via OTLP, but the telemetry backend doesn't show any data has arrived." Then it would discuss how to get additional console/log data, what are signs the data is being sent, where would errors appear, and what are some common errors.

PR #4902, opened 2023-09-29
Update examples to support net8.0

reyang:
Depends on #4904 and #4876.


docs/logs/getting-started-aspnetcore/Program.cs
```diff
@@ -56,9 +56,9 @@
 
 public static partial class ApplicationLogs
 {
-    [LoggerMessage(EventId = 1, Level = LogLevel.Information, Message = "Starting the app...")]
+    [LoggerMessage(Level = LogLevel.Information, Message = "Starting the app...")]
```
cijothomas, line 59: why EventId are removed?

docs/logs/getting-started-aspnetcore/README.md
```diff
@@ -35,7 +35,7 @@ LogRecord.SeverityText:            Information
 LogRecord.Body:                    Starting the app...
 LogRecord.Attributes (Key:Value):
     OriginalFormat (a.k.a Body): Starting the app...
-LogRecord.EventId:                 1
+LogRecord.EventId:                 225744744
```
cijothomas, line 38: ohh! "magic" commit was this, I suppose!

image

docs/logs/getting-started-aspnetcore/Program.cs
reyang: Starting with .NET 8, EventId will be automatically generated based on the hash of Function name, and this is considered the best practice. Check https://github.com/open-telemetry/opentelemetry-dotnet/pull/4821#discussion_r1316757504

docs/logs/getting-started-aspnetcore/README.md
reyang: Yup. https://github.com/open-telemetry/opentelemetry-dotnet/pull/4821#discussion_r1316757504

cijothomas: That is good to know! 
I think we need to mention that here, and call out that it is only if using v9 of m.e.logging.

docs/logs/getting-started-aspnetcore/Program.cs
noahfalk, line 59:
```suggestion
    [LoggerMessage(LogLevel.Information, "Starting the app...")]
```
I believe you can construct the attribute using positional parameters rather than naming them now. I'm guessing its obvious enough what each parameter is that we can use the more concise syntax.

noahfalk (approved):

docs/logs/getting-started-aspnetcore/Program.cs
reyang: Nice! Let me make the change here.

utpilla (approved):

PR #4917, opened 2023-10-04
Enrich DisplayName with actual route pattern for non-MVC application

### LINKED ISSUE #4525: Activity Display Name: Route Template instead of path

Want100Cookies:
Fixes #4525

## Changes

For non-MVC scenarios the Activity.DisplayName contained the whole route URI including all the parameter values. This is not correct according to the spec. Because the router needs to kick in first before you can get the route pattern I added a bit of code in the HttpInListener.OnStopActivity which gets the RouteEndpoint and uses that for the display name. I also found an existing unit test and adjusted it to match the new behaviour.


matt-hensley:
👍 this will be a nice change for minimal APIs that use routing

Kielek (approved):

matt-hensley (approved):

laupse:
Hello, we are waiting for this also because of #4525.
Seems like you were waiting to expand test suite in #4950. Since this one is merged, can we reopen this ? 
Thank you !

alanwest:
@laupse This instrumentation was released stable with version 1.6.0. This issue was resolved in that version. Also note that this instrumentation has moved to another repo. https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.Instrumentation.AspNetCore

PR #4893, opened 2023-09-27
Fix `TraceContextPropagator` behavior when trace parent flags contain…

Kielek:
Fixes test_traceparent_trace_flags_illegal_characters from w3c test suite
Design discussion issue #

## Changes

Fix `TraceContextPropagator` behavior when trace parent flags contains illegal characters.

## Merge requirement checklist

* [x] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/CONTRIBUTING.md) guidelines followed (nullable enabled, static analysis, etc.)
* [x] Unit tests added/updated
* [x] Appropriate `CHANGELOG.md` files updated for non-trivial changes
* ~~[ ] Changes in public API reviewed (if applicable)~~

PR #4937, opened 2023-10-09
W3C Trace propagator tests - switch to level 1

Kielek:
~~Fixes test_tracestate_duplicated_keys from w3c test suite~~

~~Follow up to #4893, there will be one more PR to fix random flags.~~

~~Ref: https://github.com/w3c/trace-context/blob/551a5b0855171281e98b4c2a814bf9e1f837cd53/test/test.py#L565-L567~~

## Changes

~~Duplicated keys are fully propagated now. It allows to drop creating HashSet for each propagation execution. Now only counter is created.~~

~~Opposite option, propagate only one value for given key. I think that it requires keeping HashSet.~~

Switch W3C Trace tests to level 1, based on: https://github.com/open-telemetry/opentelemetry-dotnet/pull/4937#issuecomment-1771484610


src/OpenTelemetry.Api/Context/Propagation/TraceContextPropagator.cs
```diff
@@ -294,19 +294,13 @@ internal static bool TryExtractTracestate(string[] tracestateCollection, out str
                         return false;
                     }
 
-                    // ValidateKey() call above has ensured the key does not contain upper case letters.
-                    if (!keySet.Add(key.ToString()))
```
utpilla, line 298: > expects the tracestate to be inherited, and the duplicated keys to be either kept as-is or one of them to be discarded

Could we discard the duplicate ones instead of keeping them? We are only allowed 32 keys. Why waste them by storing duplicates when we could make room for other unique keys?

Kielek: I have tried to answer in PR description on it what was the reason for the change:

> Duplicated keys are fully propagated now. It allows to drop creating HashSet for each propagation execution. Now only counter is created.

If you think that it is worth to keep it, I can change it.

utpilla: On second thought, I think we could just keep the key as-is in favor of performance as this would be exectued in the hot-path.

@alanwest @CodeBlanch @vishweshbankwar It would be great to know your thoughts as well?

utpilla (approved):

src/OpenTelemetry.Api/Context/Propagation/TraceContextPropagator.cs
vishweshbankwar: I am good to keep this as spec allows it.  
Not for this PR - It would be good to check how other languages are doing it (would be good to keep consistent behavior)

vishweshbankwar (approved):

src/OpenTelemetry.Api/Context/Propagation/TraceContextPropagator.cs
Kielek: Java is doing similar things, at the beginning checking if there is more/less than 32 attributes https://github.com/open-telemetry/opentelemetry-java/blob/57d83344178f578eb5d2f043b0ffc5c42b6719a5/api/all/src/main/java/io/opentelemetry/api/trace/propagation/internal/W3CTraceContextEncoding.java#L43-L45

Go-lang has similar issue like here, before changes: https://github.com/open-telemetry/opentelemetry-go/blob/c047088605b454f172765621d53107f80b1e6417/trace/tracestate.go#L123-L126

Python is checking duplicates: https://github.com/open-telemetry/opentelemetry-python/blob/34d11d59e9b37052eb3aa6706288641ecb3056e7/opentelemetry-api/src/opentelemetry/trace/span.py#L231-L240

I do not think that there we will be able to find common solution.

pellared:
Cross-posting https://github.com/open-telemetry/opentelemetry-go/pull/4638#discussion_r1360932855

test/OpenTelemetry.Tests/Trace/Propagation/TraceContextPropagatorTest.cs
```diff
@@ -193,10 +193,10 @@ public void Inject_WithTracestate()
     public void DuplicateKeys()
     {
         // test_tracestate_duplicated_keys
-        Assert.Empty(CallTraceContextPropagator("foo=1,foo=1"));
-        Assert.Empty(CallTraceContextPropagator("foo=1,foo=2"));
-        Assert.Empty(CallTraceContextPropagator(new[] { "foo=1", "foo=1" }));
-        Assert.Empty(CallTraceContextPropagator(new[] { "foo=1", "foo=2" }));
+        Assert.Equal("foo=1,foo=1", CallTraceContextPropagator("foo=1,foo=1"));
+        Assert.Equal("foo=1,foo=2", CallTraceContextPropagator("foo=1,foo=2"));
```
alanwest, line 197: Is the correct behavior or is the behavior in @pellared's PR for opentelemetry-go correct? That is:

```csharp
Assert.Equal("foo=1,foo=1", CallTraceContextPropagator("foo=1"));
Assert.Equal("foo=1,foo=2", CallTraceContextPropagator("foo=1"));
```

alanwest:
> Cross-posting [open-telemetry/opentelemetry-go#4638 (comment)](https://github.com/open-telemetry/opentelemetry-go/pull/4638#discussion_r1360932855)

Not sure I understand... should this PR become a draft too because the W3C spec has not yet clarified the behavior?

test/OpenTelemetry.Tests/Trace/Propagation/TraceContextPropagatorTest.cs
alanwest: Ok, nevermind. After speaking with @utpilla, sounds like both behaviors are correct and left up to the implementation.

alanwest (approved):

pellared:
> > Cross-posting [open-telemetry/opentelemetry-go#4638 (comment)](https://github.com/open-telemetry/opentelemetry-go/pull/4638#discussion_r1360932855)
> 
> Not sure I understand... should this PR become a draft too because the W3C spec has not yet clarified the behavior?

Also because W3C spec is not released.

utpilla:
> > > Cross-posting [open-telemetry/opentelemetry-go#4638 (comment)](https://github.com/open-telemetry/opentelemetry-go/pull/4638#discussion_r1360932855)
> > 
> > 
> > Not sure I understand... should this PR become a draft too because the W3C spec has not yet clarified the behavior?
> 
> Also because W3C spec is not released.

@pellared I don't follow. Are there any specific links that you're looking at to determine the readiness of the W3C spec? We are trying to find out if it's okay for us to merge the changes proposed in this PR.

alanwest:
> Also because W3C spec is not released.

I still don't understand 😆. Is there a draft of the w3c spec somewhere that clarifies the expected behavior? If so, can you share that link?

With respect to this PR, what are you saying? Should this PR be made a draft and reopened when the spec is ready? Or is this PR ready to merge?

pellared:
> > > > Cross-posting [open-telemetry/opentelemetry-go#4638 (comment)](https://github.com/open-telemetry/opentelemetry-go/pull/4638#discussion_r1360932855)
> > > 
> > > 
> > > Not sure I understand... should this PR become a draft too because the W3C spec has not yet clarified the behavior?
> > 
> > 
> > Also because W3C spec is not released.
> 
> @pellared I don't follow. Are there any specific links that you're looking at to determine the readiness of the W3C spec? We are trying to find out if it's okay for us to merge the changes proposed in this PR.

Mainly that [the specification](https://www.w3.org/TR/trace-context/) does not define this behavior. I created https://github.com/w3c/trace-context/issues/548. I propose at least wait for some feedback in the issue. I also created https://github.com/w3c/trace-context/pull/547.

CC @reyang

pellared:
> > > > Cross-posting [open-telemetry/opentelemetry-go#4638 (comment)](https://github.com/open-telemetry/opentelemetry-go/pull/4638#discussion_r1360932855)
> > > 
> > > 
> > > Not sure I understand... should this PR become a draft too because the W3C spec has not yet clarified the behavior?
> > 
> > 
> > Also because W3C spec is not released.
> 
> @pellared I don't follow. Are there any specific links that you're looking at to determine the readiness of the W3C spec? We are trying to find out if it's okay for us to merge the changes proposed in this PR.

1. The changes from https://github.com/w3c/trace-context/pull/477 are not part of the released W3C Trace stable release. See https://www.w3.org/TR/trace-context/#combined-header-value vs https://w3c.github.io/trace-context/#combined-header-value. They are only in the draft.
2. I am not sure even certain if the specification clearly defines the changes proposed in this PR. I created https://github.com/w3c/trace-context/issues/548

CC @reyang

Kielek:
I will try to add more comment here.

When the docker test image the latest commit from https://github.com/open-telemetry/opentelemetry-dotnet/blob/1ec7a635714549b8af301d2d4f44a369360434cf/test/OpenTelemetry.Instrumentation.W3cTraceContext.Tests/Dockerfile#L12 is downloaded. Then the OTel implementation of propagator is executed against latest changes from https://github.com/w3c/trace-context repository.

There are 2 versions of the specification:
* [W3C Recommendation](https://www.w3.org/standards/types#REC) 23 November 2021, Level 1: https://www.w3.org/TR/trace-context/#combined-header-value release
* [W3C Candidate Recommendation Snapshot](https://www.w3.org/standards/types#CR) 18 April 2023, Level 2: https://www.w3.org/TR/trace-context-2/#combined-header-value

Changes in tests https://github.com/w3c/trace-context/pull/477/ was merged in 2022, so after the Level 1 was released.

I see two options to proceed here:

1. Start executing tests on Level 1 whish is stable.
2. Keep executing tests for Level 2 which is in lets say RC phase.

For both cases test suited should be fetched from exact commit instead of the latest main.

utpilla:
> I will try to add more comment here.
> 
> When the docker test image the latest commit from
> 
> https://github.com/open-telemetry/opentelemetry-dotnet/blob/1ec7a635714549b8af301d2d4f44a369360434cf/test/OpenTelemetry.Instrumentation.W3cTraceContext.Tests/Dockerfile#L12
> 
> is downloaded. Then the OTel implementation of propagator is executed against latest changes from https://github.com/w3c/trace-context repository.
> There are 2 versions of the specification:
> 
> * [W3C Recommendation](https://www.w3.org/standards/types#REC) 23 November 2021, Level 1: https://www.w3.org/TR/trace-context/#combined-header-value release
> * [W3C Candidate Recommendation Snapshot](https://www.w3.org/standards/types#CR) 18 April 2023, Level 2: https://www.w3.org/TR/trace-context-2/#combined-header-value
> 
> Changes in tests [w3c/trace-context#477](https://github.com/w3c/trace-context/pull/477) was merged in 2022, so after the Level 1 was released.
> 
> I see two options to proceed here:
> 
> 1. Start executing tests on Level 1 whish is stable.
> 2. Keep executing tests for Level 2 which is in lets say RC phase.
> 
> For both cases test suited should be fetched from exact commit instead of the latest main.

I think we should aim for complying with the stable spec.

@alanwest Thoughts?

Kielek:
@utpilla, @alanwest, @vishweshbankwar, please review once more time. The PR is fully changed - tests are now executed based on Level 1 of specification (and passing on .NET7).

See: f6b2d2a8378c80556529b7cebaeac83048f6145b

pellared (approved):

Kielek:
.NET8 failure seems to be unrelated to changes. Please rerun.

test/OpenTelemetry.Instrumentation.W3cTraceContext.Tests/Dockerfile
```diff
@@ -9,7 +9,7 @@ FROM ubuntu AS w3c
 #Install git
 WORKDIR /w3c
 RUN apt-get update && apt-get install -y git
-RUN git clone https://github.com/w3c/trace-context.git
+RUN git clone --branch level-1 https://github.com/w3c/trace-context.git
```
utpilla, line 12: > There are 2 versions of the specification:

> [W3C Recommendation](https://www.w3.org/standards/types#REC) 23 November 2021, Level 1: https://www.w3.org/TR/trace-context/#combined-header-value release
[W3C Candidate Recommendation Snapshot](https://www.w3.org/standards/types#CR) 18 April 2023, Level 2: https://www.w3.org/TR/trace-context-2/#combined-header-value

@Kielek How do you determine the latest stable release of W3C? How did you find out that level 1 is the latest stable and level 2 is rc?

Kielek: Please check the headers of the pages. Statuses are pretty clear.

If we speaking about checking which branch to take. I have decided based on the commits date (in sync what is in page headers) and its names.

alanwest (approved):

utpilla (approved):

PR #4957, opened 2023-10-17
A code sample to show how to create new root activities that link to a previously current activity.

### LINKED ISSUE #4956: Need code sample to show how to create new root activities that link to a previously current activity.
# Feature Request
This is a tracking issue for a code sample (that I will be sending a PR for) to show how to create new root activities and link them to an outer activity. This can be useful in situations involving fanout operations or batched operations when there is a need to create a new trace for each outgoing fanout operation.

A code sample for the above. I will be sending a PR.

A customer could look at https://opentelemetry.io/docs/instrumentation/net/shim/#creating-new-root-spans and https://opentelemetry.io/docs/instrumentation/net/shim/#adding-links and connect the pieces together, however it will be better to give it a single sample that showcases this scenario.

kalyanaj:
Fixes #4956 

## Changes
This example shows how to create new root activities that link to an existing
activity. This can be useful in a fan-out or batched operation situation when
you want to create a new trace with a new root activity before invoking each
of the fanned out operations.


docs/trace/links-creation-with-new-activities/README.md
```diff
@@ -0,0 +1,119 @@
+# Creating new root activities that link to an existing activity: An Example
+
+This example shows how to create new root activities that link to an existing
```
cijothomas, line 3: it may not be obvious that "link to" here is referring to the OTel's concept of SpanLinks. Perhaps doc link helps?

docs/trace/links-creation-with-new-activities/Program.cs
```diff
@@ -0,0 +1,81 @@
...
+        // link it to an outer activity that happens to be the current
+        // activity.
+        for (int i = 0; i < 10; i++)
+        {
+            // Reference: https://opentelemetry.io/docs/instrumentation/net/manual/#creating-new-root-activities
+            // Since we want to create a new root activity for each of the fanned out operations,
+            // this step helps us "de-parent" it from the current activity.
+            Activity.Current = null;
+
+            // Reference: https://opentelemetry.io/docs/instrumentation/net/manual/#adding-links
```
cijothomas, line 67: Not sure if the links opetelemetry.io docs would remain same or get refactored. I'd recommend to keep links to docs within this repo, until both docs gets merged.
https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/src/OpenTelemetry.Api#activity-creation-options

docs/trace/links-creation-with-new-activities/README.md
kalyanaj: Good point, will add a sentence about it here, and will also add a reference to the span links spec.

docs/trace/links-creation-with-new-activities/README.md
```diff
@@ -0,0 +1,119 @@
...
+activity. This can be useful in a fan-out or batched operation situation when
```
cijothomas, line 4: a bit more background on what situation calls for this approach vs parenting everything to same root, would help future readers without background/context...

docs/trace/links-creation-with-new-activities/README.md
```diff
@@ -0,0 +1,156 @@
...
+[link](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/overview.md#links-between-spans)
+to an existing activity. This can be useful in a fan-out or batched operation
+situation when you want to create a new trace with a new root activity
+BEFORE invoking each of the fanned out operations, and at the same time
+you want each of these new traces to be linked to the original activity.
+
+To give an example, let's say that:
+
+- Service A receives a request for an operation that impacts 1000s of
+resources.
```
cijothomas, line 13: suggest to use some other term than "resource", as it could be confusing given OTel has its own concept of Resource!

docs/trace/links-creation-with-new-activities/Program.cs
```diff
@@ -0,0 +1,86 @@
...
+            // if/when issue https://github.com/open-telemetry/opentelemetry-dotnet/issues/984
+            // is addressed.
+            Activity.Current = null;
+
+            // Reference: https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/src/OpenTelemetry.Api#activity-creation-options
+            // Reference: https://opentelemetry.io/docs/instrumentation/net/manual/#adding-links
+            // We create a new root activity for each of the fanned out operations and link it to the outer activity.
+            using var newRootActivityForFannedOutOperation = MyActivitySource.StartActivity(
+                ActivityKind.Internal,  // Set this to the appropriate ActivityKind depending on your scenario
+                name: "FannedOutActivity",
```
utpilla, line 76: Maybe update the name to something like `$"FannedOutActivity {i + 1}"` so that it's more intuitive when reading the console output?

utpilla (approved):

docs/trace/links-creation-with-new-activities/Program.cs
kalyanaj: Thanks, updated.

docs/trace/links-creation-with-new-activities/README.md
kalyanaj: Not sure if there's an alternative good term here. Based on your feedback, I have now added a clarification about the terminology.

docs/trace/links-creation-with-new-activities/Program.cs
```diff
@@ -0,0 +1,87 @@
...
+        var previous = Activity.Current;
+        const int NumConcurrentOperations = 10;
+
+        var activityContext = Activity.Current!.Context;
+        var links = new List<ActivityLink>
+        {
+            new ActivityLink(activityContext),
+        };
+
+        // Fanning out to N concurrent operations.
```
utpilla, line 57: I don't know `concurrent` is the best term to use here as it one might conflate it with sequential/concurrent nature of execution. Maybe use something like this?
```suggestion
        // Fanning out to N related but individually unique traces.
```

kalyanaj: While I see the point about why this must be specified as concurrent (as a similar thing could apply for sequential too), I would prefer to leave it in the current form (as it is simpler to understand). My reasoning: The scenario is about a high number of fanout calls that affect individual resources, so realistically this would have to be done concurrently (as doing it sequentially for a high number of calls would make it very slow and further is not needed as each such fanout call is independent). Please feel free to let me know if you disagree.

utpilla: My take would be to just write a comment saying in a real-world case, these "related but individually unique traces" would be spawned concurrently. I believe that using the term "concurrent" could mislead a novice user to refer to the example code as a way of achieving concurrent creation of activities on C#.

If the intent here is to indeed emphasize on concurrent nature of these operations, then maybe we could update the example to use Threads/Tasks to show that instead of using a sequential for loop.

kalyanaj: Thanks @utpilla , that makes sense. 

I have now updated the example to launch each of these operations concurrently (instead of the sequential fashion). Please let me know if this looks better. 

As of now, I have not changed the comment (but can tweak it if you think it is required) - now it reflects the reality better as it is now triggering these operations concurrently (and these new activities also get created concurrently).

utpilla (approved):

jacob7395:
Hey,

I'm wondering what the side effects of this could be there were many threads. All starting traces.

I don't really understand .net threading so can't see what the issue could be, I also sore some mention of activity using thread IDs to match up the current but not sure about this.

Any thoughts?

cijothomas (approved):

utpilla:
> I'm wondering what the side effects of this could be there were many threads. All starting traces.

@jacob7395 Could you be more specific in terms what you think might be problematic with multiple threads each starting their own trace? From the standpoint of correctness, I don't see any issue with that. 

>  I also sore some mention of activity using thread IDs to match up the current but not sure about this.

I'm not sure I follow. Is there some doc that you came across which mentions this? In that case, could you share the link to that?

jacob7395:
Hi @utpilla,

I'll try and explain my thoughts here with the help of my favorite tool paint:

example

Okay so in my example we have two processes, the red process and green process, each box represents a new trace. We want the green process to have a new trance with no parent context. So during the setup of the activity we "cut" the context by setting Activity.Current, we then start our green activity and restore the context to it's previous state.

My question is what happens if while we set Activity.Current to null, our existing red process that wants to have its parent context intact. Would it end up as a new trace?

Is this situation event possible? or likely? I guess in your example what if 50% of your NumConcurrentOperations wanted to preserve the parent context?

docs/trace/links-creation-with-new-activities/Program.cs
```diff
@@ -0,0 +1,99 @@
...
+                Activity.Current = null;
+
+                // Reference: https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/src/OpenTelemetry.Api#activity-creation-options
+                // Reference: https://opentelemetry.io/docs/instrumentation/net/manual/#adding-links
+                // We create a new root activity for each of the fanned out operations and link it to the outer activity.
+                using var newRootActivityForFannedOutOperation = MyActivitySource.StartActivity(
+                    ActivityKind.Internal,  // Set this to the appropriate ActivityKind depending on your scenario
+                    name: $"FannedOutActivity {operationIndex + 1}",
+                    links: links);
+
```
jacob7395, line 85: should Activity.Current not be set back to the parent context here?

jacob7395:
To answer the sample works even if you have lots of parallel tasks, bellow is a test class to run the sample, some with parent context some without. Everything appears to work as expected.

```c#
[TestFixture]
public class NewActivity
{
    private static readonly ActivitySource MyActivitySource = new("LinksCreationWithNewRootActivities");
    
    [Test]
    public async Task TestParallelBehaviour()
    {
        using TracerProvider? tracerProvider = Sdk.CreateTracerProviderBuilder()
            .AddSource("LinksCreationWithNewRootActivities")
            .AddConsoleExporter()
            .Build();

        using (Activity? activity = MyActivitySource.StartActivity("OrchestratingActivity"))
        {
            activity?.SetTag("foo", 1);
            await DoFanoutAsync();

            using (Activity? nestedActivity = MyActivitySource.StartActivity("WrapUp"))
            {
                nestedActivity?.SetTag("foo", 1);
            }
        }
    }
    
    public async Task DoFanoutAsync()
    {
        Activity? previous = Activity.Current;

        previous.Should().NotBeNull();
        
        const int NumConcurrentOperations = 1000;

        ActivityContext activityContext = Activity.Current!.Context;
        List<ActivityLink> links = new List<ActivityLink>
        {
            new ActivityLink(activityContext),
        };

        
        List<Task> tasks = new List<Task>();

        ConcurrentBag<Activity?> withParent = new ConcurrentBag<Activity>();
        ConcurrentBag<Activity?> withoutParent = new ConcurrentBag<Activity>();
        
        Random random = new Random();
        
        for (int i = 0; i < NumConcurrentOperations; i++)
        {
            int operationIndex = i;

            Task task = Task.Run(async () =>
            {
                bool keepParent = random.Next(0, 2) == 0;

                Activity? currentParent = null;
                if (!keepParent)
                {
                    currentParent = Activity.Current;
                    Activity.Current = null;
                }

                // minor delay to try and force parallel clashes
                await Task.Delay(random.Next(0, 100));
                
                using Activity? newActivity = MyActivitySource.StartActivity(
                    ActivityKind.Internal,
                    name: $"FannedOutActivity {operationIndex + 1}",
                    links: links);

                if (!keepParent)
                {
                    withoutParent.Add(newActivity);
                    Activity.Current = currentParent;
                    return;
                }
                
                withParent.Add(newActivity);
            });

            tasks.Add(task);
        }

        await Task.WhenAll(tasks);

        foreach (Activity? activity in withoutParent)
        {
            activity.Should().NotBeNull();

            activity!.Parent.Should().BeNull();
        }
        
        foreach (Activity? activity in withParent)
        {
            activity.Should().NotBeNull();

            activity!.Parent.Should().NotBeNull();
            activity!.Parent!.Id.Should().BeEquivalentTo(previous!.Id);
        }
    }
}
```

utpilla:
@jacob7395 It's [AsyncLocal](https://learn.microsoft.com/en-us/dotnet/api/system.threading.asynclocal-1?view=net-7.0) that makes it all work because it flows with call context. `Activity.Current` is a `static AsyncLocal` [variable](https://github.com/dotnet/runtime/blob/aea45f7aaa8f73cb2a585d8b14ce66634d58cb68/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Activity.Current.net46.cs#L10).

Here's a simple code snippet which shows the behavior of `AsyncLocal`. I have used `AsyncLocal<int>` to keep it simple. You can apply this to how `Acitivity.Current` works. Hope that helps!

```csharp
public class Program
{
    private static AsyncLocal<int> asyncLocal = new AsyncLocal<int>();

    public static void Main()
    {
        asyncLocal.Value = 100;

        Console.WriteLine($"Main: asyncLocal value before update: {asyncLocal.Value}"); // 100

        var thread = new Thread(() =>
        {
            Console.WriteLine($"Thread: asyncLocal value before update: {asyncLocal.Value}"); // 100
            asyncLocal.Value = 250;
            Console.WriteLine($"Thread: asyncLocal value after update: {asyncLocal.Value}"); // 250
        });

        thread.Start();
        thread.Join();

        var task = Task.Run(() =>
        {
            Console.WriteLine($"Task: asyncLocal value before update: {asyncLocal.Value}"); // 100
            asyncLocal.Value = 500;
            Console.WriteLine($"Task: asyncLocal value after update: {asyncLocal.Value}"); // 500
        });

        task.Wait();

        Console.WriteLine($"Main: asyncLocal value after updates from Thread and Task: {asyncLocal.Value}"); // 100
    }
}
```

jacob7395:
Okay, that makes more sense now. I hadn't heard of asynlocal, I'll look into it, I really appreciate you taking the time to help. Thank you.

PR #4974, opened 2023-10-21
Benchmarks for metrics modified to have a metric reader

cijothomas:
Without metric reader, sdk does not really enable metrics, so adding to make sure we measure the actual impact.


test/Benchmarks/Instrumentation/AspNetCoreInstrumentationNewBenchmarks.cs
```diff
@@ -139,9 +139,14 @@ public void GetRequestForAspNetCoreAppGlobalSetup()
             this.StartWebApplication();
             this.httpClient = new HttpClient();
 
+            var exportedItems = new List<Metric>();
             this.meterProvider = Sdk.CreateMeterProviderBuilder()
                 .ConfigureServices(services => services.AddSingleton<IConfiguration>(configuration))
                 .AddAspNetCoreInstrumentation()
+                .AddInMemoryExporter(exportedItems, metricReaderOptions =>
```
utpilla, line 146: Could you also update the Benchmark numbers if you ran it locally?

cijothomas: no I did not. Won't have enough cycles to try that out, so we can either merge this PR as is, or wait for someone who is actively working on performance, and have them pick the changes from this PR.

utpilla (approved):

PR #5004, opened 2023-10-28
Check isSampled parameter in UpdateWithExemplar method

### LINKED ISSUE #4851: MetricPoint.UpdateWithExemplar method honors isSampled parameter only for DoubleSumIncomingDelta case
# Bug Report

Latest master:
https://github.com/open-telemetry/opentelemetry-dotnet/blob/b870ed9b0c965ec89bf6b5aedab87ff3cab8ea68/src/OpenTelemetry/Metrics/MetricPoint.cs#L444

net7.0

## Symptom

`isSampled` parameter is only checked in AggregationType.DoubleSumIncomingDelta case. That's why some exemplars are reported even when exemplar filter returns false.

`isSampled` should be checked before each this.mpComponents.ExemplarReservoir!.Offer call.

`isSampled` is only checked once in the first case.

## Additional Context

It would be great to have the ability to overwrite ExemplarReservoir. In this case, I could easily fix this bug in my codebase without waiting for bug fixes. Unfortunately, everything is internal in this library. Please consider providing an extension point to change ExemplarReservoir.

Another bug (maybe not in this library) is that SimpleExemplarReservoir can report exemplars with null trace id and trace span:
[SimpleExemplarReservoir.cs#L89)](https://github.com/open-telemetry/opentelemetry-dotnet/blob/b870ed9b0c965ec89bf6b5aedab87ff3cab8ea68/src/OpenTelemetry/Metrics/Exemplar/SimpleExemplarReservoir.cs#L89)

This breaks open-collector's open metric output and Prometheus with empty exemplars. Again, because there is no extension point I can't fix it myself, so providing it would be a great help for the dotnet community.

ImoutoChan:
Fixes #4851

## Changes

Check isSampled parameter in UpdateWithExemplar method before each call to ExemplarReservoir!.Offer

* [x] Appropriate CHANGELOG.md files updated for non-trivial changes

### DIFF
--- a/src/OpenTelemetry/CHANGELOG.md
+++ b/src/OpenTelemetry/CHANGELOG.md
@@ -11,6 +11,8 @@
   `autoGenerateServiceInstanceId` is `true`.
   ([#4988](https://github.com/open-telemetry/opentelemetry-dotnet/pull/4988))
 
+* Fixed a bug where isSampled parameter wasn't properly checked in certain cases within the `UpdateWithExemplar` method of `MetricPoint`. ([#4851](https://github.com/open-telemetry/opentelemetry-dotnet/pull/5004))
+
 ## 1.7.0-alpha.1
 
 Released 2023-Oct-16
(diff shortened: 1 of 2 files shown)


cijothomas (approved):

cijothomas:
@ImoutoChan Thanks! Could you add an entry to the changelog file as this is a bug fix?

ImoutoChan:
@cijothomas sure

cijothomas (approved):

utpilla (approved):

Issue #4983, opened 2023-10-23
AspNetCore instrumention sets Span's StatusDescription to Exception message


cijothomas:
https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs#L422 Is storing Exception.Message as Status.Description. 

The [semantic convention for http](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.md#status) does not prescribe such a behavior. Neither does the [Tracing API.](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#set-status)

cijothomas:
https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Instrumentation.AspNet/Implementation/HttpInListener.cs#L184 similar for asp.net 

https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Instrumentation.Http/Implementation/HttpHandlerDiagnosticListener.cs#L335 for httpclient. Maybe better to track separately.

vishweshbankwar:
completed in #5025

PR #5025, opened 2023-11-03
[AspNetCore] [Http] remove Activity Status Description and update unit tests

TimothyMothra:
Towards #4983

Previously, these libraries were setting both the Activity Status & StatusDescription via `Activity.SetStatus(ActivityStatusCode, Description)`. This PR removes the `string` description.

The Http spec states:
> Don't set the span status description if the reason can be inferred from http.response.status_code.
https://github.com/open-telemetry/semantic-conventions/blob/v1.23.0/docs/http/http-spans.md#status

## Changes
- Instrumentation.AspNetCore
  - Remove StatusDescription
- Instrumentation.AspNetCore.Tests
  - update unit tests
- Instrumentation.Http
  - Remove StatusDescription and light refactor.
- Instrumentation.Http.Tests
  - update unit tests
  - update json tests and data model
    - remove StatusDescription from json test cases
    - remove StatusDescription fram data model

### DIFF
--- a/src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-* Removed the Activity Status Description that was being set during unhandled
+* Removed the Activity Status Description that was being set during
   exceptions. Activity Status will continue to be reported as `Error`.
   This is a **breaking change**. `EnrichWithException` can be leveraged
   to restore this behavior.
--- a/src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
@@ -2,9 +2,8 @@
 
 ## Unreleased
 
-* On Exception, Activity Status is set to Error. The Exception Message
-  has been removed from Activity Status Description to comply with the
-  [Http spec](https://github.com/open-telemetry/semantic-conventions/blob/v1.23.0/docs/http/http-spans.md#status).
+* Removed the Activity Status Description that was being set during unhandled
+  exceptions. Activity Status will continue to be reported as `Error`.
   This is a **breaking change**. `EnrichWithException` can be leveraged
   to restore this behavior.
   ([#5025](https://github.com/open-telemetry/opentelemetry-dotnet/pull/5025))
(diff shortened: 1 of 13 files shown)


vishweshbankwar (approved):
LGTM - needs changelog.

For future:
> Don't set the span status description if the reason can be inferred from http.response.status_code

Should get clarification from spec what happens when the response code is not available. Should we be setting the description in this case? If yes, then we can update it later(Post GA) (it won't be a breaking change).

cijothomas:
> LGTM - needs changelog.

Please do mention that EnrichWithException can be leveraged to get back the old behavior. This is important as this was the behavior for a very long time, so need to smoothen upgrade/migrate disruptions.

cijothomas:
_https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#set-status
When the status is set to Error by Instrumentation Libraries, the Description SHOULD be documented and predictable. The status code should only be set to Error according to the rules defined within the semantic conventions. For operations not covered by the semantic conventions, Instrumentation Libraries SHOULD publish their own conventions, including possible values of Description and what they mean._


The above section talks about guidelines on Description. Putting exception message may not meet this requirement (as exc. message could be unpredictable/not-easy-to-document as it is not a finite list?)

src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+* On Exception, Activity Status is set to Error. The Exception Message
```
vishweshbankwar, line 5: Removed the Activity Status Description that was being set during unhandled exceptions. Activity Status will continue to be reported as `Error`

src/OpenTelemetry.Instrumentation.Http/CHANGELOG.md
```diff
@@ -2,9 +2,8 @@
 
 ## Unreleased
 
-* On Exception, Activity Status is set to Error. The Exception Message
-  has been removed from Activity Status Description to comply with the
-  [Http spec](https://github.com/open-telemetry/semantic-conventions/blob/v1.23.0/docs/http/http-spans.md#status).
+* Removed the Activity Status Description that was being set during unhandled
```
vishweshbankwar, line 5:
```suggestion
* Removed the Activity Status Description that was being set during
```

vishweshbankwar, line 5:
```suggestion
* Removed the Activity Status Description that was being set during
```

utpilla (approved):

PR #5043, opened 2023-11-10
[HttpClient & HttpWebRequest] Set network protocol version from response object

### LINKED ISSUE #4928: `network.protocol.version` - Should it be request.version or response.version?
Currently, we use `request.version` property to set `network.protocol.version`. However, with recent ootb metrics addition I noticed runtime is using `response.version` to set `network.protocol.version` tag https://github.com/dotnet/runtime/blob/7204fbe5f56d52dd3fc4238e33de45f532f05d50/src/libraries/System.Net.Http/src/System/Net/Http/Metrics/MetricsHandler.cs#L113. Opening this issue to clarify the correct value.

vishweshbankwar:
Fixes #4928 

## Changes
We are currently setting `network.protocol.version` using the version that was specified in request object. However, the final version depends on the version negotiated with the server. The PR updates to use version from response object. In case of no response, `network.protocol.version` will not be set.


antonfirsov (approved):
Conceptually LGTM, I'm not familiar with implementation details.

cijothomas:
> [5]: network.protocol.version refers to the version of the protocol used and might be different from the protocol client's version. If the HTTP client has a version of 0.27.2, but sends HTTP version 1.1, this attribute should be set to 1.1.

The above wording from https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-metrics.md#metric-httpclientrequestduration, does not mention that this attribute will not be set if no response received. Should we clarify it ? What if the otel convention is expecting to populate the client's version in such case, instead of not populating at all?

cijothomas:
> > What if the otel convention is expecting to populate the client's version in such case, instead of not populating at all?
> 
> The spec says it's _Recommended_ (vs _Required_), which in my understanding means that an implementation is allowed to omit it with an explanation.

I suggest we need to clarify it from semantic conventions. It does not say "Recommended. Populate if a response is obtained, else omit"... Given it clarifies those for certain other attributes, its best to clarify instead of assuming.

Optional attributes can cause issues like this https://github.com/open-telemetry/semantic-conventions/issues/492, so want to make this sure this is not falling into that category from sem. convention standpoint, but only from .NET implementation only.

vishweshbankwar:
> > > What if the otel convention is expecting to populate the client's version in such case, instead of not populating at all?
> > 
> > 
> > The spec says it's _Recommended_ (vs _Required_), which in my understanding means that an implementation is allowed to omit it with an explanation.
> 
> I suggest we need to clarify it from semantic conventions. It does not say "Recommended. Populate if a response is obtained, else omit"... Given it clarifies those for certain other attributes, its best to clarify instead of assuming.
> 
> Optional attributes can cause issues like this [open-telemetry/semantic-conventions#492](https://github.com/open-telemetry/semantic-conventions/issues/492), so want to make this sure this is not falling into that category from sem. convention standpoint, but only from .NET implementation only.

I have opened https://github.com/open-telemetry/semantic-conventions/issues/520 to get the clarification.

vishweshbankwar:
While we wait for the spec clarity, I think it is ok to consider getting this change merged. Because, in our current state we are not setting the version correctly and this PR is addressing that part.

cijothomas:
> While we wait for the spec clarity, I think it is ok to consider getting this change merged. Because, in our current state we are not setting the version correctly and this PR is addressing that part.

Please put a TODO in the metrics instrumentation code linking to the spec issue, and proceed with this PR.

cijothomas (approved):
Looks good. We may need adjustment based on clarifications, if any, arising from semantic conventions.

vishweshbankwar:
FYI - https://github.com/open-telemetry/semantic-conventions/issues/520#issuecomment-1812011445

src/OpenTelemetry.Instrumentation.Http/Implementation/HttpWebRequestActivitySource.netfx.cs
```diff
@@ -267,6 +269,7 @@ private static void AddExceptionTags(Exception exception, Activity activity, out
 
             if (tracingEmitNewAttributes)
             {
+                activity.SetTag(SemanticConventions.AttributeNetworkProtocolVersion, HttpTagHelper.GetProtocolVersionString(protocolVersion));
```
TimothyMothra, line 272: Is this meant to be set twice?
It's set here (line 272) in the `AddExceptionTags` and also the out parameter is used to set it again at Line 543.

It's also odd to me that it's being set in the AddExceptionTags helper method. This may be a candidate for refactor in the future.

vishweshbankwar: line 543 sets it on metric.

utpilla (approved):
`HttpWebRequestActivitySource.netfx.cs` needs to be refactored to make the code simpler to follow. Not blocking this PR.

PR #5044, opened 2023-11-13
[ASP.NET Core] Remove suppression for metrics generated from prometheus

vishweshbankwar:
Fixes #
Design discussion issue #

## Changes
Remove the special handling of metrics generated as a result of metrics pull from Prometheus endpoint. Any filtering (if at all) should be handled by generic filtering mechanism (if one is available in future).


src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
```diff
@@ -35,6 +35,9 @@ exception. The attribute value will be set to full name of exception type.
 * Fixed `network.protocol.version` attribute values to match the specification.
   ([#5007](https://github.com/open-telemetry/opentelemetry-dotnet/pull/5007))
 
+* Removed filtering of metrics generated as a result of pull from Prometheus.
```
cijothomas, line 38: This is one of the effects of the change, not the actual change!

We removed generating metrics if request path contained "metrics" - this is the change.

vishweshbankwar: updated

src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
```diff
@@ -35,6 +35,9 @@ exception. The attribute value will be set to full name of exception type.
 * Fixed `network.protocol.version` attribute values to match the specification.
   ([#5007](https://github.com/open-telemetry/opentelemetry-dotnet/pull/5007))
 
+* Removed filtering of metrics when the request path contained `/metrics`.
```
cijothomas, line 38: It won't hurt to mention that this is very common in Prometheus scraping, so hint that more measurements will be included in the final metric with this change. The "more" is likely negligible as prom's default scrap is 15secs only.

Please reword, above is just some suggestions!

rajkumar-rangaraj (approved):

src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
vishweshbankwar: Updated. is this better?

src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
```diff
@@ -35,6 +35,11 @@ exception. The attribute value will be set to full name of exception type.
 * Fixed `network.protocol.version` attribute values to match the specification.
   ([#5007](https://github.com/open-telemetry/opentelemetry-dotnet/pull/5007))
 
+* Removed filtering of metrics when the request path contained `/metrics`. This
+  change may affect prometheus pull scenario if the prometheus server sends
```
cijothomas, line 39: Probably nit picking, but the wording "may affect" may not be clear as to what is the effect? The effect of this PR is - metrics now include measurements from prometheus scrape requests. So one possible effect is - users see more count and their avge, etc might be affected. Eg: the service's latency is one-digit-ms only, but prometheus scrapes take seconds.. Now, by including prometheus scrapes as to the latency metric, it may give false impression that the service slowed! It could be immaterial, if prom scrapes are a minute fraction (quite likely.)

Not a blocker for the PR!

vishweshbankwar: I reworded first line a bit more. The point about latency is valid but I don't think it falls under the instrumentation library's scope (library simply reports metrics). I think it was a bug that we were filtering measurements and not giving the full picture.

src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
```diff
@@ -35,6 +35,12 @@ exception. The attribute value will be set to full name of exception type.
 * Fixed `network.protocol.version` attribute values to match the specification.
   ([#5007](https://github.com/open-telemetry/opentelemetry-dotnet/pull/5007))
 
+* Removed exclusion of measurements from `http.server.request.duration` metric
+  when the request path contained `/metrics`. This change may affect prometheus
```
utpilla, line 39:
```suggestion
  when the request path contained `/metrics`. This change may affect Prometheus
```

src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
```diff
@@ -35,6 +35,12 @@ exception. The attribute value will be set to full name of exception type.
...
+  pull scenario if the prometheus server sends request to the scrapping endpoint
```
utpilla, line 40:
```suggestion
  pull scenario if the Prometheus server sends request to the scraping endpoint
```

utpilla (approved):

src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
```diff
@@ -35,6 +35,12 @@ exception. The attribute value will be set to full name of exception type.
 * Fixed `network.protocol.version` attribute values to match the specification.
   ([#5007](https://github.com/open-telemetry/opentelemetry-dotnet/pull/5007))
 
+* Removed exclusion of measurements from `http.server.request.duration` metric
+  when the request path contained `/metrics`. This change may affect Prometheus
+  pull scenario if the Prometheus server sends request to the scraping endpoint
+  that contains `/metrics` in path.
+  ([#5044](https://github.com/open-telemetry/opentelemetry-dotnet/pull/5044))
```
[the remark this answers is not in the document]
vishweshbankwar: its not just `/metrics` but I see your point. Updated.

PR #5049, opened 2023-11-14
[HttpClient and ASP.NET Core] Fix failures due to changes in .NET8.0 stable release

vishweshbankwar:
Fixes #
Design discussion issue #

## Changes
Fixes couple of failures due to upgrade to .NET8.0. These are expected changes based on the issues that were fixed between rc and table release.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.Http/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.Http/CHANGELOG.md
@@ -118,7 +118,7 @@ Released 2023-Oct-26
   definition.
   ([#4990](https://github.com/open-telemetry/opentelemetry-dotnet/pull/4990))
 
-* `dns.lookups.duration` is renamed to `dns.lookup.duration`. This change
+* `dns.lookups.duration` metric is renamed to `dns.lookup.duration`. This change
   impacts users on `.NET8.0` or newer framework.
   ([#5049](https://github.com/open-telemetry/opentelemetry-dotnet/pull/5049))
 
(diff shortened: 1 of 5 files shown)


src/OpenTelemetry.Instrumentation.Http/CHANGELOG.md
```diff
@@ -118,6 +118,10 @@ Released 2023-Oct-26
   definition.
   ([#4990](https://github.com/open-telemetry/opentelemetry-dotnet/pull/4990))
 
+* `dns.lookups.duration` metric is renamed to `dns.lookup.duration`. This change
+  impacts users on `.NET8.0` or newer framework.
```
utpilla, line 122:
```suggestion
  impacts only users on `.NET8.0` or newer framework.
```

src/OpenTelemetry.Instrumentation.Http/CHANGELOG.md
```diff
@@ -118,6 +118,10 @@ Released 2023-Oct-26
   definition.
   ([#4990](https://github.com/open-telemetry/opentelemetry-dotnet/pull/4990))
 
+* `dns.lookups.duration` metric is renamed to `dns.lookup.duration`. This change
```
cijothomas, line 121: interesting...there is no code change here. Do we plan to keep updating changelog in otel instrumentation, for every change in .net repo? Or this is a one-time thing and we dont expect anything futher?

utpilla (approved):

src/OpenTelemetry.Instrumentation.Http/CHANGELOG.md
vishweshbankwar: I mentioned this one because we have a changelog entry in previous release with different name. In future, I don't think we need to mention every change. Except when the change might impact users moving from lower version to net8 or above.

utpilla: Going forward, I think we should simply update the README to point to the Microsoft official docs which lists these metrics and let users track the changes from there.

PR #5060, opened 2023-11-17
Http Span Status - Default to Unset if not in range defined by spec

vishweshbankwar:
Fixes #
Design discussion issue #

## Changes

Updates the logic to default the status to `Unset` if the httpStatusCode does not fall under the range defined by spec

https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.md#status.

Previous logic could have some unwanted side effects of setting the status to error outside of defined range.

UpperBound is defined as per https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#server_error_responses

### DIFF
--- a/src/Shared/SpanHelper.cs
+++ b/src/Shared/SpanHelper.cs
@@ -32,12 +32,13 @@ internal static class SpanHelper
     /// <returns>Resolved span <see cref="Status"/> for the Http status code.</returns>
     public static ActivityStatusCode ResolveSpanStatusForHttpStatusCode(ActivityKind kind, int httpStatusCode)
     {
-        var upperBound = kind == ActivityKind.Client ? 399 : 499;
-        if (httpStatusCode >= 100 && httpStatusCode <= upperBound)
+        var lowerBound = kind == ActivityKind.Client ? 400 : 500;
+        var upperBound = 599;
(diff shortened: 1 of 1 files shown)


utpilla (approved):

src/Shared/SpanHelper.cs
```diff
@@ -32,12 +32,13 @@ internal static class SpanHelper
     /// <returns>Resolved span <see cref="Status"/> for the Http status code.</returns>
     public static ActivityStatusCode ResolveSpanStatusForHttpStatusCode(ActivityKind kind, int httpStatusCode)
     {
-        var upperBound = kind == ActivityKind.Client ? 399 : 499;
-        if (httpStatusCode >= 100 && httpStatusCode <= upperBound)
+        var lowerBound = kind == ActivityKind.Client ? 400 : 500;
```
utpilla, line 35: The spec also says:

> or 3xx codes with max redirects exceeded), in which case status MUST be set to Error.

Do we need to check for that?

vishweshbankwar: Good catch. We might not be able to do this. In instrumentation, there is no way to detect if the response we are looking is the result of final redirect or not. I will follow up on this to see if there is some workaround we could use.

utpilla: Could you create an issue for this?

PR #5077, opened 2023-11-22
[HttpClient] Remove SDK dependency

### LINKED ISSUE #4809: Remove dependency on SDK
We should check if the instrumentation libraries really need a dependency on the SDK before we do a stable release.

vishweshbankwar:
Fixes #4809
Design discussion issue #

## Changes

Removes the SDK dependency from http client instrumentation. Note that with this change, the [SuppressDownstreamInstrumentation](https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/src/OpenTelemetry.Instrumentation.GrpcNetClient#suppressdownstreaminstrumentation) will no longer work. We can bring the dependency back for pre-release versions to support SuppressDownstreamInstrumentation.

For details on how this change impact users, see https://github.com/open-telemetry/opentelemetry-dotnet/issues/5092


src/OpenTelemetry.Instrumentation.Http/HttpClientInstrumentationOptions.cs
```diff
@@ -157,6 +138,11 @@ internal HttpClientInstrumentationOptions(IConfiguration configuration)
     /// </remarks>
     public bool RecordException { get; set; }
 
+    /// <summary>
+    /// Gets or sets a value indicating whether suppresses http instrumentation when the http call is part of grpc client call.
+    /// </summary>
+    public bool SuppressInstrumentationWhenGrpcIsPresent { get; set; }
```
alanwest, line 144: I think we should ship this new option as experimental. The option is useful as a stop gap, but I do not think it's ultimately what we want. We have an open issue with a different, more generic, proposal #4641 though it is only focused on suppressing tracing. The nice thing about our current suppression functionality is that it works to suppress all signals.

The ability to suppress downstream instrumentation is a long running [open issue](https://github.com/open-telemetry/opentelemetry-specification/issues/530) in the specification. Ultimately, I believe this will be functionality provided by the API not the SDK.

I am ok with this PR if we remove the dependency on the SDK from the HTTP instrumentation and ship this option specifically targeting gRPC as experimental.

vishweshbankwar: I removed the option proposed for now. Will bring it back after the stable release. Was hoping to use `EXPOSE_EXPERIMENTAL_FEATURES` but it is not possible to use it as is without additional changes in build. We can discuss the possible options (the one I proposed and existing one later).

test/OpenTelemetry.Instrumentation.Http.Tests/HttpClientTests.Basic.cs
```diff
@@ -278,7 +278,7 @@ public async Task InjectsHeadersAsync_CustomFormat()
         }));
     }
 
-    [Fact]
+    [Fact(Skip = "Removed for stable release of http instrumentation")]
```
cijothomas, line 281: lets add a link to a tracking gh issue.

vishweshbankwar: Yes, that is my TODO on PR desc :D

cijothomas: Also add in PR desc and/or changelog the full impact of this PR.

vishweshbankwar: updated

src/OpenTelemetry.Instrumentation.Http/CHANGELOG.md
```diff
@@ -11,6 +11,14 @@
 * Update activity DisplayName as per the specification.
   ([#5078](https://github.com/open-telemetry/opentelemetry-dotnet/pull/5078))
 
+* Removed reference to `OpenTelemetry` package. This is a **breaking change**
+  for users relying on
+  [SuppressDownstreamInstrumentation](https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/src/OpenTelemetry.Instrumentation.GrpcNetClient#suppressdownstreaminstrumentation)
```
utpilla, line 16: This should be documented in the Grpc instrumentation CHANGELOG as well as that's where the option is defined.

vishweshbankwar: done

src/OpenTelemetry.Instrumentation.GrpcNetClient/CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 ## Unreleased
 
+* **Breaking Change** :
+  [SuppressDownstreamInstrumentation](https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/src/OpenTelemetry.Instrumentation.GrpcNetClient#suppressdownstreaminstrumentation)
+  option will no longer be supported when used with certain versions of
```
utpilla, line 7: Not for this PR but we should write the exact version here when we do the release.

utpilla (approved):

PR #5097, opened 2023-11-29
[ASP.NET Core] Remove grpc server instrumentation

vishweshbankwar:
Fixes #
Design discussion issue #

## Changes
See issue https://github.com/open-telemetry/opentelemetry-dotnet/issues/5098

### DIFF
--- a/src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
@@ -11,6 +11,10 @@
 * Removed `netstandard2.1` target.
   ([#5094](https://github.com/open-telemetry/opentelemetry-dotnet/pull/5094))
 
+* Removed support for grpc instrumentation. For details, see issue
+  [#5098](https://github.com/open-telemetry/opentelemetry-dotnet/issues/5098)
+  ([#5097](https://github.com/open-telemetry/opentelemetry-dotnet/pull/5097))
+
 ## 1.6.0-beta.3
 
 Released 2023-Nov-17
--- a/src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
@@ -11,7 +11,8 @@
 * Removed `netstandard2.1` target.
   ([#5094](https://github.com/open-telemetry/opentelemetry-dotnet/pull/5094))
 
-* Removed support for grpc instrumentation. For details, see issue
+* Removed support for grpc instrumentation to unblock stable release of http
+  instrumentation. For details, see issue
   [#5098](https://github.com/open-telemetry/opentelemetry-dotnet/issues/5098)
   ([#5097](https://github.com/open-telemetry/opentelemetry-dotnet/pull/5097))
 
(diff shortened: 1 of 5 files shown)


utpilla (approved):
Approved with suggestions.

src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
```diff
@@ -11,6 +11,10 @@
 * Removed `netstandard2.1` target.
   ([#5094](https://github.com/open-telemetry/opentelemetry-dotnet/pull/5094))
 
+* Removed support for grpc instrumentation. For details, see issue
```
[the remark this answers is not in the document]
vishweshbankwar: I added the why part. Let me know if you think issue is not clear enough.

src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
```diff
@@ -11,6 +11,11 @@
 * Removed `netstandard2.1` target.
   ([#5094](https://github.com/open-telemetry/opentelemetry-dotnet/pull/5094))
 
+* Removed support for grpc instrumentation to unblock stable release of http
+  instrumentation. For details, see issue
```
alanwest, line 15: Why the decision to comment the functionality out when we could use the `EXPOSE_EXPERIMENTAL_FEATURES` directive. Seems this causes more toil for us to maintain and for users who may want to use the latest instrumentation.

utpilla: We could do that but I think it might be okay to comment this functionality out for the next rc which is supposed to mimic the stable behavior. We could see if people complain about this in the rc and decide to mark it experimental then? Otherwise, we would introduce in the first alpha after the stable release.

alanwest: I guess what I'm hearing is that we want to treat `rc` releases a bit differently than other pre-release monikers like `alpha`, `beta`, etc. I'd be fine if we set a precedent that `EXPOSE_EXPERIMENTAL_FEATURES` is only set for non-`rc` prereleases. I'm mainly looking for ways we can avoid the commenting/uncommenting toil we've had in the past.

I don't feel strongly that this is performed right now, so if you'd prefer to take the shortcut and comment it out for now then I'm ok with that. Though, I do want to see this reintroduced as soon as possible. I don't need to wait for customers to complain. I know of customers that use this instrumentation. Also, the cartservice from the [opentelemetry-demo](https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/cartservice/src/cartservice.csproj#L14) project is a gRPC service and depends on this instrumentation.

utpilla: > I guess what I'm hearing is that we want to treat rc releases a bit differently than other pre-release monikers like alpha, beta, etc.

We want to treat this particular rc that we would be releasing next differently. As this is the first time this library would have a stable release so that's why this rc is special than let's say an rc that would happen after we have had a stable release.

> Though, I do want to see this reintroduced as soon as possible. I don't need to wait for customers to complain. I know of customers that use this instrumentation.

We would have to tell them that these conventions are not stabilized yet. Ideally (taking the shortcut route), we would add it right after the first stable which should happen in ~2-3 weeks from now.

alanwest: > we would add it right after the first stable which should happen in ~2-3 weeks from now

ok

alanwest (approved):

Issue #4758, opened 2023-08-09
Reconsider Moq usage because of privacy issues


lachmatt:
Reconsider usage of `Moq` because of privacy issues described https://github.com/moq/moq/issues/1372.
For reference, `Moq` was replaced with `NSubstitute` in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/2816

Kielek:
Notes from August 15 sig meeting:
* No action necessary today since the version we use is prior to Moq adding SourceLink dependency.
* SourceLink has been removed in the latest release suggesting the maintainer is responsive to the community https://github.com/moq/moq/releases/tag/v4.20.69.
* If we do consider removing our use of Moq, we should consider attempting to eliminate using a mock library completely instead of replacing it with an alternative. Potentially a “good first issue/help wanted”.

Kielek:
@ngruson, great job here. Do you have a plan to make similar changes in https://github.com/open-telemetry/opentelemetry-dotnet-contrib ?

ngruson:
Hi @Kielek ,
I certainly can have a look in a couple of days if you want.

Kielek:
It will be great! Thanks!

PR #5106, opened 2023-11-30
Remove Moq dependency

ngruson:
Refactor unit tests in order to remove the `Moq` dependency.

## Changes

Mocks of `TextMapPropagator` have been removed.
I've introduced a `CustomTextMapPropagator` to be able to inject arbitrary key value pairs in the context of a unit test.
@vishweshbankwar : can you agree with this approach?

The dependency to the Moq package has been removed from all the test projects.
Mocking is now done using NSubstitute where needed.


ngruson:
Another option could be to use `NSubstitute` instead of `Moq` but I saw https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/2816 only after I did my changes.

ngruson:
@vishweshbankwar : To remove the dependency to NSubstitute, more work needs to be done.
It probably involves adding implementations of classes that are now mocked.
Do you want me to try that or in a different PR?

Yun-Ting:
FYI - this PR is related to this issue: https://github.com/open-telemetry/opentelemetry-dotnet/issues/4758#issuecomment-1702531106

utpilla:
> @ngruson We would like to avoid dependency on both NSubstitute and Mock. I'd suggest to tackle this one project at a time. For the current PR, lets scope this to Http one (as you had earlier) and move on to others in separate PRs.

💯

PR #5107, opened 2023-12-01
Export OpenMetrics format for prometheus exporters

robertcoltheart:
Fixes #
First part to addressing #3972 

## Changes

Export OpenMetrics format for Prometheus. See [here](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md). 

- Add `Accept` parser and set correct `Content-Type`
- Write metric timestamps in OpenMetric format


src/OpenTelemetry.Exporter.Prometheus.AspNetCore/PrometheusAspNetCoreOptions.cs
```diff
@@ -42,5 +42,14 @@ public int ScrapeResponseCacheDurationMilliseconds
         set => this.ExporterOptions.ScrapeResponseCacheDurationMilliseconds = value;
     }
 
+    /// <summary>
+    /// Gets or sets a value indicating whether to export OpenMetrics compatible scrape responses. Default value: true.
+    /// </summary>
+    public bool OpenMetricsEnabled
```
reyang, line 48: Consider a different approach which could support multiple formats/version.
For example:

```csharp
public string ExpositionFormat { get; set; } = "text/plain; charset=utf-8; version=0.0.4";
```

robertcoltheart: It is conceivable that each format may/will have specific requirements that cause the scrape format to be different, even between versions of a specified format. For example, `application/openmetrics-text` requires no extra CR characters, timestamps must be in seconds with a decimal part, and must finish with an `# EOF` line.

Given these restrictions, wouldn't it be better to allow specifying the format using an enum instead, which then lets the serializer decide the content type and all the formatting details?

Alternatively, we can expose the content type property as above, and the serializer can throw (or default back to text/plain) if the exposition format is invalid or not supported by the serializer.

A 3rd option is to expose an abstract class or interface that allows for different formats to be implemented independently. eg.

```c#
public PrometheusSerializer { get; set; } = new TextPlainPrometheusSerializer();
```

reyang: > It is conceivable that each format may/will have specific requirements that cause the scrape format to be different, even between versions of a specified format. For example, `application/openmetrics-text` requires no extra CR characters, timestamps must be in seconds with a decimal part, and must finish with an `# EOF` line.
> 
> Given these restrictions, wouldn't it be better to allow specifying the format using an enum instead, which then lets the serializer decide the content type and all the formatting details?

I think it depends. The immediate benefit of using enum is the strong type and editor/IDE support (e.g. auto completion). The potential downsides are:
1. enums are part of the public API, making it hard to deprecate anything unless we do a major version bump.
2. if the logic allows more flexibility (e.g. specifying an encoding which is not utf8), enum doesn't fit well.

> Alternatively, we can expose the content type property as above, and the serializer can throw (or default back to text/plain) if the exposition format is invalid or not supported by the serializer.

Makes sense to me if we throw exceptions when the specified exposition format is not supported.

> A 3rd option is to expose an abstract class or interface that allows for different formats to be implemented independently. eg.
> 
> ```cs
> public PrometheusSerializer { get; set; } = new TextPlainPrometheusSerializer();
> ```

I think this makes the configuration story a bit more complex. It also shares the potential issue with enums - making it hard to deprecate anything without a major version bump.

robertcoltheart: Understood, let me have a go at exposing the content type and working that into the serializer.

robertcoltheart: I've had a closer look at this, and I'm not sure allowing the content type to be configurable makes sense. The content type is switchable based on the request `Accept` headers. For example, if you request the `/metrics` endpoint from the browser, it will return `text/plain` which will display correctly, whereas forcing the `application/openmetrics-text` content type will actually try and download the file instead of displaying it (at least using Chrome). So we need to maintain the same level of usability while allowing the content type to switch for Prometheus.

From Prometheus, it will always send `application/openmetrics-text` in `Accept`, but will process either `text/plain` or `application/openmetrics-text` based on the returned `Content-Type`. 

There is some precedent for baking this into the code in this way. As far as I can tell the Java version of OTEL works the same way, and doesn't expose any method for disabling or configuring the `application/openmetrics-text` content type. I think the Go version also works this way, though I'm not a Golang expert.

For now I've removed the flags for enabling/disabling the OpenMetrics format, and just let the middleware/httplistener decide which format to send back based on the `Accept` header.

Hope this makes sense...

src/OpenTelemetry.Exporter.Prometheus.AspNetCore/PrometheusExporterMiddleware.cs
```diff
@@ -102,4 +110,28 @@ public async Task InvokeAsync(HttpContext httpContext)
...
+        if (StringValues.IsNullOrEmpty(requestAccept))
+        {
+            return false;
+        }
+
+        var acceptTypes = requestAccept.ToString().Split(',');
+
+        foreach (var acceptType in acceptTypes)
+        {
+            var acceptSubType = acceptType.Split(';').FirstOrDefault()?.Trim();
```
reyang, line 127: How much memory allocation would this introduce? Could this be improved with a zero-alloc one? (e.g. string matching instead of splitting).

reyang (approved):
LGTM with one suggestion https://github.com/open-telemetry/opentelemetry-dotnet/pull/5107/files#r1414630802.

src/OpenTelemetry.Exporter.Prometheus.AspNetCore/PrometheusExporterMiddleware.cs
robertcoltheart: Yep, this was copy/pasted from the Java version. I'll optimize this using `MediaTypeHeaderValue`.

robertcoltheart: Updated with zero allocations (hopefully)

src/OpenTelemetry.Exporter.Prometheus.AspNetCore/PrometheusExporterMiddleware.cs
```diff
@@ -102,4 +111,32 @@ public async Task InvokeAsync(HttpContext httpContext)
 
         this.exporter.OnExport = null;
     }
+
+    private bool AcceptsOpenMetrics(HttpRequest request)
+    {
+        var acceptHeader = request.Headers[HeaderNames.Accept];
```
utpilla, line 117:
```suggestion
        var acceptHeader = request.Headers.Accept;
```

src/OpenTelemetry.Exporter.Prometheus.AspNetCore/PrometheusExporterMiddleware.cs
```diff
@@ -102,4 +111,32 @@ public async Task InvokeAsync(HttpContext httpContext)
 
         this.exporter.OnExport = null;
     }
+
+    private bool AcceptsOpenMetrics(HttpRequest request)
```
utpilla, line 114:
```suggestion
    private static bool AcceptsOpenMetrics(HttpRequest request)
```

src/OpenTelemetry.Exporter.Prometheus.HttpListener/PrometheusHttpListener.cs
```diff
@@ -22,6 +22,8 @@ namespace OpenTelemetry.Exporter;
 
 internal sealed class PrometheusHttpListener : IDisposable
 {
+    private const string OpenMetricsMediaType = "application/openmetrics-text";
```
utpilla, line 25: This could be moved to the common `PrometheusExporter` class instead of both `PrometheusHttpListener` and `PrometheusExporterMiddleware` define it.

src/OpenTelemetry.Exporter.Prometheus.HttpListener/PrometheusHttpListener.cs
```diff
@@ -187,4 +193,36 @@ private async Task ProcessRequestAsync(HttpListenerContext context)
         {
         }
     }
+
+    private bool AcceptsOpenMetrics(HttpListenerRequest request)
```
utpilla, line 197:
```suggestion
    private static AcceptsOpenMetrics(HttpListenerRequest request)
```

src/OpenTelemetry.Exporter.Prometheus.HttpListener/PrometheusHttpListener.cs
```diff
@@ -187,4 +193,36 @@ private async Task ProcessRequestAsync(HttpListenerContext context)
...
+        if (string.IsNullOrEmpty(acceptHeader))
+        {
+            return string.Empty;
+        }
+
+        var separatorIndex = acceptHeader.IndexOf(';');
+
+        return separatorIndex == -1
+            ? acceptHeader
+            : acceptHeader.Substring(0, separatorIndex);
```
utpilla, line 226: Could you repurpose the no-allocation code that you introduced in the AspNetCore project here as well?

src/OpenTelemetry.Exporter.Prometheus.HttpListener/Internal/PrometheusExporter.cs
```diff
@@ -63,6 +63,8 @@ public PrometheusExporter(PrometheusExporterOptions options)
 
     internal int ScrapeResponseCacheDurationMilliseconds { get; }
 
+    internal bool? OpenMetricsRequested { get; set; }
```
utpilla, line 66: Could we use a plain `bool` instead of a nullable one? This should be simple binary decision for the exporter.

src/OpenTelemetry.Exporter.Prometheus.HttpListener/PrometheusHttpListener.cs
robertcoltheart: Created `PrometheusHeadersParser` that encapsulates all the header logic and shared it with both projects.

src/OpenTelemetry.Exporter.Prometheus.HttpListener/PrometheusHttpListener.cs
robertcoltheart: Moved all the header code to `PrometheusHeadersParser` using the no-alloc code and shared it between both projects

src/OpenTelemetry.Exporter.Prometheus.HttpListener/Internal/PrometheusCollectionManager.cs
```diff
@@ -168,11 +168,13 @@ private void WaitForReadersToComplete()
...
 
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
-    private bool ExecuteCollect()
+    private bool ExecuteCollect(bool openMetricsRequested)
     {
         this.exporter.OnExport = this.onCollectRef;
+        this.exporter.OpenMetricsRequested = openMetricsRequested;
         var result = this.exporter.Collect(Timeout.Infinite);
         this.exporter.OnExport = null;
+        this.exporter.OpenMetricsRequested = false;
```
utpilla, line 177: Is this needed? We always set `this.exporter.OpenMetricsRequested = openMetricsRequested;` before calling `Collect`.

src/OpenTelemetry.Exporter.Prometheus.HttpListener/Internal/PrometheusHeadersParser.cs
```diff
@@ -0,0 +1,60 @@
...
+// limitations under the License.
+// </copyright>
+
+namespace OpenTelemetry.Exporter.Prometheus;
+
+internal class PrometheusHeadersParser
+{
+    private const string OpenMetricsMediaType = "application/openmetrics-text";
+
+    internal bool AcceptsOpenMetrics(string contentType)
```
utpilla, line 23:
```suggestion
    internal static bool AcceptsOpenMetrics(string contentType)
```

src/OpenTelemetry.Exporter.Prometheus.HttpListener/Internal/PrometheusCollectionManager.cs
robertcoltheart: No I guess you're right, we don't need this.

src/OpenTelemetry.Exporter.Prometheus.HttpListener/PrometheusHttpListener.cs
```diff
@@ -187,4 +192,16 @@ private async Task ProcessRequestAsync(HttpListenerContext context)
         {
         }
     }
+
+    private bool AcceptsOpenMetrics(HttpListenerRequest request)
```
utpilla, line 196: You should be able to mark this method in both this class and the middleware class static once you have marked `PometheusHeadersParser` class and its method `AcceptOpenMetrics` static.

robertcoltheart: This is done

src/OpenTelemetry.Exporter.Prometheus.AspNetCore/PrometheusExporterMiddleware.cs
```diff
@@ -102,4 +107,16 @@ public async Task InvokeAsync(HttpContext httpContext)
...
+    private static bool AcceptsOpenMetrics(HttpRequest request)
+    {
+        var acceptHeader = request.Headers.Accept;
+
+        if (StringValues.IsNullOrEmpty(acceptHeader))
+        {
+            return false;
+        }
+
+        return acceptHeader.Any(PrometheusHeadersParser.AcceptsOpenMetrics);
```
utpilla, line 120: Use `foreach` instead of `IEnumerable.Any` as the latter would cause allocation.

You could write up a simple benchmark to measure the allocation. For reference, here is a benchmark that I wrote for comparing `foreach` vs `IEnumerable.Any`

| Method  | Mean     | Error    | StdDev   | Gen0   | Allocated |
|-------- |---------:|---------:|---------:|-------:|----------:|
| Any     | 53.20 ns | 1.058 ns | 1.218 ns | 0.0102 |      64 B |
| ForEach | 22.38 ns | 0.255 ns | 0.226 ns |      - |         - |

```csharp

public class EnumerationBenchmarks
{
    private static readonly StringValues StringValues = new StringValues(new string[] { "ABC", "DEF", "XYZ" });

    [Benchmark]
    public bool Any()
    {
        return StringValues.Any(DoSomething);
    }

    [Benchmark]
    public bool ForEach()
    {
        foreach (var s in StringValues)
        {
            if (DoSomething(s))
            {
                return true;
            }
        }

        return false;
    }

    private static bool DoSomething(string input)
    {
        return input.Contains("1");
    }
}
```

robertcoltheart: Roger that, updated

utpilla (approved):

test/OpenTelemetry.Exporter.Prometheus.HttpListener.Tests/PrometheusHttpListenerTests.cs
```diff
@@ -90,25 +90,53 @@ public async Task PrometheusExporterHttpServerIntegration_NoMetrics()
...
         while (retryAttempts-- != 0)
         {
             port = random.Next(2000, 5000);
             address = $"http://localhost:{port}/";
 
-            provider = Sdk.CreateMeterProviderBuilder()
-                .AddMeter(meter.Name)
-                .AddPrometheusHttpListener(options => options.UriPrefixes = new string[] { address })
-                .Build();
+            try
```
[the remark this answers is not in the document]
robertcoltheart: In this case, I'm guessing its because there might be a potential conflict with an already open port? I can remove in the next PR in any case.

utpilla:
> Thanks very much! Will crack on with the next PR.

@robertcoltheart Thanks for your contribution!

PR #5109, opened 2023-12-01
Rename HttpClientInstrumentationOptions

vishweshbankwar:
Fixes #
Design discussion issue #

## Changes

Renaming the class to make the intent more clear. It keeps the option to introduce metric options separately in future (if needed).


src/OpenTelemetry.Instrumentation.Http/HttpClientTraceInstrumentationOptions.cs
```diff
@@ -27,7 +27,7 @@ namespace OpenTelemetry.Instrumentation.Http;
 /// <summary>
 /// Options for HttpClient instrumentation.
 /// </summary>
-public class HttpClientInstrumentationOptions
+public class HttpClientTraceInstrumentationOptions
```
utpilla, line 30: Should we rename this to just `TraceInstrumentationOptions`? The namespace already has `OpenTelemetry.Instrumentation.Http`.

@alanwest What do you think?

Kielek: If you change it here to TraceInstrumentationOptions, we should made similar changes/recommendation to all other packages (here and in contrib).

utpilla: > That would cause ambiguous reference issues for users configuring options via Options API for e.g.

Ah good point!

utpilla (approved):

alanwest (approved):

src/OpenTelemetry.Instrumentation.Http/HttpClientTraceInstrumentationOptions.cs
alanwest: > Ah good point!

Agreed. Name seems good to me.

> @alanwest What do you think?

sorry... didn't see this. I used to get push notifications on my phone when @'ed, but I must of done something cause I didn't get one in this case... gonna try to fix this

PR #5110, opened 2023-12-01
Remove userAgent tag from HttpClient activity

vishweshbankwar:
Fixes #
Design discussion issue #

## Changes

User Agent is an opt-in tag so removing it https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.md#http-client and it was also a bug to use older `http.user_agent` tag instead of `user_agent.original`


cijothomas:
Do we plan to provide an option to opt-in as well? Or just removal for now, and asking to use Enrich to get user-agent?

cijothomas:
> > Do we plan to provide an option to opt-in as well? Or just removal for now, and asking to use Enrich to get user-agent?
> 
> Would suggest Enrich for any additional attributes.

That would be reasonable approach, as long we know https://github.com/dotnet/runtime/issues/93019 is committed for .NET 9. Otherwise, its better to expose APIs for opt-in fields. 
For unblocking stable release, its okay to do just-the-removal, but with a clear documentation on how users can self-mitigate.

TimothyMothra (approved):

utpilla (approved):

Issue #4376, opened 2023-04-07
Stable release of HTTP Instrumentations


cijothomas:
{Draft, still editing, adding links to other issues}

Tracking issue for covering the stable release of HTTP Instrumentations for Server and Client (i.e. Asp.Net Core, HttpClient). Most issues 
already exist, so will be linking to them. This is just to track the overall status.

* [x] Pre-req - OTel Semantic Conventions Stable.
* [x] Required/Recommended/Optional attributes - How would users enable/disable. See https://github.com/open-telemetry/opentelemetry-dotnet/issues/3373#issuecomment-1204164425 Update: https://github.com/open-telemetry/opentelemetry-dotnet/issues/4993 and https://github.com/open-telemetry/opentelemetry-dotnet/issues/4992 includes default set of attributes instrumentation will add. Ability to configure required/recommended attributes will not be provided in initial stable release.
* [x] Tracing and Metrics in one package - if Metrics convention is not stable, but Tracing - how to handle this in single package. 
**Edit on 5/17/2023** - Metrics are also expected to be stable along with Tracing. **Update 10/19** _Tracing and Metrics will be released in a single package._
* [x] Some instrumentation enables HTTP and RPC. Keep same or separate? (semantic conventions are likely moving in different pace): **update 10/19** _Grpc server instrumentation will be supported under experimental feature flag for stable release of instrumentation library._
* [x] Lifecycle management of the instrumentation libraries when there are multiple instances of SDK (especially for Metrics instrumentation)
* [x] Public API review. (Enrich* had breaking change in the past, might need similar for Filter?) Update : We will not do breaking changes for filter API.
* [x] Finalize target frameworks. (Testing netstandard2.0 code path etc.)
* [x] Triage and fix all blocking bugs
* [ ] Performance - benchmarks vs RPS drop using bombardier etc., Stress Tests to ensure no leaks, etc.
* [x] Some instrumentations have an undesired dependency on SDK, not API. Fixing this might require additional work in API/SDK. Update 11/29: SDk dependency has been removed from both the instrumentations.
* [x] Propagators from OTel vs .NET itself. How to ensure smooth/least disruptive migration to the one from DS itself. **Update 10/23** : This will be addressed in .NET9.0 when framework supports native trace instrumentation.
* [x] Document what does start and stop of Activity mean. (and there define what does the duration measures) https://github.com/open-telemetry/opentelemetry-dotnet/issues/3277
[...]

PR #5112, opened 2023-12-02
[Instrumentation.AspNetCore] Use static Meter and Histogram

utpilla:
Towards #4376 

## Changes
- Use static Meter and Histogram
- This should allow for multiple SDK registrations with instrumentation so that they can be disposed without affect each other.(This is also consistent with how we have defined ActivitySource and Meter for Http instrumentation)


vishweshbankwar (approved):

vishweshbankwar (approved):

CodeBlanch (approved):
LGTM

PR #5128, opened 2023-12-04
Removed Moq dependency from OpenTelemetry.Tests

ngruson:
Related to https://github.com/open-telemetry/opentelemetry-dotnet/issues/4758.

## Changes

- Added `TestSelfDiagnosticsConfigRefresher.cs`
- Refactored `SelfDiagnosticsEventListenerTest.cs`
- Refactored `ParentBasedSamplerTests.cs`
- Removed Moq dependency


test/OpenTelemetry.Tests/Trace/ParentBasedSamplerTests.cs
```diff
@@ -105,47 +105,28 @@ public void DoNotExamineLinks()
...
-            invokedSampler = localParentNotSampled;
-        }
-
         var expectedResult = new SamplingResult(SamplingDecision.RecordAndSample);
-        invokedSampler.Setup(sampler => sampler.ShouldSample(samplingParams)).Returns(expectedResult);
-
         var actualResult = samplerUnderTest.ShouldSample(samplingParams);
 
-        mockRepository.VerifyAll();
+        Assert.Equal(parentIsRemote && parentIsSampled, remoteParentSampled.LatestSamplingParameters.Equals(samplingParams));
```
vishweshbankwar, line 124: @ngruson I am finding these asserts bit confusing. Why cannot we simply assert like 

`Assert.True(remoteParentSampled.LatestSamplingParameters.Equals(samplingParams))`

ngruson: It depends on the inline data whether or not this assert is true.
If both params are true, the assert is true, otherwise false.

image

vishweshbankwar: got it. An alternate way could be to use if statements, but that would make the test bit longer. Ok to keep it as is.

utpilla (approved):

PR #5136, opened 2023-12-06
Only create a sibling activity when trace id, span id or trace state differ

### LINKED ISSUE #4466: `IHttpActivityFeature.Activity` and `Activity.Current` don't return the same activity
## Description

Considering the following example:
```cs
using System.Diagnostics;
using Microsoft.AspNetCore.Http.Features;

    .WithTracing(builder =>
    {
        builder
            .SetSampler(new AlwaysOnSampler()) // note here I use the AlwaysOn sampler to always record and sample activities
            .AddConsoleExporter();
    });

var app = builder.Build();

app.MapGet("/values", (HttpContext context) =>
{
    var activity = context.Features.GetRequiredFeature<IHttpActivityFeature>().Activity;
    activity.SetTag("test", "123");

    Activity.Current?.SetTag("test2", "123");

    return Results.Ok();
});

app.Run();
```

When the application receives a traceparent header with a trace-flags 00 (not sampled),`activity.Id` and `Activity.Current.Id` don't have the same id but both activities have the same name: `Microsoft.AspNetCore.Hosting.HttpRequestIn`.
In addition, only tags added through `Activity.Current.SetTag` are taken into account.

image

This seems to be a bug or an unexpected behaviour with the OpenTelemetry SDK as without the SDK both activities are the same and both tags are taken into account.

Both activities (the one from `IHttpActivityFeature.Activity` and the one from `Activity.Current`) should be the same.

Both activities are not the same.

ngruson:
Fixes #4466

Only create a sibling activity when trace id, span id or trace state differ in the extracted activity context. 
Traceflags are excluded from the comparison.

## Changes

- Changed `OnStartActivity` in `HttpInListener`
- Added unit test


ngruson:
@vishweshbankwar : Something like this?
The unit test covers the scenario where creating the new activity is skipped.
The existing unit test `SuccessfulTemplateControllerCallUsesParentContext` handles the scenario where the new activity is being created.

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs
```diff
@@ -125,9 +125,11 @@ public void OnStartActivity(Activity activity, object payload)
         if (textMapPropagator is not TraceContextPropagator)
         {
             var ctx = textMapPropagator.Extract(default, request, HttpRequestHeaderValuesGetter);
-
+            var activityContext = new ActivityContext(activity.TraceId, activity.ParentSpanId, activity.ActivityTraceFlags, activity.TraceStateString, true);
             if (ctx.ActivityContext.IsValid()
-                && ctx.ActivityContext != new ActivityContext(activity.TraceId, activity.ParentSpanId, activity.ActivityTraceFlags, activity.TraceStateString, true))
+                && ((ctx.ActivityContext.TraceId != activityContext.TraceId)
```
vishweshbankwar, line 130: could we not compare like below?

```
          if (ctx.ActivityContext.IsValid()
              && (!ctx.ActivityContext.TraceId.Equals(activity.TraceId) ||
              !ctx.ActivityContext.SpanId.Equals(activity.SpanId) ||
              !ctx.ActivityContext.TraceState.Equals(activity.TraceStateString)))
```

ngruson: The double negations confuse me tbh.
Wouldn't you rather do something like

```csharp
if (ctx.ActivityContext.IsValid()
  && !((ctx.ActivityContext.TraceId == activity.TraceId) &&
  (ctx.ActivityContext.SpanId == activity.ParentSpanId) &&
  (ctx.ActivityContext.TraceState == activity.TraceStateString)))
```

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/BasicTests.cs
```diff
@@ -1035,6 +1038,43 @@ public async Task DiagnosticSourceExceptionCallBackIsNotReceivedForExceptionsHan
         Assert.Equal(2, numberOfSubscribedEvents);
     }
 
+#if NET7_0_OR_GREATER
```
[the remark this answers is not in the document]
ngruson: I had to add the `Microsoft.Extensions.Features` package to the test app to be able to build with net6.0.
See the project file.
Should this be made conditional in the project file?
I didn't succeed right off the bat.

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs
vishweshbankwar: My snippet had a typo, Fixed it. Your suggestion also looks good to me.

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/BasicTests.cs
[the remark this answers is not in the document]
vishweshbankwar: Try this 

```
context.Features.Get<IHttpActivityFeature>()?.Activity;
```

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/BasicTests.cs
```diff
@@ -1022,6 +1022,41 @@ public async Task DiagnosticSourceExceptionCallBackIsNotReceivedForExceptionsHan
...
+        var request = new HttpRequestMessage(HttpMethod.Get, "/api/GetActivityEquality");
+        var traceId = ActivityTraceId.CreateRandom();
+        var spanId = ActivitySpanId.CreateRandom();
+        request.Headers.Add("traceparent", $"00-{traceId}-{spanId}-00");
+
+        var response = await client.SendAsync(request);
+        var result = bool.Parse(await response.Content.ReadAsStringAsync());
+
+        Assert.True(response.IsSuccessStatusCode);
+        Assert.True(result);
```
cijothomas, line 1056: nit: It'd be good to add a comment here that asserting result to be true is confirming that Activity.Current and IHttpActivityFeature activity are same.

cijothomas (approved):

ngruson:
Who needs to review this next? The PR was marked stale.

cijothomas:
> Who needs to review this next? The PR was marked stale.

Apologies. Vishwesh is travelling and will be back in Jan only. One of the other approvers will re-review this. Thanks for waiting!

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs
```diff
@@ -118,9 +118,11 @@ public void OnStartActivity(Activity activity, object payload)
         if (textMapPropagator is not TraceContextPropagator)
         {
             var ctx = textMapPropagator.Extract(default, request, HttpRequestHeaderValuesGetter);
-
+            var activityContext = new ActivityContext(activity.TraceId, activity.ParentSpanId, activity.ActivityTraceFlags, activity.TraceStateString, true);
```
utpilla, line 121: Do we even need to create an `ActivityContext` struct? We should simply compare the extracted context's Trace, SpanId, and TraceState with `activity.TraceId`, `activity.ParentSpanId` and `activity.TraceStateString`.

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs
```diff
@@ -118,9 +118,11 @@ public void OnStartActivity(Activity activity, object payload)
...
             if (ctx.ActivityContext.IsValid()
-                && ctx.ActivityContext != new ActivityContext(activity.TraceId, activity.ParentSpanId, activity.ActivityTraceFlags, activity.TraceStateString, true))
```
utpilla, line 123: `ActivityContext` equals check also to check if the extracted context's `IsRemote` field is set to `true`.

@cijothomas How do we want to handle that?

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs
```diff
@@ -118,9 +118,11 @@ public void OnStartActivity(Activity activity, object payload)
...
+                && !((ctx.ActivityContext.TraceId == activityContext.TraceId)
```
utpilla, line 123: Need to add a CHANGELOG entry for this.

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs
cijothomas: When context is extracted, it'll always be IsRemote=true....

utpilla: That depends on the implementation of the propagator, right? A user could use a custom propagator which does not set `IsRemote` to `true`.

cijothomas: That’ll be called a propagator with a remote bug 😆

utpilla: My point is that this would be a change in how we treat those buggy propagators. Up until now, we would have created a sibling activity with `parentContext.IsRemote == false` for such buggy propagators. With this change we would not do that.

Do you think we should document this change in behavior? If we think that it was bug on our side to consider `IsRemote` for determining whether a sibling activity should be created, we should call it a bug fix in the CHANGELOG.

cijothomas: We can probably add it in changelog, but I feel its going to make the changelog too complex to follow. `IsRemote`, by definition, should be `true`, when Propagators (*any*) are used. Maybe we need to explore a way to prevent IsRemote being set to false in Propagators somehow. (It should not be a priority, just mentioning.)

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs
ngruson: It seems that I made exactly that change earlier locally, but the commit was not pushed yet.
Shall I push that or do you want to keep the `ActivityContext` struct?

rajkumar-rangaraj (approved):

alanwest (approved):
@ngruson When you have a chance, please add a changelog entry https://github.com/open-telemetry/opentelemetry-dotnet/pull/5136#discussion_r1432178314 and I'll get this PR merged.

PR #5143, opened 2023-12-07
Doc fix for RecordException

### LINKED ISSUE #1569: .net definition for exception.stacktrace should be chosen and defined
# Feature Request

I was looking over the exception handling for otel.dotnet here:
https://github.com/open-telemetry/opentelemetry-dotnet/blob/master/src/OpenTelemetry.Api/Trace/ActivityExtensions.cs#L72

I was wondering why the stack trace is ex.ToString() rather than ex.StackTrace?

I see here: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/exceptions.md#stacktrace-representation
there's a suggestion this is something that google cloud does, but it's up-to-the SIG to define.  Other listed implementation suggest it's just a stack.

However, ex.ToString() will duplicate lots of extra information that's already in the exception.type and exception.message, and it's possible it doesn't even have the stack trace if the ToString is overridden in some way.

Checking reference source, suggests that ex.ToString() will by default capture the classname (which is probably a close match to 
 exception.type) the message and the stack trace.

The main difference I can see is that ToString() also goes through any InnerException, I'd have expected that inner exception should be recorded as another exception, (so it may end up recursive)

Or is it handling the scenario where a custom exception has also overridden ToString() to provide more information?

From my own perspective we tend to encourage people to log ex.ToString(), but I've heard of cases where the ex.ToString() can be very long.  So I'm not sure it's the right choice for an activity/trace.

Clarity on if stacktrace should be just the stack (as I'd expect) or the full ex.tostring()

I could create a helper routine that achieves something similar, but I'd expect confusion with having a helper and an opentelemetry api doing similar, but different things.

cijothomas:
Fixes #1569 

Given we were using `ToString()` from the 1st stable release, documenting that `ex.ToString()` is the chosen representation of stacktrace. #1569 has question about using `ToString()` vs `StackTrace`, but given this is already part of stable release, we need to keep this behavior.

It is okay to provide an option to customize this, if there is demand, or change the chosen representation in v2.0. This PR is just documenting the existing behavior.

### DIFF
--- a/src/OpenTelemetry.Api/Trace/ActivityExtensions.cs
+++ b/src/OpenTelemetry.Api/Trace/ActivityExtensions.cs
@@ -57,20 +57,26 @@ public static class ActivityExtensions
     }
 
     /// <summary>
-    /// Adds an activity event containing information from the specified exception.
+    /// Adds an <see cref="ActivityEvent"/>  containing information from the specified exception.
     /// </summary>
     /// <param name="activity">Activity instance.</param>
     /// <param name="ex">Exception to be recorded.</param>
+    /// <remarks> The exception is recorded as per https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/exceptions.md.
+    /// "exception.stacktrace" is represented using the value of https://learn.microsoft.com/dotnet/api/system.exception.tostring.
+    /// </remarks>
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
     public static void RecordException(this Activity activity, Exception? ex)
         => RecordException(activity, ex, default);
 
     /// <summary>
-    /// Adds an activity event containing information from the specified exception and additional tags.
+    /// Adds an <see cref="ActivityEvent"/> containing information from the specified exception and additional tags.
     /// </summary>
     /// <param name="activity">Activity instance.</param>
     /// <param name="ex">Exception to be recorded.</param>
     /// <param name="tags">Additional tags to record on the event.</param>
+    /// <remarks> The exception is recorded as per https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/exceptions.md.
+    /// "exception.stacktrace" is represented using the value of https://learn.microsoft.com/dotnet/api/system.exception.tostring.
+    /// </remarks>
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
     public static void RecordException(this Activity activity, Exception? ex, in TagList tags)
     {


src/OpenTelemetry.Api/Trace/ActivityExtensions.cs
```diff
@@ -57,20 +57,26 @@ public static Status GetStatus(this Activity activity)
     }
 
     /// <summary>
-    /// Adds an activity event containing information from the specified exception.
+    /// Adds an <see cref="ActivityEvent"/>  containing information from the specified exception.
     /// </summary>
     /// <param name="activity">Activity instance.</param>
     /// <param name="ex">Exception to be recorded.</param>
+    /// <remarks> The exception is recorded as per https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/exceptions.md.
```
utpilla, line 64:
```suggestion
    /// <remarks> The exception is recorded as per the <a href="https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/exceptions.md">specification</a>.
```

src/OpenTelemetry.Api/Trace/ActivityExtensions.cs
```diff
@@ -57,20 +57,26 @@ public static Status GetStatus(this Activity activity)
     }
 
     /// <summary>
-    /// Adds an activity event containing information from the specified exception.
+    /// Adds an <see cref="ActivityEvent"/>  containing information from the specified exception.
     /// </summary>
     /// <param name="activity">Activity instance.</param>
     /// <param name="ex">Exception to be recorded.</param>
+    /// <remarks> The exception is recorded as per <a href="https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/exceptions.md"/>.
```
utpilla, line 64: This sentence needs to be corrected. This would be rendered as:

The exception is recorded as per [. "exception.stacktrace" is represented using the value of ](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/exceptions.md)[Exception.ToString](https://learn.microsoft.com/dotnet/api/system.exception.tostring).

utpilla: You could change this to:

<remarks> The exception is recorded as per the <a href="https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/exceptions.md"/>specification.</a>
    /// "exception.stacktrace" is represented using the value of <a href="https://learn.microsoft.com/dotnet/api/system.exception.tostring">Exception.ToString</a> method.
   /// </remarks>

PR #5162, opened 2023-12-12
Minor refactoring for instrumentation libraries

utpilla:
## Changes
- Make methods static


src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInMetricsListener.cs
```diff
@@ -88,12 +68,12 @@ static bool TryFetchHttpContext(object payload, out HttpContext ctx)
...
     }
 
-    public void OnStopEventWritten(string name, object payload)
+    public static void OnStopEventWritten(string name, object payload)
     {
         var context = payload as HttpContext;
         if (context == null)
         {
-            AspNetCoreInstrumentationEventSource.Log.NullPayload(nameof(HttpInMetricsListener), EventName, HttpServerRequestDurationMetricName);
+            AspNetCoreInstrumentationEventSource.Log.NullPayload(nameof(HttpInMetricsListener), nameof(OnStopEventWritten), HttpServerRequestDurationMetricName);
```
utpilla, line 76: Updated this to log the name of the method instead.

vishweshbankwar (approved):

src/OpenTelemetry.Instrumentation.Http/Implementation/HttpHandlerMetricsDiagnosticListener.cs
```diff
@@ -173,4 +161,16 @@ static bool TryFetchRequest(object payload, out HttpRequestMessage request)
             return true;
         }
     }
+
+    public override void OnEventWritten(string name, object payload)
+    {
+        if (name == OnStopEvent)
```
utpilla, line 167: I have changed the order of the `if` conditions. For the common case, we don't want to waste cycles doing a string match for the exception event first.

PR #5171, opened 2023-12-14
Update CHANGELOG for 1.6.0 release of AspNetCore and Http instrumentation libraries

utpilla:
## Changes
- Update CHANGELOG for 1.6.0 release of AspNetCore and Http instrumentation libraries
- Update public API files


rajkumar-rangaraj (approved):

src/OpenTelemetry.Instrumentation.Http/.publicApi/PublicAPI.Shipped.txt
```diff
@@ -0,0 +1,24 @@
...
+OpenTelemetry.Instrumentation.Http.HttpClientTraceInstrumentationOptions.EnrichWithHttpRequestMessage.set -> void
+OpenTelemetry.Instrumentation.Http.HttpClientTraceInstrumentationOptions.EnrichWithHttpResponseMessage.get -> System.Action<System.Diagnostics.Activity, System.Net.Http.HttpResponseMessage>
+OpenTelemetry.Instrumentation.Http.HttpClientTraceInstrumentationOptions.EnrichWithHttpResponseMessage.set -> void
+OpenTelemetry.Instrumentation.Http.HttpClientTraceInstrumentationOptions.EnrichWithHttpWebRequest.get -> System.Action<System.Diagnostics.Activity, System.Net.HttpWebRequest>
+OpenTelemetry.Instrumentation.Http.HttpClientTraceInstrumentationOptions.EnrichWithHttpWebRequest.set -> void
+OpenTelemetry.Instrumentation.Http.HttpClientTraceInstrumentationOptions.EnrichWithHttpWebResponse.get -> System.Action<System.Diagnostics.Activity, System.Net.HttpWebResponse>
+OpenTelemetry.Instrumentation.Http.HttpClientTraceInstrumentationOptions.EnrichWithHttpWebResponse.set -> void
+OpenTelemetry.Instrumentation.Http.HttpClientTraceInstrumentationOptions.FilterHttpRequestMessage.get -> System.Func<System.Net.Http.HttpRequestMessage, bool>
```
utpilla, line 12: @vishweshbankwar @alanwest @CodeBlanch Should we rename the Filter APIs to use the term "With" similar to how we have named the Enrich APIs?

vishweshbankwar: "FilterWith" sounds odd to me. Probably use "FilterBy". But how does rename help here?

utpilla: > But how does rename help here?

The name should convey the intent of the API. `FilterHttpRequestMessage` sounds like an API that would filter the request message when in fact it merely uses the request message to filter out activities.

Comparing this to the Enrich APIs, how would `EnrichHttpRequestMessage` sound? We used "With" there to convey that the API would provide access to `HttpRequestMessage` to enrich the activity.

src/OpenTelemetry.Instrumentation.Http/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+## 1.6.0
+
+Released 2023-Dec-13
```
cijothomas, line 7: Mention/Highlight that this is the first stable release.

src/OpenTelemetry.Instrumentation.Http/.publicApi/PublicAPI.Shipped.txt
utpilla: I think we could rename it to something like `FilterUsingHttpRequestMessage` to make the intent more clear.

cijothomas: Agree the naming could be improved, but I suggest its better to leave it as-is to avoid another breaking change at a late stage.

vishweshbankwar: I think it depends on how one interprets it. `FilterHttpRequestMessage` implies we are filtering the data collection. Filtering HttpRequestMessage has no other meaning as per my understanding. 

`EnrichHttpRequestMessage` --> This would be confusing, because you can actually enrich the request object (like adding additional headers etc.). So `With` makes it more clear IMO. 

But I don't have strong take on it except I feel "FilterBy" is more apt if we want to rename.

utpilla: > I suggest its better to leave it as-is to avoid another breaking change at a late stage.

Yeah true ☹️

src/OpenTelemetry.Instrumentation.Http/CHANGELOG.md
utpilla: I have pushed a change. It's a little unconventional but I think it's okay considering this is a special release. Let me know what you think.

vishweshbankwar (approved):

cijothomas (approved):

src/OpenTelemetry.Instrumentation.Http/.publicApi/PublicAPI.Shipped.txt
alanwest: I also agree with the sentiment that the naming is not great, but it is not worth changing at this point in time.

alanwest (approved):

PR #5173, opened 2023-12-14
Tweaks to ActivityCreation benchmarks

cijothomas:
1. Removed parentid, kind scenarios as they don't seem to add additional value over other benchmarks.

.2 Added a benchmark to compare the perf difference between using SetTag vs AddTag. The former does [de-duplication+override](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Activity.cs#L1648-L1650), the latter does not do this. In scenarios where the tags are controlled (eg: Instrumentation Libraries from this repo), it might be faster to use AddTag instead of SetTag. This PR is simply adding benchmarks to help support that decision (in a future PR) for the instrumentations, and also give a recommendation to end users about the AddTag vs SetTag behavior.
Yes, SetTag is the spec-compliant API, but users who are in full control of their Tags might leverage AddTag to save some unwanted cpu cycles to detect duplicates. The cost increases with each attribute, as the current implementation traverses the list to look for duplicate.


test/Benchmarks/Helper/ActivityCreationScenarios.cs
```diff
@@ -13,30 +13,31 @@ public static void CreateActivity(ActivitySource source)
...
+        activity?.Stop();
+    }
+
+    public static void CreateActivityWithAddAttributes(ActivitySource source)
+    {
+        using var activity = source.StartActivity("name");
+        activity?.AddTag("tag1", "string");
+        activity?.AddTag("tag2", 1);
+        activity?.AddTag("tag3", true);
+        activity?.SetTag("tag4", "string-again");
```
utpilla, line 39:
```suggestion
        activity?.AddTag("tag4", "string-again");
```

test/Benchmarks/Helper/ActivityCreationScenarios.cs
```diff
@@ -13,30 +13,31 @@ public static void CreateActivity(ActivitySource source)
...
+        activity?.SetTag("tag5", "string-more");
```
utpilla, line 40:
```suggestion
        activity?.AddTag("tag5", "string-more");
```

test/Benchmarks/Helper/ActivityCreationScenarios.cs
utpilla, line 39: You probably would have to run the benchmarks again after correcting those changes.

test/Benchmarks/Helper/ActivityCreationScenarios.cs
```diff
@@ -13,30 +13,31 @@ public static void CreateActivity(ActivitySource source)
...
         activity?.Stop();
     }
 
-    public static void CreateActivityFromParentId(ActivitySource source, string parentId)
-    {
-        using var activity = source.StartActivity("name", ActivityKind.Internal, parentId);
-        activity?.Stop();
-    }
-
     public static void CreateActivityWithAttributes(ActivitySource source)
```
utpilla, line 22:
```suggestion
    public static void CreateActivityWithSetTag(ActivitySource source)
```

test/Benchmarks/Helper/ActivityCreationScenarios.cs
```diff
@@ -13,30 +13,31 @@ public static void CreateActivity(ActivitySource source)
...
         using var activity = source.StartActivity("name");
         activity?.SetTag("tag1", "string");
         activity?.SetTag("tag2", 1);
         activity?.SetTag("tag3", true);
+        activity?.SetTag("tag4", "string-again");
+        activity?.SetTag("tag5", "string-more");
+        activity?.Stop();
+    }
+
+    public static void CreateActivityWithAddAttributes(ActivitySource source)
```
utpilla, line 33:
```suggestion
    public static void CreateActivityWithAddTag(ActivitySource source)
```

test/Benchmarks/Trace/ActivityCreationBenchmarks.cs
```diff
@@ -55,11 +54,12 @@ public void GlobalCleanup()
     public void CreateActivity_WithParentContext_NoopProcessor() => ActivityCreationScenarios.CreateActivityFromParentContext(this.benchmarkSource, this.parentCtx);
 
     [Benchmark]
-    public void CreateActivity_WithParentId_NoopProcessor() => ActivityCreationScenarios.CreateActivityFromParentId(this.benchmarkSource, this.parentId);
+    public void CreateActivity_WithSetAttributes_NoopProcessor() => ActivityCreationScenarios.CreateActivityWithAttributes(this.benchmarkSource);
```
utpilla, line 57:
```suggestion
    public void CreateActivity_WithSetTag_NoopProcessor() => ActivityCreationScenarios.CreateActivityWithAttributes(this.benchmarkSource);
```

test/Benchmarks/Trace/ActivityCreationBenchmarks.cs
```diff
@@ -55,11 +54,12 @@ public void GlobalCleanup()
...
 
     [Benchmark]
-    public void CreateActivity_WithAttributes_NoopProcessor() => ActivityCreationScenarios.CreateActivityWithAttributes(this.benchmarkSource);
+    public void CreateActivity_WithAddAttributes_NoopProcessor() => ActivityCreationScenarios.CreateActivityWithAddAttributes(this.benchmarkSource);
```
utpilla, line 60:
```suggestion
    public void CreateActivity_WithAddTag_NoopProcessor() => ActivityCreationScenarios.CreateActivityWithAddAttributes(this.benchmarkSource);
```

test/Benchmarks/Trace/ActivityCreationBenchmarks.cs
```diff
@@ -8,20 +8,19 @@
...
 
 
-| Method                                         | Mean      | Error    | StdDev   | Gen0   | Allocated |
-|----------------------------------------------- |----------:|---------:|---------:|-------:|----------:|
-| CreateActivity_NoopProcessor                   | 307.12 ns | 5.769 ns | 6.172 ns | 0.0663 |     416 B |
-| CreateActivity_WithParentContext_NoopProcessor |  75.18 ns | 0.399 ns | 0.354 ns |      - |         - |
-| CreateActivity_WithParentId_NoopProcessor      | 156.52 ns | 1.609 ns | 1.426 ns | 0.0229 |     144 B |
-| CreateActivity_WithAttributes_NoopProcessor    | 372.34 ns | 6.215 ns | 4.852 ns | 0.0992 |     624 B |
-| CreateActivity_WithKind_NoopProcessor          | 302.24 ns | 5.859 ns | 8.402 ns | 0.0663 |     416 B |
+| Method                                         | Mean      | Error    | StdDev    | Median    |
```
utpilla, line 18: Include the memory allocation column?

utpilla (approved):

test/Benchmarks/Trace/ActivityCreationBenchmarks.cs
cijothomas: coming in next set, as i need a more stable machine !

PR #5180, opened 2023-12-14
Update README.md

utpilla:
## Changes
- Fix typo

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ libraries](https://github.com/open-telemetry/opentelemetry-specification/blob/ma
 
 * [ASP.NET Core](./src/OpenTelemetry.Instrumentation.AspNetCore/README.md)
 * [Grpc.Net.Client](./src/OpenTelemetry.Instrumentation.GrpcNetClient/README.md)
-* [HTTP clients](./src/OpenTelemetry.Instrumentation.Http/README.md)
+* [HTTP client](./src/OpenTelemetry.Instrumentation.Http/README.md)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -52,7 +52,7 @@ libraries](https://github.com/open-telemetry/opentelemetry-specification/blob/ma
 
 * [ASP.NET Core](./src/OpenTelemetry.Instrumentation.AspNetCore/README.md)
 * [Grpc.Net.Client](./src/OpenTelemetry.Instrumentation.GrpcNetClient/README.md)
-* [HTTP clients](./src/OpenTelemetry.Instrumentation.Http/README.md)
+* [HTTP client](./src/OpenTelemetry.Instrumentation.Http/README.md)
```
reyang, line 55: I don't have strong opinion on this, the actual instrumentation lib seems to support two clients?

reyang (approved):
I'm okay with either way, up to you @utpilla

README.md
utpilla: In that case, we could be more specific:

```suggestion
* [HttpClient and HttpWebRequest](./src/OpenTelemetry.Instrumentation.Http/README.md)
```

utpilla: Does that look good?

vishweshbankwar (approved):

README.md
reyang: I'm slightlying leaning towards full name (System.Net.Http.HttpClient, System.Net.HttpWebRequest) - it's the same as what we did for gRPC client. No strong opinion though.

reyang: Probably change the next line (SQL client) as well.

PR #5182, opened 2023-12-15
Scope info should only be exported for Prometheus if OpenMetrics is requested

### LINKED ISSUE #3972: Prometheus exporter handles instrumentation scope and prevents collisions
Now that https://github.com/open-telemetry/opentelemetry-specification/pull/2703 is released, we can update the prometheus exporter to make use of OpenTelemetry scope, particularly for reducing collisions between metrics with the same name.

**Describe the solution you'd like**

There are a few components:

* Add `otel_scope_name` and `otel_scope_version` attributes to all metrics based on the instrumentation scope name and version.
* Add an `otel_scope_info` metric following https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md#instrumentation-scope-1.
* Prevent collisions between TYPE, HELP, and UNIT comments following https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md#metric-metadata-1

Note that the exporter SHOULD do the above by default, but may allow disabling the behavior.

robertcoltheart:
Fixes #3972

## Changes

`otel_scope_info` was mistakenly being exported for non-OpenMetrics scrapes. This fix corrects this so that it will only be exposed for OpenMetrics.

### DIFF
--- a/src/OpenTelemetry.Exporter.Prometheus.AspNetCore/CHANGELOG.md
+++ b/src/OpenTelemetry.Exporter.Prometheus.AspNetCore/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Unreleased
 
 * Export OpenMetrics format from Prometheus exporters ([#5107](https://github.com/open-telemetry/opentelemetry-dotnet/pull/5107))
-* For requests with OpenMetrics format, scope info is automatically added 
+* For requests with OpenMetrics format, scope info is automatically added
   ([#5086](https://github.com/open-telemetry/opentelemetry-dotnet/pull/5086)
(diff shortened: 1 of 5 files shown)


reyang (approved):
Consider updating the changelog by putting this PR and the previous PR under the same entry (similar to what https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/CHANGELOG.md#050-beta2 has -  "Refactored simple export and batch export APIs (#1078 #1081 #1083 #1085 #1087 #1094 #1113 #1127 #1129 #1135)").

utpilla (approved):

PR #5193, opened 2023-12-19
AggregatorStore: don't double-count drops in doubles

NickCraver:
As funny as double counting only in doubles is, this seems like an oversight where we inadvertently call `Interlocked.Increment(ref this.DroppedMeasurements);` twice in the non-`emitOverflowAttribute` path. I noticed this digging into a related issue, so thought I'd pop up a PR while in here :)

## Changes
- Drops duplicate `DroppedMeasurements` increment in `AggregatorStore.UpdateDouble`
- Drops duplicate `DroppedMeasurements` increment in `AggregatorStore.UpdateDoubleCustomTags`


utpilla (approved):

utpilla:
@NickCraver Thanks for sending the PR!

PR #5221, opened 2024-01-13
Policheck

reyang:


Kielek (approved):

docs/diagnostics/README.md
```diff
@@ -7,7 +7,7 @@ to specific code patterns identified through static analysis.
 
 ## Experimental APIs
 
-Range: OTEL1XXX
+Range: OTEL1000 - OTEL1999
```
alanwest, line 10: This change seems to make things more clear, but What's Policheck? Was OTEL1XXX caught by some lint check or something?

alanwest (approved):
LGTM, one question though...

docs/diagnostics/README.md
reyang: > What's Policheck?

It is a tool used inside Microsoft to check potential offensive/sensitive words in codebase.

> Was OTEL1XXX caught by some lint check or something?

It was discovered by a Microsoft product which takes OpenTelemetry .NET source code. The story is similar to https://bugs.chromium.org/p/chromium/issues/detail?id=981129.

alanwest: So, you're saying OpenTelemetry .NET is now rated PG rather than XXX? 😆

PR #5249, opened 2024-01-23
Logger management guidance

reyang:


CodeBlanch (approved):
LGTM

docs/logs/README.md
```diff
@@ -103,3 +105,45 @@ logger.LogInformation("Hello from {food} {price}.", food, price);
...
+* The type of application you are building.
+* The place where you want to log.
+
+Here is the rule of thumb:
+
+* If you are building an application with [dependency injection
+  (DI)](https://learn.microsoft.com/dotnet/core/extensions/dependency-injection)
+  (e.g. [ASP.NET Core](https://learn.microsoft.com/aspnet/core) and [.NET
+  Worker](https://learn.microsoft.com/dotnet/core/extensions/workers)), in most
+  cases you should use the logger provided by DI, there are special cases when
```
cijothomas, line 126: we may want to talk about injecting logger factory and then obtaining logger from it as well,  if user desires control over category name.

reyang: Could you expand on it?

cijothomas: In DI supported apps, users can either inject ILogger<T> or ILoggerFactory to the ctor. Most folks need just ILogger<T> and it'll automatically assign LoggerCategory based on FullyQualified class name. But if they need a different category than the automatically provided one, they must instead inject ILoggerFactory to ctor, and then create logger from the loggerfactory providing the category name of their liking..
```
WeatherForecast(ILoggerFactory factory)
{
 this.Logger = factory.createLogger("my custom category name")
}
```

rajkumar-rangaraj (approved):

docs/logs/README.md
reyang: > Most folks need just ILogger and it'll automatically assign LoggerCategory based on FullyQualified class name.

I think this is the recommendation we provide, if folks want more control over the category, currently I consider this as advanced topic that doesn't apply to major audiences.

cijothomas: Can we put a note or link to
https://learn.microsoft.com/en-us/aspnet/core/fundamentals/logging/?view=aspnetcore-7.0#log-category

reyang: > Can we put a note or link to https://learn.microsoft.com/en-us/aspnet/core/fundamentals/logging/?view=aspnetcore-7.0#log-category

I've made an update and link to https://learn.microsoft.com/dotnet/core/extensions/logging?tabs=command-line#integration-with-hosts-and-dependency-injection which I consider better.

cijothomas: Well... that doc does not show injecting ILoggerFactory at all...it only shows injecting Ilogger/ilogger<t>

cijothomas (approved):

PR #5306, opened 2024-02-01
[aspnetcore-route-docs] Fix markdown to make markdownlint v15 happy

alanwest:
Our `markdownlint` CI fails when bumping `markdownlint` to version 15 (see #5290).

I did not have the time to determine the true root cause of this problem. Part of me thinks there may be a bug in `markdownlint` because the links in these readmes were perfectly functional.

It appears that `markdownlint` does not like link fragments that link to headers containing multiple `:` characters. To work around this I have place backticks around `area:exists`.


CodeBlanch (approved):
LGTM

PR #5323, opened 2024-02-07
add Kafka Exporter

lu-xiaoshuang:
add kafka exporter

test snapshot

add-kafka-exporter


Kielek:
I do not think that this is the best place for the Kafka exporter. It better fits https://github.com/open-telemetry/opentelemetry-dotnet-contrib

cijothomas (changes requested):
See https://github.com/open-telemetry/opentelemetry-dotnet/pull/5323#issuecomment-1931639048

Thats a lot of work, but unfortunately, we won't host any exporters not part of official otel spec in this repo! 

Even in contrib repo, please open an issue first to see if the maintainers are okay, before spending too much time. In general, if there is a valid use case, and someone willing to maintain, it should be acceptable there.

lu-xiaoshuang:
Hi @Kielek & @cijothomas , thanks for your comments.

We have some issues when using OpenTelemetry. (Other users probably have same issues, I guess). According to present design of OTEL, if you want a 
 message queue cluster (e.g. Apache Kafka) to store data, Applications/OTEL Agents + OTEL Collectors + MQ cluster + backend cluster shall be involved. This solution appears a little bit complicated. Many companies/teams (e.g. starups) do not has developers or resources for the OTEL collectors. They just want a simple architecture like Applications/OTEL Agents -> MQ cluster -> backend cluster.

On the other hand, why not OTLP exporter? OTLP exporter depends on HTTP/gRPC. However, both HTTP and gRPC lack the ability to store data temporarily for a duration. In contrast, a message queue can decouple data collecting and data processing. This key feature matters, especially for backend cluster maintenance. By the way, HTTP and gRPC may rely on DNS or VIPs to have high availability.

In conclusion, kafka exporter is much more important and useful. It would be great if OTEL offers kafka exporter as a built-in exporting option.

cijothomas:
@lu-xiaoshuang The only way to make OTel officially spec out and support a Kafta exporter is by proposing that to the OTel specification or most likely, to the [OTEPs repo](https://github.com/open-telemetry/oteps).

Issue #4791, opened 2023-08-21
Add retry support to otlp exporter via persistent storage


vishweshbankwar:
# Feature Request

**Is your feature request related to a problem?**

Transient server errors results in complete data loss.

**Describe the solution you'd like:**

Prevent the data loss by temporarily saving the data in persistent store so that it can be retried at a later time.

**Describe alternatives you've considered.**

There is an alternate option of retrying the data during transient errors by keeping it in memory.
This solution does not always results in successful data processing by the server and can still result in
data loss which is not a desirable state.

## Additional Context
Here is the proposed set of steps for adding retry support via persistent storage.
Each step will be one or more PRs

1) Fork in [OpenTelemetry.PersistentStorage.Abstractions](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.PersistentStorage.Abstractions) and [OpenTelemetry.PersistentStorage.FileSystem](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.PersistentStorage.FileSystem) into the otlp exporter package

   * This will be done in order to avoid adding external dependency.
   * All the APIs will be kept internal.

2) Add the retry mechanism via persistent storage. 

The general idea is to have a background thread scan for any new items in storage and re-attempt to send it to the otlp endpoint.
Here is a basic pseudo code for the background thread: The APIs pertaining to persistent storage are explained [here](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.PersistentStorage.FileSystem/README.md)

```csharp
while(_blobProvider.TryGetBlob(out var blob) && blob.TryLease(120000))
{
    blob.TryRead(out var data);
    var request = new  OtlpCollector.ExportTraceServiceRequest();
    request.MergeFrom(data);

    try
    {
	// Try re-sending request
        this.ExporClient.SendExportRequest(request);

        // Delete the blob from storage if the request was successful
        blob.TryDelete();
    }
    catch
    {
        // check if the error is retryable based on the exception/response codes
        // if yes then do not delete the blob else delete it from Storage.
    }
}
```

3) Wire up individual exporters to allow retries via persistent storage. Psuedo code:

```csharp
public ExportResult Export(...)
{
	try
	{
		// Do the export
	}
	catch (RpcException ex)
	{
		// check if error is retryable based on exception/response codes
[...]

PR #5335, opened 2024-02-09
Otlp Retry Part1 - Refactor ExportClients

vishweshbankwar:
Towards #1779 #4791 
Design discussion issue #

## Changes

This PR is part1 of multiple PRs which will bring retry functionality to otlp exporters in case of transient errors. 

1) It updates `IExportClient` interface definition to return a custom response object. Response object will be either of type `ExportClientHttpResponse` or `ExportClientGrpcResponse` based on the protocol used to send data i.e.  http protobuf or grpc.
2) Exporter implementations are updated accordingly.
3) Removing the existing `OtlpExporterTransmissionHandler` class that was added in https://github.com/open-telemetry/opentelemetry-dotnet/pull/4872. I plan to add an updated version of this in upcoming PRs.

**NOTE** This PR does not change the existing functionality of otlp exporters. For further details on what will be upcoming changes, take a look at this POC https://github.com/open-telemetry/opentelemetry-dotnet/pull/5311


CodeBlanch (approved):
LGTM

rajkumar-rangaraj (approved):
LGTM.

Tests are missing for all the properties in `ExportClientResponse`. Are you planning to cover in the follow up PRs?

alanwest (approved):
👍

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/BaseOtlpHttpExportClient.cs
```diff
@@ -34,24 +36,33 @@ protected BaseOtlpHttpExportClient(OtlpExporterOptions options, HttpClient httpC
     internal IReadOnlyDictionary<string, string> Headers { get; }
 
     /// <inheritdoc/>
-    public bool SendExportRequest(TRequest request, CancellationToken cancellationToken = default)
+    public ExportClientResponse SendExportRequest(TRequest request, CancellationToken cancellationToken = default)
     {
+        DateTime deadline = DateTime.UtcNow.AddMilliseconds(this.HttpClient.Timeout.TotalMilliseconds);
```
utpilla, line 41: Why does this not use the timeout value from the OtlpExporterOptions?

vishweshbankwar: good point - It is because user can override it with their own HttpClient instance via [HttpClientFactory](https://github.com/open-telemetry/opentelemetry-dotnet/blob/e4b08ac152fe12e88a124f099c6d7853e4c505fc/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpExporterOptions.cs#L180). So using the timeout value from the client itself and not options.

utpilla: Do we have a way to detect if the user actually configured timeout through OtlpExporter options?

vishweshbankwar: No - even if they are not using custom client. We initialize default client using timeout value set via options. https://github.com/open-telemetry/opentelemetry-dotnet/blob/e4b08ac152fe12e88a124f099c6d7853e4c505fc/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpExporterOptions.cs#L82-L88

So in the both the cases we will have the correct deadline here.

utpilla (approved):

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/BaseOtlpHttpExportClient.cs
utpilla: It would be good to add this as a comment.

vishweshbankwar: Sure, I can add it in my next PR if thats ok.

PR #5339, opened 2024-02-09
OpenTracing shim should use core MinVerTagPrefix

alanwest:
The OpenTracing shim is similar to the prometheus exporter in that it is a core component (i.e., it is declared in the OpenTelemetry specification https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/opentracing.md), but it is not yet ready for a stable release.

### DIFF
--- a/src/OpenTelemetry.Shims.OpenTracing/OpenTelemetry.Shims.OpenTracing.csproj
+++ b/src/OpenTelemetry.Shims.OpenTracing/OpenTelemetry.Shims.OpenTracing.csproj
@@ -3,6 +3,7 @@
     <TargetFrameworks>$(TargetFrameworksForLibraries)</TargetFrameworks>
     <Description>OpenTracing shim for OpenTelemetry .NET</Description>
     <PackageTags>$(PackageTags);distributed-tracing;OpenTracing</PackageTags>
+    <MinVerTagPrefix>core-</MinVerTagPrefix>
     <IsPackable>true</IsPackable>
     <!--
     TODO: Disable this exception, and actually do document all public API.


src/OpenTelemetry.Shims.OpenTracing/OpenTelemetry.Shims.OpenTracing.csproj
```diff
@@ -3,6 +3,7 @@
     <TargetFrameworks>$(TargetFrameworksForLibraries)</TargetFrameworks>
     <Description>OpenTracing shim for OpenTelemetry .NET</Description>
     <PackageTags>$(PackageTags);distributed-tracing;OpenTracing</PackageTags>
+    <MinVerTagPrefix>core-</MinVerTagPrefix>
```
utpilla, line 6: For this package, we had made a decision to only have beta releases as we previously considered it a non-core component. With this change, we would be redefining the stability readiness of this package as it would now begin to have alpha, beta, and rc releases (similar to Prometheus Exporter packages).

We even had an entry describing this decision in the [CHANGELOG](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Shims.OpenTracing/CHANGELOG.md#150-beta1):

> The pre-release identifier has also been changed from rc to beta as we believe this more accurately reflects the status of this package. We believe the rc identifier will be more appropriate as semantic conventions reach stability.

Do we want to change the above-mentioned expectation set for the users?

alanwest: 🤷 I have thoughts and opinions, but let's discuss in our next SIG meeting. I do not plan on releasing a new version of this package today anyways.

PR #5322, opened 2024-02-07
Release instrumentation packages  1.7.1/1.7.0-beta.1

Kielek:
## Changes

Request release for all instrumentation packages. It is especially needed for AspNetCore for AutoInstrumentation due to #5252.
Changes in other packages are also worth to release. The only package with only minor fix is HTTP.

Before merge please check if `csproj` files for SqlClient and GrpcNetClient shouldn't be updated. In these files there is no definition of `MinVerTagPrefix`.

## Merge requirement checklist

* [x] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/CONTRIBUTING.md) guidelines followed (license requirements, nullable enabled, static analysis, etc.)
* [ ] Unit tests added/updated
* [x] Appropriate `CHANGELOG.md` files updated for non-trivial changes
* [ ] Changes in public API reviewed (if applicable)

PR #5340, opened 2024-02-09
[GrpcNetClient] Clarify impact to SuppressDownstreamInstrumentation

alanwest:
We should probably decide what we're actually going to do with `SuppressDownstreamInstrumentation`, but for the purpose of getting the next release out, I just wanted to clean up our advisory.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.GrpcNetClient/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.GrpcNetClient/CHANGELOG.md
@@ -2,13 +2,15 @@
 
 ## Unreleased
 
-* **Breaking Change** :
+* **Breaking Change**:
+  Please be advised that the
   [SuppressDownstreamInstrumentation](https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/src/OpenTelemetry.Instrumentation.GrpcNetClient#suppressdownstreaminstrumentation)
-  option will no longer be supported when used with certain versions of
-  `OpenTelemetry.Instrumentation.Http` package. Check out this
+  option no longer works when used in conjunction with the
+  `OpenTelemetry.Instrumentation.Http` package version `1.6.0` or greater.
+  This is not a result of a change in the `OpenTelemetry.Instrumentation.GrpcNetClient`
+  package therefore this affects versions prior to this release. See this
   [issue](https://github.com/open-telemetry/opentelemetry-dotnet/issues/5092)
   for details and workaround.
-  ([#5077](https://github.com/open-telemetry/opentelemetry-dotnet/pull/5077))
 * Removed support for the `OTEL_SEMCONV_STABILITY_OPT_IN` environment variable
   which toggled the use of the new conventions for the
   [server, client, and shared network attributes](https://github.com/open-telemetry/semantic-conventions/blob/v1.23.0/docs/general/attributes.md#server-client-and-shared-network-attributes).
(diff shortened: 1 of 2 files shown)


reyang:
> We should probably decide what we're actually going to do with `SuppressDownstreamInstrumentation`, but for the purpose of getting the next release out, I just wanted to clean up our advisory.

Would this PR description help on the direction? https://github.com/open-telemetry/opentelemetry-dotnet/pull/960#issue-670039806

alanwest:
> Would this PR description help on the direction?

Yes, this is an interesting thought to consider. I believe what you're saying is there's no reason instrumentation needs an SDK dependency to check `otel.suppress_instrumentation` using `RuntimeContext` from `OpenTelemetry.Api` directly. We'd add the check back in the HTTP instrumentation as well.

Does anyone have strong feelings that we pursue this fix prior to releasing 1.7.1 today (#5322)? (@utpilla, @CodeBlanch, @vishweshbankwar)

My opinion is, since this breaking change was introduced last November that we merge this PR with a clear advisory, release 1.7.1 today, and then follow up with this idea.

reyang:
> > Would this PR description help on the direction?
> 
> Yes, this is an interesting thought to consider. I believe what you're saying is there's no reason instrumentation needs an SDK dependency to check `otel.suppress_instrumentation` using `RuntimeContext` from `OpenTelemetry.Api` directly. We'd add the check back in the HTTP instrumentation as well.

Correct, it's just a convention/suggestion (instead of a promise) that anyone can choose to follow (e.g. if there is another telemetry SDK, it can choose to respect this flag as well).

src/OpenTelemetry.Instrumentation.GrpcNetClient/CHANGELOG.md
```diff
@@ -6,13 +6,15 @@
...
-* **Breaking Change** :
+* **Breaking Change**:
+  Please be advised that the
   [SuppressDownstreamInstrumentation](https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/src/OpenTelemetry.Instrumentation.GrpcNetClient#suppressdownstreaminstrumentation)
-  option will no longer be supported when used with certain versions of
-  `OpenTelemetry.Instrumentation.Http` package. Check out this
+  option no longer works when used in conjunction with the
+  `OpenTelemetry.Instrumentation.Http` package version `1.6.0` or greater.
+  This is not a result of a change in the `OpenTelemetry.Instrumentation.GrpcNetClient`
+  package therefore this affects versions prior to this release. See this
```
utpilla, line 15:
```suggestion
  package therefore this also affects versions prior to this release. See this
```

utpilla:
> Does anyone have strong feelings that we pursue this fix prior to releasing 1.7.1 today (#5322)?

No. 

> My opinion is, since this breaking change was introduced last November that we merge this PR with a clear advisory, release 1.7.1 today, and then follow up with this idea.

I'm fine with that.

utpilla (approved):

vishweshbankwar:
> > > Would this PR description help on the direction?
> > 
> > 
> > Yes, this is an interesting thought to consider. I believe what you're saying is there's no reason instrumentation needs an SDK dependency to check `otel.suppress_instrumentation` using `RuntimeContext` from `OpenTelemetry.Api` directly. We'd add the check back in the HTTP instrumentation as well.
> 
> Correct, it's just a convention/suggestion (instead of a promise) that anyone can choose to follow (e.g. if there is another telemetry SDK, it can choose to respect this flag as well).



> > > Would this PR description help on the direction?
> > 
> > 
> > Yes, this is an interesting thought to consider. I believe what you're saying is there's no reason instrumentation needs an SDK dependency to check `otel.suppress_instrumentation` using `RuntimeContext` from `OpenTelemetry.Api` directly. We'd add the check back in the HTTP instrumentation as well.
> 
> Correct, it's just a convention/suggestion (instead of a promise) that anyone can choose to follow (e.g. if there is another telemetry SDK, it can choose to respect this flag as well).

I believe this is a good option to consider for offering `SuppressDownstreamInstrumentation` as an experimental feature. Experimental because we also need to think about the case when instrumentation will be done natively by HttpClient (starting with `.NET 9.0`) and eventually by Grpc.Net.Client as well.

I am good with going ahead with the release for now.

vishweshbankwar (approved):

Issue #5342, opened 2024-02-10
Semantic Convention not fully working


smoms:
Why standard exception fields like [exception.message](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/attributes-registry/exception.md) and [exception.stacktrace](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/attributes-registry/exception.md). are NOT visible in the following output?

I have the ILogger configured for using OTel.
When I log this error on Console:

```
try
{
...
int num1 = 5, num2 = 0, division_res = 0;
division_res = num1 / num2;
}
catch (Exception ex)
{
    logger.LogError(ex, "You divided by 0");
}
```
I get on Console:

```
fail: Program[0]
      "You divided by 0
      System.DivideByZeroException: Attempted to divide by zero.
         at Program.<<Main>$>g__HandleRollDice|0_4(ILogger`1 logger, String player) in C:\...\Program.cs:line 85
LogRecord.Timestamp:               2024-02-06T07:57:44.2418249Z
LogRecord.TraceId:                 d73bd4a40d4c4ca764ac87b0aa392b04
LogRecord.SpanId:                  b735f0c40f6c7243
LogRecord.TraceFlags:              Recorded
LogRecord.CategoryName:            Program
LogRecord.Severity:                Error
LogRecord.SeverityText:            Error
LogRecord.Body:                    "You divided by 0
LogRecord.Attributes (Key:Value):
    OriginalFormat (a.k.a Body): Y"You divided by 0
LogRecord.Exception:               System.DivideByZeroException: Attempted to divide by zero.
```

I do not see anywhere those attributes?

Furthermore, the same beahaviour appears with OTel Exporter too:

```
2024-02-09 12:15:47.005	
{
  "body": "You shall not divide by 0",
  "traceid": "a9b25c950fcadf6bc37ad7a88af58cfc",
  "spanid": "67687d1a30e805fd",
  "severity": "Error",
  "flags": 1,
  "resources": {
    "service.instance.id": "38033f53-0fb8-490c-8261-c2c050a0def9",
    "service.name": "myService",
    "service.namespace": "myNs",
    "service.version": "1.0.0",
    "telemetry.sdk.language": "dotnet",
    "telemetry.sdk.name": "opentelemetry",
    "telemetry.sdk.version": "1.7.0"
  },
  "instrumentation_scope": {
    "name": "Program"
  }
}
```

cijothomas:
[...]

PR #5351, opened 2024-02-12
OpenTelemetry.Exporter.Console: Print exception attributes according to SemConv

gregkalapos:
Follow up from https://github.com/open-telemetry/opentelemetry-dotnet/issues/5342#issuecomment-1937110524 - and https://github.com/open-telemetry/semantic-conventions/issues/689

## Changes

`ConsoleLogRecordExporter` prints exception attributes according to the [SemConv spec](https://opentelemetry.io/docs/specs/semconv/exceptions/exceptions-logs/). With this PR, `ConsoleLogRecordExporter` behaves similarly to [`OtlpLogRecordTransformer`](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/OtlpLogRecordTransformer.cs#L144-L149) from `OpenTelemetry.Exporter.OpenTelemetryProtocol`.

## Open questions

- [ ] I'm not familiar with previous design choices, but I really wonder why this is not done directly in [/src/OpenTelemetry/Logs](https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/src/OpenTelemetry/Logs)? E.g. in [LogRecord.cs ](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/Logs/LogRecord.cs). With that, all exporters would benefit from this and the responsibility of filling the attributes according to the spec would not be the responsibility of the exporter.

Since I'm fairly new to the code-base I wanted to ask this before I do any such changes - let me know if you think it'd make sense to explore this and I'm happy to update the PR - or open a follow up.

### DIFF
--- a/src/OpenTelemetry.Exporter.Console/ConsoleLogRecordExporter.cs
+++ b/src/OpenTelemetry.Exporter.Console/ConsoleLogRecordExporter.cs
@@ -94,7 +94,8 @@ public class ConsoleLogRecordExporter : ConsoleExporter<LogRecord>
                         // See https://github.com/open-telemetry/opentelemetry-dotnet/pull/3182
                         // for explanation.
                         var valueToTransform = logRecord.Attributes[i].Key.Equals("{OriginalFormat}")
-                            ? new KeyValuePair<string, object>("OriginalFormat (a.k.a Body)",
+                            ? new KeyValuePair<string, object>(
+                                "OriginalFormat (a.k.a Body)",
                                 logRecord.Attributes[i].Value)
                             : logRecord.Attributes[i];
 
@@ -166,6 +167,17 @@ public class ConsoleLogRecordExporter : ConsoleExporter<LogRecord>
         return ExportResult.Success;
     }
 
+    protected override void Dispose(bool disposing)
+    {
+        if (!this.disposed)
+        {
+            this.disposed = true;
+            this.disposedStackTrace = Environment.StackTrace;
+        }
+
+        base.Dispose(disposing);
+    }
+
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
     private void PrintAttribute(string key, string value)
     {
@@ -176,16 +188,5 @@ public class ConsoleLogRecordExporter : ConsoleExporter<LogRecord>
         {
             this.WriteLine($"{string.Empty,-4}{result}");
         }
-    }
-
-    protected override void Dispose(bool disposing)
-    {
-        if (!this.disposed)
-        {
-            this.disposed = true;
-            this.disposedStackTrace = Environment.StackTrace;
-        }
-
-        base.Dispose(disposing);
     }
 }
--- a/src/OpenTelemetry.Exporter.Console/ConsoleLogRecordExporter.cs
+++ b/src/OpenTelemetry.Exporter.Console/ConsoleLogRecordExporter.cs
@@ -1,9 +1,11 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
+using System.Runtime.CompilerServices;
 using OpenTelemetry.Internal;
 using OpenTelemetry.Logs;
 using OpenTelemetry.Resources;
+using OpenTelemetry.Trace;
 
 namespace OpenTelemetry.Exporter;
 
@@ -81,22 +83,33 @@ public class ConsoleLogRecordExporter : ConsoleExporter<LogRecord>
                 this.WriteLine($"{"LogRecord.Body:",-RightPaddingLength}{logRecord.Body}");
             }
 
-            if (logRecord.Attributes != null)
+            if (logRecord.Attributes != null || logRecord.Exception != null)
             {
                 this.WriteLine("LogRecord.Attributes (Key:Value):");
-                for (int i = 0; i < logRecord.Attributes.Count; i++)
+                if (logRecord.Attributes != null)
                 {
-                    // Special casing {OriginalFormat}
-                    // See https://github.com/open-telemetry/opentelemetry-dotnet/pull/3182
-                    // for explanation.
-                    var valueToTransform = logRecord.Attributes[i].Key.Equals("{OriginalFormat}")
-                        ? new KeyValuePair<string, object>("OriginalFormat (a.k.a Body)", logRecord.Attributes[i].Value)
-                        : logRecord.Attributes[i];
-
-                    if (ConsoleTagTransformer.Instance.TryTransformTag(valueToTransform, out var result))
+                    for (int i = 0; i < logRecord.Attributes.Count; i++)
                     {
-                        this.WriteLine($"{string.Empty,-4}{result}");
+                        // Special casing {OriginalFormat}
+                        // See https://github.com/open-telemetry/opentelemetry-dotnet/pull/3182
+                        // for explanation.
+                        var valueToTransform = logRecord.Attributes[i].Key.Equals("{OriginalFormat}")
+                            ? new KeyValuePair<string, object>("OriginalFormat (a.k.a Body)",
+                                logRecord.Attributes[i].Value)
+                            : logRecord.Attributes[i];
+
+                        if (ConsoleTagTransformer.Instance.TryTransformTag(valueToTransform, out var result))
+                        {
+                            this.WriteLine($"{string.Empty,-4}{result}");
+                        }
                     }
+                }
+
+                if (logRecord.Exception != null)
+                {
+                    this.PrintAttribute(SemanticConventions.AttributeExceptionType, logRecord.Exception.GetType().Name);
+                    this.PrintAttribute(SemanticConventions.AttributeExceptionMessage, logRecord.Exception.Message);
+                    this.PrintAttribute(SemanticConventions.AttributeExceptionStacktrace, logRecord.Exception.ToInvariantString());
                 }
             }
 
@@ -151,6 +164,18 @@ public class ConsoleLogRecordExporter : ConsoleExporter<LogRecord>
         }
 
         return ExportResult.Success;
+    }
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    private void PrintAttribute(string key, string value)
+    {
+        var valueToTransform =
+            new KeyValuePair<string, object>(key, value);
+
+        if (ConsoleTagTransformer.Instance.TryTransformTag(valueToTransform, out var result))
+        {
+            this.WriteLine($"{string.Empty,-4}{result}");
+        }
     }
 
     protected override void Dispose(bool disposing)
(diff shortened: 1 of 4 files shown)


gregkalapos:
> > * [ ]   I'm not familiar with previous design choices, but I really wonder why this is not done directly in [/src/OpenTelemetry/Logs](https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/src/OpenTelemetry/Logs)? E.g. in [LogRecord.cs ](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/Logs/LogRecord.cs). With that, all exporters would benefit from this and the responsibility of filling the attributes according to the spec would not be the responsibility of the exporter.
> > 
> > Since I'm fairly new to the code-base I wanted to ask this before I do any such changes - let me know if you think it'd make sense to explore this and I'm happy to update the PR - or open a follow up.
> 
> Perf. For exporters which do not need the exception string, we don't want the SDK to put pressure on GC heap.

Yes, perf is definitely a valid point. However, I was [reading the SemConv spec](https://opentelemetry.io/docs/specs/otel/logs/data-model/#errors-and-exceptions), which says:
> Additional information about errors and/or exceptions that are associated with a log record MAY be included in the structured data in the Attributes section of the record. If included, they MUST follow the OpenTelemetry [semantic conventions for exception-related attributes](https://opentelemetry.io/docs/specs/semconv/exceptions/exceptions-logs/). 

My interpretation of the `If included, they MUST follow the OpenTelemetry semantic conventions` was that if an exception is captured, then it's always in Attributes - as specified above.

But this is not yet part of this PR anyways, so I'm happy close this question.

src/OpenTelemetry.Exporter.Console/ConsoleLogRecordExporter.cs
```diff
@@ -81,23 +83,35 @@ public override ExportResult Export(in Batch<LogRecord> batch)
                 this.WriteLine($"{"LogRecord.Body:",-RightPaddingLength}{logRecord.Body}");
             }
 
-            if (logRecord.Attributes != null)
+            if (logRecord.Attributes != null || logRecord.Exception != null)
```
utpilla, line 86: No need to add this check.

src/OpenTelemetry.Exporter.Console/ConsoleLogRecordExporter.cs
```diff
@@ -81,23 +83,35 @@ public override ExportResult Export(in Batch<LogRecord> batch)
...
+                            : logRecord.Attributes[i];
+
+                        if (ConsoleTagTransformer.Instance.TryTransformTag(valueToTransform, out var result))
+                        {
+                            this.WriteLine($"{string.Empty,-4}{result}");
+                        }
                     }
                 }
+
+                if (logRecord.Exception != null)
```
utpilla, line 109: Move this `if` block outside of the `if (logRecord.Attributes != null)` check.

test/OpenTelemetry.Exporter.Console.Tests/ConsoleLogRecordExporterTest.cs
```diff
@@ -0,0 +1,57 @@
...
+namespace OpenTelemetry.Exporter.Console.Tests;
+
+public class ConsoleLogRecordExporterTest
+{
+    [Fact]
+    public void VerifyExceptionAttributesAreWritten()
+    {
+        using var writer = new StringWriter();
+        System.Console.SetOut(writer);
+        using var tracerProvider = Sdk.CreateTracerProviderBuilder()
```
utpilla, line 19: Remove traces related code?

test/OpenTelemetry.Exporter.Console.Tests/ConsoleLogRecordExporterTest.cs
```diff
@@ -0,0 +1,57 @@
...
+            .AddSource("*")
+            .ConfigureResource(resource =>
+                resource.AddService(
+                    serviceName: "Test_VerifyExceptionAttributesAreWritten",
+                    serviceVersion: "1.0.0"))
+            .AddConsoleExporter()
+            .Build();
+
+        using ILoggerFactory factory = LoggerFactory
+            .Create(builder => builder.AddOpenTelemetry(n =>
```
utpilla, line 29: nit: Use more intuitive variable names.
```suggestion
            .Create(builder => builder.AddOpenTelemetry(logging =>
```

test/OpenTelemetry.Exporter.Console.Tests/ConsoleLogRecordExporterTest.cs
```diff
@@ -0,0 +1,57 @@
...
+            {
+                n.AddConsoleExporter();
```
utpilla, line 31:
```suggestion
                logging.AddConsoleExporter();
```

test/OpenTelemetry.Exporter.Console.Tests/ConsoleLogRecordExporterTest.cs
```diff
@@ -0,0 +1,57 @@
...
+        using var tracerProvider = Sdk.CreateTracerProviderBuilder()
+            .AddSource("*")
+            .ConfigureResource(resource =>
+                resource.AddService(
+                    serviceName: "Test_VerifyExceptionAttributesAreWritten",
+                    serviceVersion: "1.0.0"))
+            .AddConsoleExporter()
+            .Build();
+
+        using ILoggerFactory factory = LoggerFactory
```
utpilla, line 28: Since you have been using `var at other places, you could just use it for LoggerFactory and Logger as well to be consistent.

test/OpenTelemetry.Exporter.Console.Tests/ConsoleLogRecordExporterTest.cs
```diff
@@ -0,0 +1,57 @@
...
+            .Create(builder => builder.AddOpenTelemetry(n =>
+            {
+                n.AddConsoleExporter();
+            }));
+
+        ILogger logger = factory.CreateLogger("Program");
+        try
```
utpilla, line 35: Instead of this setup, you could simply log an exception with a custom message to test this. Something like:

```csharp
logger.LogError(default(EventId), new Exception("Exception Message"), null);
```

src/OpenTelemetry.Exporter.Console/ConsoleLogRecordExporter.cs
gregkalapos: The idea here is to treat exception attributes like any other attributes, since [SemConv defines them as attributes](https://opentelemetry.io/docs/specs/semconv/exceptions/exceptions-logs/).

With this, the output is something like this:
```
LogRecord.Attributes (Key:Value):
    OriginalFormat (a.k.a Body): You divided by 0
    exception.type: DivideByZeroException
    exception.message: Attempted to divide by zero.
    exception.stacktrace: System.DivideByZeroException: Attempted to divide by zero.
   at OpenTelemetry.Exporter.Console.Tests.ConsoleLogRecordExporterTest.VerifyExceptionAttributesAreWritten() in /Users/gregkalapos/repos/gregkalapos_opentelemetry-dotnet/test/OpenTelemetry.Exporter.Console.Tests/ConsoleLogRecordExporterTest.cs:line 39
LogRecord.Exception:               System.DivideByZeroException: Attempted to divide by zero.
   at OpenTelemetry.Exporter.Console.Tests.ConsoleLogRecordExporterTest.VerifyExceptionAttributesAreWritten() in /Users/gregkalapos/repos/gregkalapos_opentelemetry-dotnet/test/OpenTelemetry.Exporter.Console.Tests/ConsoleLogRecordExporterTest.cs:line 3
```

If we move the logging of exception attributes outside this block, then the `LogRecord.Attributes (Key:Value):` may not be printed (if no other attribute is present, and I'm not sure we can assume we'll always have any other attribute). Also the indention of these attribute also matches other attribute's indention on purpose.

src/OpenTelemetry.Exporter.Console/ConsoleLogRecordExporter.cs
gregkalapos: Same reasoning as above: https://github.com/open-telemetry/opentelemetry-dotnet/pull/5351#discussion_r1494981092

gregkalapos:
Thanks for the review @utpilla.

Full agreement on changes in `ConsoleLogRecordExporterTest.cs` - I pushed an update to that file.

Regarding the suggestions in `ConsoleLogRecordExporter.cs`, I added some context: https://github.com/open-telemetry/opentelemetry-dotnet/pull/5351#discussion_r1494981092

src/OpenTelemetry.Exporter.Console/ConsoleLogRecordExporter.cs
reyang: I'm concerned about the readability of the exception callstack, create a tracking issue with my suggestions #5372.

utpilla: I agree with the concerns raised by @reyang here. YAML has better support for multi-line strings. @gregkalapos Would you be interested in working on updating the `ConsoleExporte`r to print the data in `yaml` format?

gregkalapos: Yeah, I'd be happy to look into that - I also think it's a good idea. Let's handle it in the issue.

What do we do with this PR?

The issue with the readability of the exception callstack is not new, `LogRecord.Exception` is already printed a few lines below with this output (that's not changed in the PR):

```
LogRecord.Exception:               System.DivideByZeroException: Attempted to divide by zero.
   at OpenTelemetry.Exporter.Console.Tests.ConsoleLogRecordExporterTest.VerifyExceptionAttributesAreWritten() in /Users/gregkalapos/repos/gregkalapos_opentelemetry-dotnet/test/OpenTelemetry.Exporter.Console.Tests/ConsoleLogRecordExporterTest.cs:line 39
```

I see https://github.com/open-telemetry/opentelemetry-dotnet/issues/5372 and this PR orthogonal.

utpilla: > The issue with the readability of the exception callstack is not new,
> I see https://github.com/open-telemetry/opentelemetry-dotnet/issues/5372 and this PR orthogonal.

I don't see them as orthogonal. The issue clearly talks about new things getting added with no well-defined structure. Just because the ConsoleExporter already has a readability issue, doesn't mean we keep making it worse.

> What do we do with this PR?

I would say we should close it and invest in making ConsoleExporter better and extensible for newer changes in a well-defined way. I don't think anyone is blocked on ConsoleExporter not printing out the exception information with Semantic Convention attributes. As you previously mentioned, the exception information is already printed to console in some other format.

utpilla:
https://github.com/open-telemetry/opentelemetry-dotnet/pull/5351#discussion_r1501269285

Issue #5377, opened 2024-02-21
HttpProtobuf protocol not working


srinivasmummareddy:
# Bug Report

List of [all OpenTelemetry NuGet
packages](https://www.nuget.org/profiles/OpenTelemetry) and version that you are
using (e.g. `OpenTelemetry 1.0.2`):
		<PackageReference Include="OpenTelemetry" Version="1.7.0" />
		<PackageReference Include="OpenTelemetry.AutoInstrumentation" Version="1.2.0" />
		<PackageReference Include="OpenTelemetry.Exporter.Console" Version="1.7.0" />
		<PackageReference Include="OpenTelemetry.Exporter.OpenTelemetryProtocol" Version="1.7.0" />
		<PackageReference Include="OpenTelemetry.Extensions.Hosting" Version="1.7.0" />
		<PackageReference Include="OpenTelemetry.Instrumentation.AspNetCore" Version="1.7.0" />
		<PackageReference Include="OpenTelemetry.Instrumentation.Http" Version="1.7.0" />
		<PackageReference Include="OpenTelemetry.Instrumentation.Runtime" Version="1.7.0" />

Runtime version (e.g. `net462`, `net48`, `netcoreapp3.1`, `net6.0` etc. You can
find this information from the `*.csproj` file):

net8.0

## Symptom
Whenever we send telemetry data using http protocol its not getting sent to collector. But,the same working fine with grpc

**What is the expected behavior?**
Should send data with httpprotobuf protocol but not working

## Reproduce
https://github.com/srinivasmummareddy/TelemetryHttpTester

## Additional Context
Everything is working as expected with grpc but its not working fine with http protobuf protocol. I have added opentelemetry collector config for your reference. I have also cross checked with grafana agent as well but issue persists.

vishweshbankwar:
@srinivasmummareddy - Could you try changing the endpoint to `http://otel-collector:4318/v1/traces`

srinivasmummareddy:
@vishweshbankwar  Didn't expect that we should the path as well as its taken care by default for grpc...It would be great if we add some example for this

MarlonnLima:
that was helpful, thanks!

PR #5385, opened 2024-02-23
[otlp] Add readme toc and clarify endpoint confusion

vishweshbankwar:
Fixes #
Design discussion issue #

## Changes

Adds toc to otlp exporter Readme as it is getting bigger. Readme needs more organization, planning to do that separately. Added a note about endpoint confusion when using HttpProtobuf https://github.com/open-telemetry/opentelemetry-dotnet/issues/5377


src/OpenTelemetry.Exporter.OpenTelemetryProtocol/README.md
```diff
@@ -6,6 +6,26 @@
...
 implementation.
 
+<!-- markdownlint-disable MD033 -->
+<details>
+<summary>Table of Contents</summary>
+
+* [Installation](#installation)
+* [Enable Trace Exporter](#enable-trace-exporter)
+* [Enable Metric Exporter](#enable-metric-exporter)
+* [Enable Log Exporter](#enable-log-exporter)
```
reyang, line 16: Any reason behind this specific ordering?

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/README.md
```diff
@@ -169,6 +189,11 @@ appBuilder.Services.AddOptions<OtlpExporterOptions>()
   `OtlpExportProtocol.Grpc` and "localhost:4318" for
   `OtlpExportProtocol.HttpProtobuf`.
 
+> [!NOTE]
+> When using `OtlpExportProtocol.HttpProtobuf`, the full URL MUST be
+> provided, including the signal-specific path v1/{signal}. For example, for
+> traces, the full URL will look like <http://your-custom-endpoint/v1/traces>.
```
reyang, line 195: This is an invalid link (we don't expect folks to click it), change to backquotes.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpExporterOptions.cs
```diff
@@ -97,6 +97,10 @@ public OtlpExporterOptions()
     /// * http://localhost:4317 for <see cref="OtlpExportProtocol.Grpc"/>
     /// * http://localhost:4318 for <see cref="OtlpExportProtocol.HttpProtobuf"/>.
     /// </summary>
+    /// <remarks>
+    /// When using <see cref="OtlpExportProtocol.HttpProtobuf"/>, the full URL MUST be provided, including the signal-specific path v1/{signal}.
```
reyang, line 101: Not a blocker for this PR, what if the user doesn't specify a full URL or doesn't include signal-specific path - do we validate the input and give the caller an exception?

reyang (approved):

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpExporterOptions.cs
vishweshbankwar: We do not validate the input. The Uri is used as is. The only way they would know is via sdk internal logs when the request fails. Exception would be a better option, will look in to it.

rajkumar-rangaraj (approved):

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/README.md
```diff
@@ -6,6 +6,26 @@
 [The OTLP (OpenTelemetry Protocol) exporter](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md)
 implementation.
 
+<!-- markdownlint-disable MD033 -->
+<details>
+<summary>Table of Contents</summary>
```
utpilla, line 11: This renders as a collapsed list. That makes the toc not so visible. Could we simply use a list of links (something similar to [this](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/README.md?plain=1#L6-L14)) instead of using `<details>`?

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpExporterOptions.cs
utpilla: The [documentation comments](https://github.com/open-telemetry/opentelemetry-dotnet/pull/5385/files#diff-1e7c29e7cbb12b62aa8515d38a5bb21af38f73d4594559d323bc180d49090fefR98) still mention the default value for `OtlpExportProtocol.HttpProtobuf` to be `http://localhost:4318`. Could we update that to the actual default values that would be used by the exporter? That is, `http://localhost:4318/v1/traces` for traces and so on?

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/README.md
TimothyMothra: `<details open>` will render as open by default.
https://gist.github.com/pierrejoubert73/902cc94d79424356a8d20be2b382e1ab#31-expand-by-default

utpilla: > I don't think there's a wrong answer here.

It all depends on what we want the user experience for our docs to be. Do we want the TOC to be noticeable? We could also have it collapsed and still make it noticeable by making the font bold, bigger in size etc. Whatever we decide to do, we should be consistent as there are other docs in the repo which have TOC:
- [OpenTelemetry API](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Api/README.md)
- [OpenTelemetry SDK](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/README.md)

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpExporterOptions.cs
utpilla: This was resolved in #5434

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/README.md
utpilla: It looks like we are already not consistent with other docs in the repo: https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/docs/metrics

It's okay to use the `details` tag in this PR for now. We could address the issue of using the same type of table across all docs later.

utpilla (approved):

PR #5026, opened 2023-11-03
Fix Activity http.route and name

alanwest:
Fixes #3485
Fixes #3461
Fixes #2967 
Fixes #3771

#### This PR fixes a number of items with respect to the `Activity.DisplayName` and `http.route` attribute:

* `Activity.DisplayName` now conforms to the specification
    * It is set to  `{http.method} {http.route}` when there is a low-cardinality `http.route`
    * It is `{http.method}` when there is not a low-cardinality `http.route`
* `http.route` was missing for some attribute-based routing scenarios
* `http.route` was missing when using minimal API
* `http.route`, when present, is now the same for both the Activity and metric emitted

#### There remain scenarios not fixed by this PR and also changes their behavior. You can review [this readme](https://github.com/open-telemetry/opentelemetry-dotnet/blob/2ce8425902ee64cf555703efd69bb55dbb40cccb/test/OpenTelemetry.Instrumentation.AspNetCore.Tests/RouteTests/README.net8.0.md) for the scenarios that remain unresolved. In summary:

* When using conventional-based routing, `http.route` is set to the matched route template. This does not accurately reflect the actual route when the controller/action invoked are not the defaults. For example:
    * See this [test case](https://github.com/open-telemetry/opentelemetry-dotnet/blob/2ce8425902ee64cf555703efd69bb55dbb40cccb/test/OpenTelemetry.Instrumentation.AspNetCore.Tests/RouteTests/README.net8.0.md#conventionalrouting-non-default-action-with-route-parameter-and-query-string).
        * The path invoked is `/ConventionalRoute/ActionWithStringParameter/2?num=3`
        * `http.route` is set to `{controller=ConventionalRoute}/{action=Default}/{id?}`
        * Ideally, `http.route` would be something like `ConventionalRoute/ActionWithStringParameter/{id?}`.
* When using Razor pages and invoking the root path (e.g. http://example.com/), `http.route` is empty.
    * See this [test case](https://github.com/open-telemetry/opentelemetry-dotnet/blob/2ce8425902ee64cf555703efd69bb55dbb40cccb/test/OpenTelemetry.Instrumentation.AspNetCore.Tests/RouteTests/README.net8.0.md#razorpages-root-path)
        * The path invoked is '/'
        * `http.route` attribute is not set
        * Ideally, `http.route` would be `/Index` because that is the default page.

These scenarios are unfortunate because it limits users from meaningfully querying metrics and faceting them by `http.route`. See #5056 and #5057.

PR #5402, opened 2024-02-29
Allow root action rewrites

PascalSenn:
# Explanation 
Since https://github.com/open-telemetry/opentelemetry-dotnet/pull/5026 the activity name of the asp net core span is set on activity end. This means that any change that was done to the name of this span will be overwritten. Many monitoring solutions ( elastic, sentry, datadog et al) use the name of the root span to group similar spans. A tool to control this was to rename the root activity, which is the asp net core activity if you have it enabled. As #5026 overwrites any changes to the root activity on transaction end, this control is gone. 

## Example

The path (or route pattern) for a GraphQL request is always the same. For example `/graphql`. The open telemetry integration of HotChocolate rewrites the name of the root span to the operation that you are execution. Specficially into something like `query FindUserById { userById }`, `mutation updateUserName { updateUserName }` ... . As of #5026, all graphql operations are reported as `POST /graphql/{**slug}` which makes it impossible to properly trace different operations. 

The specific treatment of GRPC in the HttpInListener shows that it would have the same issue, it just happens to be included by default.  

## Why the hooks is not enough
There is a hook which you can use to change the properties of the span. Frameworks like HotChocolate (or others) that want to overwrite the naming of the AspNetCore trace could just use this hook. But this is not a good solution:

1. It requires all frameworks to have a dependency on `OpenTelemetry.Instrumentation.AspNetCore`.
1. It would require the user to add additional configuration to the `AddAspNetCoreInstrumentation`  registration, which leads to confusion. 

## Changes

Before overwriting the activity name on activity end, the `DisplayName` is compared to the one that is set on activity start. If the values are not the same, the user ( or a library), is interested in renaming the name of the span. In this case the name of the span in not modified on activity end. 

### DIFF
--- a/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs
+++ b/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs
@@ -241,7 +241,13 @@ internal class HttpInListener : ListenerHandler
                     context.GetEndpoint() as RouteEndpoint)?.RoutePattern.RawText;
             if (!string.IsNullOrEmpty(routePattern))
             {
-                activity.DisplayName = GetDisplayName(context.Request.Method, routePattern);
+                // only override the display name if it was not set by the user
+                var previousDisplayName = GetDisplayName(context.Request.Method);
+                if (activity.DisplayName == previousDisplayName)
+                {
+                    activity.DisplayName = GetDisplayName(context.Request.Method, routePattern);
+                }
+
                 activity.SetTag(SemanticConventions.AttributeHttpRoute, routePattern);
             }
 #endif
--- a/test/OpenTelemetry.Instrumentation.AspNetCore.Tests/RouteTests/RoutingTests.cs
+++ b/test/OpenTelemetry.Instrumentation.AspNetCore.Tests/RouteTests/RoutingTests.cs
@@ -85,9 +85,12 @@ public class RoutingTests : IClassFixture<RoutingTestFixture>
 
         // Activity.DisplayName should be a combination of http.method + http.route attributes, see:
         // https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.md#name
-        var expectedActivityDisplayName = string.IsNullOrEmpty(expectedHttpRoute)
-            ? testCase.HttpMethod
-            : $"{testCase.HttpMethod} {expectedHttpRoute}";
+        // unless the user has overriden the route name
+        var expectedActivityDisplayName = string.IsNullOrEmpty(testCase.ExpectedDisplayName)
+            ? string.IsNullOrEmpty(expectedHttpRoute)
+                ? testCase.HttpMethod
+                : $"{testCase.HttpMethod} {expectedHttpRoute}"
+            : testCase.ExpectedDisplayName;
 
         Assert.Equal(expectedActivityDisplayName, activity.DisplayName);
 
--- a/test/OpenTelemetry.Instrumentation.AspNetCore.Tests/RouteTests/README.net6.0.md
+++ b/test/OpenTelemetry.Instrumentation.AspNetCore.Tests/RouteTests/README.net6.0.md
@@ -23,6 +23,7 @@
 | :green_heart: | MinimalApi | [Action without parameter](#minimalapi-action-without-parameter) |
 | :green_heart: | MinimalApi | [Action with parameter](#minimalapi-action-with-parameter) |
 | :green_heart: | ExceptionMiddleware | [Exception Handled by Exception Handler Middleware](#exceptionmiddleware-exception-handled-by-exception-handler-middleware) |
+| :broken_heart: | ConventionalRouting | [Overridden display name](#conventionalrouting-overridden-display-name) |
 
 ## ConventionalRouting: Root path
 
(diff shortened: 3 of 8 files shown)


vishweshbankwar:
@PascalSenn Do you mind elaborating bit more on

> The specific treatment of GRPC in the HttpInListener shows that it would have the same issue, it just happens to be included by default.

PascalSenn:
@vishweshbankwar  

The DisplayName of the activity is set on activity end here:
https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs#L244

After this code, there is a special treatment for GRPC:
https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs#L251-L254

Which then overwrites the activity name to the RPC spectific semantic convention:
https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs#L351-L354

If GRPC would not be handled in `OpenTelemetry.Instrumentation.AspNetCore` there would be no way for a GRPC component further down the middleware pipeline to set correct root span name for the RPC semantic convention unless you would register a specific  enricher with`EnrichWithHttpResponse` on the AspNetCore instrumentation

vishweshbankwar:
> @vishweshbankwar
> 
> The DisplayName of the activity is set on activity end here: https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs#L244
> 
> After this code, there is a special treatment for GRPC: https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs#L251-L254
> 
> Which then overwrites the activity name to the RPC spectific semantic convention: https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs#L351-L354
> 
> If GRPC would not be handled in `OpenTelemetry.Instrumentation.AspNetCore` there would be no way for a GRPC component further down the middleware pipeline to set correct root span name for the RPC semantic convention unless you would register a specific enricher with`EnrichWithHttpResponse` on the AspNetCore instrumentation

@PascalSenn - Thanks for explaining. I think it's acceptable to implement this minor adjustment in the instrumentation library. However, it's important to note that the library cannot handle any other sequencing issues comprehensively. It primarily enhances activities according to the OTel specification. For any other requirements, users can utilize the `Enrich` API to customize existing activities.

PascalSenn:
@vishweshbankwar 
Do you mean with 

> I think it's acceptable to implement this minor adjustment in the instrumentation library

the GRPC code which is already in place or the change in this PR?

vishweshbankwar:
> @vishweshbankwar Do you mean with
> 
> > I think it's acceptable to implement this minor adjustment in the instrumentation library
> 
> the GRPC code which is already in place or the change in this PR?

I meant the change in this PR.

PascalSenn:
@vishweshbankwar 

Very well, i would assume it's also what most frameworks with custom semantic conventions (like RPC) would want to do. Additional properties are also not a problem. They most important part is the name of the root activity.

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs
```diff
@@ -241,7 +241,13 @@ public void OnStopActivity(Activity activity, object payload)
                     context.GetEndpoint() as RouteEndpoint)?.RoutePattern.RawText;
             if (!string.IsNullOrEmpty(routePattern))
             {
-                activity.DisplayName = GetDisplayName(context.Request.Method, routePattern);
+                // only override the display name if it was not set by the user
```
vishweshbankwar, line 244: @PascalSenn - Could you add a unit test for this? These tests are more focused on `http.route` tag value.

PascalSenn: You mean instead of the one in RoutingTests?

vishweshbankwar: ok to add this in addition to unit test. @alanwest - Do you have any preference?

alanwest:
Before we move forward with this PR, there's a few things we should discuss.

I captured some general thoughts/questions on #5417 which may be best to discuss in our next SIG meeting.

With respect specifically to this AspNetCore instrumentation, if we move forward with this PR, we should confirm how and if this can be solved in the event that ASP.NET 9 introduces native trace instrumentation (see https://github.com/dotnet/aspnetcore/issues/52439)

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs
alanwest: I think it makes sense to test this new behavior separate from the tests for http.route.

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/RouteTests/RoutingTests.cs
```diff
@@ -85,9 +85,12 @@ public async Task TestHttpRoute(RoutingTestCases.TestCase testCase)
 
         // Activity.DisplayName should be a combination of http.method + http.route attributes, see:
         // https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.md#name
-        var expectedActivityDisplayName = string.IsNullOrEmpty(expectedHttpRoute)
-            ? testCase.HttpMethod
-            : $"{testCase.HttpMethod} {expectedHttpRoute}";
+        // unless the user has overriden the route name
```
OptimusPi, line 88: replace typo `overriden` with `overridden` to satisfy ms-misspell PR check.

OptimusPi (changes requested):
Fix typo here to satisfy PR misspell bot.

vishweshbankwar:
@PascalSenn - We discussed this during yesterday's SIG meeting. Before proceeding with committing this change, we'd like to wait and gather more information on how https://github.com/dotnet/aspnetcore/issues/52439 will address these scenarios. This is to ensure that we don't encounter similar issues when users upgrade to a new .NET version where instrumentation is natively implemented within ASP.NET Core.

vishweshbankwar:
Closing this for now. See https://github.com/open-telemetry/opentelemetry-dotnet/pull/5402#issuecomment-1995982094

Issue #5430, opened 2024-03-08
Shared files are out of sync between `opentelemetry-dotnet` and `opentelemetry-dotnet-contrib`


TimothyMothra:
Some of the files in the shared directory are out of sync between the `opentelemetry-dotnet` and `opentelemetry-dotnet-contrib` repos.
Opening this issue to track syncing these files.

- `ActivityInstrumentationHelper.cs`  
    - Dotnet-Contrib version is using compiled Linq Expressions.
        ![image](https://github.com/open-telemetry/opentelemetry-dotnet/assets/28785781/2be7666f-8adc-4e1b-920e-5fa6691725ce)

- `ResourceSemanticConventions.cs`
     - Namespaces differ. `OpenTelemetry.Resources` vs `OpenTelemetry.Trace`
        ![image](https://github.com/open-telemetry/opentelemetry-dotnet/assets/28785781/1ac2cc24-c5dd-4ce7-96ce-596a1cbc70e4)
    - Dotnet-Contrib version has extra constants
        ![image](https://github.com/open-telemetry/opentelemetry-dotnet/assets/28785781/9f0cebae-02cb-462a-b82e-2e543f130430)

- `SemanticConventions.cs`
    - Dotnet version has extra constants
        ![image](https://github.com/open-telemetry/opentelemetry-dotnet/assets/28785781/a8a2adab-9d0c-4ded-a0bc-878f9b25246d)

PR #5432, opened 2024-03-09
sync changes in shared files

TimothyMothra:
Towards #5430

Some of the Shared files between `opentelemetry-dotnet` and `opentelemetry-dotnet-contrib` are out of sync.
This PR syncs the contents of those files.
If this is approved, I'll make the corresponding changes in the contrib repo.

## Changes
- ActivityInstrumentationHelper
  remove `#pragma warning restore`. There was no disable in this file. I assume it was left over from some other refactoring.
- ResourceSemanticConventions
  Contrib had a few extra constants. I synced those here and alphabetized.
- SemanticConventions
  Dotnet only had a 1.21.0 section, where as Contrib had both 1.21.0 and 1.23.0. I decided to keep both sections. I verified that most of these constants still exist in 1.23.0 and moved them under that section.


CodeBlanch (approved):
LGTM

Issue #4929, opened 2023-10-06
Test stability - OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests.MockCollectorIntegrationTests.TestRecoveryAfterFailedExport


reyang:
![image](https://github.com/open-telemetry/opentelemetry-dotnet/assets/17327289/aed61c36-c38d-4653-ae9d-e24260f8121d)

reyang:
Failed again https://github.com/open-telemetry/opentelemetry-dotnet/actions/runs/6499900795/job/17654192628?pr=4933

![image](https://github.com/open-telemetry/opentelemetry-dotnet/assets/17327289/530a3b14-a8fa-46eb-b252-580429b637f8)

reyang:
https://github.com/open-telemetry/opentelemetry-dotnet/actions/runs/7834756766/job/21378670233?pr=5329

```text
[xUnit.net 00:00:09.06]     OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests.MockCollectorIntegrationTests.TestRecoveryAfterFailedExport [FAIL]
[xUnit.net 00:00:09.06]       Assert.Equal() Failure: Values differ
[xUnit.net 00:00:09.06]       Expected: Failure
[xUnit.net 00:00:09.06]       Actual:   Success
[xUnit.net 00:00:09.06]       Stack Trace:
[xUnit.net 00:00:09.07]         D:\a\opentelemetry-dotnet\opentelemetry-dotnet\test\OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests\MockCollectorIntegrationTests.cs(84,0): at OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests.MockCollectorIntegrationTests.TestRecoveryAfterFailedExport()
[xUnit.net 00:00:09.07]         --- End of stack trace from previous location ---
  Failed OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests.MockCollectorIntegrationTests.TestRecoveryAfterFailedExport [6 s]
  Error Message:
   Assert.Equal() Failure: Values differ
Expected: Failure
Actual:   Success
  Stack Trace:
[...]

PR #5433, opened 2024-03-09
[otlp] Add Retry Handler

vishweshbankwar:
Towards #1779 
Design discussion issue #

## Changes

Adds a retry transmission handler for retrying transient errors. It covers both http and grpc protocol.

Planning to do a follow up PR which will add the ability to opt-in for retries via an experimental environment variable. 


src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/Transmission/OtlpExporterRetryTransmissionHandler.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+    }
+
+    protected override bool OnSubmitRequestFailure(TRequest request, ExportClientResponse response)
+    {
+        var nextRetryDelayMilliseconds = OtlpRetry.InitialBackoffMilliseconds;
+        while (RetryHelper.ShouldRetryRequest(request, response, nextRetryDelayMilliseconds, out var retryResult))
+        {
+            // Note: This delay cannot exceed the configured timeout period for otlp exporter.
+            // In case if the backend responds with delay(via RetryAfter) that would result in exceeding the configured timeout period
```
utpilla, line 23:
```suggestion
            // If the backend responds with `RetryAfter` duration that would result in exceeding the configured timeout period
```

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/Transmission/RetryHelper.cs
```diff
@@ -0,0 +1,32 @@
...
+
+namespace OpenTelemetry.Exporter.OpenTelemetryProtocol.Implementation.Transmission;
+
+internal static class RetryHelper
+{
+    internal static bool ShouldRetryRequest<TRequest>(TRequest request, ExportClientResponse response, int retryDelayMilliseconds, out OtlpRetry.RetryResult retryResult)
+    {
+        if (response is ExportClientGrpcResponse)
+        {
+            if (response.Exception is RpcException rpcException
```
utpilla, line 15: Could we move this check inside `TryGetGrpcRetryResult` method? It makes the `TryGetGrpcRetryResult` method more complete and we wouldn't have to worry about any pre-requisites before calling this method. It would also be consistent with the `TryGetHttpRetryResult` method that way.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/Transmission/OtlpExporterRetryTransmissionHandler.cs
```diff
@@ -0,0 +1,37 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#nullable enable
+
+using OpenTelemetry.Exporter.OpenTelemetryProtocol.Implementation.ExportClient;
+
+namespace OpenTelemetry.Exporter.OpenTelemetryProtocol.Implementation.Transmission;
+
+internal class OtlpExporterRetryTransmissionHandler<TRequest> : OtlpExporterTransmissionHandler<TRequest>
```
utpilla, line 10:
```suggestion
internal sealed class OtlpExporterRetryTransmissionHandler<TRequest> : OtlpExporterTransmissionHandler<TRequest>
```

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/MockCollectorIntegrationTests.cs
```diff
@@ -91,6 +98,197 @@ public async Task TestRecoveryAfterFailedExport()
...
 
+    [Theory]
+    [InlineData(true, ExportResult.Success, Grpc.Core.StatusCode.Unavailable)]
+    [InlineData(true, ExportResult.Success, Grpc.Core.StatusCode.Cancelled)]
+    [InlineData(true, ExportResult.Success, Grpc.Core.StatusCode.Aborted)]
+    [InlineData(true, ExportResult.Success, Grpc.Core.StatusCode.OutOfRange)]
+    [InlineData(true, ExportResult.Success, Grpc.Core.StatusCode.DataLoss)]
+    [InlineData(true, ExportResult.Failure, Grpc.Core.StatusCode.Internal)]
+    [InlineData(true, ExportResult.Failure, Grpc.Core.StatusCode.InvalidArgument)]
+    [InlineData(true, ExportResult.Failure, Grpc.Core.StatusCode.Unimplemented)]
```
utpilla, line 109: Retrying for unimplemented status code seems wrong.

How was this list compiled? The [spec](https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#failures) also mentions this status code to be non-retriable:

|gRPC Code|Retryable?|
|---------|----------|
|CANCELLED|Yes|
|UNKNOWN|No|
|INVALID_ARGUMENT|No|
|DEADLINE_EXCEEDED|Yes|
|NOT_FOUND|No|
|ALREADY_EXISTS|No|
|PERMISSION_DENIED|No|
|UNAUTHENTICATED|No|
|RESOURCE_EXHAUSTED|Only if the server can recover (see below)|
|FAILED_PRECONDITION|No|
|ABORTED|Yes|
|OUT_OF_RANGE|Yes|
|UNIMPLEMENTED|No|
|INTERNAL|No|
|UNAVAILABLE|Yes|
|DATA_LOSS|Yes|

vishweshbankwar: > The [spec](https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#failures) also mentions this status code to be non-retriable

Correct and that's what the test confirms. The test is set up in a way that the first response would be one of the status codes from the list and the second one would be `Ok`. 

For e.g. take a look at this case

`InlineData(true, ExportResult.Failure, Grpc.Core.StatusCode.InvalidArgument)`

When the status code is retryable then the expected result should be `ExportResult.Success`. But in this case we assert that the result is `ExportResult.Failure` i.e. the request will not be retried.

This ^ is different than the following case

`[InlineData(true, ExportResult.Success, Grpc.Core.StatusCode.Unavailable)]`

utpilla: Oh okay. I got confused because of the parameter name `shouldRetry`. Could you change it to `useRetryTransmissionHandler`? That makes the intent clear.

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/MockCollectorIntegrationTests.cs
```diff
@@ -91,6 +98,197 @@ public async Task TestRecoveryAfterFailedExport()
         await host.StopAsync();
     }
 
+    [Theory]
+    [InlineData(true, ExportResult.Success, Grpc.Core.StatusCode.Unavailable)]
```
utpilla, line 102: Add a test case for `StatusCode.DeadlineExceeded` as well.

vishweshbankwar: see https://github.com/open-telemetry/opentelemetry-dotnet/issues/5436. I added the comment

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/Transmission/RetryHelper.cs
vishweshbankwar: done

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/OtlpRetryTests.cs
```diff
@@ -186,16 +188,6 @@ public static IEnumerable<object[]> GetGrpcTestCases()
                     },
                     expectedRetryAttempts: 9),
             };
-
-            yield return new[]
```
vishweshbankwar, line 190: @alanwest - I did not quite understand the significance of this test. It was testing that we would retry if the deadline is not set and server responds with a very high delay. We should not run in to this case as we would always have the deadline set, let me know if I am missing something here.

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/MockCollectorIntegrationTests.cs
utpilla: We should still add a test case for it and expect the result to be `ExportResult.Failure` always?

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/OtlpRetry.cs
```diff
@@ -83,9 +83,16 @@ public static bool ShouldHandleHttpRequestException(Exception? exception)
         return true;
     }
 
-    public static bool TryGetGrpcRetryResult(StatusCode statusCode, DateTime? deadline, Metadata trailers, int retryDelayMilliseconds, out RetryResult retryResult)
+    public static bool TryGetGrpcRetryResult(ExportClientGrpcResponse response, int retryDelayMilliseconds, out RetryResult retryResult)
     {
-        return TryGetRetryResult(statusCode, IsGrpcStatusCodeRetryable, deadline, trailers, TryGetGrpcRetryDelay, retryDelayMilliseconds, out retryResult);
+        retryResult = default;
```
utpilla, line 88: nit: Move this to the end of the method before returning `false`. The `TryGetRetryResult` method would anyway set some value to `retryResult`.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/Transmission/OtlpExporterRetryTransmissionHandler.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        var nextRetryDelayMilliseconds = OtlpRetry.InitialBackoffMilliseconds;
+        while (RetryHelper.ShouldRetryRequest(request, response, nextRetryDelayMilliseconds, out var retryResult))
+        {
+            // Note: This delay cannot exceed the configured timeout period for otlp exporter.
+            // If the backend responds with `RetryAfter` duration that would result in exceeding the configured timeout period
+            // we would fail fast and drop the data.
+            Thread.Sleep(retryResult.RetryDelay);
+
+            nextRetryDelayMilliseconds = retryResult.NextRetryDelayMilliseconds;
```
utpilla, line 27: Move this to the end of the loop. Why waste cycles on setting `nextRetryDelayMilliseconds` when it might not even be needed if `TryRetryRequest` succeeds?

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/Transmission/RetryHelper.cs
```diff
@@ -0,0 +1,30 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+using OpenTelemetry.Exporter.OpenTelemetryProtocol.Implementation.ExportClient;
```
utpilla, line 4: Use `#nullable enable`.

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/MockCollectorIntegrationTests.cs
vishweshbankwar: Added. It will be hard to show #5436 behavior here. The test would work as expected because there isn't actual delay in the request processing.

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/OtlpRetry.cs
```diff
@@ -83,9 +83,16 @@ public static bool ShouldHandleHttpRequestException(Exception? exception)
...
+    public static bool TryGetGrpcRetryResult(ExportClientGrpcResponse response, int retryDelayMilliseconds, out RetryResult retryResult)
     {
-        return TryGetRetryResult(statusCode, IsGrpcStatusCodeRetryable, deadline, trailers, TryGetGrpcRetryDelay, retryDelayMilliseconds, out retryResult);
+        if (response.Exception is RpcException rpcException)
+        {
+            return TryGetRetryResult(rpcException.StatusCode, IsGrpcStatusCodeRetryable, response.DeadlineUtc, rpcException.Trailers, TryGetGrpcRetryDelay, retryDelayMilliseconds, out retryResult);
+        }
+
+        retryResult = default;
+
```
utpilla, line 94: nit: Probably no need of new line here.

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/MockCollectorIntegrationTests.cs
```diff
@@ -91,6 +98,201 @@ public async Task TestRecoveryAfterFailedExport()
...
+    [InlineData(true, ExportResult.Success, Grpc.Core.StatusCode.Aborted)]
+    [InlineData(true, ExportResult.Success, Grpc.Core.StatusCode.OutOfRange)]
+    [InlineData(true, ExportResult.Success, Grpc.Core.StatusCode.DataLoss)]
+    [InlineData(true, ExportResult.Failure, Grpc.Core.StatusCode.Internal)]
+    [InlineData(true, ExportResult.Failure, Grpc.Core.StatusCode.InvalidArgument)]
+    [InlineData(true, ExportResult.Failure, Grpc.Core.StatusCode.Unimplemented)]
+    [InlineData(true, ExportResult.Failure, Grpc.Core.StatusCode.FailedPrecondition)]
+    [InlineData(true, ExportResult.Failure, Grpc.Core.StatusCode.PermissionDenied)]
+    [InlineData(true, ExportResult.Failure, Grpc.Core.StatusCode.Unauthenticated)]
+    [InlineData(true, ExportResult.Success, Grpc.Core.StatusCode.DeadlineExceeded)]
```
utpilla, line 115: Shouldn't we expect `ExportResult.Failure` here?

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/MockCollectorIntegrationTests.cs
```diff
@@ -18,6 +22,9 @@ namespace OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests;
 
 public sealed class MockCollectorIntegrationTests
 {
+    private static int gRPCPort = 4317;
```
utpilla, line 25: Why use `static`? I think we could get rid of the `Interlocked.Increment` statements if we mark these integer as instance fields. Every `Fact` or `Theory` should create its own instance of `MockCollectorIntegrationTests`.

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/MockCollectorIntegrationTests.cs
```diff
@@ -91,6 +98,201 @@ public async Task TestRecoveryAfterFailedExport()
...
+                               var codes = responseCodesCsv.Split(",").Select(x => int.Parse(x)).ToArray();
+                               collectorState.SetStatusCodes(codes);
+                           });
+
+                       endpoints.MapGrpcService<MockTraceService>();
+                   });
+               }))
+           .StartAsync();
+
+        var httpClient = new HttpClient() { BaseAddress = new Uri($"http://localhost:{testHttpPort}") };
```
utpilla, line 161:
```suggestion
        using var httpClient = new HttpClient() { BaseAddress = new Uri($"http://localhost:{testHttpPort}") };
```

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/MockCollectorIntegrationTests.cs
vishweshbankwar: We have this issue currently: https://github.com/open-telemetry/opentelemetry-dotnet/issues/4929. I am trying to see if this fixes the issue. We had a similar issue with asp.net core tests and having separate port helped fixing it.

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/MockCollectorIntegrationTests.cs
```diff
@@ -91,6 +98,201 @@ public async Task TestRecoveryAfterFailedExport()
...
+        var otlpExporter = new OtlpTraceExporter(exporterOptions, new(), transmissionHandler);
+
+        var activitySourceName = "otel.grpc.retry.test";
+        using var source = new ActivitySource(activitySourceName);
+
+        using var tracerProvider = Sdk.CreateTracerProviderBuilder()
+            .AddSource(activitySourceName)
+            .Build();
+
+        var activity = source.StartActivity("GrpcRetryTest");
```
utpilla, line 194:
```suggestion
        using var activity = source.StartActivity("GrpcRetryTest");
```

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/MockCollectorIntegrationTests.cs
```diff
@@ -91,6 +98,201 @@ public async Task TestRecoveryAfterFailedExport()
...
+        activity.Stop();
+        var batch = new Batch<Activity>([activity], 1);
```
utpilla, line 196:
```suggestion
        using var batch = new Batch<Activity>([activity], 1);
```

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/MockCollectorIntegrationTests.cs
utpilla: Okay interesting. In that case, `Interlocked.Increment` should ensure that none of tests are using the same port.

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/MockCollectorIntegrationTests.cs
```diff
@@ -91,6 +98,201 @@ public async Task TestRecoveryAfterFailedExport()
...
+                           var state = ctx.RequestServices.GetRequiredService<MockCollectorHttpState>();
+                           ctx.Response.StatusCode = (int)state.NextStatus();
+
+                           await ctx.Response.WriteAsync("Request Received.");
+                       });
+                   });
+               }))
+           .StartAsync();
+
+        var httpClient = new HttpClient() { BaseAddress = new Uri($"http://localhost:{testHttpPort}") };
```
utpilla, line 255:
```suggestion
        using var httpClient = new HttpClient() { BaseAddress = new Uri($"http://localhost:{testHttpPort}") };
```

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/MockCollectorIntegrationTests.cs
vishweshbankwar: See https://github.com/open-telemetry/opentelemetry-dotnet/pull/5433#discussion_r1520568932

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/MockCollectorIntegrationTests.cs
utpilla: We should be able to do it by tweaking the test to selectively return failure for both attempts, right? Something like this:

```csharp
var codes = new[] { initialStatusCode, Grpc.Core.StatusCode.OK };

if (initialStatusCode == StatusCode.DeadlineExceeded)
{
    codes[1] = initialStatusCode;
}
```

Anyway, I think if we decide that we would never retry on Deadline Exceeded, it's better to change the code to return `false` for that code instead of updating the test here.

https://github.com/open-telemetry/opentelemetry-dotnet/blob/adc89d94249446cbd153b99330f274a76a33ca09/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/OtlpRetry.cs#L199-L207

test/OpenTelemetry.Exporter.OpenTelemetryProtocol.Tests/MockCollectorIntegrationTests.cs
```diff
@@ -91,6 +101,201 @@ public async Task TestRecoveryAfterFailedExport()
...
+    [InlineData(false, ExportResult.Failure, Grpc.Core.StatusCode.FailedPrecondition)]
+    [InlineData(false, ExportResult.Failure, Grpc.Core.StatusCode.DeadlineExceeded)]
+    public async Task GrpcRetryTests(bool useRetryTransmissionHandler, ExportResult expectedResult, Grpc.Core.StatusCode initialStatusCode)
+    {
+        var testGrpcPort = Interlocked.Increment(ref gRPCPort);
+        var testHttpPort = Interlocked.Increment(ref httpPort);
+
+        using var host = await new HostBuilder()
+           .ConfigureWebHostDefaults(webBuilder => webBuilder
+                .ConfigureKestrel(options =>
```
utpilla, line 135: nit: We could speed up the tests by clearing the default logging providers that get added when building the hosts.

```csharp
.ConfigureLogging(loggingBuilder => loggingBuilder.ClearProviders())
```

utpilla (approved):

src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/Transmission/OtlpExporterRetryTransmissionHandler.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        var nextRetryDelayMilliseconds = OtlpRetry.InitialBackoffMilliseconds;
+        while (RetryHelper.ShouldRetryRequest(request, response, nextRetryDelayMilliseconds, out var retryResult))
+        {
+            // Note: This delay cannot exceed the configured timeout period for otlp exporter.
+            // If the backend responds with `RetryAfter` duration that would result in exceeding the configured timeout period
+            // we would fail fast and drop the data.
+            Thread.Sleep(retryResult.RetryDelay);
+
+            if (this.TryRetryRequest(request, response.DeadlineUtc, out response))
```
rajkumar-rangaraj, line 27: Line 44 in `OtlpExporterTransmissionHandler.cs` calls `this.OnSubmitRequestFailure(request, response)`, which means it is recursively calling this method `OnSubmitRequestFailure` on Line 17. If that happens, it always gets `OtlpRetry.InitialBackoffMilliseconds`, and in certain cases, we may end up in an endless loop. Is that correct?

rajkumar-rangaraj (approved):

PR #5472, opened 2024-03-25
[Exporter.Console] Add support for ActivitySource.Version

Kielek:
Fixes #
Design discussion issue #

## Changes

[Exporter.Console] Add support for ActivitySource.Version

* ~~[ ] Unit tests added/updated~~ No tests for similar functionalities.


cijothomas (approved):

rajkumar-rangaraj (approved):

src/OpenTelemetry.Exporter.Console/ConsoleActivityExporter.cs
```diff
@@ -32,6 +32,11 @@ public override ExportResult Export(in Batch<Activity> batch)
             }
 
             this.WriteLine($"Activity.ActivitySourceName: {activity.Source.Name}");
+            if (!string.IsNullOrEmpty(activity.Source.Version))
+            {
+                this.WriteLine($"Activity.ActivitySourceVersion: {activity.Source.Version}");
```
reyang, line 37: Why do we want `Activity.ActivitySourceVersion` instead of `Activity.Source.Version`?

Kielek: Same convention as for ActivitySourceName

cijothomas: It maybe beneficial to use ActivitySourceVersion(aka Tracer version) to make it easier to realize that ActivitySource and Tracer are the same concept.
AND/OR use InstrumentationScope wording?

Kielek: I am less concerned about wording as it is mostly for debugging purposes. It should be fine to have it this value in any form.

I can make additional changes in this PR on follow up PR if we have agreement. It will be great to include this fix on 1.8.0 release so I would like to avoid long discussion.

cijothomas: This is a non-blocking comment. We do have another task to revisit console output format, so any additional improvements can be addressed as part of that.

PR #5510, opened 2024-04-04
Minor refactoring for AggregatorStore and MetricPoint

utpilla:
## Changes
- Extract the logic for MetricPoint reclaim to its own method
- Add code comments
- Rename a method to make its intent clear

### DIFF
--- a/src/OpenTelemetry/Metrics/MetricPoint.cs
+++ b/src/OpenTelemetry/Metrics/MetricPoint.cs
@@ -1197,9 +1197,10 @@ public struct MetricPoint
     }
 
     /// <summary>
-    /// Denote that this MetricPoint is reclaimed.
+    /// This method sets the member object references of MetricPoint to `null`.
+    /// This is done to have them collected faster by GC.
     /// </summary>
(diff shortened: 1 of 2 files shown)


src/OpenTelemetry/Metrics/MetricPoint.cs
```diff
@@ -1197,9 +1197,10 @@ internal void TakeSnapshotWithExemplar(bool outputDelta)
     }
 
     /// <summary>
-    /// Denote that this MetricPoint is reclaimed.
+    /// This method sets the member object references of MetricPoint to `null`.
```
vishweshbankwar, line 1200: number of object references?

vishweshbankwar (approved):

CodeBlanch (approved):
LGTM

src/OpenTelemetry/Metrics/MetricPoint.cs
utpilla: I used "**member** object references" here to denote any fields/ properties of `MetricPoint` struct that are reference types.

PR #5577, opened 2024-04-26
Use struct in the complex type logging demo

reyang:
Encourage the readers to write performant code.

### DIFF
--- a/docs/logs/complex-objects/FoodRecallNotice.cs
+++ b/docs/logs/complex-objects/FoodRecallNotice.cs
@@ -1,7 +1,7 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
-public class FoodRecallNotice
+public struct FoodRecallNotice
 {
     public string? BrandName { get; set; }
 
--- a/docs/logs/complex-objects/Program.cs
+++ b/docs/logs/complex-objects/Program.cs
@@ -34,5 +34,5 @@ internal static partial class LoggerExtensions
     [LoggerMessage(LogLevel.Critical)]
     public static partial void FoodRecallNotice(
         this ILogger logger,
-        [LogProperties(OmitReferenceName = true)] FoodRecallNotice foodRecallNotice);
+        [LogProperties(OmitReferenceName = true)] in FoodRecallNotice foodRecallNotice);
 }
--- a/docs/logs/complex-objects/README.md
+++ b/docs/logs/complex-objects/README.md
@@ -30,16 +30,12 @@ dotnet add package Microsoft.Extensions.Telemetry.Abstractions
 Define a new complex data type, as shown in [FoodRecallNotice.cs](./FoodRecallNotice.cs):
 
 ```csharp
-public class FoodRecallNotice
+public struct FoodRecallNotice
 {
     public string? BrandName { get; set; }
-
     public string? ProductDescription { get; set; }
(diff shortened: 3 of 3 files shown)


Kielek (approved):

rajkumar-rangaraj (approved):

docs/logs/complex-objects/Program.cs
```diff
@@ -34,5 +34,5 @@ internal static partial class LoggerExtensions
     [LoggerMessage(LogLevel.Critical)]
     public static partial void FoodRecallNotice(
         this ILogger logger,
-        [LogProperties(OmitReferenceName = true)] FoodRecallNotice foodRecallNotice);
+        [LogProperties(OmitReferenceName = true)] in FoodRecallNotice foodRecallNotice);
```
CodeBlanch, line 37: `in` with mutable structs can be tricky. If the compiler can't determine stuff is `readonly` it will make defensive copies. I think the auto-properties shown here are fine (IIRC the getters get marked as `readonly` automatically for auto-properties), but users could run into trouble if they handroll more. If we want to recommend `in` we may also want to recommend the `struct`s be `readonly` too? `readonly struct` + `in` work always.

reyang: Build failed, seems it is only available in .NET Core. I'll revert the change.

docs/logs/complex-objects/complex-objects.csproj
```diff
@@ -1,6 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <SuppressTfmSupportBuildWarnings>true</SuppressTfmSupportBuildWarnings>
+    <EmitCompilerGeneratedFiles>true</EmitCompilerGeneratedFiles>
```
reyang, line 4: This would put the generated source under `.\obj\Debug\net8.0\generated\Microsoft.Gen.Logging\Microsoft.Gen.Logging.LoggingGenerator\Logging.g.cs` which helps to understand things behind the scene.

docs/logs/complex-objects/FoodRecallNotice.cs
```diff
@@ -1,15 +1,15 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
-public class FoodRecallNotice
+public struct FoodRecallNotice
```
cijothomas, line 4: would you show the benchmarks results from this change?

reyang: |           Method |     Mean |   Error |   StdDev |  Gen 0 | Allocated |
|----------------- |---------:|--------:|---------:|-------:|----------:|
| LogPayloadClass  | 122.1 ns | 3.45 ns | 9.67 ns  | 0.0088 |      56 B |
| LogPayloadStruct | 122.6 ns | 3.60 ns | 10.44 ns |      - |         - |

reyang: I think the only noticeable difference is heap vs. stack alloc (CPU is almost the same).

docs/logs/complex-objects/complex-objects.csproj
CodeBlanch: Just FYI in case you weren't aware you can see them in IDE under the analyzer:

image

PR #5593, opened 2024-05-06
[Instrumentation.GrpcNetClient] Move to contrib repository

Kielek:
Towards: #5526
Related to: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1704

## Changes

[Instrumentation.GrpcNetClient] Move to contrib repository


.github/workflows/ci.yml
```diff
@@ -87,6 +87,7 @@ jobs:
       project-name: './build/InstrumentationLibraries.proj'
       project-build-commands: '-p:RunningDotNetPack=true -p:ExposeExperimentalFeatures=false'
       code-cov-name: 'Instrumentation-Stable'
+      tfm-list: '[ "net6.0", "net7.0", "net8.0" ]'
```
reyang, line 90: Curious, why is this needed here? Can this info be deduced from existing configuration? Ideally, TFM info should be in a central place so adding future TFM and retiring legacy TFM would only touch one single file.

examples/Console/Examples.Console.csproj
```diff
@@ -21,12 +21,12 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
     <PackageReference Include="OpenTelemetry.Instrumentation.Http" />
+    <PackageReference Include="OpenTelemetry.Instrumentation.GrpcNetClient" />
```
reyang, line 24: I vote for removing this dependency and the corresponding example code entirely from this repo.

.github/workflows/ci.yml
Kielek: It is temporary change. Steps ` build-test-instrumentation-stable:` and `build-test-instrumentation-experimental` will be fully removed from the settings when Instrumentation.AspNetCore package is also migrated to the contrib. I would not care about this change so much.

If we speaking more generally, tfm's  are defined in two places right now:
1. Common.props - for .NET definitions what to build an how to pack production projects
2. Build definitions (mostly Component.BuildTets.yml) to define frameworks to execute tests in the pipeline.

Adding one more layer on top of it, to unify this two things could be artificial and potentially require introducing custom parsing. As long as we have _only 2 places_ it should be fine to keep it as is.

Expected changes happen once per year (together with .NET releases).

examples/Console/Examples.Console.csproj
Kielek: It GrpcNetClient example is tightly coupled with AspNetCore example. I will try to cleanup this after all instrumentation packages are moved to the contrib.

```
        // Prerequisite for running this example.
        // In a separate console window, start the example
        // ASP.NET Core gRPC service by running the following command
        // from the reporoot\examples\GrpcService\.
        // (eg: C:\repos\opentelemetry-dotnet\examples\GrpcService\)
        //
        // dotnet run

        // To run this example, run the following command from
        // the reporoot\examples\Console\.
        // (eg: C:\repos\opentelemetry-dotnet\examples\Console\)
        //
        // dotnet run grpc
```

reyang (approved):
LGTM.

.github/workflows/ci.yml
reyang: Got it, thanks!

vishweshbankwar (approved):

rajkumar-rangaraj (approved):

PR #5643, opened 2024-05-22
[sdk] Provide better concurrency modes

reyang:
This is following the same design as push/pull metrics exporter: https://github.com/open-telemetry/opentelemetry-dotnet/blob/3cff6dbfd15ed0b539c289a4444198fe437a5ebf/src/OpenTelemetry/Metrics/ExportModesAttribute.cs#L9-L10

Exporter (also processors, samplers, etc.) authors can optionally provide additional hints so the SDK can better serve the need:

```csharp
[ConcurrencyModes(ConcurrencyModes.Multithreaded | ConcurrencyModes.Reentrant)]
internal class MyExporter : BaseExporter<LogRecord>
{
    ...
}
```

In addition, I envision that Console Exporters can leverage this to provide synchronization across multiple instances, so a log exporter and metrics exporter won't have race condition and cause garbled text in stdout:

```csharp
[ConcurrencyModes(ConcurrencyModes.Global)]
internal class ConsoleExporter<T> : BaseExporter<T>
{
    ...
}
```


src/OpenTelemetry/SimpleExportProcessor.cs
```diff
@@ -21,11 +22,32 @@ public abstract class SimpleExportProcessor<T> : BaseExportProcessor<T>
...
+        {
+            var attr = (ConcurrencyModesAttribute)attributes[attributes.Length - 1];
+            this.supportedConcurrencyModes = attr.Supported;
+        }
     }
 
     /// <inheritdoc />
     protected override void OnExport(T data)
     {
+        if (this.supportedConcurrencyModes.HasFlag(ConcurrencyModes.Multithreaded))
```
reyang, line 37: @CodeBlanch do you have some idea how to make this code less ugly? Currently there are lots of duplicated stuff.

src/OpenTelemetry/SimpleExportProcessor.cs
```diff
@@ -21,11 +22,32 @@ public abstract class SimpleExportProcessor<T> : BaseExportProcessor<T>
     protected SimpleExportProcessor(BaseExporter<T> exporter)
         : base(exporter)
     {
+        var exporterType = exporter.GetType();
+        var attributes = exporterType.GetCustomAttributes(typeof(ConcurrencyModesAttribute), true);
```
rajkumar-rangaraj, line 26: Would `GetType()` or `GetCustomAttributes()` cause any issues with AOT trimming?

reyang: I think no 
image

src/OpenTelemetry/SimpleExportProcessor.cs
CodeBlanch: Most simple thing would probably be like this...

```csharp
public abstract class SimpleExportProcessor<T> : BaseExportProcessor<T>
    where T : class
{
    private readonly object syncObject = new();
    private readonly ConcurrencyModes supportedConcurrencyModes;

    /// <summary>
    /// Initializes a new instance of the <see cref="SimpleExportProcessor{T}"/> class.
    /// </summary>
    /// <param name="exporter">Exporter instance.</param>
    protected SimpleExportProcessor(BaseExporter<T> exporter)
        : base(exporter)
    {
        var exporterType = exporter.GetType();
        var attributes = exporterType.GetCustomAttributes(typeof(ConcurrencyModesAttribute), true);
        if (attributes.Length > 0)
        {
            var attr = (ConcurrencyModesAttribute)attributes[attributes.Length - 1];
            this.supportedConcurrencyModes = attr.Supported;
        }
    }

    /// <inheritdoc />
    protected override void OnExport(T data)
    {
        if (this.supportedConcurrencyModes.HasFlag(ConcurrencyModes.Multithreaded))
        {
            this.Export(data);
            return;
        }

        lock (this.supportedConcurrencyModes.HasFlag(ConcurrencyModes.Global)
            ? GlobalLock.Instance
            : this.syncObject)
        {
            this.Export(data);
        }
    }

    private void Export(T data)
    {
        try
        {
            this.exporter.Export(new Batch<T>(data));
        }
        catch (Exception ex)
        {
            OpenTelemetrySdkEventSource.Log.SpanProcessorException(nameof(this.OnExport), ex);
        }
    }
}

internal static class GlobalLock
{
    public static object Instance { get; } = new();
}
```

* Using `MethodImplOptions.AggressiveInlining` so it doesn't introduce an extra call in there.
* Added in the global lock because I was curious what that might look like.

reyang: Thanks! For "Global", I was thinking of something like named mutex - which works across multi processes.
For example, if OpenTelemetry .NET is used by dotnet-cli (e.g. dotnet.exe) during build, there could be multiple processes running concurrently and outputting metrics/logs/traces to stdout, and we should make sure these are coordinated. This might require more thinking and experiments, most likely I'll scope down this PR by removing "Global" for now.

reyang: Scoped down this PR.

src/OpenTelemetry/ConcurrencyModes.cs
```diff
@@ -0,0 +1,36 @@
...
+    | | +------------- (reserved)
+    | +--------------- (reserved)
+    +----------------- (reserved)
+    */
+
+    /// <summary>
+    /// Reentrant, the component can be invoked recursively without resulting
+    /// a deadlock or infinite loop.
+    /// </summary>
+    Reentrant = 0b1,
```
CodeBlanch, line 29: We removed `Global` for now should we also remove `Reentrant`? Doesn't seem to be used at the moment.

reyang: My gut feeling is no. Multithreaded and Reentrant are closely related, the existing implementation (which uses `lock(obj)`) implies that the exporter supports reentrancy but not multithreading (which is buggy IMHO).

reyang: No strong feelings though, I can scope this out and add it later once we figured out how to correctly handle reentrancy.

CodeBlanch: Interesting. So if the attribute is not specified and defaults to 0 our implied default is essentially `SingleThreaded | Reentrant`?

Should the defined `Reentrant` flag then negate the default behavior?

```csharp
public enum ConcurrencyModes : byte
{
    /// <summary>
    /// Nonreentrant, the component cannot be invoked recursively without resulting
    /// in a deadlock or infinite loop.
    /// </summary>
    Nonreentrant = 0b1,

    /// <summary>
    /// Multithreaded, the component can be invoked concurrently across
    /// multiple threads.
    /// </summary>
    Multithreaded = 0b10,
}
```

reyang: > Interesting. So if the attribute is not specified and defaults to 0 our implied default is essentially `SingleThreaded | Reentrant`?

Maybe. I personally consider this as a bug in the current SDK implementation.

> Should the defined `Reentrant` flag then negate the default behavior?

I guess no, better to treat it as a bug, then figure out how to fix it in a non-breaking way.

reyang: My thinking - if the exporter doesn't have the attribute at all, fall back to the old (and buggy) behavior, if the exporter has the attribute, start to enforce the correct behavior.

eerhardt:
@reyang @CodeBlanch - I see this got closed due to inactivity. Do we have a plan for removing the private reflection in the GenevaExporter?

reyang:
> @reyang @CodeBlanch - I see this got closed due to inactivity. Do we have a plan for removing the private reflection in the GenevaExporter?

@CodeBlanch would you follow up?

CodeBlanch:
@eerhardt Still working on the plan. I've been exploring some other possible APIs/alternatives. Hoping to have something to propose soon and then we'll pick a direction to have in place for the next release.

PR #5647, opened 2024-05-23
[baggage] validate key and values during baggage injection/extraction

### LINKED ISSUE #5479: [BaggagePropagator] Do not encode and decode keys
According to the spec only the value should be encoded. 

The key is not expected to be percent encoded. More:
- https://www.w3.org/TR/baggage/#key
- https://github.com/w3c/baggage/issues/131
- https://github.com/open-telemetry/opentelemetry-go/issues/4946

If the key is not W3C Baggage spec compliant then key-value pair must not be propagated and a warning should be emitted.

lachmatt:
Fixes #5479 
Design discussion issue #

## Changes

- Baggage item key is no longer encoded and decoded - it is validated against `token` requirement from the [spec](https://w3c.github.io/baggage/#key) instead
- Baggage item value is validated during extraction
- Only invalid baggage items are rejected during injection
- All of the baggage items are rejected during extraction if any of them is invalid (more strict than injection)

NOTE:
Amount of allocations (e.g in extract part) can be reduced, I'd prefer to address it in a separate PR.


src/OpenTelemetry.Api/Internal/OpenTelemetryApiEventSource.cs
```diff
@@ -108,4 +108,16 @@ public void FailedToInjectBaggage(string format, string error)
...
     }
+
+    [Event(12, Message = "Baggage item key is invalid, key = '{0}'", Level = EventLevel.Warning)]
+    public void BaggageItemKeyIsInvalid(string key)
+    {
+        this.WriteEvent(12, key);
+    }
+
+    [Event(13, Message = "Baggage item value is invalid, value = '{0}'", Level = EventLevel.Warning)]
+    public void BaggageItemValueIsInvalid(string value)
```
Kielek, line 119: Q: Should we inform what was the key for the given value? It might be useful for debugging.

Kielek (approved):

vishweshbankwar:
@lachmatt - Could you please clarify where these requirements are coming from?

* Only invalid baggage items are rejected during injection
* All of the baggage items are rejected during extraction if any of them is invalid (more strict than injection)

vishweshbankwar:
> > @lachmatt - Could you please clarify where these requirements are coming from?
> > 
> > * Only invalid baggage items are rejected during injection
> > * All of the baggage items are rejected during extraction if any of them is invalid (more strict than injection)
> 
> This is not a requirement, but a choice made during implementation (I'm open for suggestions how to improve it).
> 
> For the injection part , Baggage API allows arbitrary strings as [names](https://github.com/open-telemetry/opentelemetry-specification/blob/f3aeadbd152ce50c4d2cc9bc94902613bb49cffd/specification/baggage/api.md?plain=1#L38). This is less restrictive than `key` requirements from the [W3C Baggage specification](https://www.w3.org/TR/baggage/#key). What would be the preferred behavior for situation when one of the names previously accepted in Baggage API `Set` operation doesn't meet `token` requirement from the [W3C Baggage specification](https://www.w3.org/TR/baggage/)? I assumed it might be beneficial to propagate valid items in that case.
> 
> For the extraction part, current behavior is consistent with e.g. how invalid items are handled during `tracestate` [extraction](https://github.com/open-telemetry/opentelemetry-dotnet/blob/b9d56aa64ecbad7a56c61941a760abd210051590/src/OpenTelemetry.Api/Context/Propagation/TraceContextPropagator.cs#L223).
> 
> As far as I know, invalid values are handled differently by different language implementations, e.g.:
> 
> * in Java, only invalid baggage items are ignored, during both injection and extraction (valid items are propagated)
> * in Go, all baggage items are ignored during extraction if invalid item is recognized

@lachmatt - Shouldn't this be coming from spec? 

Looking at the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/baggage/api.md#:~:text=However%2C%20the%20specific%20Propagators%20that%20are%20used%20to%20transmit%20baggage%20entries%20across%20component%20boundaries%20may%20impose%20their%20own%20restrictions%20on%20baggage%20names.%20For%20example%2C%20the%20W3C%20Baggage%20specification%20restricts%20the%20baggage%20keys%20to%20strings%20that%20satisfy%20the%20token%20definition%20from%20RFC7230%2C%20Section%203.2.6)

> However, the specific Propagators that are used to transmit baggage entries across component boundaries may impose their own restrictions on baggage names. For example, the [W3C Baggage specification](https://www.w3.org/TR/baggage/#key) restricts the baggage keys to strings that satisfy the token definition from [RFC7230, Section 3.2.6](https://tools.ietf.org/html/rfc7230#section-3.2.6)

`May` impose restrictions - This seems to suggest it is not a MUST?

Also, looking at this https://www.w3.org/TR/baggage/#mutating-baggage. Looks like there is already a suggested approach for values.

> If a system determines that the value of a baggage entry is not in the format defined in this specification, it MAY remove that entry before propagating the baggage header as part of outgoing requests.

But a different approach is proposed [here](https://github.com/open-telemetry/opentelemetry-dotnet/blob/b557142e9f1ba66c80a0d4dae27d243113af0402/src/OpenTelemetry.Api/Context/Propagation/BaggagePropagator.cs#L171)?

xiang17:
> @vishweshbankwar I created an issue in the [spec](https://github.com/w3c/baggage/issues/137)

I think the spec says to remove **that (baggage) entry** if conditions are met, rather than dropping the entire baggage.
> If a system determines that the value of a baggage entry is not in the format defined in this specification, it MAY remove **that entry** before propagating the baggage header as part of outgoing requests.

src/OpenTelemetry.Api/Context/Propagation/BaggagePropagator.cs
```diff
@@ -116,7 +131,8 @@ internal static bool TryExtractBaggage(string[] baggageCollection, out Dictionar
 
             if (string.IsNullOrEmpty(item))
             {
-                continue;
+                baggage = null;
+                return false;
```
xiang17, line 135: It feels very strict to ignore the entire baggage collection if any entry is invalid. I'm not so familiar with the baggage spec, but the [tracestate spec](https://www.w3.org/TR/trace-context/#tracestate-header-field-values) specifically calls out that empty list members are allowed:

> Empty and whitespace-only list members are allowed. Vendors MUST accept empty tracestate headers but SHOULD avoid sending them. Empty list members are allowed in tracestate because it is difficult for a vendor to recognize the empty value when multiple tracestate headers are sent. Whitespace characters are allowed for a similar reason, as some vendors automatically inject whitespace after a comma separator, even in the case of an empty header.

src/OpenTelemetry.Api/Context/Propagation/BaggagePropagator.cs
```diff
@@ -84,20 +84,35 @@ public override void Inject<T>(PropagationContext context, T carrier, Action<T,
...
                 KeyValuePair<string, string> item = e.Current;
-                if (string.IsNullOrEmpty(item.Value))
+
+                if (!ValidateKey(item.Key))
                 {
                     continue;
                 }
 
-                baggage.Append(Uri.EscapeDataString(item.Key)).Append('=').Append(Uri.EscapeDataString(item.Value)).Append(',');
+                if (baggage == null)
```
xiang17, line 97: Is there a need to differentiate null vs empty? This would get checked in every iteration of the loop.

lachmatt: There isn't a need to differentiate between these, I will modify it.

src/OpenTelemetry.Api/Context/Propagation/BaggagePropagator.cs
lachmatt: According to [W3C Baggage spec](https://w3c.github.io/baggage/#definition) empty entries are invalid - entry needs to at least contain a `key` satisfying a [`token`](https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.6) requirement (note at least one char from `tchar` range) followed by a `=`.

src/OpenTelemetry.Api/Context/Propagation/BaggagePropagator.cs
```diff
@@ -131,33 +147,146 @@ internal static bool TryExtractBaggage(string[] baggageCollection, out Dictionar
...
+    private static bool ValidateValue(string encodedValue)
+    {
+        var index = 0;
+        while (index < encodedValue.Length)
+        {
+            var c = encodedValue[index];
+
+            if (c == '%')
+            {
+                if (!ValidatePercentEncoding(index, encodedValue))
```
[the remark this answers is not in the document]
xiang17: I've realized this is assuming the string is encoded value. You can ignore above comment's case.

However, the next `if` branch checks `ValidateValueCharInRange` which is assuming the string is already decoded. The two different assumptions here smells not good. i.e. Could there be a char that's invalid to be a `baggage-octet` but won't be detected here by `ValidateValueCharInRange`?

src/OpenTelemetry.Api/Context/Propagation/BaggagePropagator.cs
```diff
@@ -131,33 +147,146 @@ internal static bool TryExtractBaggage(string[] baggageCollection, out Dictionar
 
                 if (pair.IndexOf('=') < 0)
                 {
-                    continue;
+                    baggage = null;
+                    return false;
                 }
 
                 var parts = pair.Split(EqualSignSeparator, 2);
                 if (parts.Length != 2)
```
xiang17, line 155: List-member could contain multiple `=` signs. For example, 
1. in property.

> list-member            =  key OWS "=" OWS value *( OWS ";" OWS property )
property               =  key OWS "=" OWS value
property               =/ key OWS

2. [A value can also contain equal signs.](https://github.com/open-telemetry/opentelemetry-go/blob/59bbe11641bb8af130ecf0a4bb48a2b11479285c/baggage/baggage.go#L285C23-L285C48)

test/OpenTelemetry.Api.Tests/Trace/Propagation/BaggagePropagatorTest.cs
```diff
@@ -118,50 +117,100 @@ public void ValidateLongBaggageExtraction()
...
-        Assert.True(actualBaggage.ContainsKey("key4"));
-        Assert.Equal(" !\"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]^_`abcdefghijklmnopqrstuvwxyz{|}~", actualBaggage["key4"]);
+    [Theory]
+    [InlineData("key(=value")]
+    [InlineData(@"key=\\")]
+    [InlineData("key=val%")]
+    [InlineData("key=%gg")]
+    [InlineData("key=val%2")]
+    [InlineData("key=val ue")]
+    [InlineData("key=val%IJK")]
```
xiang17, line 182: Add a case which has valid percent coding triplets, but invalid as value in list-member once URI escaped?

src/OpenTelemetry.Api/Context/Propagation/BaggagePropagator.cs
xiang17: I see. They are indeed different on whether allowing list-member to be OWS.

src/OpenTelemetry.Api/Context/Propagation/BaggagePropagator.cs
lachmatt: Ad 1. Properties are not handled/recognized properly currently, this is something that needs to be addressed, but as it requires changes in both Baggage API and propagator, I'd prefer to address it in a separate PR.
Ad 2. This will be handled properly - see test case added in https://github.com/open-telemetry/opentelemetry-dotnet/pull/5647/commits/6b88989ce3243593c4db0beea6774f039ea4222e

lachmatt: I created [issue](https://github.com/open-telemetry/opentelemetry-dotnet/issues/5677) to track handling of properties.
Currently, properties attached to a baggage entry are treated as a part of a value.
With baggage entry value validation added in this PR (`ValidateValue`), such entries will be considered invalid due to the existence of `;` char. This might need to be adjusted.

